PR #20, opened 2018-09-29
Build: Compile with current HEAD of vstgui

kmatheussen:
The version of VSTGUI currently in the project doesn't compile with Linux, so this pull request changes the code to compile with current HEAD of vstgui. (I've gotten the the plugin to run on Linux now, although it crashes when launching the GUI.)

Note that I wasn't patient enough to figure out how to add changes to the "vst3sdk" submodule, so this pull request doesn't include that.

Also, I don't really understand the logic in the HEAD of vstgui with regards to the k* constant. I think the code is correct as I have done it, but it seems flaky to OR toghether constants pulled out of different enums. But it seems like you are supposed to do that now.


Jorgeelalto:
I tried the following:
- Clone the Surge repo
- Update the submodules
- Apply the diff patch
- Remove vst3sdk folder and clone the vst3sdk repo from Steinberg media
- Download the VST_GUI folder from steinberg.net and copying the vst2sdk to the previous folder like usual
- Compile with premake and everything

But now it throws me again the error of <vstgui/vstgui.h> not found. Any ideas?

GeorgeNs:
you will need to clone the sdk with --recursive,otherwise your files will be empty!
Also it sounds crazy but you will need to re-update/re-clone the vstgui folder.
You can see here all the steps:
[https://github.com/kmatheussen/surge/commit/ac5efff384f886f27c00a1518296f59599f14790](url)

Even if you build the plugin you will have many limitations as I described above.
Kjetil too,he said it's not going to work with ardour

mickski56:
@tank-trax wine has a poor implementation of d2d which i think is the problem with the wine version.
I get many fixme:d2d:d2d_ messages and Surge looks the same as your screenshots. I don't think this is going to be fixed in wine anytime soon.
Is there an asio implementation for virtualbox ?

tank-trax:
I only ran it in VirtualBox to see the GUI and also in the hopes that using Process Monitor could provide me a clue to get the GUI looking proper in WINE. As far as I know there is no ASIO for VirtualBox.

mickski56:
Still I'm optimistic that there will be a native LInux version sooner rather than later.

GeorgeNs:
> Still I'm optimistic that there will be a native LInux version sooner rather than later.
@mickski56 

HaHa,me too, but don't expect too much it's quite troublesome for many reasons.
If you haven't try already,there are many other updated, robust and with more DSP going on synthsss
such as ZynAddSubFX, Helm, Rack that work perfect on linux. Cheers

premake5.lua
```diff
@@ -81,10 +81,16 @@ elseif (os.istarget("linux")) then
 
    defines { "WINDOWS=0" }
    
-   buildoptions { "-std=c++17", "-I/home/kjetil/SDKs/VST_SDK/VST2_SDK" }
+   buildoptions {
+                "-std=c++17",
+                "-I/home/kjetil/SDKs/VST_SDK/VST2_SDK",
```
abique: What is this hard path?

tank-trax: replace your username with kjetil

GeorgeNs: Actually it's not important,you do not need to replace anything since you are going to merge the CONTENTS of vst2_sdk folder (OLD version needed!) inside the vst3sdk folder

abique:
Thank you. :)

GeorgeNs:
Yes, of course,sure, but now It Does!Haha

esaruoho:
@kmatheussen i'd love to hear more about this PR and whether segments of it could be isolated into more mergable parts? @abique had trouble building, so he reverted this PR, but there are so many commits in this, surely some of it could be isolated, PR'd and then merged?

tank-trax:
Perhaps a long run solution (although probably the most difficult and time consuming on the short term) would be to modify the Linux version so that is does not use the app's built in GUI and instead uses the generic GUI provided by whichever DAW (like Digits) the synth would be loaded in.

Jorgeelalto:
> Perhaps a long term solution (although probably the most difficult and time consuming on the short term) would be to modify the Linux version so that is does not use the app's built in GUI (like Digits) and instead uses the generic GUI provided by whichever DAW the synth would be loaded in.

Definitely, I'm now trying to compile it again and even making a GUIless version will be enough for now.

esaruoho:
Yeah totally. If GUI-less is a low-hanging fruit then surely getting that running would be a good step forwards!

esaruoho:
people are now working on headless #502 after getting the GUI to work on macOS + Windows and Linux.

baconpaul:
Oh jeez you terrified my @esaruoho by putting a comment on here and having it show up in my notifications. I thought this was live and swept today!!! Ha. 

Back to coffee and lowering blood pressure.

Yeah PR #502 will introduce a gui-less (but not guileless) version of surge once @jsakkine does a review today.

PR #33, opened 2018-11-24
WIP of OSX instructions

esaruoho:
After running these instructions, Xcode says:

`surge-au`  `plugin.h file not found (called by aulayer.h, in file aulayer_presets.cpp:1:)`.

`surge-vst3` `Cannot initialize a parameter of type 'wchar_t *' with an lvalue of type 'Steinberg::Vst::String128' (aka 'Steinberg::Vst::TChar [128]')

`surge-vst2` will succeed - IF you have disabled `Enable Strict Checking of objc_msgSend Calls` by switching it from Yes to No. However, the surge.vst actually does nothing, at least not on my computer. I'm not even able to have it display with Renoise.

however, at least we're closer than ever before to getting Surge running with OSX.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,7 @@
 # Surge
 
-This is the synthesizer plug-in Surge which I previously sold as a commercial product as the company [vember audio](http://vemberaudio.se). 
+This is the synthesizer plug-in Surge which I previously sold as a commercial product as the company [vember audio](http://vemberaudio.se).
+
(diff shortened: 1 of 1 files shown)


esaruoho:
@JurekR you might find this PR amusing :)

JurekR:
It is good somebody takes care now :)

esaruoho:
added kvr forum discussion link to it, i hope there's some actual discussion there instead of "download how please thx plox"

abique:
Thank you :-)

PR #35, opened 2018-11-27
Linux build fix for VST2

jsakkine:
src/common/gui/SurgeBitmaps.cpp:6:8: error: ‘atomic’ does not name a type; did you mean ‘atoi’?
 static atomic<int> refCount = 0;
        ^~~~~~
        atoi


abique:
Thank you :-)

PR #36, opened 2018-11-28
Fix remaining Linux compilation issues

jsakkine:
There is still lot of non-standard cruft that causes warnings (because MSVC++ does not have a good reputation to follow standards at least in the past). These commits should at least fix the compilation errors both for VST2 and VST3.


abique:
Thank you, I had a doubt about the reinterpret_cast, but it is correct. ;-)

jsakkine:
Yeah, had to scratch my head for a while because last time wrote C++ was pre-C++11 times :-) One thing I've always wonder in C++ is that why can use `static_cast` with float as in data format level integers and floats basically share nothing but that is another story...

Issue #38, opened 2018-11-28
VST does not load with Linux version of Renoise


jsakkine:
I launch Renoise like this:

```
VST_PATH=$PWD renoise
```
And get this to the terminal when doing rescanning plugins:

```
Renoise LOG> VstPlugs: Instantiate FAILED (failed to open the DLL: '<XXX>/target/vst2/Release/Surge.so: undefined symbol: _ZN6VSTGUI8soHandleE') !!!
```
With a tool such as `nm` it is easy to define that, yes, it is an undefined symbol. It is also a symbol that is unused by the plugin, which can be easily deduced with `git grep`.

jsakkine:
Reported here: https://github.com/steinbergmedia/vstgui/issues/68

GeorgeNs:
I have tried many forks and even if I could build the vst there is no DAW that can detect it!
So, I just made a fork that works for linux in any DAW as long as you are not going to use the surge's default GUI . See INSTALL instructions if interested
Cheers

PR #39, opened 2018-11-28
Define soHandle when on Linux

whydoubt:
Much like hModule is defined when on Windows, it is necessary to define
VSTGUI::soHandle when on Linux.  This keeps loading the plugin from
raising an undefined symbol error.

Fixes: #38


abique:
Thank you. :-)

PR #40, opened 2018-11-28
Flag SurgeStorage::load_vt_wav() only for Windows

jsakkine:
Right now SurgeStorage::load_vt_wav() is flagged ambiguously. This commit
flags it consistently only for Windows. Without this change Renoise's
plugin loader will complain about undefined symbol.


esaruoho:
@abique @kurasu hi, looks like these two PRs are waiting to be merged, and the guys doing them are itching to commit more changes to the repo to take this closer to compile - can ya guys help with a quick review + merge?

src/common/SurgeStorage.cpp
```diff
@@ -356,8 +356,10 @@ void SurgeStorage::load_wt_wt(string filename, Wavetable* wt)
    wt_header wh;
    memset(&wh, 0, sizeof(wt_header));
 
-   fread(&wh, sizeof(wt_header), 1, f);
-   if (wh.tag != vt_read_int32BE('vawt'))
+   if (fread(&wh, sizeof(wt_header), 1, f) != 1)
+      fprintf(stderr, "SurgeStorage: reading the wavetable header failed.\n");
```
abique: Maybe it needs to further deal with the error here? Could it return an error code or throw an exception? Sorry I'm not too familiar with that code myself.

src/common/SurgeStorage.cpp
```diff
@@ -371,7 +373,8 @@ void SurgeStorage::load_wt_wt(string filename, Wavetable* wt)
       ds = sizeof(float) * vt_read_int16LE(wh.n_tables) * vt_read_int32LE(wh.n_samples);
 
    data = malloc(ds);
-   fread(data, 1, ds, f);
+   if (fread(data, 1, ds, f) != ds)
+      fprintf(stderr, "SurgeStorage: reading the wavetable data failed.\n");
```
abique: Same here, is fprintf(err) enough?

src/common/SurgeStorage.cpp
jsakkine: I would suggest at this point a log message and a `FIXME` comment to implement proper error handling later. At least now we are aware if this happens (which should be rarely). It fixes the compiler warning and makes the error visible.

Things like exception handling and RTTI should be decided in the project level. For example, **WebKit** uses neither for maximum portability. Personally I think they are good for UI and application level code and not that good for engine level code but I don't think this is right time to nail that. For me here my personal preferences are less important than project wide conventions.

src/common/SurgeStorage.cpp
jsakkine: `fprintf()` does not set `error`. You query for EOF with `feof()` and _some_ error with `ferror()` but it doesn't give any better granularity. Also here I would suggest updating the patch with FIXME comment.

src/common/SurgeStorage.h
```diff
@@ -494,7 +494,9 @@ class SurgeStorage
    void load_wt(int id, Wavetable* wt);
    void load_wt(string filename, Wavetable* wt);
    void load_wt_wt(string filename, Wavetable* wt);
+#if WINDOWS
```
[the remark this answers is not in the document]
jsakkine: Yes. absolutely. At this point it would make sense to have a log message message here and `#warning FIXME`.

src/common/SurgePatch.cpp
```diff
@@ -741,7 +740,7 @@ void SurgePatch::load_patch(const void* data, int datasize, bool preset)
    patch_header* ph = (patch_header*)data;
    ph->xmlsize = vt_read_int32LE(ph->xmlsize);
 
-   if ((int)ph->tag == vt_read_int32BE('sub3'))
+   if (!memcmp(&ph->tag, "3bus", 4))
```
whydoubt: After taking all the endian-mangling into account, I believe this actually should be "sub3".

src/common/SurgePatch.cpp
```diff
@@ -785,7 +784,7 @@ unsigned int SurgePatch::save_patch(void** data)
    // void **xmldata = new void*();
    void* xmldata = 0;
    patch_header header;
-   header.tag = vt_write_int32BE('sub3');
+   memcpy(&header.tag, "3bus", 4);
```
whydoubt: "sub3"

src/common/SurgeStorage.cpp
```diff
@@ -356,8 +356,10 @@ void SurgeStorage::load_wt_wt(string filename, Wavetable* wt)
    wt_header wh;
    memset(&wh, 0, sizeof(wt_header));
 
-   fread(&wh, sizeof(wt_header), 1, f);
-   if (wh.tag != vt_read_int32BE('vawt'))
+   if (fread(&wh, sizeof(wt_header), 1, f) != 1)
+      fprintf(stderr, "SurgeStorage: reading the wavetable header failed.\n");
+
+   if (!memcmp(wh.tag, "twav", 4))
```
whydoubt: "vawt"

jsakkine:
Updated my pull request.

jsakkine:
Based on @whydoubt's remark I removed now the updates for tag comparisons and since the cpuid stuff can be fully done for macOS and Linux now (does not require any major GUI stuff) I did that too.

jsakkine:
@whydoubt, @abique: I updated now the commit so that the WINDOWS compilation flag is only used inside `SurgeStorage.cpp` as suggested by @whydoubt. It is cleaner as it trashed the source code less. I also added a remark about Linux and macOS to the commit message.

jsakkine:
Rebased to the latest master.

jsakkine:
BTW, I looked at  the WAV loading code and I can definitely make it fully working on all platforms but I need to see the plugin running on my desktop first. Does not make sense to code that type of functionality blind :)

esaruoho:
@kurasu & @abique d'you have bandwidth to look at this? would be great to get this and the other PR merged in eventually. they seem to be itching to push more changes and fixes in.

esaruoho:
@abique what do you think, could this land?

esaruoho:
@kurasu what's your take on whether this PR could land?

esaruoho:
@jsakkine d'you have the time to fix the conflicting `README.md`? It's probably a minor fix :)

esaruoho:
@jsakkine I've isolated the `README.md` fix in this PR to PR #65 so if you edit yours out, I guess this PR will be ready for review, huh?

jsakkine:
Rebased my PR. When we can except them to be pulled?

PR #50, opened 2018-12-08
add OBJC_OLD_DISPATCH_PROTOTYPES=1 to macOS

esaruoho:
as per discussion on https://github.com/kurasu/surge/issues/49#issuecomment-445348253
this is needed to evade the first fail-error in building this on OSX
```
 in Xcode 10 / macOS 10.14, the default prototype of objc_msgSendSuper has 
changed from the 3 args to a no arg type.
```

thanks for the assist, @baconpaul!

### DIFF
--- a/package-vst3.sh
+++ b/package-vst3.sh
@@ -1,0 +1,34 @@
+#!/bin/bash          
+
+# input config
+RES_SRC_LOCATION="resources"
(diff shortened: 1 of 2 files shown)


esaruoho:
seems like i screwed up this PR so gonna remake it, so it's only that bit.

esaruoho:
recreated at #56

PR #52, opened 2018-12-08
Correct the include path in the AU version of SurgeGUIEditor.h

baconpaul:
Hi! I'm chipping away on making the AU version work and in doing so found this small typo in a header file inside the #ifdef au block. Seems like a small enough change that I figure I'd fire in a PR to help other folks.

If you would rather have larger PRs rather than smaller ones, please just let me know and I can adjust.

Thanks for making this open source!


esaruoho:
@abique @kurasu any chance this could land? seems like a minor fix?

baconpaul:
I have a larger PR coming which actually gets it running in logic so closing this one.

Issue #58, opened 2018-12-08
Getting Audio Unit to link on macos


baconpaul:
Hi!

So I just got the audio unit code to link on macOS. Don't have it packaged but have it linking. There are a few larger things I had to do to do this.

The code I have isn't really mergeable yet I don't think. (I did some radical things like remove the VST target in my premake for now for instance). And I have no idea if it works. But if you want to see what I did the whole thing is pushed on this fork (https://github.com/baconpaul/surge/tree/au-build) and you can just check that out by cloning my fork and hopping on to branch au-build.

There's a few bit things which were wrong in addition to the minor bits and bobs I already put in other issues (like the OBJC_PROTOTYPE and the missing header I did a PR for). The amounted to 2 enormous things

* In order to build an AU V2 you need the AUPublic and PublicUtility code from the audio toolkit available. This is actually referenced in the premake5.lua (look https://github.com/kurasu/surge/blob/c6ad95dfa051649f5647b0064fe8b0548680c08a/premake5.lua#L460 at that line) but the source wasn't in lib. The source actually isn't all that easy to find but luckily other GitHub projects have it included so I did the following

     * Looked at the repo "AudioKit/AudioKit" where they have this code staged in several sub projects (like https://github.com/AudioKit/AudioKit/tree/master/Developer/AKSampler/Mac%20AUv2%20Plugin/AUPublic/OtherBases)
     * Checked it out, undid the AudioKit subdirectorying to put it flat
     * Added it in libs, added it as a target to my Xcode (since I had already generated with premake)
     * and voila. The #include "MusicDeviceBase.h" and all the ComponentResult stuff works

* So great now we have all the code we need to build an AUV2. But the other problem was that class names had changed since src/au too today. So go through and replace all the sub3_synth with SurgeSynthesizer and all the names_like_this with namesLikeThis, comment out one place I couldn't figure out in presets for a link error and put in a fix me, and

voila! A linking surge-au.

So anyway I'll keep picking on this some but if others are working I thought they may want to know how close I got and look at the code in my branch.

Hope this is helpful. Welcome all comments.

baconpaul:
Oh and I meant to tag @esaruoho in this since he and I were chatting about it in last few days

esaruoho:
@baconpaul could you provide an image or a list of the files you grabbed from @getdunne's folder in AudioKit? Would help immensely. :)

esaruoho:
[...]

PR #59, opened 2018-12-09
Build and run an Audio Unit on mojave with logic 10.4 and get sound, presets, but no UI

baconpaul:
Hi!

OK so this is a pretty massive pull request and I'm just some random dude on the internet. But this PR when applied means that on macOS 10.14 with Xcode 10.whatever and Logic 10.4 I get Surge running, responding to parameter changes, making sound, loading presets, etc... Basically everything *except* showing the UI.

Let me add a bit of explanation on what I had to do, a lot of which is detailed in #58 

* I had to load in the apple AUPublic classes needed to run an AUV2. I grabbed them from AudioKit but could also have grabbed them from juce or whatnot. These are the new files in libs/AUPublic and libs/PublicUtility. They are unmodified.
     * I then modified premake5.lua to include them properly

* The StorageLayer needed some work
      * There were some bugs in the filesystem implementation around null pointers and directory paths I fixed
     * There were a bunch of non-conservative null checks in the XML parsing in surge storage

* The au-layer hadn't been updated since SurgeSynthesizer was named sub3_synth so I did that including all the this_case to thatCase and all the name changes

* I fixed and left FIXMES with some things which were stopping it from running

* I fixed the include paths in premake5.lua and also added the OBJC_PROTOCOLS flag to make sure that objc_mesgSendSuper had the right protytpe

* I added a new install script for AU which runs auval and installs the presets and what not and added an au packaging script to make the item. I updated the Info.plist to be correct.

* I updated the README.md

So this looks like a big change - mostly because all that libs/ code comes in - but a lot of the changes are small chip aways.

I'd love it if we could merge this. I think it means that other people can work on AU. Along the way I also got the vst linking, but I have no idea if it works. I tried really hard to make sure I did nothing which will touch windows or linux (where I don't have a build environment).

Thanks to @esaruoho for chipping in ideas and testing as I went.

Let me know if you are interested in merging! I'm going pencils down now, happy that it works as well as it does!

Best


esaruoho:
@abique d'you have the bandwidth to look at this? :)

baconpaul:
Heya! I just realized overnight that a couple of these diffs are not exactly what I want so I'm going to add a minor commit to this PR in the next 24 hours which does two things

1. Remove the fprintf stderr I left in source outside of au (sorry about that)
2: Backs out the THIS_AND change I made to the XML parser which is why I started having to do those non conservative checks in the XML parse before ->ToDocument, and back out the associated changes I had do to in storage for the crashes. (The THIS_AND is really annoying in tinyxml but since the methods are non virtual it sort of allows dispatch on a null pointer which is gross and weird but hey).

The meat of the changes will still be in place but if you do get to review, please know these two will also be coming along before the PR is fully ready.

baconpaul:
OK so that commit backs down the changes I made along the way which aren't needed, and makes the total footprint of diffs much smaller (basically libs/ and src/au and one thing in SurgeSynth which is in an if block and some make shenanigans). Look forward to any and all feedback!

I'm now going to branch off of au-build to make the UI work. Have some initial progress there too but no promises if I'll do it this week.

esaruoho:
@abique @kurasu @kzantow hi guys, do you have the time to look at this? it's pretty hot stuff, getting an AU to build AND run AND play. Just missing the UI - surely someone well-versed in the Carbon<->Cocoa conversion of UIs might be inspired by seeing how far @baconpaul has progressed this?

baconpaul:
I’ve got a start on the cocoa ui also. Will start another issue for that if I get anywhere

esaruoho:
thanks heaps @kurasu

baconpaul:
Yes thanks so much @kurasu this is fantastic!

Last night I got the start of the Cocoa UI at least calling into objective C. I'll pull this and make a fresh branch for that work. Really appreciate it!

PR #70, opened 2018-12-14
premake: Define VST2 boilerplate cpp-files explictly

jsakkine:
If the boilerplate cpp-files are not enumerated explicitly in premake5.lua,
and if any of them is missing, the plugin will compile without errors but
obviously won't load.


jsakkine:
Probably a good practice/balance with premake5 would be to use explicit paths for any external boilerplate files and wildcards for the internal files.

esaruoho:
@asimilon - query, are bits in this PR something that help with https://github.com/asimilon/surge/commits/macOS_VST2_build ?

esaruoho:
@jsakkine all this has been added to #92 so you might wanna close this PR..

PR #68, opened 2018-12-13
Readme.MD, Premake5, Swedish comment translation to English in code, au-package.sh quickfix, wavetable rmb quickfix, alert on mac if no configuration.xml

esaruoho:
This PR updates
- `README.md` (more links, moved to the bottom)
- `Premake5.lua` (spaces replaced with tabs, beautification)
- tons of files (Swedish comments translated into English by Google Translate and by hand).
- fixes `au-package.sh` text so it talks about AudioUnit and not AU
- takes @baconpaul 's quickfix for #29 and does that too (RMB to open wavetable dropdown menu)
- takes @baconpaul 's quickfix for #89 for mac (alert user if configuration.xml not present in `~/Library/Application Support/Surge` -folder
- takes @kzantow 's quickfix for #77 for mac (hides `.` and `..`)
- Merged Surge.pdf PR #74 into this PR

@kurasu @kzantow @abique thoughts appreciated.

heads up, @baconpaul - not sure if these help but just on the offchance they do, have fun!

Fixes #67  
Fixes #29  (Thanks @baconpaul!)
Fixes #89 (Thanks @baconpaul)
Fixes #77 (Thanks @kzantow)

PR #74, opened 2018-12-14
Add Surge pdf manual to repo

esaruoho:
Now that @baconpaul got an AudioUnit (AU) created with a GUI - I think it's high time this repo have Surge.pdf, right?

@kurasu  hope that's fine with you?


jsakkine:
Prefer not to add binaries to GIT repositories unless that is the last straw. Are there sources available for this. If not, would keep it out of the GIT as we can never update it anyway.

jsakkine:
Better idea would be to redo it in ReStructedText/Sphinx and add a build target for it.

esaruoho:
@jsakkine what's a ReStructuredText/Sphinx? I guess I could do copypasta work but maybe @kurasu has the originals of this?

jsakkine:
http://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html

jsakkine:
Could be other options. That is the only that what I'm familiar with because it is used in the Linux kernel.

jsakkine:
"Kernel documentation with Sphinx, part 1: how we got here"
https://lwn.net/Articles/692704/
"Kernel documentation with Sphinx, part 2: how it works"
https://lwn.net/Articles/692705/

baconpaul:
I agree we should write docs in a format which is editable in emacs (or vi or whatever) and makes web pages. I was thinking “markdown and pngs in a docs directory” but happy for whatever the doc writer chooses as long as we choose something!

esaruoho:
@baconpaul i found it at http://vemberaudio.se/download/surge.pdf

basically, i think, i could type it up to some markdown format, but would rather not, as it is 38 pages long. but if @kurasu tells us: the original master document does not exist, i can type it up no problem. It'll just take me some time.

I'm mostly interested in there being a reasonably opensourced format for this, so would edge towards markdown. something that we can bake a PDF out of, for sure. as i'd want it to not be _only_ online. i need to dump it onto an iPad for instance. or have it available for offline use as a downloadable PDF.

esaruoho:
posted a question at #78 - let's hope Claes answers it!

esaruoho:
merged into #68, closing

PR #69, opened 2018-12-14
Cocoa UI on MacOS in Audio Unit with Zoom, LFO display, and several other fixes.

baconpaul:
Hi

This diff is the PR we need to build and correctly run the gui on MacOS. Nowhere near as big as the one to get the synth running, but quite a lot harder, mostly because I'm not really a cocoa expert and the documentation on the internet just sorta stinks. But I figured it out eventually and here we go.

The discussion of this is in #61 

Thanks to @esaruoho for feedback and stuff. 

Once this is done, I think the only step left to actually being able to do a 64 bit audio unit mac release of surge is making a packager rather than the hack job scripts I have together to build this now.

Appreciate any feedback, corrections, changes, concerns, or critiques.

Issue #85, opened 2018-12-15
Create a rudimentary MacOS packager


baconpaul:
@baconpaul will do this tomorrow, and it matches some of the issues in #83 and #84. 

To install Surge you need to

1. move Surge.component to the plugins folder
2. move the resources to the Application Support area
3. if Surge has never been installed, remover the audio unit cache 

This should be done properly in a .pkg but that’s a bit of a pain. So I propose that for now I just add a script which

1. makes a dmg with the appropriate name
2. Puts the component and resources and a script and a readme in there
3. The readme says run the script but don’t if you don’t know how to debug your Mac
4. The script does the above from the images in the dmg

Then we can at least share dmgs for alpha releases more easily

Again, I’m away from my coding today but wanted to memorialize this as I think it will help get us some more Mac alpha testers, even before all the code is merged. (Like: I could run this from my zoom-ui branch and get all my bug fixes in while waiting for @kurasu to review the stream of PRs over the next few days which will come if/once #69 is merged).

kzantow:
Some discussion on #87, but summarizing: what if we just provide one of those "drag this to this folder symlink" type of installers, and I (or someone else) could whip up some code that runs at startup to sync the `/Library/Application Support/Surge/...` required files from things included in the bundle? The biggest question I have about this is the `configuration.xml` -- someone could hose this and surge would not work. In terms of adding presets to the waveform and effects sections, would it make more sense to just merge customizations in with a locally stored version containing the bare minimum from the bundle, so even without any files there Surge would still work?

baconpaul:
The permissions thing is solved by using ~/Library.

Yes you can hose ~/Library/Application Support and break surge. I think some better error reporting in that case is useful. Here's the current code in SurgeStorage.cpp

```
   string snapshotmenupath = datapath + "configuration.xml";

   if (!snapshotloader.LoadFile(snapshotmenupath.c_str())) // load snapshots (& config-stuff)
   {
#if !MAC && !__linux__
      MessageBox(::GetActiveWindow(), "Surge is not properly installed. Please reinstall.",
                 "Configuration not found", MB_OK | MB_ICONERROR);
#endif
   }
```

which explains why the MAC and __linux__ users aren't seeing this error! I'll open a separate issue for that. 

But I don't want to change things like configuration file behavior and split it and what not now; I think we may be better focused in getting a regular engine which can build, run, and package macOS with few feature changes. Of course I broke this rule by implementing zoom!

[...]

Issue #83, opened 2018-12-15
Surge-AU UI: Waveform + LFO Waveform are not displayed


esaruoho:
As per https://www.youtube.com/watch?v=JdRZpDTi5QM

and as per a screenshot isolated from it:
<img width="1440" alt="fullscreen_15_12_2018__10_10" src="https://user-images.githubusercontent.com/4966687/50040726-d22c9980-0051-11e9-9cd8-d94c7baa2566.png">

@baconpaul you might find this interesting - it appears that the waveform on the top left corner and the LFO waveform on the bottom center is not rendered with the macOS 64-bit AudioUnit for now.

baconpaul:
Thanks!

In my build I see the one in the top left corner and that changes when I do things like change wave shape or wave table. But the bottom one is definitely missing.

I’m not coding today, but one thing I was thinking was I should put together at least a small hacky “make an alpha release” script which just makes a dmg with an install shell script, the component, and the resources. Then I can build that from my branch so we can get other folks testing more easily. I think I have a bit of time to do that tomorrow.

Also: I really appreciate all the building, looking, host checking, and work you’ve been doing here @esaruoho . As I told a friend over breakfast yesterday “isnt’ it cool that some random dude from the states (me) and some random dude from Sweden (you) are talking on the internet to get open source things working using github”. So just sending a positive thought your way for a good weekend!

esaruoho:
I've modified the ticket to be about the LFO waveform only. btw I'm not Swedish, instead Finnish.

Thanks for taking the time to write about the `~/Library/Application Support/Surge` and how `resources/data` needs to be copied there before the waveform on the top left corner is displayed :) Much appreciated!

Yeah it's really awesome work you've been doing with the audiounit, I'm just happy to be along for the ride!

I hope @kurasu is, too.

baconpaul:
Oh so sorry! Didn't mean to mistake your country of origin. But still a pleasure hacking along with you!!

esaruoho:
no worries, @baconpaul :)
i wonder where those lfo waveforms reside.. weird!

baconpaul:
So the LFO Waveform display is super clever. It actually configures an LFO and runs it, sampling the data, drawing the graph. So this problem is not a missing resource problem, its a "bug in the drawing code on mac" problem.

Just letting you know to make sure we don't conflate this LFO display issue with all the ~/Library/Application Support issues kicking around in last couple of days (which #89 makes a bunch better).

[...]

PR #87, opened 2018-12-16
[WiP] Include configuration in bundle

kzantow:
Some drop-downs didn't work because they were expecting a file in `/Library/Application Support/Surge/configuration.xml`, which I assume was put there from an installer. This just moves it into the bundle and loads it from there. Also builds off of #69

### DIFF
--- a/libs/filesystem/filesystem.cpp
+++ b/libs/filesystem/filesystem.cpp
@@ -12,53 +12,67 @@
 #include "filesystem.h"
 
 namespace std::experimental::filesystem {
-        path::path():
-        path("")
-        {}
-        
-        path::path(std::string filePath):
-        p(filePath)
-        {}
-        
-        path::operator std::string() {
-            return p;
-        }
-        
-        void path::append(std::string s) {
-            p.append("/");
-            p.append(s);
-        }
-        
-        const char* path::c_str() {
-            return p.c_str();
-        }
-        
-        std::string path::generic_string() {
-            return p;
-        }
-        
-        path path::filename() {
-            auto idx = this->p.find_last_of("/");
-            path p(this->p.substr(idx+1));
-            return p;
-        }
-        
-        std::string path::extension() {
-            auto idx = this->p.find_last_of(".");
-            return p.substr(idx);
-        }
+    // path class:
+    path::path():
+    path("")
+    {}
     
-        file::file(std::string filePath):
-        p(filePath)
-        {}
-        
-        file::operator class path() {
-            return p;
-        }
-        
-        path file::path() {
-            return p;
-        }
+    path::path(std::string filePath):
+    p(filePath)
+    {}
+    
+    path::operator std::string() {
+        return p;
+    }
+    
+    void path::append(std::string s) {
+        p.append("/");
+        p.append(s);
+    }
+    
+    const char* path::c_str() {
+        return p.c_str();
+    }
+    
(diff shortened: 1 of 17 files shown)


baconpaul:
Hi!

Thanks so much! But im not totally sure this is the right approach. Lots of things like patches and waves also come from lib/app support

#85 would fix this but today the post build local install script installer-local-au.sh sets up the app support dir

Since surgery uses filestystem heirarchg as category and has writeable patches, things will be outside the bundle so we just gotta write an installer.

Also check out #83 

Glad to be wrong! And super appreciate you jumping in!

kzantow:
@baconpaul right, as indicated this is a WiP and shouldn't be merged as-is. Even without an installer, the bundle could include the bare minimum to at least get access to very basic presets for effects and get access to the waveform selector, otherwise the synth is fairly limited. Perhaps this could be written in code (I just added the `configuration.xml` for expediency.

For user-provided additional content, that could (and still is) be read from `/Library/Application Support/Surge/...`, especially things like additional waveforms for the wavetable oscillator, etc.. I did not even see the `installer-local-au.sh` 🤦‍♂️  but it would be nice to have VST2/3 work consistently with the AU version as well. I don't use AU ever, so I'm not familiar with the expectation from end users of having an installer vs. copying the `<plugin>.component`. For VST, quite a few simply provide the `<plugin>.vst` and let users just copy it to the right spot.

esaruoho:
@kzantow well, let's see what `installer-local-au.sh` actually does and whether it could be modified into `installer-local-vst2andvst3.sh`..

kzantow:
@esaruoho even if the "right" thing to do is put everything in `/Library/Application Support/...` I'd much rather just write a bit that runs at plugin startup and automatically does the install (copies the xml file, makes any directories that don't exist, etc). Having users run a script to set it up seems silly to me, especially since it's really just making sure one file is available at the beginning. Maybe this will change as patches/waveforms/etc. are added to the distribution, but it would be pretty easy to just include those in the bundle and copy/sync them to the `Application Support` location. I'm happy to write the code that does this, then we could very easily (I expect) alter the bundle to do one of those installers that is just "drag this file into this folder symlink". WDYT? cc. @baconpaul @esaruoho @kurasu

esaruoho:
@kzantow hi, i pushed a PR #88 with (duplicated from installer-local-au.sh content modified to fit) `installer-local-vst2.sh` and `installer-local-vst3.sh` and modified the already existing `package-vst3.sh` .  any thoughts would be much appreciated.

kzantow:
@baconpaul agree; I'll update #87  with the approach for review

esaruoho:
@kzantow could this PR itself be closed if everyone's focusing on #87? we don't want to confuse @kurasu with too much work.. :)

kzantow:
@esaruoho yeah, I'd say we close this one and focus on #87 ; I'm almost done with changes we discussed

esaruoho:
@kzantow hmm .. did you just say that we close PR 87 and focus on #87 (same one)? :).. sorry, i'm easily confused. just trying to figure out where the "dump `.` and `..`" commit might be :)

kzantow:
@esaruoho haha oops, i meant close #88

esaruoho:
@kzantow ok, i closed #88

kzantow:
Closing as superseded by #92

PR #92, opened 2018-12-16
Cumulative fixes and zoomable ui

kzantow:
This may subsume #68 #69 #87 and other fixes/enhancements including a zoomable UI for AU

@baconpaul @esaruoho please have a look

PR #96, opened 2018-12-17
Project target/template for a standalone app

jsakkine:
Add a new target to premake5.lua called app that builds a standalone
application. Surge (and VSTGUI) files  are linked to the final
executable but it does not do anything useful yet. This just the initial
plumbing needed.

In addition, the target is only available on Linux. This is done
purpose. It is better to make it first work on one platform, and after
it is fully working, scale it to the other platforms. They are easy to
add once app itself is not a moving platform.

Linux is the most reasonable choice for this because on that platform
the standalone app is most desperately needed for development as plugin
targets do not work properly yet. Then, at least one target will support
Linux and probably the plugin issues get more easily fixed.

### DIFF
--- a/premake5.lua
+++ b/premake5.lua
@@ -494,3 +494,39 @@ if (os.istarget("macosx")) then
 	postbuildcommands { "./package-au.sh" }
 	
 end
+
+if (os.istarget("linux")) then
+    project "surge-app"
+    kind "WindowedApp"
+
+    defines
+    {
+        "TARGET_APP=1"
+    }
+
+    plugincommon()
+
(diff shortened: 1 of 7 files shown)


baconpaul:
This is great; but there will be a lot of merge pain if this gets merged before #92 - is there any way you could pull your changes against that branch or try and pull @kzantow's thread as a remote so you can see if there's a conflict?

jsakkine:
Rebased. I'll continue with GUI code afer this. Plan is to stub audio up until GUI is working and then use rtaudio as the backend.

PR #98, opened 2018-12-18
Build using XCode9 and homebrew llvm + Packages based installer.

asimilon:
I have added the necessary changes to `premake5.lua` for everything to build with XCode9 with homebrew llvm (tested with XCode 9.2 on macOS 10.12.6).
Also created a Packages based installer which should work fine for XCode10 as long as everything is building OK.


baconpaul:
This is fantastic! Thank you.

esaruoho:
what does this let us do? bake 32-bit audiounits? or use xcode 9 and macos sierra (maybe high sierra)? :)
sorry, i'm not very xcode-savvy.

baconpaul:
We definitely don’t want to bake a 32 bit AU. Mostly it lets developers who aren’t on bleeding edge work. But the Packages stuff is really what I was excited about. That’s cool.

esaruoho:
@baconpaul are we quite sure about https://github.com/kurasu/surge/blob/01d2741da60e38ec3fe3ae9b5dacb2a978a35d93/installer_osx/mv_resources.sh

where it copies `/Users/Surge`. then does some chown work, then actually rm -rf's `/Users/Surge` 

i would not expect any installer to be doing this kind of stuff. what if my username is actually Surge?

baconpaul:
Yeah that looks wrong to me, unless it is run in some weird chrooted environment or something.

baconpaul:
Also 
https://github.com/kurasu/surge/pull/98/commits/2052f60b9336439e45933e266bd47944344a8d91 Is already in master so I think @asimilon needs to rebase this PR?

asimilon:
I wasn't quite sure what was going on there with those 2 commits.  I fetched and merged upstream, and it says "no conflicts" here, but not sure what I'm supposed to do, some advice would be appreciated.
Regards `mv_resources.sh`, that is a work around for putting stuff into the `/Users` folder, as I never managed to work out how to install into there with Packages.  I can update that to use something more sensible, didn't occur to me the possibility that a user may be called Surge.

esaruoho:
@asimilon did you see what @kzantow just posted in this thread about the data/resources ? https://github.com/kurasu/surge/issues/85#issuecomment-448346653 <- looks like there's no need to do the copying or create these users. hope this is helpful

esaruoho:
@kzantow any chance you could take a look at this PR, maybe?

premake5.lua
```diff
@@ -337,6 +338,15 @@ if VST24SDK then
 
 	if (os.istarget("macosx")) then
 
+		if XCODE9 then
+			xcodebuildsettings
+			{
+				["CC"] = "/usr/local/opt/llvm/bin/clang";
```
kzantow: I'm not sure any of this sort of build stuff is necessary anymore. I've [implemented enough of the filesystem API](https://github.com/kurasu/surge/blob/master/libs/filesystem/filesystem.cpp) that you can just use the normally included clang with XCode 9+.

installer_osx/mv_resources.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+cp -R /Users/Surge ~/Library/Application\ Support/
```
kzantow: This is wacky; does the installer actually make a USER named Surge?

kzantow:
@asimilon it doesn't look like you've merged the current master branch, there are tons of fixes in there, including almost everything that this PR is doing (not a packager, but just already including the resources in the `.component` and `.vst` and automatically populating `~/Library/Application Support/Surge` with the right stuff. (caveat: only if the directory doesn't exist)

kzantow:
hm, maybe you started this branch off the latest master, d'oh sorry for not noticing that. It definitely compiles without the homebrew llvm. What's the error you get from current master?

asimilon:
The error I get from XCode 9.2 (highest version I can get to with Sierra) is
```error: invalid value 'c++17' in '-std=c++17'```
setting that to c++14 builds at least (haven't tested the result as yet).

Regards the
> automatically populating ~/Library/Application Support/Surge with the right stuff

that is not working for the VST2, it is dropping everything into `~/Library/Application Support/Surge/data` and consequently I get:
image

image

I'm not entirely sure that this is something the plugin should be doing at run-time, this is the installers job really.  Like this there is no need to have 3 copies distributed (compression will make that a moot point admittedly), or up to 4 copies left on the disk after installation.

**Comment by baconpaul** [comment]
Hmm ... lot going in in this PR.

I wonder if we should split it into

1: Changes for Xcode 9 (which are fine as long as they don’t interfere with Xcode 10) and

2: Build a packager (which perhaps we hash out on the issue where we discussed that as opposed to the thread). I agree that in an ideal world the packager would install the ~/Library/App Support stuff but I think @kzantow solution was a cleverer one than my requirement that you run a shell script or don’t even get an error. (At least now you get an error!)

I think if we want to end up with a “Release Build” (where we have a packager which is signed that normal people can use) we probably want to move towards a proper packager. If we can do that with system standard tools (that is, figure out the pkgbuild making which we discussed over in #85) that would be super best.  But to be honest, I’m never going to be the person creating a distribution package from this software! So my opinion on this isn’t that strong.

On C++ spec:  I’ll definitely use c++-14-isms in my regular coding and others may use c++-17-isms also since those seem further along on windows. But I’m also running the head of Xcode and macOS so can’t test other configurations.

Hope this helps!

**Comment by kzantow** [comment]
Hm. That must be a recently introduced bug if it's moving resources to `~/.../data`, I'll retest and make sure this works reliably.

I don't really know for sure, but most people would probably just download one version that works for their DAW, it really doesn't make sense to download and install AU, VST, and VST3. Even so, I've proposed what I think is a better solution than copying resources to that folder anyway and just merging any user uploaded files from there with the contents of the package. That way it's easy to update and redistribute updated `configuration.xml`, wavetables and such, and nothing really ever needs to be installed that could potentially be messed up by users editing it directly.

**Comment by baconpaul** [comment]
@asimilon yeah so generally “never work on master” is a good rule of thumb, even if you are forked. The easiest way to undo your master changes may just be to re-fork but if you are really committed to your repo you would do something like

```
Git remote add upstream http://github/Kurasu/surge
Git fetch upstream
Git reset —hard upstream/master
Git push origin
```

And it will reset your fork and repush that to github. The URL in the first one needs fixing of course and I am typing from memory on my ipad so ... uhhh ... check things before you do it! But the trick is to have two remotes, origin (which should be git@your-repo) and upstream (which should be http://.../Kurasu)

Then generally what I do is a branch-per-feature and generate prs from them. So once you’ve done that

```
Git checkout -b Xcode-9
(Code code code)
Git add / commit
Git push origin xcode-9
```

Then your github has the branch and you can PR from there.

PR #102, opened 2018-12-18
Fix VST2 project creation and packaging

kzantow:
VST2 project creation was erroneously including some VST3 files, causing compilation failures.

Additionally, as [reported here](https://github.com/kurasu/surge/pull/98#issuecomment-448370669) the scripts had the possibility of including the resources in the `data` subdirectory, `rsync` fixes that and makes it more efficient.

PR #103, opened 2018-12-18
Xcode9.2 and packaging fixes

asimilon:
Default clang for XCode 9.2 (last version possible in Sierra 10.12.6) complains about C++17.
It is possible to build without error if that is rolled back to C++14, however using Homebrew clang works fine and doesn't impose that restriction.

I updated the packaging scripts to copy the resource files and not the whole `data` folder into the bundle, as this was then being copied by the plugin into `~/Library/Application Support/Surge` as is (i.e. `configuration.xml` was at `~/Library/Application Support/Surge/data/configuration.xml` and not the parent where it should be, causing error when loading the plugin).

I also fixed the `premake5.lua` to remove the `targetextension` directive for VST3, as this was being honoured by XCode10, yet the package script expects the target file to be named `Surge.dylib`.  As XCode9.2 is *not* honouring this directive and outputting `Surge.dylib` regardless, I thought this is the best solution.


baconpaul:
Nice. Can you look at the comment I made over on the very similar #102 to @kzantow then we can proceed? Thanks!

kzantow:
wait, sorry. this needs more discussion.

using stock Xcode 9, without any homebrew `llvm` version, I can compile the current code just fine. I'm having a hard time understanding why any of this is needed?

kzantow:
```
$ clang --version
Apple LLVM version 9.1.0 (clang-902.0.39.2)
Target: x86_64-apple-darwin18.2.0
Thread model: posix
InstalledDir: /Applications/Xcode 9.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
```
What version of macOS are you on @asimilon ?

esaruoho:
@kzantow if i understood correctly, he's on macos sierra.. which one are you on?

baconpaul:
OK I will pause on merging any of them then!

asimilon:
Yep, Sierra 10.12.6 XCode 9.2
```
$ clang --version
Apple LLVM version 9.0.0 (clang-900.0.39.2)
Target: x86_64-apple-darwin16.7.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
```
As noted above, with the default it *immediately* complains about C++17 not being supported, to support it could set that down to C++14, but that seems a bit ridiculous just for someone like me to be able to build, or can add this small conditionals to the `premake5.lua` so I can continue to build.
I am not at all interested in upgrading to High Sierra or Mojave until absolutely necessary, since my system works absolutely fine without at the moment, but mostly because I won't be able to upgrade XCode to 9.4 (which is I presume which version you are using @kzantow), it would skip straight to XCode10, and that is full of bugs for the use case of my day job.

baconpaul:
If llvm 9.1.0 works and llvm 9.0.0 doesn't, can premake check on that version difference? Or put another way: How does that Xcode env variable get set on which you switch? I wonder if it we could just fix it by having the switch variable be XCODE9_SIERRA and update the readme?

I think it's great to make it so it builds for you but lets make sure the smallest number of people try to brew install llvm if their native compiler will work. That's roughly I think the goal we should shoot for.

By the way, for those of us on the latest greatest, Xcode 10 has some great code analysis which seemed to just install when I upgraded last night. Full of bugs, sure, but interesting software. But I've used as much emacs as I have Xcode on this project in the last week so ...

kzantow:
sorry, I had a brain fart -- totally understand now that this is contingent on the environment variable. but at the same time, it seems fine to just set c++14. I just tested both AU and VST2 work fine with c++14. can you compile #102 ?

asimilon:
@baconpaul It's much the same as the VST2 environment variable: set it up before running `premake5 xcode4` or `build-osx.sh`.  I explicitly called it XCODE92 for that reason that @kzantow mentioned he can build with XCode9 (again assuming it's 9.4 he has).  I did consider that maybe it should be mentioned as a "less than XCode 9.3" in the README, as I should imagine that it will work with prior versions of XCode, then maybe something like BREWBUILD would be a more sensible env name.

@kzantow I'm not sure it's a great idea for us to unilaterally decide to use C++14, there might be some other useful C++17 `std::` things to use going forward, who knows. :-D  I think though that it was C++17 purely to get the `filesystem` stuff available to aid with Linux port, but that is moot now.

Going to test those #102 changes now, and report back.

baconpaul:
@asimilon I think your points are correct. And yeah you did name it XCODE92.

I agree backing down from c++17 seems riskier than having an env variable and documentation, especially since we aren't sure how far up the standards the default linux and windows compilers will be, and we don't want cross compile issues.

@kzantow any thoughts?

asimilon:
#102 builds fine, I hadn't noticed you'd updated the c++17 setting there.

I think perhaps the best plan would be change the README.md accordingly:

## Building with an XCode that doesn't support C++17

Use [homebrew](https://brew.sh/) to install llvm

```brew install llvm```

and set environment variable `BREWBUILD` to "true", eg:

```export BREWBUILD="true"```


and then of course update the `premake5.lua` with the new env name.

**Comment by baconpaul** [comment]
Works for me and seems like the right approach. If you are in there say "Xcode 9.4 and higher support it" or some such maybe too?

Thanks for this! Sorry it took a little while to figure it out but want to make sure we agree before I do my first merge. Appreciate the caution and dialog.

@asimilon if you update on your branch and repush it will come into your PR automatically and then I can merge.

**Comment by kzantow** [comment]
@baconpaul @asimilon it's hard to say, I just updated the build for AU, VST2, and VST3 and tested them all with the c++14 setting, everything seems to work just fine in #102 . I think the c++17 was to get `<filesystem>` access, which of course doesn't work really out of the box on XCode <= 10

**Comment by baconpaul** [comment]
Yeah my thinking is if linux and windows are on 17 we should be too in mac land. I kinda like @asimilon's approach with the modified doc.

Although I just approved @esaruoho README.md changes (which were great) without thinking that they will end up with a conflict here. I'm sorry about that!

**Comment by asimilon** [comment]
#3 was having trouble with 9.4.1, so I'm a little bit worried about giving any specific versions, maybe I'll add something about the error you will get thrown by XCode.

I'm not sure what the problem is with building VST3 on macOS, all 3 are building fine for me here :)

What the best way to resolve the conflict?

**Comment by baconpaul** [comment]
OK I'm happy with this; it seems to contain some of the #102 fixes, makes it clear the problem is with 17, keeps us on 17, and is generally good. I fixed the README merge conflict and am going to merge it in unless i hear a vigorous "no" from @kzantow in the next 30 minutes or so.

Back to debugging MPE until then...

File: README.md [hunk 1]
```diff
@@ -129,6 +129,18 @@ An example of setting the environment variable `VST2SDK_DIR` would be:
 
 ***NOTE***: This environment variable needs to be set _before_ running `premake5 xcode4` - which generates projects / and is part of the `build-osx.sh` script.
 
+## Building a Surge.vst3 (VST3) with OSX/macOS
```
asimilon: I have no trouble building/packaging/running VST3, I think the reason that was added is rendered moot by this PR.

baconpaul: OK wanna just fix it in place in this PR? You can edit the README directly and I'll approve it. Thanks for the careful eyes!

And does vst3 work for you? If so I can look at zoom this week!!

asimilon: Actually I've just done a little more testing, it's a bit odd:

Live10 : VST2 works, AU won't load at all (VST3 not supported).
Reaper : VST3 works, AU loads in some weird bridge mode and I don't get sound, VST2 doesn't show up at all.
Cubase 9.5 : VST2 works, VST3 won't load at all (AU not supported).
Cubase 8.5 : VST3 loads but doesn't produce sound, no sign of VST2.
Logic Pro X 10.4.1 : AU works (VST of course not supported)

esaruoho: sure, no worries. it's just that i haven't seen anyone dump a .vst3 and go "well this definitely works". if it just works right off the bat (clone repo, submodules, build-osx, xcode and package) then that's great, just add it in as text :)

baconpaul: OK well that sounds roughly like it could be described as "here be dragons" for now. Why don't I merge this so we can close this thread then if/when we get a vst working across a couple of places we can update the build instructions with that sweep.

Sorry this took so long!

PR #111, opened 2018-12-19
Duplicate of #70

jsakkine:
I closed #70 because these were supposed to be part of a bigger PR but did not find them so here they are again.


esaruoho:
@baconpaul any cjance this could land?-)

baconpaul:
Sure sorry we missed that merge. Pulling it in. I can't test it since I don't build VST2 but it looks like an equiv to me. If there's a VST2 error as a result let me flag @asimilon and @kzantow who are building that.

PR #117, opened 2018-12-20
Pass AU Validation (move Surge.component to /Library and not ~/Library )

esaruoho:
@baconpaul @kzantow @asimilon you guys might find this of interest.

Problem: If running `installer-local-au.sh`, the AU Validation fails. Time to lit hair on fire and run around in a state of absolute panic. Something's wrong. Well, what could it be?

Upon closer inspection, the AU Validation is done based on components available in /Library/Audio/Plug-Ins/Component - if the file is in ~/Library/Audio/Plug-Ins/Component - it'll time out and you'll feel like something is wrong.

Simple fix: remove two `~`'s from the scrip.  Result: script now deletes the Surge.component in /Library - and moves the Surge.component to /Library. This means that the AU Validation will succeed instead of failing.

Pros: most apps will read /Library (global plugin folder) and might not even offer the possibility of having a secondary plugin folder (set to ~/Library )

Cons: here's yet another PR..

### DIFF
--- a/installer-local-au.sh
+++ b/installer-local-au.sh
@@ -3,8 +3,8 @@
 #touch src/au/aulayer.cpp
 xcodebuild build -configuration Release -project surge-au.xcodeproj || exit 1
 ./package-au.sh
-rm -rf ~/Library/Audio/Plug-Ins/Components/Surge.component
-mv products/Surge.component ~/Library/Audio/Plug-Ins/Components
+rm -rf /Library/Audio/Plug-Ins/Components/Surge.component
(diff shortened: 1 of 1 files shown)


baconpaul:
No that's not right. This is the problem I referred to on the other PR where if you run the packager it installs in /Library but when you develop that will mask ~/Library so you need to remember to delete /Library before you use the local install scripts.

The right check is to check for a component in /Library in install and issue a warning message.

So we should not merge in this PR. It's not the right solution. Sorry!

baconpaul:
Also that script can't have possible worked run as a normal user could it? You don't have permissions on /Library?

Issue #73, opened 2018-12-14
premake5 modification so that it grabs the repo submodule dependencies?


esaruoho:
Is there some way premake5.lua could be used to get the submodule dependencies in the GitHub repo?

esaruoho:
@baconpaul @kzantow @asimilon guys, d'ya think this is somehow possible? a lot of people, testing the github repo for the first time, seem to screw this up and forget to grab the submodule dependencies. if that could be wrapped into the premake5.lua, surely that'd be helpful?

asimilon:
I'm a bit salty when it comes to this sort of thing, people need to RTFM! ;-)

asimilon:
Less salt:

[test for file existence](https://stackoverflow.com/questions/4990990/check-if-a-file-exists-with-lua) in lua

execute command in [premake5 alpha 12](https://github.com/premake/premake-core/wiki/os.execute)

Would need to test if `os.execute("git submodule update --init --recursive")` is portable across platforms, one would hope so.

esaruoho:
i..err..hurrdurr.. i tried to "do stuff" and promptly failed at it.

tried both of these but i just couldn't..
```
if os.isfile(vst3sdk/LICENSE.txt) then
echo "hello"
else echo "hoo"
end

##function file_exists(vst3sdk/LICENSE.txt)
##   local f=io.open(name,"r")
##   if f~=nil then io.close(f) echo hallo return true else return false echo end end
##end
```
.. i guess i'm pretty c`LUA`less

baconpaul:
I think the way to do this is to do a check in the top of build-OS X.sh which ejects an error message with the line to copy and paste then does an exit 1. Safest possible approach and easy to code in shell.

I’d approve a PR which did that. You know like

```
If [ ! (-f vst/blah (]; then
  Echo “you don’t have vst/blah so you probably need Submodules. Try this:
  Echo git -blah 
  Exit 0
Fi
````

esaruoho:
attempting this now, @baconpaul

esaruoho:
fixed by #118

baconpaul:
Tx! Merged

PR #118, opened 2018-12-20
check for submodules, prompt to get them.

esaruoho:
check for submodules, if `vst3sdk/LICENSE.txt` not found, tell the user to get the submodules.
if the user has the submodules, start building as usual.
discussion from #73 
@baconpaul thanks for the suggestions


esaruoho:
@baconpaul thanks for the minor fixes, btw!

Issue #121, opened 2018-12-20
Root of the Repo


baconpaul:
In #116 @asimilon said

> I did think that maybe a good idea would be to clean up the root of the repo, and move any scripts into relevant sub-directories, eg:
> 
> ```
> kurasu/surge
>   - scripts
>     - windows
>     - osx
>     - linux
> ```
> 
> perhaps leaving the build scripts in the root would be a good idea though.
> _Originally posted by @asimilon in https://github.com/kurasu/surge/pull/116#issuecomment-448983103_

and I said

> Yeah I was thinking the same.
> 
> I was also thinking of making build-osx.sh a bit of a meatier script with functions and options. So you could do
> 
> ./build-osx.sh  # current behaviour
> ./build-osx.sh --install-local-au  # basically what my install local au does
> --install-local-vst
> --install-local
> --clean-builds
> --remove-installs
> --build-pkg
> you get the idea. Then we can just add functions to that on switches. Easy enough.

so then I created this issue rather than talking about a change on a now merged and closed PR.

baconpaul:
OK @asimilon said in the PR he liked the idea so I'll take a shot at it today. I'm tired of looking at vstgui and midi binary code so it will be a nice shift!

baconpaul:
OK all merged. The mac stuff is now all ensconced in subdirectories with just a beefier build-osx.sh

PR #116, opened 2018-12-19
macOS installer

asimilon:

PR #122, opened 2018-12-20
A New build.osx

baconpaul:
As discussed in issue #121 I have rewritten build.osx and move around the top of the repo to clean things up. build.osx now contains functions and switches for all the things that were becoming scripts.

I tested it all and it seems to work but I don't have VST2. @asimilon or @kzantow could you take a look and offer a review? Also @kzantow this may change some of your CI stuff, but I don't think all that much since build-osx.sh without args still does the same thing.

Would love another set of eyes on this though.

Thank you.


build-osx.sh
```diff
@@ -1,17 +1,208 @@
...
+run_build()
+{
+    flavor=$1
+    xcodebuild build -configuration Release -project surge-${flavor}.xcodeproj
+}
+
+default_action()
+{
+    run_premake
+    if [ -n "$VST2SDK_DIR" ]; then
```
asimilon: Since we're in a green field, I think would be prudent to check for the project directory
```
if [ -d surge-vst2.xcodeproj ]; then
```
everywhere instead of looking for the `env` variable

build-osx.sh
```diff
@@ -1,17 +1,208 @@
...
+
+prerequisite_check()
+{
+    if [ ! -f vst3sdk/LICENSE.txt ]; then
+        echo You have not gotten the submodules required to build Surge. Run the following command to get them.
+        echo git submodule update --init --recursive
+        exit 1
+    fi
+
+    # TODO: Check premake is installed
```
asimilon:
```
if [ ! $(which premake5) ]; then
```

asimilon:
Looks good reading through, will check out your branch to test shortly.

build-osx.sh
```diff
@@ -1,17 +1,208 @@
...
+    --premake)
+        run_premake
+        ;;
+    --build)
+        run_all_builds
+        ;;
+    --install-local)
+        run_install_local
+        ;;
+    --build-and-install)
```
kzantow: nice, this is what I was looking for; seems the most generally useful case for development

build-osx.sh
```diff
@@ -1,17 +1,208 @@
...
+{
+    flavor=$1
+    xcodebuild build -configuration Release -project surge-${flavor}.xcodeproj
+}
+
+default_action()
+{
+    run_premake
+    if [ -n "$VST2SDK_DIR" ]; then
+        run_clean "vst"
```
asimilon: This should be vst2
```
xcodebuild: error: 'surge-vst.xcodeproj' does not exist.
```

asimilon:
Error building VST2, see comment above.

asimilon:
I have another suggestion:
the output of the build is incredibly verbose, perhaps we could redirect stdout to `warnings-${flavor}.txt`
Something like:
```
xcodebuild ... > warnings-${flavor}.txt
if [ ! $(grep "BUILD SUCCEEDED" warnings-${flavor}.txt) ]; then
    echo "Build errors encountered whilst building ${flavor}!  See warnings-${flavor}.txt for details."
else
    echo "Build of ${flavor} succeeded, check warnings-${flavor}.txt for any warnings."
fi
```

baconpaul:
OK these are all super comments. Let me merge them in.

baconpaul:
Oh and @kzantow the discussion which started this over in the PR linked to the issue (#116 maybe?) had me suggesting a makefile before we ended up here. In the end I didn't go with a makefile because

1. We really have almost no dependencies and we have lots of file moving stuff so bash seemed easier. The actual code dependency management is done inside Xcode.
2: I know people won't have a screwed up install of #!/bin/sh. They shouldn't have a broken make either but...
3: For the last 22 years I have been looking at that page which reminds me the difference between $@ and $% and $* and so didn't feel like continuing the streak.

Ha! OK the last one is a joke. But I think we could have used either tool so I settled on simple sh for now.

build-osx.sh
baconpaul: Done!

build-osx.sh
baconpaul: Thanks.

kzantow:
@baconpaul this looks excellent as-is, it was more of a question whether make was more ubiquitous and could double so the build instructions are the same across platforms; I've only used make a bit, so don't really know the C++ ecosystem there; I'll certainly delegate to your decisions here

baconpaul:
@asimilon and @kzantow I pushed your code changes. Great suggestions all.

The idea of putting build output in a log is super smart. I went a bit beyond and actually grep out errors but it is already way way nicer.

Can you pull and take a look if it works with VST2 and then if it does I'll merge this in?

Thanks!

asimilon:
Tested and working fine here.  Good to merge.

build-osx.sh
```diff
@@ -148,7 +178,7 @@ run_clean_builds()
 run_clean_all()
 {
     run_clean_builds
-    rm -rf Surge.xcworkspace *xcodeproj target product
+    rm -rf Surge.xcworkspace *xcodeproj target products build_logs
```
asimilon: If wanted to be thorough should remove `obj` as well?

baconpaul: Oh yeah. I'll add that in and merge. Thanks!

asimilon:
Great work @baconpaul 👍

PR #124, opened 2018-12-20
Correct windows.wt wavetable header

whydoubt:
The file header indicates that there are sixteen waves contained in the
file when there are actually only nine.  This results in allocating more
memory than necessary, and attempting to read more data than the file
contains.

Modify the header to indicate the correct number of waves in the file.


baconpaul:
Hi @whydoubt - can you help me understand how you'd test this? This is a bit of the code I don't know that well.  Is it basically that wh.n_tables will end up being 12 rather than 9 and then we will puke out, and so now it is 9 and we don't? I understand the overalloc and failed read is bad but these are the wavetables that show up in the classic oscillator?

Sorry to ask a mundane question but I don't want to merge something I don't understand. Most probably this is one for @kurasu and not for me right now, since I'm still new in that part of the code.

whydoubt:
I don't what to suggest for testing this.  At one point there was a check on the return value from fread in the load_wt_wt method, but that seems to have been lost in a merge somewhere.

Note: windows.wt file size is 18444.

Currently: 12-byte header is read. Header contains value 16 for n_tables, 1024 for n_samples, flag for int16 is set.  ds = 2 * 16 * 1024 = 32768.  A 32768-byte buffer is allocated to contain the data.  An attempt to read 32768 bytes from the file is made.  Only 18432 bytes are read.  The return value from fread is not checked (as I mentioned above).  The rest of the buffer remains uninitialized, but I expect the code does not try to access it.

With this PR: 12-byte header is read. Header contains value 9 for n_tables, 1024 for n_samples, flag for int16 is set.  ds = 2 * 9 * 1024 = 18432.  A 18432-byte buffer is allocated to contain the data.  An attempt to read 18432 bytes from the file is made.  18432 bytes are successfully read.

Surge user documentation mentions something about the 9 windows waveforms as well.  Also, I imported the file into Audacity as raw 16-bit LSB values, and visually confirmed that it is actually 9 distinct waveforms.

baconpaul:
Well that's great info I didn't have! It helps me just do a pre-merge check. Thanks!

Since this is a binary file let me put a quick log of what I inspected in your diff.

```
struct wt_header
{
   char tag[4];
   unsigned int n_samples;
   unsigned short n_tables;
   unsigned short flags;
};
```

this is what we read

here's the current file, as you point out

```
paul:~/dev/music/surge$ od -d resources/data/windows.wt  | head -1
0000000     24950   29815    1024       0      16       4      14      42
paul:~/dev/music/surge$ od -c resources/data/windows.wt  | head -1
0000000    v   a   w   t  \0 004  \0  \0 020  \0 004  \0 016  \0   *  \0
paul:~/dev/music/surge$ od -x resources/data/windows.wt  | head -1
0000000      6176    7477    0400    0000    0010    0004    000e    002a
paul:~/dev/music/surge$ ls -l resources/data/windows.wt 
-rw-r--r--+ 1 paul  staff  18444 Dec  8 16:44 resources/data/windows.wt
```

So there's that 16 and 1024 aligned with the item in the read. 9*1024*2 + 12 is the file size.

```
paul:~/dev/music/surge$ git checkout whydoubt/windows_wavetable
HEAD is now at 5685a84 Correct windows.wt wavetable header
paul:~/dev/music/surge$ od -d resources/data/windows.wt  | head -1
0000000     24950   29815    1024       0       9       4      14      42
```

here's the 9 in your diff.

and finally that's the only change

```
paul:~/dev/music/surge$ cp resources/data/windows.wt ..
paul:~/dev/music/surge$ git checkout master
Previous HEAD position was 5685a84 Correct windows.wt wavetable header
Switched to branch 'master'
Your branch is up to date with 'origin/master'.
paul:~/dev/music/surge$ od ../windows.wt > wpull_od
paul:~/dev/music/surge$ od resources/data/windows.wt > wmaster.od
paul:~/dev/music/surge$ diff wpull_od wmaster.od 
1c1
< 0000000    060566  072167  002000  000000  000011  000004  000016  000052
---
> 0000000    060566  072167  002000  000000  000020  000004  000016  000052
```

editorial: Would be great if GitHub did that style diff for us on binaries!

Finally I made sure all the wavetables loaded and displayed in the mac AU.

So great. Yes that's super useful! I'll do the merge.

Issue #127, opened 2018-12-20
User patches don't save properly on MacOS


baconpaul:
Now I'm getting MPE working, I started making some patches for MPE which I was going to eventually move into the surge factory patches. But I discovered that on mac the save of user patches doesn't work.

I suspect it is something to do with this in SurgeStorage.cpp

```
      datapath = path;
      datapath += "/Surge/";
   }

   userDataPath = "~/Documents/Surge";

#elif __linux__
```

that ~ probably needs to be `getenv( "HOME" )`-ed. But put an issue here.

I will try and build and merge #125 before PR this.

PR #129, opened 2018-12-20
Users can save patches on MacOS now

baconpaul:
As mentioned in #127 user patches weren't saving. Now they are.

The first thing - expand ~ with getenv( "HOME" ) was pretty easy

But there's a second thing I'd like someone to peek at. In the stub of experimental::filesystem create_directories (which unlike create_directory is supposed to recurse; it is mkdir -p) was not recursing.

So I made it recurse using a hacked version of a C++ recursive make from stack overflow.

Could someone look at this before I merge it? @kzantow you've been in this code.

### DIFF
--- a/libs/filesystem/filesystem.cpp
+++ b/libs/filesystem/filesystem.cpp
@@ -10,8 +10,6 @@
 #ifdef TARGET_OS_MAC
 
 #include "filesystem.h"
-#include <cstring>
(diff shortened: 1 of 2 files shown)


baconpaul:
Thanks @kzantow. Going to merge.

I'm so close to having Surge working as a third MPE synth in my rig.  So close...

PR #131, opened 2018-12-21
Mpe with a ROLI Seaboard in Logic and AU

baconpaul:
OK so the MPE support in Surge is really quite complete but it wasn't activating properly with the Roli SEABOARD. And this PR has a lot of commits because in all the shuffles of PRs in the last few days I actually screwed up this branch. But @kurasu could I ask you to look at this one? (Look at the diffs not the commits sorry. If the commit splattage really bugs you I can recreate the branch).

I fixed one problem and added one feature

The feature is simple. As well as learning a modulation control you can directly assign it to a CC with a menu. Simple. I thought I would need that. I didn't but lets leave it in here since it is handy with other controllers.

But the real fix is that the ROLI Seabord 49 with firmware 1.1.7 sends an MPE Initialize message which doesn't match the MPE spec. It is close but not quite. So what I did was add a big honking comment and check for the ROLI case also and then turn on MPE.

With this, the unit sounds fantastic. If you are OK with this patch I will also put together a series of MPE patches in the new year (now that I can save patches!) and add those to the Factor folder.

I'm not going to merge this myself without your say-so @kurasu since it is adding a specific device behavior in. But it is a very nice and popular MPE device so I hope you are OK with this solution.

### DIFF
--- a/src/common/SurgeSynthesizer.cpp
+++ b/src/common/SurgeSynthesizer.cpp
@@ -689,6 +689,33 @@ void SurgeSynthesizer::sendParameterAutomation(long index, float value)
 
 void SurgeSynthesizer::onRPN(int channel, int lsbRPN, int msbRPN, int lsbValue, int msbValue)
 {
+    /* OK there are two things we are dealing with here
+     
+     1: The MPE specification v1.0 section 2.1.1 says the message format for RPN is
+     
+        Bn 64 06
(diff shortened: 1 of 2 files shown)


baconpaul:
Oh those commits are bugging me too much. I wouldn't approve this myself. 

Hold on let me make a fresh branch and a new PR

Issue #132, opened 2018-12-21
Add flag to output build logs with build-osx.sh


kzantow:
When building in CI, logs written to a file are lost when the VM terminates; it is much better to just output to terminal so results can be viewed. It would be fine to add a flag to make this behavior non-default.

baconpaul:
Easy enough to do. Also easy to do with an environment variable. Which do you prefer? Perhaps if you set a uniform ci build env var we can use that elsewhere too?

kzantow:
I don't really have a preference, though I guess an environment variable would be the most consistent if there are scripts all over using across platforms. I'd like to see what the CMake stuff looks like #128 .

baconpaul:
I just tossed in a PR which adds a --verbose flag. Check it out and if it works I'll merge it.

PR #134, opened 2018-12-21
Add a verbose option to build-osx.sh

baconpaul:
Hey @kzantow can you eyeball and try this for me? This lets you add a --verbose and get both the build log and stdout for the build. Think it's what you want in #132.

If you review it and it works for you I'll merge it over. Thanks!

### DIFF
--- a/build-osx.sh
+++ b/build-osx.sh
@@ -23,7 +23,11 @@ To do a simple local build which is not installed do
 
    ./build-osx.sh
 
-Options are:
+Other usages take the form of
+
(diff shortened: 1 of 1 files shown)


baconpaul:
Yeah it's bash-y. Kinda wish someone would just do this all with CMake or something. Chuckle.

Let me know if it works for you. I think it's fine but a second test would be great.

kzantow:
Just tried with/without `--verbose` and seems good, thanks @baconpaul 👍

PR #135, opened 2018-12-21
VST3: InitDll/ExitDll should already be defined

whydoubt:
InitDll and ExitDll should only be needed on Windows, and there they
should be defined in dllmain.cpp from the VST SDK.

If someone could test this on Windows, that would be great.
This should work fine, but if there is a problem, I could instead
just wrap the code with #if WINDOWS


baconpaul:
I'm sorry I have no windows environment in which I can test this PR.

Issue #130, opened 2018-12-20
Linux VST2: undefined references for __cpu_model, __cpu_indicator_init


whydoubt:
As a quick-n-dirty check for undefined references on Linux, I run
```
ld -o /dev/null target/vst2/Debug/Surge-Debug.so
```

I get `__cpu_model` and `__cpu_indicator_init` reported as undefined.  I found that it seems to be related to the use of `__builtin_cpu_supports`.  It can be fixed in one of two ways: link with gcc instead of g++, or add libgcc and libgcc_s to the libraries to link.

I wondered if anyone else had seen this?

This is on current master (5e6a7351) with no changes to it or vst3sdk,
```
rm -rf *.make Makefile obj target
VST2SDK_DIR=... premake5 gmake
make surge-vst2
```
It applies to the release version as well.
I read that this was supposed to be fixed in gcc 6, but I am using gcc/g++ version 7.3.1 on fedora 27.

#19

baconpaul:
That ld trick is cute. I always used a version of nm -g and grep and un-doing the name mangle in my head.

Anyway I can't build VST2 but I can build VST3 and then I get __cpu_mode and moduleHandle being undefined as well as all the stuff in SurgeVst3EditController which didn't build on mac.

baconpaul:
Apologies for screenshots. My VM isn't copy and pasting to my browser properly.

With these three diffs

First stub out SurgeVST3EditController just like I did on mac (two diffs)

<img width="856" alt="screen shot 2018-12-20 at 7 30 42 pm" src="https://user-images.githubusercontent.com/13387561/50318228-d90e3e80-048d-11e9-9734-22fbf6a18fca.png">

Second add -lgcc_s and -lgcc to premake5.lua

<img width="633" alt="screen shot 2018-12-20 at 7 32 03 pm" src="https://user-images.githubusercontent.com/13387561/50318263-0824b000-048e-11e9-98dd-732263a62ac0.png">

your LD trick shows me 


<img width="896" alt="screen shot 2018-12-20 at 7 33 04 pm" src="https://user-images.githubusercontent.com/13387561/50318277-2094ca80-048e-11e9-9daf-ac827b7eb8f2.png">

Without the premake change I also see __cpu_blah and without the ifdef a bunch of SurgeVST3Blah

Can you try with those diffs?

whydoubt:
[...]

PR #136, opened 2018-12-21
Clean up more undefined symbols in VST3 on Linux

whydoubt:


whydoubt:
See also #130

baconpaul:
Yup looks great thanks

PR #138, opened 2018-12-22
ppc: Remove architecture

jsakkine:
Does not make sense to carry code that nobody uses nor maintains
properly. Makes our life easier to just drop it.


jsakkine:
Probably better to peer check the diff carefully to make sure that I haven't removed anything that should not be removed.

jsakkine:
@esaruoho kindly helped me to test this change. Still compiles and loads  to various DAWs.

src/common/vt_dsp/vt_dsp_endian.h
```diff
@@ -6,88 +6,41 @@
...
 }
 
 __forceinline int vt_write_int32BE(int t)
 {
-#if !PPC
-   return swap_endian(t);
+#if (__linux__ || MAC || _M_X64)
+   return __builtin_bswap32(t);
 #else
-   return t;
```
baconpaul: This seems like it’s an actually diff, not just a PPC removal. Not super clear to me from inspection it is the same. Seems the old behaviour was just “return t” and now we have two branches of bswap32.

jsakkine:
I did also rewrite this function in `vt_dsp_endian.h`:

```
__forceinline int vt_write_int32BE(int t)
{
#if (__linux__ || MAC || _M_X64)
   return __builtin_bswap32(t);
#else
   return _byteswap_ulong(t);
#endif
}
```

src/common/vt_dsp/vt_dsp_endian.h
jsakkine: Hold on, checking...

jsakkine: Yeah, well there is only one call site for `swap_endian()` after the PPC code has been removed so I open coded it to `vt_write_int32BE()` and use compiler builtins instead of just copy-pasting the original code. Should probably mention this detail in the commit message.

jsakkine:
AFAIK the most recent POWER based machines that are not running in the server room were Playstation 3 and Wii U and for those there would much much more to do than just having SIMD code for oscilators and filters.

baconpaul:
Oooh you should merge to head of master! The new osx build scripts ar way less painful. Useful if you are doing the CMake port.

I can still test with the 4-day-ago starting point of course. I got a clean build. Will report more shortly.

jsakkine:
Sure. Will do!

baconpaul:
Anyway I got a clean mac build AU and VST3 both and both worked, and I visually inspected the code, so I think this is good to merge. I'll do so now.

jsakkine:
OK, thank you!

PR #142, opened 2018-12-23
No more need for demo mode

whydoubt:
Having a separate demo version of the plugin seems of negligible value at this stage.


jsakkine:
+1 same as with PPC

Issue #141, opened 2018-12-23
Step Sequencer Display on MacOS


baconpaul:
In COscillatorDisplay and CLFOGui I had to change the colmap to make sure it used the correct raw mac RGBA ordering. Look at the #if mac sections or at PR #140 after @esaruoho asked a question in #139 

Grepping before bed for if anyone else has a coltable with a similar bug, I see CStepSequencer does.

Here's how the manual says it should look

<img width="570" alt="screen shot 2018-12-22 at 9 18 30 pm" src="https://user-images.githubusercontent.com/13387561/50380233-24138780-062f-11e9-8593-a2ba9beee63b.png">

and here's how it does look

<img width="435" alt="screen shot 2018-12-22 at 9 19 16 pm" src="https://user-images.githubusercontent.com/13387561/50380235-3db4cf00-062f-11e9-9eee-b7557a0ef26d.png">

so we just need to scrub that code looking for everywhere an RGBA color is made into an int and make sure the mac has the right order, starting with coltable but probably looking elsewhere in the step seq .h and .cpp file.

Issue #91, opened 2018-12-16
Vector Drawing in the UI


baconpaul:
Now that we have issue #81 implementing a zoom (AU Mac only right now, I'm sorry; If you want to work on making it work VST or Win/Lin tag me on an issue and I'll give you guidelines if it's not clear from that issue), I am thinking about replacing elements with vector elements. Putting this issue in place to be the placeholder for my plan in case anyone has feedback.

In issue #23 @cyanit has shared SVG version of the bitmap (actually PNG) assets. I'm wondering if I can use them.

I'm going to start by seeing if I can just replace the horizontal background on a fader. But there's a lot of work to make that happen since the CBlahDeBlah world of vstgui doesn't have an SVG renderer.

But it does have canvas components with a ::draw( CDrawContext * ) method; and as we know from the screenshots in #81 if you call the draw methods directly they scale with vector style, as you would expect.

So here's my rough plan

1. Implement a super-tiny subset of SVG (just paths and groups probably) using tinyxml and make a common/gui/CSVG.cpp and h which kinda like Bitmap holds a drawable asset

2. For one of the bitmaps in CSurgeSlider - maybe even just a handle - replace the bitmap with my svg thing, have the svg thing also be in the resources bundle, and hack away until it works.

3. commit || give up

4. If (! give up ) figure out a plan to distribute the vectorizing work. Gulp!

The weird thing I notice is that the bitmap assets contain multiple views. I haven't quite figured out how you pan around the views in the UI yet. For instance bmp00154.png (which is IDB_FADERH_BG) has all of these in it. So also want to dig around on how pTray moves around the bmp before it paints.

![bmp00154](https://user-images.githubusercontent.com/13387561/50058263-b3a6cb00-0143-11e9-825d-469472c59c8e.png)

Welcome any comments. I'm going to do this off on a branch so as to not mess up au-cocoaui-build and so on while @esaruoho and @kzantow try to get the big pull request together.

baconpaul:
Oh the answer to how you move around is pretty simple. Look around like 158 of CSurgeSlider.cpp and you see that the one you choose to draw is controlled by typex and typey which move you around the image space before drawing. So that should be pretty easy to keep in place if the SVG draw image which I am imagining does the same thing. (Big if)

kurasu:
This kind of assets could be done with draw commands rather than SVGs. That way we also have control over the pixel grid so we could do non-integer scaling without going to blurry-town. But not all of them are this easy.. 

[...]

PR #145, opened 2018-12-23
Fix the paint bit order issues for the step sequencer

baconpaul:
Hey if someone could just eyeball this for me when you have a moment (@jsakkine @kzantow @esaruoho @asimilon maybe?) it fixes the step sequencer UI on mac, as described in #141. Wouldn't mind a quick once over before I push it just to make sure I didn't screw up the constants, even though I'm pretty confident.

Thanks!

### DIFF
--- a/src/common/gui/CLFOGui.cpp
+++ b/src/common/gui/CLFOGui.cpp
@@ -68,7 +68,7 @@ void CLFOGui::draw(CDrawContext* dc)
        int loopRegionLo = PIX_COL( 0xffb6d9b4, 0xb4d9b6ff );
        int noLoopHi = PIX_COL( 0xffdfdfdf, 0xdfdfdfff );
        int noLoopLo = PIX_COL( 0xffcfcfcf, 0xcfcfcfff );
-       int grabMarker = PIX_COL( grabMarker, 0x007f08ff ); // Surely you can't mean this to be fully transparent?
+       int grabMarker = PIX_COL( 0x00087f00, 0x007f08ff ); // Surely you can't mean this to be fully transparent?
        // But leave non-mac unch
        
       for (int i = 0; i < n_stepseqsteps; i++)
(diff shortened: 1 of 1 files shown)


jsakkine:
I would prefer 

```
#ifdef MAC
// declarations
#else
// declarations
#endif 
```

About the same amount of lines. No reason to be too smart here. Maybe in the long run this could be handled with some kind of theming...

baconpaul:
Yeah I almost did that - the other places I fixed this bug I did - but the reason I didn't was to avoid adding or changing a color in one platform and not the other. I figured the higher risk was "we change mac and not windows" not that "we have a bit of splattage". So no way we can add a variable in "only one branch". Also then I can visually check those strings are the same just transposed line by line.

If you feel really strongly (I don't feel really strongly - it was a 55/45 type thing with me) I can change it.

I agree in the long run this needs to be themed. That's "a lot of work" but not as much as porting to juce :)

jsakkine:
Not really a biggie for me. Either way will work. We cannot have anything solid for this right now. I just said what I had in mind because you asked for it :)

baconpaul:
Sanity checking is why we do reviews! Appreciate it. If it falls in the camp of not a biggie and looks ok otherwise I’ll merge it for now and we will revisit when we think skins and there’s (see #91 for some on that)

Thanks for the eyeball!’

asimilon:
> `#ifdef MAC`

should use `#if MAC`, as is the convention throughout rest of code, also I'm not sure about the VS project, but in the XCode project the preprocessor macros are: 

image

so, if the same applies to the VS project, `#ifdef` is going to evaluate true always.

baconpaul:
Oh thank you! I will patch this.

asimilon:
I had a look at `premake5.lua` and it looks to me like `MAC=1` should only actually be in the XCode project, so it's most likely not an issue.

baconpaul:
Yeah maybe. But I patched it anyway for consistency.

And now I'm really off for the day :)

Thanks!

Issue #146, opened 2018-12-23
LFO FreeRun doesn't in AU Host


baconpaul:
So fixing the step sequencer I put it in freerun mode so that it would, well, freerun and sync. But it still key triggered.

Looking at the code it seems nothing ever sets the run mode to `lm_freerun`

```
paul:~/dev/music/surge$ grep -r keytrigger src
src/common/SurgeStorage.h:   lm_keytrigger,
src/common/gui/CLFOGui.cpp:         tp[lfodata->trigmode.param_id_in_scene].i = lm_keytrigger;
src/common/dsp/LfoModulationSource.cpp:      case lm_keytrigger:
paul:~/dev/music/surge$ grep -r freerun src
src/common/SurgeStorage.h:   lm_freerun = 0,
src/common/dsp/LfoModulationSource.cpp:      case lm_freerun:
src/common/dsp/LfoModulationSource.cpp:   /*if (lfo->trigmode.val.i == lm_freerun)
paul:~/dev/music/surge$ 
```

so another bug to fix I think. It may be just that the GUI doesn't set the flag properly and then it all works (so it is in CLFOGui.cpp) but that commented out bit in the modulation source may also be part of it.

inigokennedy:
Not sure if it is directly related, FreeRun does work (at least in some cases) but it also resets rather than genuinely free runs so if you are running in short loops it constantly restarts. From the code it looks like it is tied to _songpos_ which would explain that - not desirable for genuine free run in my opinion (and one of my pet peeves over the years) but not sure if it can be counted as a bug.

baconpaul:
Right. I will dig more but the thing that made me say this was I put a step sequencer in free run mode and bound it to pitch. I press c and it beeps and boops. When I hold down an e in a bit with the c depressed I expected them to be synced - if the sequencer was free running the keypress on the new voice should change the lfo start position - but that was not what happened. Rather it retriggered for the new note

So I put in the bug. When I have a bit of time I’ll read the code and see what it is doing.

Curious what sort of configuration shows the behavior you see? Or do you have a patch file?

Thank you!

inigokennedy:
[...]

Issue #94, opened 2018-12-17
A way to log messages and errors


baconpaul:
So the only way I managed to get the whole thing working was by writing a dumb little class in the audio unit layer called "AULOG" which has one method (::log( char* format, ... )) which fprintfs to stderr and to a log file in the right place on macOS.

My dev cycle looks like "include this where I'm debugging, insert printfs, remove it from where I'm debugging"

So this is fine and all, but wondering what devs on other platforms like linux and windows think about having this be something more central in Surge land. So have src/common/SurgeLog.h and src/common/SurgeLog.cpp which (basically) look like my AULOG for now but is available everywhere and has the right OS-specific things for stuff like where to cache log files and so on. We could also use some external library or do something like proper log date and time formatting in the future if we had such a hook.

I am fine with not doing this. I can just whack away on mac with my little hack in AU only. But since I know some of you (like @jsakkine) are looking at linux and standalone also, I thought having 'one way to do it' would be nice, and also reduces the risk that I accidentally leave an AULOG in non-mac code and what not.

asimilon:
It would be good to wrapper in a macro that is wiped out of existence for Release builds.

baconpaul:
Agreed. My aulog has that mostly (the function becomes a noop which any optimized will optimize away if you set the logger off with an ifdef)

Just we need to pick a name and place and put it there. 

Gotta say I would not have been able to make it work without a reliable printf to log handler though - so I think it is important.

And also, I applaud your optimism in contemplating a Mac or Linux release build! :) I think this will be beta for a while.

jsakkine:
I don't whether it is good or not but many projects favor http://www.pantheios.org/. Worth of looking into anyway just because of the popularity. Here's something about its performance: http://www.pantheios.org/performance.html

asimilon:
Things that 100% need to be avoided on the audio thread:
 - allocations
 - disk i/o

else risk dropouts.  That stuff needs to be done on a separate thread with communication between the two.

baconpaul:
I completely agree with that!

Like I said, my AULOG is a gross hack I used so I wasn't blind (and now I'm in vst I have a not-for-commit similar gross hack). It's really tough and slow to attach a debugger to logic. Much easier to rebuild with a printf, alas. (I have an entire side commentary on the history of debuggers).

Pantheios looks very interesting. Thanks for the link.

jsakkine:
[...]

PR #148, opened 2018-12-26
Integrate spdlog

baconpaul:
As mentioned in #94 I banged together a small logging class which helped me a lot with getting AU running. After chatting with @jsakkine and @asimilon about how to get a useful logger, we thought about integrating someone else's written fast C++ logger. 

I chose spdlog https://github.com/gabime/spdlog

With this PR, spdlog works and you can use it anywhere if you `#include "SurgeLogger.h"` and then use their native `spdlog::info()` or what not.

And I gotta say, spdlog works great. I accidentally put a log into the render loop on every process block while debugging #146 and while my log file got big, my sound didn't waver.

@jsakkine @asimilon @kzantow and @kurasu interested in review or opinion from any of you. Especially note I added a new git submodule. I hope you like it, because wow, this is super handy for me squashing bugs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -52,6 +52,7 @@ products/
 installer_osx/installer
 installer_osx/Install_Surge_*.dmg
 build_logs/
+.DS_Store
 
 # IntelliJ IDEA
 .idea
--- a/.gitignore
+++ b/.gitignore
@@ -52,7 +52,6 @@ products/
 installer_osx/installer
 installer_osx/Install_Surge_*.dmg
 build_logs/
-.DS_Store
 
 # IntelliJ IDEA
 .idea
--- a/build-osx.sh
+++ b/build-osx.sh
@@ -61,7 +61,7 @@ NC=`tput init`
 
 prerequisite_check()
 {
-    if [ ! -f vst3sdk/LICENSE.txt ]; then
+    if [ ! -f vst3sdk/LICENSE.txt ] || [ ! -f spdlog/LICENSE ]; then
         echo
         echo ${RED}ERROR: You have not gotten the submodules required to build Surge. Run the following command to get them.${NC}
         echo
@@ -248,8 +248,11 @@ case $command in
     --uninstall-surge)
         run_uninstall_surge
         ;;
-    *)
+    "")
         default_action
+        ;;
+    *)
+        help_message
         ;;
 esac
 
--- a/build-osx.sh
+++ b/build-osx.sh
@@ -248,11 +248,8 @@ case $command in
     --uninstall-surge)
         run_uninstall_surge
         ;;
-    "")
-        default_action
-        ;;
     *)
-        help_message
+        default_action
         ;;
 esac
 
--- a/src/au/aulayer.cpp
+++ b/src/au/aulayer.cpp
@@ -1,4 +1,5 @@
 
+#include "SurgeLogger.h"
 #include "aulayer.h"
 #include <gui/SurgeGUIEditor.h>
 #include <AudioToolbox/AudioUnitUtilities.h>
@@ -6,10 +7,6 @@
 #include "aulayer_cocoaui.h"
 
 typedef SurgeSynthesizer sub3_synth;
-
-#ifdef GENERATE_AU_LOG
-FILE* AULOG::lf = NULL;
-#endif
 
 //----------------------------------------------------------------------------------------------------
 
@@ -112,14 +109,11 @@ void aulayer::InitializePlugin()
 {
 	if(!plugin_instance) 
 	{
-          AULOG::log( "SURGE:>> Constructing new plugin\n" );
-          AULOG::log( "     :>> BUILD %s on %s\n", __TIME__, __DATE__ );
-          //sub3_synth* synth = (sub3_synth*)_aligned_malloc(sizeof(sub3_synth),16);
-          //new(synth) sub3_synth(this);
+          spdlog::info( "Constructing new Surge AU Instance" );
+          spdlog::info( "     built {0} on {1}", __TIME__, __DATE__ );
           
           // FIXME: The VST uses a std::unique_ptr<> and we probably should here also
           plugin_instance = new SurgeSynthesizer( this );
-          AULOG::log( "     :>> Plugin Created\n" );
     }
 	assert(plugin_instance);
 }
--- a/src/common/SurgeLogger.cpp
+++ b/src/common/SurgeLogger.cpp
@@ -1,0 +1,76 @@
+#include "SurgeLogger.h"
+#include "spdlog/sinks/basic_file_sink.h"
+#include "spdlog/sinks/stdout_sinks.h"
+
+#include "spdlog/logger.h"
+#include "spdlog/async.h"
+#include <memory>
+#include <filesystem.h>
+
+/*
+** This file exists to have a single global static which is initialized at library open time
+** thereby making sure that the log is set up as we want when it runs. 
+**
+** Two parts. The "init" function below sets up the logger in the OS reasonable way.
+** For now it is stderr+~/Library/Logs on macos and stderr only on windows and linux.
+**
+** The second part is the static variable which is initialized with the function call
+*/
+
+namespace SurgeLogInternal
+{
+
+typedef std::shared_ptr< spdlog::logger > log_t;
+namespace fs = std::experimental::filesystem;
+
+log_t internalSurgeLogInit()
+{
+    std::vector<spdlog::sink_ptr> sinks;
+    
+    /*
+    ** Where to put the logfile is actually OS dependent. 
+    ** The approach we take here is to add a pair of sinks
+    ** to the logger, the first to stderr
+    */
+    auto stderrLogger = std::make_shared<spdlog::sinks::stderr_sink_mt>();
+    sinks.push_back( stderrLogger );
+    
+    /*
+    ** And if someone has implemented this for their OS, the second to a standard
+    ** logfile location
+    */
+
+
+#if MAC 
+    char lfb[ PATH_MAX ];
+    sprintf(lfb, "%s/Library/Logs/Surge.log", getenv("HOME"));
+    auto fileSink = make_shared<spdlog::sinks::basic_file_sink_mt> (lfb);
+    sinks.push_back( fileSink );
+#endif
+    
+#if WINDOWS
+    // FIXME: Where's the default log location on windows?
+#endif
+    
+#if __linux__
+    char lfb[ PATH_MAX ];
+    sprintf(lfb, "%s/.cache/Surge", getenv("HOME"));
+    fs::create_directories(fs::path(lfb));
+    
+    sprintf(lfb, "%s/.cache/Surge/Surge.log", getenv("HOME"));
+    auto fileSink = make_shared<spdlog::sinks::basic_file_sink_mt> (lfb);
+    
+    sinks.push_back( fileSink );
+#endif
+    
+    
+    auto logger = std::make_shared<spdlog::logger>("surge", sinks.begin(), sinks.end());
+    spdlog::register_logger(logger);
+    spdlog::set_default_logger(logger);
+    
+    return logger;
+}
+    
+static log_t internal_surge_log_instance = internalSurgeLogInit();
+
+}
--- a/src/common/SurgeLogger.h
+++ b/src/common/SurgeLogger.h
@@ -1,0 +1,10 @@
+/*
+ * SurgeLogger - quick wrapper to hold reference to and bootstrap the spd logger
+ */
+
+#ifndef __SURGELOGGER_INCLUDE__
+#define __SURGELOGGER_INCLUDE__
+
+#include "spdlog/spdlog.h"
+
+#endif // __SURGELOGGER__INCLUDE__
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,6 @@
 [submodule "vst3sdk"]
 	path = vst3sdk
 	url = https://github.com/steinbergmedia/vst3sdk.git
+[submodule "spdlog"]
+	path = spdlog
+	url = https://github.com/gabime/spdlog.git
--- a/premake5.lua
+++ b/premake5.lua
@@ -132,6 +132,7 @@ includedirs {
 	"src/common/thread",
 	"vst3sdk/vstgui4",
 	"vst3sdk",
+	"spdlog/include",
 	"libs/"
 	}
 
--- a/src/au/aulayer.h
+++ b/src/au/aulayer.h
@@ -12,38 +12,6 @@ class SurgeGUIEditor;
 class SurgeSynthesizer;
 typedef SurgeSynthesizer plugin;
 
-#define GENERATE_AU_LOG
-struct AULOG
-{
-#ifdef GENERATE_AU_LOG
-    static FILE* lf;
-    
-    static void log( const char* format, ... )
-    {
-        if( lf == NULL )
-        {
-	    char fname[ 1024 ];
-            sprintf( fname, "%s/Library/Logs/Surge.log", getenv( "HOME" ) );
-            lf = fopen( fname, "a" );
-        }
-        va_list args;
-        va_start( args, format );
-        vfprintf( stderr, format, args );
-        va_end( args );
-        if( lf != NULL )
-        {
-            va_start( args, format );
-            vfprintf( lf, format, args );
-            va_end( args );
-            fflush( lf );
-        }
-        
-    }
-#else
-    static void log( const char* format, ... ) { }
-#endif
-};
-
 //-------------------------------------------------------------------------------------------------------
 const CFStringRef rawchunkname = CFSTR("VmbA_chunk");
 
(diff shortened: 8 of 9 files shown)


src/common/SurgeLogger.cpp
```diff
@@ -0,0 +1,57 @@
...
+  sinks.push_back( filesink );
+#endif
+
+#if WINDOWS
+  // FIXME: Where's the default log location on macos?
+#endif
+
+#if __linux__
+  // FIXME: Where's the default log location on linux?
+#endif
```
jsakkine: ~/.cache/Surge.log

jsakkine: Perhaps, own subdirectory would not be a bad idea, though.

jsakkine: ~/.cache/Surge/Surge.log

src/common/SurgeLogger.cpp
```diff
@@ -0,0 +1,57 @@
...
+#include "spdlog/async.h"
+#include <memory>
+
+
+typedef std::shared_ptr< spdlog::logger > log_t;
+
+log_t internal_surge_log_init()
+{
+  std::vector<spdlog::sink_ptr> sinks;
+
```
jsakkine: Not really a review comment but should we just move to "more standard" 4 character tabs for new code :D

src/au/aulayer.cpp
```diff
@@ -1,4 +1,7 @@
 
+#include "SurgeLogger.h"
+
+
```
jsakkine: Why two extra newlines here?

jsakkine: Do we have any convention how to order includes?

src/common/SurgeLogger.cpp
jsakkine: To get more coheren codebase I would use CamelCase for all type (e.g. `Log`) and camelCase for or variables and function (e.g. `internalSurgeLogInit()`). Now this contributes to the porridge of different styles and conventions. Lets get moro rigid.

jsakkine: Not really a review comment but maybe we could have

```
namespace Log {
    Log init()
    {
    }
}
```

We could possibly have few namespaces e.g. Log, GUI, Synth (non-GUI engine). I don't really get why you wan't to prefix with "Surge" and for me anything that does not have static keyword is not internal. It is exported symbol.

jsakkine: Being programmed with only C for past 8-9 C++ is a mystery box but didn't recent standard versions have some kind lambda type of thing, or am I wrong? Then you could initialize the log by using such?

src/common/SurgeLogger.cpp
```diff
@@ -0,0 +1,57 @@
...
+{
+  std::vector<spdlog::sink_ptr> sinks;
+  
+  /*
+  ** Where to put the logfile is actually OS dependent. 
+  ** The approach we take here is to add a pair of sinks
+  ** to the logger, the first to stderr
+  */
+  auto stderr_logger = std::make_shared<spdlog::sinks::stderr_sink_mt>();
+  sinks.push_back( stderr_logger );
```
jsakkine: Is that a convention that is used widely in Surge i.e. extra space after `(` and before `)`? Also, I'm puzzled about variable naming i.e. you use underscores to separate words.

src/common/SurgeLogger.cpp
jsakkine: Should `filesink` be `file_sink` or perhaps `fileSink`?

src/common/SurgeLogger.cpp
```diff
@@ -0,0 +1,57 @@
...
+
+#if __linux__
+  // FIXME: Where's the default log location on linux?
+#endif
+
+
+  /* auto logger = std::make_shared<spdlog::async_logger>("surgelogger", sinks.begin(), sinks.end(),
+                                                       spdlog::thread_pool(),
+                                                       spdlog::async_overflow_policy::block); */
+
```
jsakkine: I don't like cruft code to be added to repository. These will just give only WTF moments some time from now.

src/common/SurgeLogger.h
```diff
@@ -0,0 +1,7 @@
+/*
+ * SurgeLogger - quick wrapper to hold reference to and bootstrap the spd logger
+ */
+
+#pragma once
+
```
jsakkine: `#pragma once` is a non-standard convention that should not be used. Rather have just plain old ifdef cage.

build-osx.sh
```diff
@@ -61,7 +61,7 @@ NC=`tput init`
 
 prerequisite_check()
 {
-    if [ ! -f vst3sdk/LICENSE.txt ]; then
+    if [ ! -f vst3sdk/LICENSE.txt ] || [ ! -f spdlog/LICENSE ]; then
         echo
```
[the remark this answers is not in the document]
baconpaul: This one is here because we had more than one “why doesn’t it build” question from people who didn’t run git submodules so a polite check with a clearer error seemed friendly.

src/common/SurgeLogger.cpp
baconpaul: Yeah c++ 11 lambdas are cool - I used them in zoom - but they aren’t the right construxtnfor this sort of initializarion. 

Like the namespace idea. Will do that. And will clean up all the casing and stuff. A lot of this got copied from the spdlog manual which has yet another coding style.

The mush mash of styles and tools is driving me crazy too. We should adopt a standard!

baconpaul: Oh the only one of your comments I'm not integrating is the type rename. stl has a pretty clear convention that types are named like `thething_t` so I think `typedef thisWhatzIt<foobaz> mything_t` is very idiomatic. Just an FYI for when you read the next version!

src/au/aulayer.cpp
baconpaul: The newlines are sloppy and I will remove them but the whole allayed.cpp code is a mix of carbon and cocoa and ancient and new and this tab and that tab, so I will do a separate PR to generally clean it up.

src/common/SurgeLogger.h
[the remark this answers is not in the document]
jsakkine: It's a convention used in Linux, and some other open source projects have inherited it: https://www.kernel.org/doc/html/v4.17/process/submitting-patches.html (sections 11, 12 and 13). Not something that I proactively push to this project. Just want to know what kind of conventions we are going to use.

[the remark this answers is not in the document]
baconpaul: Oh gotcha. I’ve never worked on the kernel. (Btw: thanks for working on the kernel!!) But many projects I’ve worked on which have succeeded have well defined rules like that. We should probably adopt them also and make a pull request template.

As you have seen, I’ve squashed all my commits. For this project, that’s a fine thing to do. And the “-i” option to rebase makes it easy enough. In doing so, I think I improved the commit log buy making sure the commits link to github issue and have a clear English explanation. Perfectly reasonable thing to expect.

Do you see any other problems with the PR in the current state? Since it adds a new dep, I’ll still wait for one of @asimilon  @kurasu @kzantow or @whydoubt to take a look I think, just in case they have orthogonal suggestions!

Oh and maybe lets open a fresh issue for “developer documentation” which includes a new doc altogether so we can include things like how to PR, how the branches work, how to set up emacs properly (sorry about those 2 space tabs! Do’h) and so on.  Like I’ve said several times, I’m happy to work any way we want. I think we probably need less heavyweight tools than the Linux kernel but a standard would be lovely.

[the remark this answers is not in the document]
jsakkine: Sounds like a plan :)

jsakkine:
Looking into your update. What was your take on indentation BTW?

baconpaul:
> Looking into your update. What was your take on indentation BTW?

I went to 4 spaces no tabs and set emacs to use Linux-style rather than gnu-style brace positioning which roughly matches what Xcode does by default too. I changed `if( you_need_space )` to if(you_need_space)` but muscle memory will have me getting that one wrong forever I fear. 

And I chose to not indent the first-level namespace. So

```
Namespace foo
{

Class don’t indent
{
}

Namespace bar
{ 
    Class in_foo_bar gets indented one
}.
```

Which was in the standard you circulated earlier.

src/common/SurgeLogger.cpp
```diff
@@ -0,0 +1,76 @@
...
+*/
+
+namespace SurgeLogInternal
+{
+
+typedef std::shared_ptr< spdlog::logger > log_t;
+namespace fs = std::experimental::filesystem;
+
+log_t internalSurgeLogInit()
+{
```
jsakkine: Still puzzling around this:

- Could it be just `Log`, the typedef I mean?
- I guess you could use `static` here.

baconpaul: I could use static on the function declaration but would it make a difference? It’s a symbol which appears in no headers. I forget what static does on a standalone function in modern C++ anymore.

I made a comment on your review about Log vs log_t. STL standard for typedefs is really `typedef foo bar_t` and so I followed that.  I’d really rather keep it that way. Or I could also just kill the typedef. Originally I had a class which people used so SurgeLogger::log_t is the idiomatic name and when I stopped exporting that class in my dev I kept the standard-ish name

baconpaul: On the static thing: internalSurgeLogInit is not a member function of any class. So it is behaves just like a static member function of a single function class.

src/common/SurgeLogger.h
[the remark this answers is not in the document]
baconpaul: Great!

src/common/SurgeLogger.h
```diff
@@ -0,0 +1,10 @@
+/*
+ * SurgeLogger - quick wrapper to hold reference to and bootstrap the spd logger
+ */
+
+#ifndef __SURGELOGGER_INCLUDE__
```
whydoubt: AFAICT, ifndef guards are used about six times in surge.  'pragma once' is used fifty-four.
I have nothing against ifndef guards, but I usually try to follow a project's conventions.

src/au/aulayer.cpp
```diff
@@ -110,18 +110,16 @@ CFURLRef aulayer::GetIconLocation ()
...
-	{
-          AULOG::log( "SURGE:>> Constructing new plugin\n" );
-          AULOG::log( "     :>> BUILD %s on %s\n", __TIME__, __DATE__ );
-          //sub3_synth* synth = (sub3_synth*)_aligned_malloc(sizeof(sub3_synth),16);
-          //new(synth) sub3_synth(this);
-          
-          // FIXME: The VST uses a std::unique_ptr<> and we probably should here also
-          plugin_instance = new SurgeSynthesizer( this );
-          AULOG::log( "     :>> Plugin Created\n" );
+  if(!plugin_instance)
```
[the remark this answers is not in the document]
baconpaul: Sure! Just did that and pushed. Thanks! As long as @jsakkine doesn't now go and suggest I take it to 4. Chuckle.

aulayer.cpp and .h (the files) are a mess. I'll put in another issue to retab and remove cruft and so on.

src/common/SurgeLogger.h
baconpaul: Ha! OK so I had #pragma once for that reason, and @jsakkine suggested I replace it to be more standardized, so I did. Then I squashed my commits so you can't see that back and forth.

I think we need a clear style guide we can point to. Does it have pragma once or ifdef guards? I am completely neutral and will do either! But unfortunately we can't have both.

.gitignore
```diff
@@ -52,6 +52,7 @@ products/
 installer_osx/installer
 installer_osx/Install_Surge_*.dmg
 build_logs/
+.DS_Store
```
whydoubt: I don't think this relates to spdlog

build-osx.sh
```diff
@@ -248,8 +248,11 @@ case $command in
     --uninstall-surge)
         run_uninstall_surge
         ;;
-    *)
+    "")
```
whydoubt: Is this actually spdlog-related. I honestly can't tell.

baconpaul: Good eyes, and very sorry it is not; that got swept into this as I fixed that bug (which was basically you could use build-osx --anything and it worked) while developing spdlog.

If you want, I can pull it out and put it in a separate PR.

.gitignore
baconpaul: That is correct; it relates to clean up mac build, which should be a separate PR as I commented below.

build-osx.sh
baconpaul: OK I removed this and the .gitignore from this PR. I will squash all the commits when done with any other comments and before I merge. I'll do a separate PR for the .DS_Store and the "" vs * in the build.

baconpaul: OK I pushed these two diffs into PR #153 rather than this one. Thanks!

src/common/SurgeLogger.cpp
jsakkine: OK, right, sorry I missed that one but so you did. I like clean binaries even if no other function (e.g. nm output and such).

src/common/SurgeLogger.h
jsakkine: If no other reason, at least ifndef cage is part of the standard whereas `#pragma once` is not. We are now IMHO defining the conventions (like the use of 4 space tabs, which is a relief :))

src/au/aulayer.cpp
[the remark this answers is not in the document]
jsakkine: @whydoubt, Well, if we want to define any kind of conventions, it is an unnecessary evil to do this. It is better to do it now when the codebase is not yet otherwise mature.

src/common/SurgeLogger.h
baconpaul: Yeah I agree we need standards and perhaps another issue on them and how to get to them. Are we going to do a one day “super-retab”? Are we going to have mixed conventions in PRs? Do we have a tool which auto formats for us? Etc... All those things are great to discuss.

But I’d sooner not use the review of spdlog to resolve them :)

So can we decide whether, given the state of the code, these diffs are pushable? I’m fine either way of course! That’s why I asked for a review. But realizing that given the code is a bit here and there on tabs and spaces and names and ifdefs so we will incrementally improve, is this an incremental improvement? Especially now I’ve stripped out the stuff accidentally swept along as @whydoubt found and made the code way more consistent thanks to @jsakkine‘s eagle eyes!

Thanks!

src/au/aulayer.cpp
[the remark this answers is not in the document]
baconpaul: Cool. @whydoubt is better that we should keep the diffs minimal even if it means the tabs are gross. I actually made a tiny change to the pan2 PR to meet that goal since it had the same problem. Thanks folks!

src/common/SurgeLogger.h
jsakkine: OK, lets do it that way! Lets have the fight about of space in an issue and then one atomic update to the code base :D

src/common/SurgeLogger.cpp
```diff
@@ -0,0 +1,76 @@
+#include "SurgeLogger.h"
+#include "spdlog/sinks/basic_file_sink.h"
+#include "spdlog/sinks/stdout_sinks.h"
+
+#include "spdlog/logger.h"
+#include "spdlog/async.h"
+#include <memory>
+#include <filesystem.h>
```
kzantow: should this be `"filesystem.h"` ?

src/common/SurgeLogger.cpp
```diff
@@ -0,0 +1,76 @@
...
+    /*
+    ** And if someone has implemented this for their OS, the second to a standard
+    ** logfile location
+    */
+
+
+#if MAC 
+    char lfb[ PATH_MAX ];
+    sprintf(lfb, "%s/Library/Logs/Surge.log", getenv("HOME"));
+    auto fileSink = make_shared<spdlog::sinks::basic_file_sink_mt> (lfb);
```
kzantow: Should this be opt-in instead of always on? Is the log always appended to - is there some auto-truncation size? What happens when multiple processes might be accessing the same log? Seems to me logging to the console by default is much preferred and opting in to dump to a log if necessary is a nicer default behavior. Just my 2c.

baconpaul: If I had any idea at all what happened to stderr or stdout ejected from a plugin in logic, au lab, and hosting au, I would be fine to skip this. Do you know?

That said: SPD has an auto rotating log. I could use that. Or I could comment this section out and just leave it on in my development area inside an #ifdef log-to-file or some such.

kzantow: @baconpaul can you run Logic from the command line? If I run Reaper, Live, and Bitwig from command line, I get all the `stdout` dumped to the terminal; don't have Logic to see if it's the same, but I assume so?

kzantow: (an aside: I usually set up REAPER as the app to run in XCode with a project that just loads the plugin and opens it by default, in this case, the terminal output is just dumped to the XCode console)

baconpaul:
Huh. Yeah of course that works. I am simply amazed I didn't think of that. Like really amazed.

Well that's mortally embarrassing. None of this code is needed. I'm going to close this PR and delete the code. I can just use printf when I'm debugging.

Sorry for the noise.

kzantow:
@baconpaul now we can argue about tabs vs. spaces and such vastly important programming topics ;)

baconpaul:
@kzantow. Sure! spaces.
Oh, and emacs.

Issue #119, opened 2018-12-20
"rename" on mod controllers doesn't work in Mac AU


baconpaul:
So I'm close to getting MPE all squared away, and in doing so added an explicit menu on the mod controllers to set a mod source to, say, midi CC 13. You can see all that in my mpg-routing branch which is underway.

Along the way I checked the other menu items and it turns out that "rename" doesn't work.

```
else if (command == id_rename)
            {
                AULOG::log( "rename\n" );
               spawn_miniedit_text(synth->storage.getPatch().CustomControllerLabel[ccid], 16);
               ((CModulationSourceButton*)control)
                   ->setlabel(synth->storage.getPatch().CustomControllerLabel[ccid]);
               //((gui_slider*)metaparam[ccid])->setLabel(synth->storage.getPatch().CustomControllerLabel[ccid]);
               synth->updateDisplay();
            }
```

seems `spawn_miniedit_text` in the AU at least ... well ... doesn't spawn! So let me log this bug here so we remember it. I'm not going to fix it now.

baconpaul:
Huh. This code in PopupEditorDialog.cpp certainly explains why!

```
#elif MAC_COCOA

void spawn_miniedit_text(char* c, int maxchars)
{
   // TODO
}
```

Ha. Well I guess at least we know now what the problem is.

baconpaul:
I have this fixed in baconpaul/surge on branch rename-popup-119. It's behind #148 so I'll hold off on the PR until I get feedback on that one. But no need for anyone to work on this.

baconpaul:
Fixed at head. Closing.

PR #150, opened 2018-12-26
Implement miniedit on Cocoa (and thus modulator rename #119)

baconpaul:
A simple NSAlert based implementation of the miniedit; hook into
the PopupDialog for cocoa; and correct invalidation when a change
occurs. Means rename works fixing #119 

I'll merge this tomorrow unless there are comments. Thanks!

### DIFF
--- a/src/common/gui/SurgeGUIEditor.cpp
+++ b/src/common/gui/SurgeGUIEditor.cpp
@@ -1566,6 +1566,8 @@ int32_t SurgeGUIEditor::controlModifierClicked(CControl* control, CButtonState b
                spawn_miniedit_text(synth->storage.getPatch().CustomControllerLabel[ccid], 16);
                ((CModulationSourceButton*)control)
                    ->setlabel(synth->storage.getPatch().CustomControllerLabel[ccid]);
+                control->setDirty();
+                control->invalid();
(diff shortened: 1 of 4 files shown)


src/common/gui/SurgeGUIEditor.cpp
```diff
@@ -1566,6 +1566,8 @@ int32_t SurgeGUIEditor::controlModifierClicked(CControl* control, CButtonState b
                spawn_miniedit_text(synth->storage.getPatch().CustomControllerLabel[ccid], 16);
                ((CModulationSourceButton*)control)
                    ->setlabel(synth->storage.getPatch().CustomControllerLabel[ccid]);
+                control->setDirty();
+                control->invalid();
                //((gui_slider*)metaparam[ccid])->setLabel(synth->storage.getPatch().CustomControllerLabel[ccid]);
```
jsakkine: Can we remove the commented out code or is desperately needed?

jsakkine: Not really any type of review comment for this. Just wondering.

baconpaul: So since the code is kinda historied and we are learning about it I’ve decided to not remove commented code in case it holds some nugget of past wisdom we need in the future. Thanks for the reviews on these issues!

jsakkine: Makes sense, agreed.

PR #151, opened 2018-12-26
Remove useless __forceinline and forceinline macros

jsakkine:
These macros look fairly useless so just replace the uses with the
"regular" inline.

### DIFF
--- a/src/common/precompiled.h
+++ b/src/common/precompiled.h
@@ -4,5 +4,4 @@
 
 #include "globals.h"
 
-#define forceinline __forceinline
 //#include <vt_util/vt_string.h>
--- a/premake5.lua
+++ b/premake5.lua
@@ -35,7 +35,6 @@ if (os.istarget("macosx")) then
 	defines
 	{
 		"_MM_ALIGN16=__attribute__((aligned(16)))",
-		"__forceinline=inline",
 		"_aligned_malloc(x,a)=malloc(x)",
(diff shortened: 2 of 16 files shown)


jsakkine:
Came up while doing cmake build. Does make much sense to duplicate those definitions...

src/common/precompiled.h
```diff
@@ -4,5 +4,5 @@
 
 #include "globals.h"
 
-#define forceinline __forceinline
+#define inline inline
```
whydoubt: Isn't this is redundant?
The line should probably just be removed.

jsakkine: It is. I made this commit by

```
$ find -regex ".*\.cpp\|.*\.h"|xargs sed -i 's/__forceinline/inline/'
$ find -regex ".*\.cpp\|.*\.h"|xargs sed -i 's/forceinline/inline/'
```
Thanks for catching that one up. Will update.

jsakkine:
Updated.

src/common/precompiled.h
baconpaul: Might be reading diffs wrong, but it looks like you got this in vstquartzplug.h but not in precompiled.h.

baconpaul:
I confirmed your master builds AU and VST3 mac both with these changes, and @whydoubt and I both reviewed, so I am going to merge them. I'll rebase my logging branch after that.

PR #153, opened 2018-12-26
Small fixes for mac build

baconpaul:
Two small fixes I accidentally swept into the spdlog PR. These
ignore .DS_Store which pops up eventually if you use finder,
and makes sure build-osx gets a valid input or no arg, but
doesn't treat an invalid flag as a no-arg


jsakkine:
Reviewed-by: Jarkko Sakkinen <jarkko.sakkinen@iki.fi>

PR #155, opened 2018-12-26
Use alignas() for alignment

jsakkine:
Define a generic ALIGN() macro albeit there is only 16-byte aligns. Just
makes the source code more cleaner and obvious. Use _MSC_VER to detect
MSVC vs rest of the world (probably could use that in most of the cases
as the differentiator as 90% of time the same shit works for clang and
GCC). As a side-effect, this will remove some cruft from premake5.lua.

### DIFF
--- a/src/vst2/vstplugsquartz.h
+++ b/src/vst2/vstplugsquartz.h
@@ -27,7 +27,6 @@
 #include <vstgui/vstgui.h>
 #include "globals.h"
 
-#define _MM_ALIGN16 __attribute__((aligned(16)))
 #define stricmp strcmp
 #define _aligned_malloc(x, y) malloc(x)
 #define _aligned_free(x) free(x)
--- a/premake5.lua
+++ b/premake5.lua
@@ -34,7 +34,6 @@ if (os.istarget("macosx")) then
 	
 	defines
 	{
-		"_MM_ALIGN16=__attribute__((aligned(16)))",
 		"_aligned_malloc(x,a)=malloc(x)",
 		"_aligned_free(x)=free(x)",
 		"stricmp=strcasecmp",
@@ -70,7 +69,6 @@ elseif (os.istarget("linux")) then
 	
 	defines
 	{ 
-		"_MM_ALIGN16=__attribute__((aligned(16)))",
 		"_aligned_malloc(x,a)=malloc(x)",
 		"_aligned_free(x)=free(x)",
 		"stricmp=strcasecmp",
--- a/src/common/AbstractSynthesizer.h
+++ b/src/common/AbstractSynthesizer.h
@@ -66,8 +66,8 @@ public:
    virtual unsigned int saveRaw(void** data) = 0;
 
 public:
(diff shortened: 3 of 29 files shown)


src/vst2/vstplugsquartz.h
```diff
@@ -27,7 +27,7 @@
 #include <vstgui/vstgui.h>
 #include "globals.h"
 
-#define _MM_ALIGN16 __attribute__((aligned(16)))
+#define ALIGN(16) __attribute__((aligned(16)))
 #define stricmp strcmp
```
jsakkine: UGH, will remove this one. Time to do a git grep again.

baconpaul:
I pulled jsakkine:master

```
paul:~/dev/music/surge$ git remote -v | grep jsakkine
jsakkine	https://github.com/jsakkine/surge (fetch)
jsakkine	https://github.com/jsakkine/surge (push)
paul:~/dev/music/surge$ git fetch jsakkine
paul:~/dev/music/surge$ git checkout jsakkine/master
```

Then did a `./build-osx.sh --clean-all` and a `./build-osx.sh` 

The `build-osx.sh` standalone built a vst3 but gave an error on the au.

Using upstream/master (which is the kurasu master) I get a clean build.

The problem seems to be at the ALIGN(16) in SurgeStorage.h at line 54. 

So I think this isn't mergeable right now. Do you see the same?

kurasu:
c++11 added the alignas keyword for this, that is what we should be using instead of defining something ourselves. Back when I wrote this that obviously didn't exist yet.

kurasu:
The reason why alignment is used so much in this project is that the DSP code is written using SSE intrinsics, and the insructions they generate will fail/crash if they're used with memory adresses which are not 16-byte aligned. 

This might not be immediately obvious to those who haven't used these intrinsics before so I though it would be good to mention it.

baconpaul:
> c++11 added the alignas keyword for this, that is what we should be using instead of defining something ourselves. Back when I wrote this that obviously didn't exist yet.

Thanks! @jsakkine perhaps you want to close this PR and instead use https://en.cppreference.com/w/cpp/language/alignas appropriately? Also I'll not merge this one, I will let @kurasu review the changes.

jsakkine:
@kurasu, @baconpaul: I'll update this. I suck at C++ ATM so please say if I'm reinventing the wheel. I'm familiar with SSE2 constraints though in a very different context (crypto). Most of x86 ISA extensions require so called natural aligment (128 bit XMM registers and same size data blobs from system memory i.e. 16 byte alignment).

jsakkine:
Before converting rest, is that the correct usage pattern for `alignas()` ?

jsakkine:
Just rebased to the latest mainline.

jsakkine:
Divided this to three more or less self-contained commit to make more digestable. Still WiP to test it with macOS (and at least compile with Windows).

jsakkine:
Awesome, thank you.

PR #163, opened 2018-12-27
Resolve my logging issue by just not having anything.

baconpaul:
I have no idea why the obvious @kzantow idea of run logic from the command
line to look at stdout didn't occur to me. But it didn't.  Now it has, through him, I don't
need any of this code or any discussion in #94. Apologies!

### DIFF
--- a/src/au/aulayer.cpp
+++ b/src/au/aulayer.cpp
@@ -7,10 +7,6 @@
 
 typedef SurgeSynthesizer sub3_synth;
 
-#ifdef GENERATE_AU_LOG
-FILE* AULOG::lf = NULL;
-#endif
-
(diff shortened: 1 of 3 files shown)


kzantow:
This is easily the best PR of the day: removing code is the best! MEGA-:+1:

baconpaul:
> Maybe it would be good to add a note to the README for the next person who might not think to run a DAW from the command line?

Yeah we should add that to "docs/DEVELOPER.md" which doesn't exist yet. @jsakkine keeps threatening to start it. When the doc exists I will add this tip. Trust me, I won't forget how dumb I feel now for not thinking of this for a while!

Issue #160, opened 2018-12-26
CHSwitch2 don't visually update when modified externally (was Osc1 Octave bug)


esaruoho:
Steps to replicate:
1. Move Osc1 Octave from -3 to +3
2. Add a controller for Osc1 Octave to DAW
3. Move slider in DAW to control Osc1 Octave

Expected results:
Osc1 Octave changes from -3 to +3 in real time

Current result:
Osc1 Octave does not change visually, but audio does change.

OTOH, mapping a Cutoff Filter to the same DAW will result in graphical update.

here is a gif
![surge_buggy updates](https://user-images.githubusercontent.com/4966687/50456033-491f3780-095a-11e9-9555-ad5321e86be4.gif)

@baconpaul you might find this intriguing

baconpaul:
Anyway able to reproduce this in logic pro with the AU. Will look.

baconpaul:
Just as a note to myself, since I now have to go back down my needless logging code (d'oh) the difference between a filter cutoff and the oscillator octave is that the filter cutoff has a parameter in param[ j ] around line 300 of SurgeGUIEditor and the octave does not. So we need some other "if" in there for discrete controls that I don't know yet.

baconpaul:
OK so this happens for any of the visual bitmap switches. Filter type, octave, etc... The problem is that since these aren't "modulatable parameters" they don't save themselves in SurgeGUIEditor params api. That doesn't mean they don't get their value updated. 

The code path when you wiggle the external knob in logic or whatever is

```
aulayer::SetParameter
SurgeSynthesizer::setParameter01 with external = true
update value 
add item to refresh_queue

// then on the idle thread
SurgeGUIEditor::idle
Traverse the refresh queue (SurgeGUIEditor around line 294)
If bound to param[ ] then also graphically invalidate the parameter
```

this explains why things like cutoff work (they are bound to a param) and why things like filter type and osc octave don't (they aren't).

So the question is: How do you get a reference to the graphical item around like 305 of the idle queue if you aren't in the param loop. The item is created around like 877 of SurgeGUIEditor and added to the frame and even bound by tag id. But I need to get back that reference to a control by the parameter id to invalidate it in this circumstance.

Note that some of the controls preserve the reference to the control. Check out filter subtype at line 673, which is preserved just for invalidation later on. So clearly we need to do the same thing here.

[...]

PR #179, opened 2018-12-27
Make switches and so on invalidate in GUI

baconpaul:
When you have external (like DAW) controlled items
which are not continuous sliders they didn't change.
See #160. To fix this we need to keep a reference to
them in the refresh_params update idle look. This
implements that and keeps a reference on a few key controls.

This is a partial fix to the bug. See my request in #160 for help finding the next set of controls which don't work.


jsakkine:
Looking into this today if I have time or latest tomorrow...

baconpaul:
Thanks! Lots of details in the issue on how I approached it and what the code flow is.

PR #180, opened 2018-12-27
Fix macOS installer, remove duplicated data in bundles, terminology

kzantow:
Removed bundling of resources in AU, VST, VST3 since installer has separate resources for it, fixed some terminology, aligned the macOS build naming with windows, always show customization step (I really don't ever want to install AU, personally).


esaruoho:
@kzantow query, why is there a .pkg there?

kzantow:
I built the installers with this branch with an updated cloud build script, they're available here: https://kzantow.github.io/surge/ to test

baconpaul:
Any reason surge-1.0.6b4.pkg is included in this PR? I think that’s noit right.

kzantow:
@baconpaul @esaruoho oops, yeah, i've been doing releases all day with this to iron out the bugs, overlooked that slipping in, but fixed now

VERSION
```diff
@@ -0,0 +1 @@
+1.6
```
baconpaul: What’s this file for?

kzantow: Experiment with releasing; I have the build scripts set up to auto-increment the latter part of the version with each release.

baconpaul: Got it. I worry I won’t be the only person to ask!

I wonder if we’d be better to have something like “scripts/BuildConfig.sh” which had

```
# a comment
VERSIONBASE=1.6
```

Type thing so we don’t end up with one file per feature and we can explain it a bit?

I guess it all depends on how you are reading VERSION but you get the idea!

kzantow: I don't much care. Just need some way to auto-increment a version number. Are there any standard ways of doing this in the c++ world? I'm spoiled by [Maven](maven.apache.org/maven-release/maven-release-plugin/)

baconpaul: Not standard ones no. Where do you use the file?

baconpaul: Ok cool so you do something like cat it with backticks?

Thinking we will have other info we want to pass to the auto build scripts (I’m thinking things like signature files and stuff in the future) would it be possible to have a file which just defines surge version base (using = assigned to a name) and then include that so we could set other things later and include a comment? I worry the single file called version with a single number 1.6 could be a little bit confusing to a future person figuring this out.

baconpaul:
Oh a thought @kzantow - once this is in can we also kill the recursive copy at startup in c++ to move the resources in the synth? That code won’t work anymore and won’t be needed. Separate from this issue but something I just thought of.

Other than the version comment above this diff looks good to me now although as I said I haven’t been able to test the installers.

kzantow:
@baconpaul yeah, I'll kill the recursive copy as part of this, it already removes the files from the bundles.

esaruoho:
what's happening with this PR, btw? seems like the conflicting License.txt file is fine (just use the newer one that is at master/head)..

esaruoho:
PR comment test

esaruoho:
one more pr comment test

kzantow:
@esaruoho yeah, I've been busy for a few days; just merged and rebuilt

kzantow:
@baconpaul @kurasu @jsakkine okay, I just ran another build and it all looks pretty good.

the only real question with this PR was the `VERSION` file, which I just removed as I realized it really has nothing to do with this at all. There are plenty of ways to deal with incrementing the version, but if we'd like to move the installer and some PR builds to this I'll have to make a few more tweaks to the build script anyway and maybe we can figure out what to do then.

baconpaul:
OK great; I pulled this and did a clean build and got an installer that seems to work. Thank you! I'll merge this now

Issue #183, opened 2018-12-29
Suggestion: fix sorting of patches in submenus to be A->Z?


esaruoho:
the patch loading submenus are not A-to-Z, instead whatever goes wherever. Surely something like that could be fixed, or what do you think @inigokennedy ?

<img width="456" alt="fullscreen_29_12_2018__12_56" src="https://user-images.githubusercontent.com/4966687/50537516-3cbcf980-0b69-11e9-8f7f-0efa7842c870.png">

baconpaul:
They are just in whatever order readdir returns. A sort is a good idea and easy enough.

baconpaul:
@jsakkine you were looking for a way to get into the ui. This one is pretty easy. There’s a vector ctge assembled in CPatchBrowser.cpp in the onmousedown. Apply the right flavor of std::sort to it before the subgroup iteration and this should be fixed.

baconpaul:
Or I can do it in the new year. But just thought this may be an easier approach to Mac dev and test than the other harder one with parameter invalidation

baconpaul:
Also looking at that screenshot we may want to adjust the split to submenu threshold

baconpaul:
Oh and don’t forget to remap indices. Alternately you could sort the entire structure when it is loaded in surgestorage. That way you can sort categories too. But then you need to adjust when you save one as a user. Not sure which I would do - would want to stare at code a minute more

jsakkine:
@baconpaul, what is split to submenu threshold?

jsakkine:
I think `SurgeStorage` is the appropriate choice, and instead of sorting, I'd just put categories to `std::map`.

jsakkine:
Actually, `std::set` should do as `struct patchlist_category` contains only a `std::string` member.

jsakkine:
Ugh, did not understand how deep in the code category array indexes are use as global identifiers, so yeah, doing something simpler while still trying to also clean up thinks just a bit.

jsakkine:
Commits done. No, PR yet because I only know that they compile at this point. If you want to take risk, go ahead and try (and give feedback).

jsakkine:
PR done. Next time I will do something to this will after NYE.

jsakkine:
Thanks for linking that! Just starting work on this.

esaruoho:
fingers crossed!

jsakkine:
@baconpaul, just detailing a bit #183 to check that I've not missed something.

[...]

PR #187, opened 2018-12-31
#183 ("Suggestion: fix sorting of patches in submenus to be A->Z?")

jsakkine:


jsakkine:
Please actually review the code changes. Simple stuff but touches category indices, which are spread all around the code base so better be careful here.

baconpaul:
Ok I can’t compile this until into 19 but did a scan with one immediate question. 

I’m not sure a partial change from this_name to thatName names the code more readable either. I will ponder that

jsakkine:
@baconpaul, That is how the coding style is usually changed in large open source projects. You apply the new style when you do something more useful in order to keep commit diffs nice and compact. It is well considered why I did not rename anything else. I'd guess camelCase for member variables is way to go?

baconpaul:
Oh yeah the two var array is a good change. I was wondering why you eliminated the little single element structure in my review

I think you are right in naming changes. But we do need a standard! Otherwise we will keep discussing it in prs :)

jsakkine:
For this particular class we could do one "rename commit" soonish e.g. `load_wt()` to `loadWaveTable()` and so forth. It has its tentacles everywhere. Much more important to have clean code than lets say in the code of one purpose UI component.

jsakkine:
I eliminated the one element structure essentially because it brings no value and map or vector of strings is easier data structure for the last change.

baconpaul:
Hmm ... well it adds symmetry and extensibility - if/when we add more info about categories we have a place to hang it. I rather liked the little structure for these reasons!

jsakkine:
OK, fair enough :) But I'll hold up until the current version is tested before changing anything. Would suggest then rename it simply as `struct Category`.

jsakkine:
Does it work out, though?

jsakkine:
My stand on removing that struct. If it is not used, then it is not used. If we need to have something fancier than just `std::string`, we should introduce it in that point. Now that struct only introduces complexity and does not bring value. Only when you need to extend something you'll know what you actually need. "just in-case" things usually don't work out. I could understand it if that was part of API/ABI but we don't have such constraints. Not going to argue hugely about it and can update the patch set not to remove it but does it make any sense?

baconpaul:
Still no laptop 

But I have a use for the strict to fix some of the saving issues kicking around. Would be good to be able to ask things like if a category is factory. So if it works I will restore the struct squash and push. But not until I get a machine with a compiler!

jsakkine:
@baconpaul, That is why I was asking for code review to get feedback what to do with it. I was wondering why the original issue was asking to only sort categories but I just realized that I've misread issue title. Also, it was not obvious if over-length was part of the issue or not.

Can you also tell what is wrong how the categories are sorted wrong at the moment when you say that "it sorts some of the categories"? It does sort all of the categories in three groups (factory, 3rd party and user).

I rework with this but yeah I was on hold for feedback for some time (no one had tried it before you).

jsakkine:
And keeping that microstruct is not like a show stopper for me. I can take it back but if we have it maybe it would not be a bad idea to have all data that describes a category in it. Then it would probably be more future-proof e.g.

```
struct Category {
   int id;
   std::string name;
}
```
This can be more easily kept in different types of data hierarchies and do same queries (linked list, vector, map in sorted or unsorted form).

jsakkine:
Yes, I get that design.

Here's what I propose:

1. Lets close this PR with the commit from my just rebased master with just the accepted commit.
2. I'll create a feature branch called `menusort` to my tree and work out something that I see appropriate based on your suggestion.
3. Then we figure out the rest.

baconpaul:
Great. Tested it all built so merging it in now. Thanks!

PR #200, opened 2019-01-04
Macro cleanup

whydoubt:


baconpaul:
Hi! Sorry but this change doesn't build on mac.

I think the problem is that you blanket changed #if AU to #if TARGET_AUDIO_UNIT but #if AU evaluates to false in the audio unit build, whereas #if TARGET_AUDIO_UNIT evaluates to true. So a bunch of old code which needs to be investigated activates, and doesn't build.

Did it build for you on mac? When I did this

```
  git fetch whydoubt
  git checkout whydoubt/macro_cleanup
  ./build-osx.sh --clean-all
  ./build-osx.sh --build-and-install
```

I got an error right away. (It would be super great if you could check mac PRs on mac if you don't, by the way).

I think the answer is to restore the #if AU and then we should go and read that code to see if it is needed in some other form. 

Thanks!

baconpaul:
@jsakkine it actually doesn’t build on mac! AU != TARGET_AUDIO_UNIT (in that in the AU build, TARGET_AUDIO_UNIT is defined and AU is not). See my comment above.

baconpaul:
Thanks for the force push @whydoubt. Builds for me now. I'll merge this - looks good.

Issue #157, opened 2018-12-26
Update Surge-plugin About screen?


esaruoho:
mainly talking about the image below..
<img width="898" alt="surge_50-surge" src="https://user-images.githubusercontent.com/4966687/50454143-3b63b500-094e-11e9-9bac-1de932de4d06.png">

says 2005-2015, Vember Audio, and Version is 1.6.0.. where are these texts coming from?
I tried doing git greps but was unable to come up with anything.

whydoubt:
https://github.com/kurasu/surge/blob/master/resources/bitmaps/bmp00158.png

baconpaul:
Yeah this is just an image blatted on.

In an ideal world, our build pipeline would create some file which had some defined strings we could paint on the screen. But it doesn't yet. @jsakkine is it possible cmake could construct "versions.h" or some such at build time so we could do this?

kurasu:
I'm not so sure either but we definitely need something automatic that we just render at runtime. 

I think we at least want

- Version or revision & branch
- The contributors
- Link to here
- Copyright notices as required by the VST2, 3 and Audio Units SDKs

Not sure if copyright is relevant, but the GPL license might be.

jsakkine:
Does SVG exist for the Surge logo?

kurasu:
No, it's a 3D render, but I should have a higher res version somewhere.. But it could also be time for a new logo, it's been >10 years.

baconpaul:
So the reason people appear twice - or at least I do - is that my GitHub id has proper name "paul" and my mac git global configuration has my full name.  If you use "%aN %ae" you can see the fact that at least for me the "paul" is baconpaul@users.noreply.github.com and Rich and Richie are the same email address.

I agree with @kurasu list of what should go in the screen. I would also include a build time and build date.

jsakkine:
I could write easily a Python script using GitPython that grabs the authors list and filters out broken names. Seems like the best option really. Once it is done, it is done.

esaruoho:
yeah, auto-generated "all that's been discussed in the comments above" feels like the right way to go.

incidentally, there's a TON of lines in the code to the tone of

```
Copyright 2005 Claes Johanson & Vember Audio
```

or

```
Copyright 2006 Claes Johanson & Vember Audio
```

or

```
Copyright 2005-2006 Claes Johanson & Vember Audio
```

or

```
Copyright 2004 Claes Johanson
```

or even

```
Copyright 2005 Claes Johanson & vember|audio
```

[...]

PR #204, opened 2019-01-04
More explanatory about box.

baconpaul:
This initial commit rewires the about box so the text painted on screen
comes from code, and also includes a build date and time. See issue #157 for
a discussion of the final state; this is an interim step to at least get correct information
and a build date into our releases.


src/common/gui/CAboutBox.cpp
```diff
@@ -29,6 +35,26 @@ void CAboutBox::draw(CDrawContext* pContext)
    if (value)
    {
       _aboutBitmap->draw(pContext, getViewSize(), CPoint(0, 0), 0xff);
+
+      int strHeight = infoFont->getSize(); // There should really be a better API for this in VSTGUI
+      std::vector< std::string > msgs = { {
+              std::string( "Version 1.6.0 beta (build: " ) + __DATE__ + " " + __TIME__ + ")",
+              "Released under the GNU General Public License, v3",
```
jsakkine: Should be a file lets say `VERSION` in root directory where the version would be grabbed?

src/common/gui/CAboutBox.cpp
```diff
@@ -29,6 +35,26 @@ void CAboutBox::draw(CDrawContext* pContext)
...
+              "VST Plug-in technology by Steinberg, AU Plugin Technology by Apple Computer"
+          } };
+
+      int yMargin = 5;
+      int yPos = toDisplay.getHeight() - msgs.size() * (strHeight + yMargin); // one for the last; one for the margin
+      int xPos = strHeight;
+      pContext->setFontColor( kWhiteCColor );
+      pContext->setFont( infoFont );
+      for( auto s : msgs )
+      {
```
jsakkine: Cosmetic: hinking about the white-spacing here and also in function calls. I use a different syntax. Which one Surge originally uses?

src/common/gui/CAboutBox.cpp
```diff
@@ -92,4 +118,4 @@ void CAboutBox::unSplash()
          getFrame()->setModalView(NULL);
       }
    }
-}
\ No newline at end of file
+}
```
jsakkine: Yep looking at `setModalView()` call, it uses a different as your commit.

src/common/gui/CAboutBox.cpp
baconpaul: Well we should have a built artifact with all our information! That's why this is an incomplete diff. For now I just hardcoded 1.6 so I didn't have to deal with cross platform resource loads but to properly finish this we need to create the XML file we were discussing which should include the version and so on and then not have this here.

My thinking was "hardcoded as string is better than as png" and "getting the build date and correct license" was valuable which is why I did this slightly but still slightly less tacky interim step.

src/common/gui/CAboutBox.cpp
baconpaul: You are 100% correct. That convention will never leave my fingers. I will go and fix it and squash in.

baconpaul: OK I pushed this change. Thanks for the careful look!

PR #205, opened 2019-01-04
Allow GUI invalidation for automation

baconpaul:
For almost every item the invalidation by the DAW now correctly invalidates
the GUI as described in #160. Only envelope mode is left unworking, since
that invalidates both the switch and another element but isn't hooked up in
the multi-element invalidation code right now. I'll open a separate low priority
issue for that and close this one.


baconpaul:
Absent comments I'll merge this later today. This just adds the rest of the places where my fix has to go, per the feedback on the first fix by @kurasu

PR #207, opened 2019-01-04
Assets directories

baconpaul:
Set up directories so the team in #23 can start doing PRs on various assets which we can use to get to a scalable UI one day.


baconpaul:
Closing this while folks talk about what's easy for affinity to export file and directory wise then I'll put to back together again!

PR #209, opened 2019-01-04
use wcsicmp instead of _stricmp where wchar_t* is provided (on windows)

kurasu:
master currently doesn't build on Windows due to the usage of _stricmp in SurgeStorage.cpp (lines 307, 348, 357)

_stricmp expects const char* but f.path().extension().c_str() is of type const wchar_t*.

The equivalent method to use for wchar_t* is wcsicmp (or _wcsicmp). One issue with a simple redefinition of _stricmp like we've done for mac & linux before is that the second argument is also a wide string, so it needs the L-prefix for string constants. So this needs to be adapted somehow to work on mac/linux as well.

### DIFF
--- a/src/common/SurgeStorage.cpp
+++ b/src/common/SurgeStorage.cpp
@@ -304,7 +304,7 @@ void SurgeStorage::refreshPatchlistAddDir(bool userDir, string subdir)
 
          for (auto& f : fs::directory_iterator(p))
          {
-            if (_stricmp(f.path().extension().c_str(), ".fxp") == 0)
+            if (wcsicmp(f.path().extension().c_str(), L".fxp") == 0)
             {
                patchlist_entry e;
                e.category = category;
@@ -345,7 +345,7 @@ void SurgeStorage::refresh_wtlist()
 
          for (auto& f : fs::directory_iterator(p))
          {
-            if (_stricmp(f.path().extension().c_str(),".wt") == 0)
+            if (wcsicmp(f.path().extension().c_str(), L".wt") == 0)
             {
                patchlist_entry e;
                e.category = category;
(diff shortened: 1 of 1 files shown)


baconpaul:
Sigh. So I think the problem here is that our implementation of filesystem on macOS is incorrect.

https://en.cppreference.com/w/cpp/experimental/fs/path/extension says that .extension() has type path; and path.c_str() has the native type https://en.cppreference.com/w/cpp/experimental/fs/path/native which on windows will be a wchar.

So what we need to do here is (I think) make extension() return a path as it should on mac and then do a _stricmp against path().extension().generic_string().c_str(). There's no guarantee that the native type is a wchar alas, but there is a guarantee that generic_string is non-wide.

I was trying to get my windows environment working this morning, and now I'm just going to try and do that to see if I can test this change. I think your change will break mac.

That being said, if you want to break mac at master so you can keep pushing on windows, please do. I can clean this up.

Apologies again. I won't push PRs until I can get a windows env working reliably.

baconpaul:
Yeah that's definitely it. On windows extension has type path() and so c_str() doesn't do what you'd expect. Let me try a different patch to fix it correctly OK?

baconpaul:
@kurasu I'm getting a windows VM which will actually build put together.

But I think the correct fix here is

https://github.com/baconpaul/surge/commit/16c3c6e640efb1f5b86c01302b10053fa6563f15

which fixes filesystem and uses .generic_string. 

I will see if I can get a build env on windows which builds surge. 

Apologies again.

kurasu:
yeah that commit works. thanks

no worries, would be great if we eventually get azure pipeline working though.

PR #213, opened 2019-01-05
Adding vector GUI sprite sheets.

itsmedavep:
I have now created these assets as vectors and generated the scaled .png files:

bmp00 112 (Filter Config)
bmp00 137 (FX Routing)
bmp00 151 (FM Routing)
bmp00 108 (Filter Type)
bmp00 145 (Envelope Stuff)


baconpaul:
Hi @itsmedavep - you actually named the directory "origional" not original.

But I changed my mind overnight. We shouldn't call this new skin "original" but should call it "classic-vector". Then we can preserve "original" for the non-scaled bitmaps we already have.

If you would make that change, I would sweep this. Also if you want to grab the rendered assets from @cyanit google drive and check them in, I would sweep that also.

Sweeping all to "asset" branch though not "master" branch for now. Good idea with that.

itsmedavep:
This has been done now.

> Hi @itsmedavep - you actually named the directory "origional" not original.
> 
> But I changed my mind overnight. We shouldn't call this new skin "original" but should call it "classic-vector". Then we can preserve "original" for the non-scaled bitmaps we already have.
> 
> If you would make that change, I would sweep this. Also if you want to grab the rendered assets from @cyanit google drive and check them in, I would sweep that also.
> 
> Sweeping all to "asset" branch though not "master" branch for now. Good idea with that.

 🤦‍♂️ 

Nice catch on the spelling. Thanks. 

Sounds good to me. I'll download Leo's scaled .png's and add them. Adding in assets seemed like Claes intent when he made it. I thought I should roll with it.

After getting some sleep and reading your comments Im proposing this. 

Id love you and Leo to weigh in. 

I have created this now:

assets/classic-vector
assets/classic-vector/exported
assets/classic-vector/exported/original
assets/classic-vector/exported/@1.5x
assets/classic-vector/exported/@2x
assets/classic-vector/exported/@3x
assets/classic-vector/master

1) We are having Affinity Designer spit out the scaled .png's filenames with this structure 
2) Mapping the dir structure the same way will make it a bit more human understandable at a glance. 

or 

As a designer, if I am looking at that in my build and comparing it against my slices output I'd "get it" 

Ohh we are generating these files with these naming conventions. Once built, they land in exported/@(whatever)

itsmedavep:
If you want entirely flat. You got it!

My thinking was that I was trying to save the necessity of pragmatically creating dirs for all this. Sounds unnecessary and painful even lol. Sorry about that. 

Done and done. 

FYI nope I couldn't get Affinity Designer to do that. I just sort by name and the select and move. 🤷‍♂️

PR #228, opened 2019-01-07
Add VERSION and AUTHORS files

jsakkine:
Add VERSION and AUTHORS files so that this information is easily available both for human readers inspecting the source code and scripting to bundle this to the application itself.

Part of the solution for #157. Probably will work well almost forever for a project of this scale but at least this is sufficient for 1.6.0.


baconpaul:
What's the mechanism by which these stay up to date?

I pushed back on @kzantow when he added a file called "VERSION" also. Is that some sort of standard? Using a single file for that seems so odd to me.

jsakkine:
@baconpaul, yeah it is fairly common. Nice to be able to check the version from the root of the source tree. For master it is often just the next version. Then you can easily pull the version number to various scripts such as one generating the about screen, packaging etc.

I would manually update AUTHORS once per version. You cannot get it reliably from git log because of inconsistencies and we might even get authors in future that do not show in git log for substantial changes (lets say university group contributing some cool DSP or whatever). Some things just should not be automated.

baconpaul:
OK fine you and @kzantow convinced me that a file called VERSION is reasonable and we can hand update AUTHORS. We will all notice when a new person asks for a PR, at least for now.

Issue #188, opened 2018-12-31
Clean fallback from memory management errors


jsakkine:
**Problem**

Currently we need to use `_aligned_alloc()`  together with the `new`-operator in order to allocate 16-byte aligned memory for use with the SSE2 instructions. For Linux and macOS this function is substituted to `malloc()`, which is always gives 16-byte aligned on these platforms.

It'd be better to use the `new`-operator for all allocations because this will cause `std::bad_alloc` to be thrown when allocations fail, which means that we always get a clean exit instead of 
crashing with a null pointer dereference. 

**Solution**

C++17 provides language level support for allocations for over-aligned data [1].

Example program:

```
#include <iostream>
#include <iomanip>

class alignas(256) Foo {
    unsigned long data;
};

int main()
{
    Foo *foo = new Foo();
    std::cout << std::internal << std::hex << std::setw(18) << std::setfill('0')
              << foo << std::endl;

    return 0;
}
```
If I compile with `-faligned-new`:

```
$ g++ -faligned-new -std=c++14 -o test test.cpp 

$ ./test 
0x0000557b8ff08f00
```

Compiler support:
- GCC: since 7.0 [2]. Distributions currently ship with 8.x.
- Clang: since 4 0 [3]. Isn't this quite old already?
- Visual Studio: since 15.5 [4] (with `/Zc:alignedNew`). Not sure how recent this is.

[1] https://www.bfilipek.com/2017/06/cpp17-details-clarifications.html#dynamic-memory-allocation-for-over-aligned-data
[2] https://www.gnu.org/software/gcc/projects/cxx-status.html
[3] https://clang.llvm.org/cxx_status.html
[4] https://docs.microsoft.com/en-us/cpp/build/reference/zc-alignednew?view=vs-2017

baconpaul:
Is there any reasonable clean fallback we can take if this happens? So many os host conbos.  Is it a legit concern?

jsakkine:
@baconpaul: I think it is important just to recognize the problem at this point. Not something that there is a "quick fix". Changing `malloc()` calls to use `new`-operator in the code base is a good start because at least the the behavior is consistent. Would not try rush with this.

baconpaul:
Ooh malloc and new are different on purpose. 

I guess my view is “if you are out of memory you will crash”. That’s not ok in the kernel but in Logic Pro i don’t think there’s anything smart a plug-in can do!

jsakkine:
[...]

PR #229, opened 2019-01-07
Update Windows compiler requirements and enable over-aligned allocations

jsakkine:
Be explicit on minimum version for Visual Studio. 15.5 is a good entry level version for us because it is the first version of Visual Studio that has support for over-aligned allocations. Update premake5.lua to
enable this support. This is needed for SSE2 opcodes that require 16-byte aligned data.

For Linux and macOS we need not to do anything because they align all allocations at least to 16 bytes.

This provides starting point for implementing #188 changes. The use of `_aligned_alloc()` is scattered around the code base. Once the build support over-aligned allocations is landed, the work can be more easily distributed to multiple people.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ Prerequisites
 
 * [Git](https://git-scm.com/downloads)
 * [Premake 5](https://premake.github.io/download.html#v5) for generating project files
-* [Visual Studio 2017](https://visualstudio.microsoft.com/downloads/)
+* [Visual Studio 15.5 (at least)](https://visualstudio.microsoft.com/downloads/)
 * [Inno Setup](http://jrsoftware.org/isdl.php) for building the installer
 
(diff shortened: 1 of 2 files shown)


baconpaul:
I think this is OK but I don't have a windows environment and I don't know what the span of build platforms we have is, so I'd really like some feedback from @kzantow and @kurasu on this, if that's OK.

baconpaul:
(I will have a windows environment soon; I have bit the bullet and making it part of my workflow)

jsakkine:
@baconpaul, at least seems to pass the Windows build according to the checks.

kurasu:
@jsakkine actually those are just dummy checks to verify the integration with azure pipelines, but @kzantow should have a proper build yml-scripts ready soon.

But it builds, and works.

PR #232, opened 2019-01-07
Basic powershell scripts to build.

baconpaul:
So I got windows building and installing using a PowerShell script which works for me. May work for other devs also with modification. I put it in scripts/win/ rather than the top level but it does build and results in me being able to run a VST2.

I put a big disclaimer in the comments that this is not production software. But I would love it in master so I can use it to quickly test builds on various other branches.

Anyone object if I sweep this? I will let it sit here for a day or two in case there are strenuous objections.


esaruoho:
@baconpaul in it goes

PR #241, opened 2019-01-08
Read version from the file VERSION to the about dialog

jsakkine:
Define a macro called SURGE_VERSION in premake5.lua that has the version string from the file VERSION and use this macro in CAboutBox to print the correct version number. Define a macro called STR(x) that stringifies a macro value.

### DIFF
--- a/premake5.lua
+++ b/premake5.lua
@@ -10,7 +10,8 @@ VSTGUI = "vst3sdk/vstgui4/vstgui/";
 
 defines
 {
-	"VSTGUI_ENABLE_DEPRECATED_METHODS=0"
+    "VSTGUI_ENABLE_DEPRECATED_METHODS=0",
+    "SURGE_VERSION=" .. string.gsub(io.readfile("VERSION"), "\n$", "")
 }
 
 floatingpoint "Fast"
--- a/src/common/globals.h
+++ b/src/common/globals.h
@@ -31,6 +31,9 @@ static inline int _stricmp(const char *s1, const char *s2)
 }
 #endif
 
+#define _SURGE_STR(x) #x
+#define SURGE_STR(x) _SURGE_STR(x)
+
 const int namechars = 64;
 const int block_size = 32;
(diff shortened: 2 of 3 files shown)


jsakkine:
Partial solution for #157.

jsakkine:
In https://github.com/premake/premake-core/issues/1216 there is some speculation that it might even be a bug in premake5. The method used in this commit is more or less guaranteed to work in all circumstances so I would go with this still.

jsakkine:
Well, if there ever is conflict, then we'll just rename it :) It is a macro so you cannot namespace it. If you want to be extra careful, it could be renamed `SURGE_STR()`. I would not bind the name to version as it can be used to convert anything to a string.

baconpaul:
That rename makes sense. If you do that I’ll happily merge. Thanks!

jsakkine:
Yeah, and I think then we have sufficient ingredients for closing about #157 (finally) at least for 1.6 release. In future, it would be cool to do some kind of maybe demo/intro style easter egg to the about dialog  that shows full credits :) Maybe even with some music played by Surge.

jsakkine:
Updated.

baconpaul:
Great. Thanks. Sweeping and I agree we can close #157.

jsakkine:
Finally. Who would have guessed that about-dialog could be such as daunting task :D

baconpaul:
Yeah! Well maybe we were a bit ambitious. It would be great to have gotten the head git hash on the screen. I'm thinking about poor @esaruoho hanging out in kvr forum with people with random installers asking questions. Having a clear version of a build a user can find seemed key. So I think it was worth it. But the build date and time plus non hardcoded version meet all those goals.

Now to get the installers working...

jsakkine:
Getting git hash is actually easy just in premake5.lua. If you want it, open an issue for me.

baconpaul:
Ok! I might (but will tag it low Pri)

Issue #250, opened 2019-01-10
Display the error recovery from the missing configuration.xml on Linux


jsakkine:
With Linux VST2 I get this:

```
SurgeStorage: Unable to load Surge configuration file "/home/jsakkine/.local/share/Surge/configuration.xml"
```

A file missing from the home folder (any file) **is not** an error condition. If a configuration file is missing from the home folder, the plug-in should either load or be able to create it.

What is so essential in this file that it cannot be created by the plug-in itself (the first version)?

jsakkine:
And what are snapshots and how they relate to the configuration?

Could this be instead:

```
if (snapshotloader.LoadFile(snapshotmenupath.c_str())) 
   snapshotloader.SaveFile();
```

jsakkine:
And there is null pointer dereference here:

```
TiXmlElement* e = snapshotloader.FirstChild("autometa")->ToElement();
```

Assumes that the configuration file was found even when it wasn't

jsakkine:
- I think SurgeStorage should throw an exception on error and not do GUI at all. Now it continues on error, which is just wrong.
- Whoever creates SurgeSynthesizer should catch this exception and display an error message.

I would suggest that we create a single exception class called `SurgeError` that has integer codes (e.g. enum) for different errors and maybe mapping to messages for those codes. We can start with a single error `CONFIG_FILE_NOT_FOUND`.

I think it would make sense to always allocate `SurgeStorage` from heap to make the whole creation process better controlled.

esaruoho:
@jsakkine can ya try and see what happens if the folder data/resources is copied manually to the path the configuration.xml error mentions?

jsakkine:
Yes, of course I tested it, but plug-in should not crash no matter what. It is in home folder so this is just broken behavior. After that it crashes in different location but I'll fix this one first.

jsakkine:
Ideally plug-in should be able to create that file.

jsakkine:
@baconpaul, BTW why OS specific code is used for the error dialog? Couldn't you just derive something from `CControl`?

esaruoho:
@jsakkine the configuration.xml missing error is pretty much the only way to know that:
- factory presets not present
- waveform display will be empty
- no wavetables present.

jsakkine:
What are snapshots? State of controls? Could those be stripped out fro the default? Can the plug-in re-create them?

jsakkine:
[...]

PR #251, opened 2019-01-10
Throw SurgeError when configuration.xml is not found

jsakkine:
Add a custom exception class SurgeError that is used to fail SurgeStorage constructor when configuration.xml is not found.

Not a perfect fix for #250 but exception is better than a crash for the VST host.

### DIFF
--- a/src/common/SurgeError.cpp
+++ b/src/common/SurgeError.cpp
@@ -1,0 +1,12 @@
+#include "SurgeError.h"
+
+SurgeError::SurgeError(const std::string &message)
+   : message(message)
+{
+}
+
+
+const std::string &SurgeError::getMessage()
+{
+   return message;
+}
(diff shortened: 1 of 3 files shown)


jsakkine:
After this fix Renoise at least fails cleanly and does not eat all memory. This commit really helps to continue debugging with Renoise because you don't have to always wait that it eats all your memory and crashes (takes a few minutes). I really hope it gets pulled.

jsakkine:
image

Much much nicer end result.

jsakkine:
Perhaps SurgeError should just take the message in its constructor because we are only going to just display a message anyway? I.e. I could make it even more dummier. The class is mainly for situations that will anyway end up the plugin being closed so I think that would be perfectly adequate.

baconpaul:
builds and runs mac; totally standard c++; sweeping with the caveat that I bet this will change again in some regard

jsakkine:
@baconpaul, yeah that is why I did not even try to over-engineer the exception class. Something dead simple we can refine as we go :)

Issue #248, opened 2019-01-09
Poly count fluctuates and flickers to 0 while notes are playing (VST/AU)


esaruoho:
Logic AU + Live VST
![poly freakout2](https://user-images.githubusercontent.com/4966687/50925762-f5cdde80-145b-11e9-83ab-3a0d0cd4c753.gif)

i tried to up the framerate of the gif but not sure if it shows properly

baconpaul:
I was able to see this in Hosting AU + AU also. The code looks correct and clear as day so I wonder what it is. Will require some debugging. Not an instant fix.

baconpaul:
Oh this is a threading problem

Idle polls on ui thread
Value is updated on audio thread
Value is zeroed out and reincremenred each process
We grab in the middle

Easy to fix. Don’t update in place etc

PR #254, opened 2019-01-10
Polydisplay Doesn't Flicker

baconpaul:
synth->polydisplay was updated by inplace set to 0 and increment
per voice in the audio thread, then read in idle from the ui thread.
thus the ui thread ocassionaly caught an interim value and flicked.
Change this by
1. make a local voices count and assign to polydisplay once
2. make polydisplay have atomic assignment

fixes issue #248


esaruoho:
@baconpaul are you waiting for someone to codereview this or is it just waiting for testing?

Issue #226, opened 2019-01-06
Towards Scalable UI: Retina Display on Mac


baconpaul:
So a full blown scalable UI is work, but thanks to the excellent work of @cyanit and @itsmedavep we now have assets and thanks to @esaruoho we are tweaking and cleaning them up in the 1x view available today on mac. Making the code work to scale properly will be tricky.

But a less tricky first step, on the mac at least, is to load the 2x assets and paint them scaled at 0.5. On retina displays and the like this will give an immediate improvement in the look, and it starts plumbing together the code.

My plan is roughly similar to the one I used in my abandoned SVG painting approach namely

1. Make a subclass of CBitmap which loads multi-scale
2. Make it load the scaled assets on Mac and return the underlying on Win and Lin
3. Have the ::draw mechanism do something different; in this case it would push a set scale to .5; super draw; and then pop.
4. Adjust the packaging scripts to include the scaled assets in the resources folder
5. See if it all works.

gulp.

baconpaul:
So hey I just got this kinda-working. Not ready for a PR or anything but on baconpaul branch named retina-bitmaps-226 it will use the 2x bitmaps at 0.5x scale.

I program with a nice big 5k monitor so use mac scaling. Don't know what my DPI is but the difference is immediate. Here's a screen grab of the 1x assets

<img width="907" alt="screen shot 2019-01-07 at 8 17 55 pm" src="https://user-images.githubusercontent.com/13387561/50803887-a2238f80-12b9-11e9-8ea6-d1c500035425.png">

and here's a screen grab of the 2x assets at 0.5 scale so macOS interpolates to use the intermediate pixels.

<img width="914" alt="screen shot 2019-01-07 at 8 17 16 pm" src="https://user-images.githubusercontent.com/13387561/50803907-bbc4d700-12b9-11e9-9ae6-7c723f5d8406.png">

So that's pretty cool!

PR #259, opened 2019-01-11
Retina Bitmaps

baconpaul:
@kurasu and others - I would really appreciate a review of this. While it works, I made design choices you may not like, so please take a look. I will not sweep it without a positive ack from @kurasu. 

And note that in default configuration this does nothing.

If scalable bitmaps are present we can begin to implement zoom.
Step one is to implement Mac Retina display. We do that here
by loading 2x bitmaps if available and drawing them at 0.5x
scale. On retina displays this looks great. On regular displays
it is a no-op basically. See #226 for more.

None of this is activated if the skins aren't installed and
by default it falls back to the standard bitmaps.

### DIFF
--- a/scripts/macOS/package-au.sh
+++ b/scripts/macOS/package-au.sh
@@ -37,7 +37,9 @@ cp -R "$BUNDLE_RES_SRC_LOCATION" "$BUNDLE_DIR/Contents/Resources"
 if [[ -z "$SURGE_USE_VECTOR_SKIN" ]]; then
     cp $BITMAP_SRC_LOCATION/* "$BUNDLE_DIR/Contents/Resources/"
 else
-    rm "$BUNDLE_DIR/Contents/Resources/bmp?????.png"
+    rm "$BUNDLE_DIR/Contents/Resources/bmp*.png"
     cp $VECTOR_BITMAP_SRC_LOCATION/bmp?????.png "$BUNDLE_DIR/Contents/Resources/"
+    mkdir "$BUNDLE_DIR/Contents/Resources/scalable"
+    cp $VECTOR_BITMAP_SRC_LOCATION/*png "$BUNDLE_DIR/Contents/Resources/scalable"
 fi
 
--- a/src/common/gui/CScalableBitmap.cpp
+++ b/src/common/gui/CScalableBitmap.cpp
@@ -1,0 +1,63 @@
+#include "CScalableBitmap.h"
+#if MAC
+#include <CoreFoundation/CoreFoundation.h>
+#endif
+
+CScalableBitmap::CScalableBitmap( CResourceDescription desc ) : CBitmap( desc )
+{
+    int id = 0;
+    if( desc.type == CResourceDescription::kIntegerType )
+        id = (int32_t)desc.u.id;
+    
+    scales = {{ 100, 200 }};
+    scaleFilePostfixes[ 100 ] = "";
+    scaleFilePostfixes[ 200 ] = "@2x";
+
+    //fprintf( stderr, "\n" );
+    for( auto sc : scales )
+    {
+        auto postfix = scaleFilePostfixes[ sc ];
+        char filename [PATH_MAX];
+        sprintf (filename, "scalable/bmp%05d%s.png", id, postfix.c_str() );
+            
+        CBitmap *tmp = new CBitmap( CResourceDescription( filename ) );
+        //fprintf( stderr, "%s, Scale=%d, Dim=(%lf x %lf)\n", filename, sc, tmp->getWidth(), tmp->getHeight() );
+
+        if( tmp->getWidth() > 0 )
+        {
+            scaledBitmaps[ sc ] = tmp;
+        }
+        else
+        {
+            scaledBitmaps[ sc ] = NULL;
+        }
+        
+    }
+}
+
+void CScalableBitmap::draw (CDrawContext* context, const CRect& rect, const CPoint& offset, float alpha )
+{
+    /*
+    ** For now this is a fixed 'retina' style draw of using 2x bitmaps at 0.5 scale if they exist
+    */
+    if( scaledBitmaps[ 200 ] != NULL )
+    {
+        CGraphicsTransform tf = CGraphicsTransform().scale( 0.5, 0.5 );
+        CGraphicsTransform invtf = tf.inverse();
+        
+        CDrawContext::Transform tr( *context, tf );
+
+        // Have to de-const these to do the transform alas
+        CRect ncrect = rect;
+        CRect nr = invtf.transform( ncrect );
+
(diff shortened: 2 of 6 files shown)


scripts/macOS/package-au.sh
```diff
@@ -37,7 +37,9 @@ cp -R "$BUNDLE_RES_SRC_LOCATION" "$BUNDLE_DIR/Contents/Resources"
 if [[ -z "$SURGE_USE_VECTOR_SKIN" ]]; then
     cp $BITMAP_SRC_LOCATION/* "$BUNDLE_DIR/Contents/Resources/"
 else
-    rm "$BUNDLE_DIR/Contents/Resources/bmp?????.png"
+    rm "$BUNDLE_DIR/Contents/Resources/bmp*.png"
```
jsakkine: What is this change BTW?

src/common/gui/CScalableBitmap.cpp
```diff
@@ -0,0 +1,63 @@
...
+CScalableBitmap::CScalableBitmap( CResourceDescription desc ) : CBitmap( desc )
+{
+    int id = 0;
+    if( desc.type == CResourceDescription::kIntegerType )
+        id = (int32_t)desc.u.id;
+    
+    scales = {{ 100, 200 }};
+    scaleFilePostfixes[ 100 ] = "";
+    scaleFilePostfixes[ 200 ] = "@2x";
+
```
jsakkine: Or?

```
    scales = {{100, 200}};
    scaleFilePostfixes[100] = "";
    scaleFilePostfixes[200] = "@2x";
```

src/common/gui/CScalableBitmap.cpp
```diff
@@ -0,0 +1,63 @@
...
+    //fprintf( stderr, "\n" );
+    for( auto sc : scales )
+    {
+        auto postfix = scaleFilePostfixes[ sc ];
+        char filename [PATH_MAX];
+        sprintf (filename, "scalable/bmp%05d%s.png", id, postfix.c_str() );
```
jsakkine: `sprintf()` vs `sprintf ()`?

jsakkine: BTW, why not just `std::string` for filename.

src/common/gui/SurgeBitmaps.cpp
```diff
@@ -72,7 +80,11 @@ void SurgeBitmaps::addEntry(int id)
 {
    assert(bitmap_registry.find(id) == bitmap_registry.end());
 
+#ifdef USE_SCALABLE_BITMAPS
```
jsakkine: BTW, could add some prefix for all _feature flags_ i.e. compilation flags that enable/disable a feature. Actually `USE_` is quite common and would be fine but can be take it as a convention?

src/common/gui/CScalableBitmap.cpp
```diff
@@ -0,0 +1,63 @@
...
+    if( desc.type == CResourceDescription::kIntegerType )
+        id = (int32_t)desc.u.id;
+    
+    scales = {{ 100, 200 }};
+    scaleFilePostfixes[ 100 ] = "";
+    scaleFilePostfixes[ 200 ] = "@2x";
+
+    //fprintf( stderr, "\n" );
+    for( auto sc : scales )
+    {
```
jsakkine: `for (auto sc : scales)` ?

jsakkine: Weird spacing is everywhere in this commit...

src/common/gui/SurgeBitmaps.cpp
```diff
@@ -72,7 +80,11 @@ void SurgeBitmaps::addEntry(int id)
...
+   VSTGUI::CBitmap *bitmap = new CScalableBitmap(CResourceDescription(id));
+#else
    VSTGUI::CBitmap* bitmap = new VSTGUI::CBitmap(CResourceDescription(id));
+#endif
```
jsakkine: Couldn't you always use CScalableBitmap even if it does not make difference? Do not understand this complexity...

src/common/gui/CScalableBitmap.cpp
jsakkine: I would just have something like

```
int scaleFilePostfixes[2];
VSTGUI::CBitmap *scaleBitmaps[2];
```

The need to use `std::map` comes only from the fact that you have factored your scales.

scripts/macOS/package-au.sh
baconpaul: That change makes sure you blow away bitmaps which also have form 00100@2x.png names

src/common/gui/CScalableBitmap.cpp
baconpaul: I don't have 2 sizes; I have 4 or 5 (1, 1.5, 2, 3, 4 at least). I chose an int because the 1.5 size is a float and I didn't want to hash float. When we do zoom properly this will be a longer list so I built it to scale to that.

src/common/gui/CScalableBitmap.cpp
baconpaul: You mean "correct spacing is everywhere in this commit and the rest of the world uses weird spacing" surely ;)

That spacing will never leave my fingers. I'll clean it up when I clean up other review comments.

src/common/gui/CScalableBitmap.cpp
```diff
@@ -0,0 +1,63 @@
...
+    int id = 0;
+    if( desc.type == CResourceDescription::kIntegerType )
+        id = (int32_t)desc.u.id;
+    
+    scales = {{ 100, 200 }};
+    scaleFilePostfixes[ 100 ] = "";
+    scaleFilePostfixes[ 200 ] = "@2x";
+
+    //fprintf( stderr, "\n" );
+    for( auto sc : scales )
```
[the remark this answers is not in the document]
baconpaul: Yeah I left those in since I want them back for debugging when I go to 1.5x 3x and 4x

src/common/gui/SurgeBitmaps.cpp
baconpaul: The "if mac" has to go somewhere until we sort out resources and loading them on windows and linux. I put it outside the class. I could put it inside the class. Outside the class seemed cleaner.

src/common/gui/CScalableBitmap.cpp
baconpaul: Per slack convo I will add a comment here explaining a bit more

baconpaul:
OK I incorporated @jsakkine review comments and had a lengthy conversation with him about it on slack. As a step on the way and as optionally off, he understood. So I'll merge this so I can put skin switchers in now we have @itsmedavep skin.

Issue #247, opened 2019-01-09
Automation Hosts learn the wrong names (Live, Bitwig, VST2, AU)


esaruoho:
Steps to replicate:
1. Launch Ableton Live
2. Load Surge (VST or AU)
3. Click on "Configure"
4. Move a slider in Surge

Expected result:
the slider's slider will appear in Ableton Live

Current Result:
the slider's slider appears, but when you let go of tweaking the slider, another "twin" slider appears.

![surge_ableton_live](https://user-images.githubusercontent.com/4966687/50925370-d6828180-145a-11e9-9a4b-5ec0197573fa.gif)
![surge_ableton_livev2](https://user-images.githubusercontent.com/4966687/50925373-d8e4db80-145a-11e9-8be7-a3bb8ab67424.gif)

for instance,
`Filter EG Decay` "twin" is `LFO4 Deform`
`F2 Cutoff` "twin" is `LFO 3 Rate`
`FX1 Send` "twin" is `F2 Envmod`
`Output Pan` "twin" is `F2 Resonance`
`F2 Resonance` "twin" is `LFO3 Magnitude`
`F2 Keytrack` "twin" is `LFO3 Unipolar`

and so on and so on.

baconpaul:
Any way to get a demo version of live? I’ve never used it.

esaruoho:
@baconpaul https://www.ableton.com/en/trial/

baconpaul:
I think this and #193 sound super similar right?

esaruoho:
@baconpaul well it could be related, if bitwig is a kind of "only one can get bound with a click" (i.e. it discards the first one of the "two" bindings)..

esaruoho:
@baconpaul to know for sure, i'd have to have bitwig installed..

baconpaul:
OK so in bitwig I just automated Filter EG Decay and got the binding to only one thing, which was LFO4 Deform. So bitwig and Live are doing the same thing.

Which means this and #193 are the same problem.

And which means I can debug it because I can use bitwig!

Note that the vst3 in bitwig basically totally doesn't work on automation.

So definitely something interesting here.

baconpaul:
OK I'm going to consolidate into one issue. From #193 

> From aMUSEd on kvraudio:
[...]

PR #262, opened 2019-01-11
Fix VST2 and AU Learned Automation

baconpaul:
@kurasu very interested in your comments on this diff also. See #247 for the problem. This solution seems the smallest one.

This diff fixes automation in VST2 in Live, Bitwig, Reason etc...
The basic problem is that the display control ids differ from the logical
control ids. The synth remap does logical remapping but the display
remapping - which is implemented by "- start_paramtag" or "+ start_paramtag"
in the code doesn't flow to events uniformly. Specifically it didn't
flow to beginEdit and endEdit in VST2. So expose that "-" operation as
a well named function, override begin and end edit in vst2 host,
and voila.

We will definitely need to do the same with vst3 and au somehow.
Still figuring out those codepaths though.

### DIFF
--- a/src/common/gui/SurgeGUIEditor.cpp
+++ b/src/common/gui/SurgeGUIEditor.cpp
@@ -2356,4 +2356,9 @@ void SurgeGUIEditor::zoomInDir(int dir)
    zoom_callback(this);
 }
 
+long SurgeGUIEditor::applyParameterOffset(long id)
+{
+    return id-start_paramtags;
+}
+
 //------------------------------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


esaruoho:
serious LGTM on this, with #263 being a continuation of "just a few left"..

baconpaul:
Oh hey the AudioUnit fix was super easy. Learn now works in logic, so let me sweep that into this PR also for when @kurasu gets an opportunity to look.

kurasu:
So AU is tested? I will build VST 2 & 3 and make sure they work

esaruoho:
@kurasu AU + VST2 tested on macOS and worked fine.

kurasu:
Tested on windows 10 in Bitwig

**VST3**

- Writes automation properly, but doesn't repaint sliders on automation playback

**VST2**

- begin/endEdit seems to be called from the wrong parameter indexes
- No automation gets written when moving parameters from within Surge
- Automation writes properly from the Bitwig GUI & then plays correctly

PR #276, opened 2019-01-12
Vector Source: Adding Sketch Source file. Attempted to match 1:1 the look of the old plug

itsmedavep:
Created a Sketch vector source file
Attempted to match as closely as possible the original look
Wanted a known state of usability
Rendered all slices
Identified any issues
Re-rendered all files 
* About screen still needs to be fixed

I went a bit nuts with this. Lots of it is essentially pixel by pixel BUT with vectors instead. Props to whoever did the pixel by pixel font in the first place. I did a whole bunch of that type of work once upon a time. 

Regardless, I converted all fonts to vector outlines to help with legibility. I have not included the brushed aluminum look. I still need to fix the about screen. I have spent a fair amount of time debugging and making sure everything was there. This all looks correct to me.


baconpaul:
Dave this is great but by putting it in the classic vector folder it overrides Leo’s work right? Maybe make an original-vector folder side by side. If you want to run it I can make the installers pick easily.

Look forward to trying but don’t want to do an overwrite merge, disk space is cheap!

esaruoho:
Thanks! Any chance of attaching a screenshot (or the sketch file, i happen to actually have the app)

itsmedavep:
Hello Everyone:

Awww. It was not my intent to overwrite anything or anyone's work with this. 

I thought by putting it thusly:
assets/classic-vector/master/sketch
assets/classic-vector/master/sketch/master
assets/classic-vector/master/sketch/exported

That it would only be an add and wouldn't disturb other work. 

Regardless, I wasn't sure where it _should_ land. So, all of that is to say that I'm happy to move.

So the structure is now:
assets/original-vector/exported
assets/original-vector/master

esaruoho:
@itsmedavep hi, here's one thing that has gone missing, these little "brighter gray" boxes between Scene, Route and Modulation:
fullscreen_12_01_2019__17_14

p.s. hurrah this is a great set of assets!

itsmedavep:
> @itsmedavep and the corners are more round in old (new on left old on right)
> vember_audio__surge_and_renoise__intel_64bit_


Ah that. That is one thing I did take liberty with. Early decision and I didnt even consider it was different. I just went "Oh they didnt think about this"  in the original and moved on.  The OG bothered me because the outer rounded rect's radius seemingly was not 2x as big as the inner. Thus creating a bulging at the corner. Its an easy fix if we want to go there.  The way it is now I have become accustom to. 

> also the `M` coloring is less visible for reading on the new than on the old :)
> 
> would love to see what the skin looks like when the M / S are activated - but is that done by code, @baconpaul ?

This is something I noticed as well. Ill take a look at fixing it up.

baconpaul:
Great
But Dave your diff still removes all the svgs in classic. That’s the overwrite.

I have 10 minutes today to make skin pickanle at build time for Mac. Just not right now. Can get it together before end of day

baconpaul:
So basically those gone svg do a checkout st master commit push and then I can sweep

Or alternately close this pr make a fresh branch and just add original vector as a new pr. That may be easier

baconpaul:
And very excited

itsmedavep:
> Great
> But Dave your diff still removes all the svgs in classic. That’s the overwrite.
> 
> I have 10 minutes today to make skin pickanle at build time for Mac. Just not right now. Can get it together before end of day

Ahh. Got it. Ill fix. Ill close this. Make a freshie and just add this stuff back.

Issue #268, opened 2019-01-11
"Clear All" is too heavyhanded - will rename LFO + Velocity etc to -


esaruoho:
As of now, if you learn a modulation for LFO1, Velocity, Keytrack, Poly AT, Chan AT, Pitchbend, Modwheel, Amp EG, Filter EG, Timbre etc and then right-click and choose `Clear All` - these will be named to `-`

![clear all bug](https://user-images.githubusercontent.com/4966687/51051178-b62e0080-15db-11e9-8f6a-fdb410ad545d.gif)

baconpaul:
if (within_range(ms_ctrl1, modsource, ms_ctrl1 + n_customcontrollers - 1))

Is the condition

baconpaul:
OK this is fixed but I'm going to combine it with a PR which also gets #261

Issue #261, opened 2019-01-11
[bug] when binding two parameters, refresh control-box to parameter#2-name if parameter#1 is unbound


esaruoho:
steps to replicate:
1. activate learning of Control-box (blue box)
2. bind Cutoff
3. bind Resonance

control-box now says Cutoff
4. Remove Cutoff
expected result:
control-box now says Resonance
current result:
control-box still says Cutoff
![second half](https://user-images.githubusercontent.com/4966687/51040342-e5cf0f80-15bf-11e9-87b7-d764f41dac14.gif)

third half:
bind Cutoff, bind Resonance, clear Cutoff, clear Resonance = still says Cutoff

esaruoho:
@baconpaul requested i reference the commit with the `clear all`
b0fa50560d4da4e23a3797756d0930de89a8d5f2

PR #277, opened 2019-01-12
Fix two menu reset problems

baconpaul:
As reported in #268, the menu reset i put was too heavy handed.
Make it so it cant reset "Timbre and the like".

As reported in #261, if you are in auto-name mode and remove
your first mapping you shoudl change the name. Comment this since
the logic required more than 10 seconds thought.


esaruoho:
probably codereview required, but tester's side is lgtm

jsakkine:
Looking into this. Hold on.

jsakkine:
Simple question: what is EID?

baconpaul:
Eid is the counter menus use as they build in current code. Element id perhaps? I just kept the name in the tracking variable

jsakkine:
Already merged, but looked good to me (except spacing and maybe static_cast<> instead of C style cast).

baconpaul:
I thought I fixed all the spacing! Thanks for looking tho!

Yeah static cast in new code is better. I was sloppy copying the old cast here.

Issue #231, opened 2019-01-07
Store patch dialog: Clicking on Name/Category can result in dialog forgetting changes


esaruoho:
Logic Pro X has the best bugs:

Steps to replicate:
1. Launch Surge
2. Click Store
Store Dialog opens.
3. Click on `Name`.
4. Type banana on `Name`.
5. Click on `Category`.
Expected result:
banana stays
Current result:
banana disappears
6. Type mango on `Category`
7. Click on `Name`
Expected result:
mango stays
Current result:
mango disappears.

Modified:
Type banana on `Name`.
Press Tab.
Type mango on `Category`
Press Tab.
Type avocado on `Creator`
Press Tab.
Type cucumber on `Comment`
Click `Name`

expected result:
cucumber stays
current result:
cucumber disappears

esaruoho:
![surge_storer3](https://user-images.githubusercontent.com/4966687/50794640-57fbd780-12d4-11e9-8f0f-a8c612aea5f3.gif)
here's a better one.

kurasu:
This code is new for 1.6, so its not been widely tested. The save dialog was previously implemented natively for both windows & osx but I converted it to just use the VSTGUI widgets to ease portability.

baconpaul:
Interestingly, and terrifyingly, this is host dependent. 

For instance if I run it in logic what you say happens

But when I do the same in Hosting AU I get this

![hosting-au](https://user-images.githubusercontent.com/13387561/50798693-3d5f3980-12a7-11e9-9dda-20399ced9359.gif)

"cool".

ugh must be some intercept happening deep inside the vstgui which differentiates tab from return in some host contexts on mac. 

ugh.

esaruoho:
@baconpaul further detail on Logic Pro X saving:

1. Press Store
2. Click on Name
3. Change name
4. Press OK

Result: Name is not changed.

Solution:

1. Press Store
2. Click on Name
3. Change name
4. Press Tab
5. Click OK

result: name is changed.

baconpaul:
Yeah I took a minute last night to look at the ctext implementation. There is some focus handling stuff which looks like it may be the issue.

This one will be tricky to debug I think. It will stay open for a bit. I certainly won’t get to it soon

PR #278, opened 2019-01-12
Focus issues in save

baconpaul:
The vstgui and various mac hosts have wierd focus change issues
which don't update the value in some cases, which are both
host specific and wierd. Instead of fixing that underlying
scary issue, we can just put the editors in immediate mode
instead and get a working dialog.

This fixes the issue reported in #231


baconpaul:
OK @esaruoho now you are good to test. Thanks!

esaruoho:
i am able to "go crazy" on the VST2 and AU save dialog, can type to Name, click on Category, type, click on Author, type, click on comments, type, then click Ok -> saved.
I can also just click on Store -> type a name -> click Ok = the correct name is saved.

This with Renoise (VST2) and Logic (AU). Good stuff @baconpaul

Issue #269, opened 2019-01-11
Next/Previous Patch/Category/Wavetable do not conform to A-Z sorting


esaruoho:
it would appear that the @jsakkine A-Z sorting for patches, is ignored when using the +/- to select next/previous patch.

First the Next/Previous Patch:

Steps to replicate:
1. Select `Bass` -> `Attacky` (first one)
2. Click on `+` 

Expected result:
`Bass 10`  is shown and ready to play

Current result:
`WTbass` is shown and ready to play.

3. Click on `+` again. 
Expected result:
the first Chord-category preset is loaded.
Current result:
`Bass` -> `Stable` is shown and ready to play.

`+` again -> `Pad` -> `Pad 5`

and so on.
![patch-change](https://user-images.githubusercontent.com/4966687/51051669-2721e800-15dd-11e9-9625-62d34a0ec5e6.gif)

Second:
Next/Previous Category:
Steps to replicate
1. Select a Bass patch
2. Click "+". on Category.
Expected result:
Chord
Current result:
Pad

Press +, get Percussion. Press +, get Polysynth. Press +, get Pad (choir), 
`+` = Vocoder.
`+` = Pad (animated)
`+` = Pluck (fast)
`+` = Sequence
`+` = Chord
`+` = Keyboard
`+` = inigo_kennedy_01
`+` = Argitoth
`+` = Giana-Brotherz
`+` = Bluelight
`+` = Noisegeek
`+` = Rozzer

and so on.

jsakkine:
Really appreciate the level of detail that you go with your reports!

esaruoho:
@jsakkine my pleasure :)

baconpaul:
Great! Around line 1965 of surgeguieditor.cpp is where the event gets handled in case you haven’t traced through the event code yet.

esaruoho:
@baconpaul you were right. the wavetables next/back also doesn't work accurately.
update:
the arrow right in wavetables, when your selected wavetable is "pmfreq" and you press arrow right, you get "sine windowed".. not "primes".


<img width="707" alt="fullscreen_13_01_2019__16_02" src="https://user-images.githubusercontent.com/4966687/51086228-c74a4f00-174c-11e9-8667-2bbef8ccb6c9.png">

jsakkine:
OK was a bit busy weekend. I'll look into this tonight.

jsakkine:
[...]

PR #297, opened 2019-01-14
Fix for #269

jsakkine:
See the commit message.


jsakkine:
Hey, if this looks otherwise good, I will replace the enum with a boolean parameter `nextPrev`. Using enum for that is an overkill. But before doing this I'll hold for feedback.

jsakkine:
Started to distract me enough, so I did that change already.

baconpaul:
Hey thanks for doing this! This is a big diff and I won’t have time to look at it closely for a day or two but I do see it here. Curious why you chose true false as opposed to +1/-1 thought? The rest scans as correct but involved in a very quick read but like I said want to spend a bit more time with it

esaruoho:
verified as working:
VST2: Renoise
AU: Logic Pro X.

I can switch both Categories, and Presets, and they work as expected (incremental change of +1 or -1, not +31 -31). LGTM

jsakkine:
@esaruoho, thank you. 

@baconpaul, I'll do another PR for wavetables after this has been merged.

esaruoho:
One thing that could be fixed with the -/+ Category behaviour, is when you cross over from "regular factory presets" to "regular factory 3rd party presets" to "user presets".

Steps to replicate:
Be in `Bass` -category ( first category ). press `-` 

Expected result:
you are immediately in the lowest category of the user-preset categories.

Current result:
nothing happens. press `-` again. nothing happens. press `-` again.. -> you are in the lowest category of the user-preset categories.

== this means that to cross the "boundary" of the preset categories, you need to press `-` three times.

Now, go to the "first user preset category", and press `-`. again, nothing happens. press `-` again, and `-` again. now you are in the "last 3rd party preset category". get to "first 3rd party preset category", press `-` , again nothing happens, press `-` two times more, you are in the `last factory preset category`.

Same with `+` . so "a little extra oomph" required currently to "cross the "threshold""

jsakkine:
@esaruoho, thanks does not sound like that there is anything seriously wrong in the code change. I'll look into this after work.

jsakkine:
Just rebased the existing code so that I can just concentrate on fixing the issue that @esaruoho reported once I have time.

jsakkine:
@esaruoho, I have _FM bass 1_ and I press '-'. I tried also your other example (from user categories to 3rd party categories by using '-'). It works too as expected. I also look at the code change and hard to see anything in it that could possibly cause that kind of behaviour.

Maybe I did not fully understand the example.

jsakkine:
@baconpaul, maybe you could try the scenario that @esaruoho describes in the comment section (when you really have the time, does not make sense to rush with big changes). Tried to reproduce it numerous times but do not experience any problems.

baconpaul:
Hmm OK so I got this to build and when I start it from scratch it doesn't show the @esaruoho but when I start it with a saved state in my AU test host, I get a reliable core dump. I'm just in a coffee shop on my laptop trying to check quickly, but clearly something odd here. 

I'll try and get a better report but yeah I think we want to hold this one for a minute while I figure out what is making my env give a segfault.

esaruoho:
well guys, i couldn't replicate the behaviour renoise/vst2 logic/au at home, so i'm inclined to think something else was wrong. i'll build one on the other computer tomorrow morning and see if it works like it should (like you guys are seeing it work, and me at home), or if it doesn't work like it should on this one specific machine, for some unknown reason :)

Issue #295, opened 2019-01-13
orange boxes and the text therein


esaruoho:
@baconpaul requested i post a ticket about this (instead of just posting it on #design on slack), so here it is:

"kinda looks like the text is right at the bottom corner of the orange boxes, which makes them feel like they're bugging just a bit"
<img width="161" alt="fullscreen_13_01_2019__17_45" src="https://user-images.githubusercontent.com/4966687/51089384-e8bf3100-1774-11e9-85c7-dbdd750ae41c.png">

and
![text going above orange box](https://user-images.githubusercontent.com/4966687/51089390-efe63f00-1774-11e9-915f-ae8addca5ebf.gif)

itsmedavep:
Ive fixed these in the buttons now.
![bmp00161](https://user-images.githubusercontent.com/11078351/51094906-5ddf3600-1776-11e9-8bb7-076eb8baaa4f.png)

Im off for the night but, Ill get a PR pulled together for this in the AM.

baconpaul:
In fact you can see that in this screenshot, it is aligned at the drop point (the bottom of the p) not the baseline (the bottom of the m). But that font looks a pt or two too big for that display area I think is the real problem.

baconpaul:
that's lucida grande 9 on mac; and microsoft sans serif 9 on windows

getting a similar screenshot from prod on windows would be great.

esaruoho:
@itsmedavep hi, looks like these could be a bit more related to eachother - what i'm getting at is that now the sine+triangle+square+ramp+noise+s&h+envelope+stepseq orange looks like it's "lower" in the "box area" than keytrigger+freerun+random.
(i.e. feels like the orange box is "align to bottom" rather than "align to center". dunno if that makes sense.

<img width="947" alt="fullscreen_15_01_2019__0_52" src="https://user-images.githubusercontent.com/4966687/51146664-2947a900-1860-11e9-85a2-597844a63582.png">

baconpaul:
The orange box hasn't moved; that is painted with code in explicit coordinates. If you see that it means the gray area has shifted "up"

itsmedavep:
> The orange box hasn't moved; that is painted with code in explicit coordinates. If you see that it means the gray area has shifted "up"


Yup. The orange didn't move but I must have managed to nudge the box somehow. Ill fix this up!

Sorry I introduced this BTW.

sense-amr:
windows version here.. 
as requested by @esaruoho 

[...]

PR #299, opened 2019-01-14
Original Skin: Update to source and rendered files - Centering button…

itsmedavep:
… labels, Fixing routing Diagram, Fixing ADSR env drawing

Per an ongoing conversation WRT #295 

I have done my part to better center the text in several buttons.

Centering text in several buttons. Warm, Neutral, Bright, About, Store

I also discovered that the routing diagram was missing the triangle and other bits for the L-R diagram. This update fixes that as well.

The ADSR envelop should draw correctly now as well. It was off by a pixel.


baconpaul:
Ok that’s way better! Merging

Issue #217, opened 2019-01-05
Uniform way to show errors to users


baconpaul:
if you `grep -r MessageBox` in src you see it shows up some, and as I'm writing wav file reader on mac, the condition which causes us to call it may show up more. In the SurgeStorageLoadWavetable even better it defines a function called error_msg which does nothing with errors. There's a lot of fprintf(stderr on mac and linux as well which doesn't help get to most users.

So I think we need to do something like have an entry point like "displayErrorToUser( title, string )" which pops a reasonable modal error message with an OK button and then go cleanup code.

Probably the way I would do this, but I'm happy to listen to other suggestions, is in SurgeSynthesizer add something like

```
public:
   static void displayErrorToUser( const char* title, const char* msg );
   static void setDisplayErrorHandler( std::function<...> f );
private:
   static std::function<...> s_displayErrorHandler
```

and have the default displayErrorHandler be fprintf(stderr). Then in the SurgeGUIEditor constructor set a display handler that does GUI stuff. That way if we want to use Surge in a headless-y mode sometime in the future we can still see user errors.

Not coding this now. Comments on design welcome.

jsakkine:
Some questions on top of my head: 

- Why they need to be inside a class? I don't see any special association to `SurgeSynthetizer`? Alternative would be to have something like `Error.h` and global functions.
- Why `setDisplayErrorHandler()` is needed? Why not just have the platform magic inside `displayErrorToUser()`?
- Why `displayErrorToUser()` and not just `displayError()`? Doesn't everything displayed go to user?

baconpaul:
The reason to have a handler is part of me is thinking about a GUI-less surge and so being able to pick your handler matters. 

The rest all make sense to me.

jsakkine:
@whydoubt, how does VSTGUI expose `GtkWindow`? Would be needed to use `GtkMessageDialog`: https://developer.gnome.org/gtk3/stable/GtkMessageDialog.html

jsakkine:
@baconpaul, @kurasu: proposal for new plug-in creation sequence

1. `new SurgeGUIEditor()` (does not take synth)
2. `new SurgeSynthesizer()`
3. `editor->setSynthesizer(synthesizer)`

This way we can use VSTGUI straight from the epoch.

whydoubt:
@jsakkine Regarding VSTGUI and GtkWindow: FWIW, note that VSTGUI has been stripping out uses of Gtk and is targeting lower-level libraries instead.

[...]

PR #301, opened 2019-01-15
Implement a User Reporting class - ready for second review

baconpaul:
There are lots of times when we want to pop messages to users,
and as we find more cases where we want to tell users things, the
code was going to be littered with #if mac this #else windows that.

So make a class which has some basic functions for reporting an error
and doing an OK cancel; implement it on our platforms under the covers;
and use it in two places (configuration.xml missing and re-save a patch).

Get it working on Mac and Windows, compiles and uses stderr on linux. And voila

@jsakkine I am especially interested in your design review of this (as well as the overall code review). Lets leave aside the throw vs prompt in place in SurgeStorage for a little - that's a separate issue - and let me know if you think this API is the right one for this function and so on.

The issue discussing is #217

### DIFF
--- a/src/common/SurgeError.h
+++ b/src/common/SurgeError.h
@@ -2,13 +2,21 @@
 
 #include <string>
 
-class SurgeError {
-public:
-   SurgeError(const std::string &message);
-   SurgeError() = delete;
+namespace Surge
+{
 
-   const std::string &getMessage();
+class Error {
+public:
+   static const std::string DEFAULT_TITLE;
+   Error(const std::string &message, const std::string &title = DEFAULT_TITLE );
+   Error() = delete;
+
+   const std::string &getMessage() const;
+   const std::string &getTitle() const;
 
 private:
    std::string message;
+   std::string title;
+};
+
 };
--- a/src/common/SurgeError.cpp
+++ b/src/common/SurgeError.cpp
@@ -1,12 +1,23 @@
 #include "SurgeError.h"
 
-SurgeError::SurgeError(const std::string &message)
-   : message(message)
-{
-}
 
-
-const std::string &SurgeError::getMessage()
+namespace Surge
 {
-   return message;
-}
+    const std::string Error::DEFAULT_TITLE = "An error has occured";
+
+    Error::Error(const std::string &message, const std::string &title)
+        : message(message), title(title)
+    {
+    }
+    
+    
+    const std::string &Error::getMessage() const
+    {
+        return message;
+    }
+    
+    const std::string &Error::getTitle() const
+    {
+        return title;
+    }
+};
--- a/src/common/SurgeStorage.cpp
+++ b/src/common/SurgeStorage.cpp
@@ -4,6 +4,7 @@
 #include "DspUtilities.h"
 #include "SurgeError.h"
 #include "SurgeStorage.h"
+#include "UserInteractions.h"
 #include <set>
 #include <numeric>
 #include <vt_dsp/vt_dsp_endian.h>
@@ -198,27 +199,12 @@ SurgeStorage::SurgeStorage()
 
    if (!snapshotloader.LoadFile(snapshotmenupath.c_str())) // load snapshots (& config-stuff)
    {
-#if MAC
-      SInt32 nRes = 0;
-      CFUserNotificationRef pDlg = NULL;
-      const void* keys[] = {kCFUserNotificationAlertHeaderKey, kCFUserNotificationAlertMessageKey};
-      const void* vals[] = {
-          CFSTR("Surge is not properly installed"),
-          CFSTR("Unable to open configuration.xml from ~/Library/Application Support/Surge")
-
-      };
-
-      CFDictionaryRef dict =
-          CFDictionaryCreate(0, keys, vals, sizeof(keys) / sizeof(*keys),
-                             &kCFTypeDictionaryKeyCallBacks, &kCFTypeDictionaryValueCallBacks);
-
-      pDlg = CFUserNotificationCreate(kCFAllocatorDefault, 0, kCFUserNotificationStopAlertLevel,
-                                      &nRes, dict);
-#elif __linux__
-      throw SurgeError("configuration.xml was not found from " + snapshotmenupath);
+      Surge::Error exc("Cannot find 'configuration.xml' in path '" + datapath + "'. Please reinstall surge.",
+                       "Surge is not properly installed.");
+#if __linux__
+      throw exc;
 #else
-      MessageBox(::GetActiveWindow(), "Surge is not properly installed. Please reinstall.",
-                 "Configuration not found", MB_OK | MB_ICONERROR);
+      Surge::UserInteractions::promptError(exc);
 #endif
    }
 
--- a/src/common/SurgeSynthesizerIO.cpp
+++ b/src/common/SurgeSynthesizerIO.cpp
@@ -14,6 +14,7 @@
 #endif
 #include <fstream>
 #include <iterator>
+#include "UserInteractions.h"
 
 namespace fs = std::experimental::filesystem;
 
@@ -187,26 +188,6 @@ string SurgeSynthesizer::getLegacyUserPatchDirectory()
 #endif
 }
 
-bool askIfShouldOverwrite()
-{
-#if MAC
-   CFOptionFlags responseFlags;
-   CFUserNotificationDisplayAlert(0, kCFUserNotificationPlainAlertLevel, 0, 0, 0,
-                                  CFSTR("Overwrite?"),
-                                  CFSTR("The file already exists, do you wish to overwrite it?"),
-                                  CFSTR("Yes"), CFSTR("No"), 0, &responseFlags);
-   if ((responseFlags & 0x3) != kCFUserNotificationDefaultResponse)
-      return false;
-#elif __linux__
-   printf("Implement me\n");
-#else
-   if (MessageBox(::GetActiveWindow(), "The file already exists, do you wish to overwrite it?",
-                  "Overwrite?", MB_YESNO | MB_ICONQUESTION) != IDYES)
-      return false;
-#endif
-
-   return true;
-}
 
 void SurgeSynthesizer::savePatch()
 {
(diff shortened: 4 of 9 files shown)


src/common/SurgeMessage.cpp
```diff
@@ -0,0 +1,99 @@
...
+                  << "------------------------------------------------------------\n"
+                  << url << "\n"
+                  << "============================================================\n"
+                  << std::flush;
+        return;
+
+    }
+};
+
+SurgeMessage *s_impl;
```
baconpaul: Re-reading my code now, this should be s_impl = NULL.

src/common/gui/SurgeGUIEditor.cpp
```diff
@@ -126,6 +120,8 @@ SurgeGUIEditor::SurgeGUIEditor(void* effect, SurgeSynthesizer* synth) : super(ef
 
    // ToolTipWnd = 0;
 
+   // Before we do much of anything else install our message handlers
```
baconpaul: This comment should go; I will just revert this file back to master since it ended up unchanged at the end of all my commits. Sorry!

jsakkine:
Maybe landed bit too fast.

jsakkine:
Ugh, I misread that this was already merged :) Can you hold with this for a while? I'm bit busy with my paid job right now but would like to give feedback. Now that my other changes are pushed I can continue working on with error dialog (namely VST2 rework to be more robust on errors).

baconpaul:
Oh yeah I have no intent of merging this without your feedback! This is too big a class to do unilaterally for sure.

I force pushed a new rev last night since I saw some things which I knew you would say when looking. But it isn't merged at all.

baconpaul:
And zero rush. I don't have any serious dev time for surge until Thursday anyway. Appreciate your review and good luck with your work!

src/common/SurgeError.h
```diff
@@ -5,10 +5,13 @@
 class SurgeError {
 public:
    SurgeError(const std::string &message);
+   SurgeError(const std::string &title, const std::string &msg);
```
jsakkine: You have an abbreviation `msg` here. Please just write it as `message` (same name as the member variable has). Also, the parameters are in wrong order.

I would define here:

```
static const std::string DEFAULT_TITLE ;

SurgeError(const std:;string& message, const std::string &title = DEFAULT_TITLE);
```
And declare in the CPP-file:

```
static const std::string SurgeError::DEFAULT_TITLE = "An error has occured";
```
Having two constructors for such a simple class is an overkill.

src/common/SurgeMessage.cpp
```diff
@@ -0,0 +1,99 @@
...
+#if MAC
+#include "MacSurgeMessage.h"
+#endif
+
+#if WINDOWS
+#include "WinSurgeMessage.h"
+#endif
+
+class stderrMessage : public SurgeMessage
+{
```
jsakkine: `class StdErrMessage`

src/common/SurgeError.h
```diff
@@ -5,10 +5,13 @@
 class SurgeError {
```
jsakkine: Lets rename this as just Error and put it to namespace Surge.

src/common/SurgeMessage.h
```diff
@@ -0,0 +1,55 @@
...
+**  or what not. SurgeMessage is guaranteed not to be null once statics are initialized.
+*/
+
+#pragma once
+
+#include <string>
+#include <atomic>
+#include "SurgeError.h"
+
+class SurgeMessage
```
jsakkine: The class name should be `ErrorMessage` (here also lets start putting stuff to Surge namespace instead of prefixing).

jsakkine: Obviously, the class name needs to be changed to something else. `Surge::Error` is the message object and this is something else.

jsakkine: Maybe it could be `ErrorReporter`. Nothing better comes to mind.

jsakkine:
OK, I've given my feedback to this version. Probably won't be the last review round. Just pointed out some bigger issues.

src/common/SurgeMessage.cpp
```diff
@@ -0,0 +1,99 @@
...
+void SurgeMessage::installPlatformHandler()
+{
+#if MAC
+    if( s_impl ) delete s_impl;
+    s_impl = new MacSurgeMessage();
+#elif WINDOWS
+    if( s_impl ) delete s_impl;
+    s_impl = new WinSurgeMessage();
+#else
+    std::cerr << "No impl of platform handler on this platform yet" << std::endl;
```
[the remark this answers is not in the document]
baconpaul: Oh right and it is stderr message. Just I print out that it isn’t. I will kill that. If you look in geti it uses stderr if nothing is set

src/common/SurgeError.h
baconpaul: Agree. And will do so

baconpaul:
So the o my comment I have a question about is your last one. I do already have a win and Mac subclass which are picked by premake. The only difference is do they implement an interface and get dropped on or do they just be copies of the same signature and get differentially included in the header. That is do we put the if mac in a cpp file then instsntiate or in a h file then include

I chose the route I did for three reasons

1. You can see the signature of the message thingy by looking st one header in common and ignoring the implementation 
2. We make sure all platforms implement the same signature (so avoid build breaks if you compile only on Mac or what not)
3. We preserve the ability to run this whole thing headless on Mac which I rather want and still see messages because we can have a headless host which doesn’t install the GUI handler

See what I mean? Like if I legitimately want the Mac ui version sometimes and the stderr version some other times in Mac I will be happy to have this structured like it is

Agree I should rename the windows and Mac classes to have GUI in their name of course.

As to the name of this class I think of it more as “user interactions” not just “errors” since it contains the suite of things we want to say to users. So I’m not sure what to call it. I agree surgemessage is not a great name. How about surge::userprompts or some such?

Apologies for capitalizations - on my phone

baconpaul:
Oh and thank you as always for the thoughtful review!

jsakkine:
In addition, this will reduce the amount of flagging in the CPP-files. Now you need to select either macOS or Windows class in the source code. With my suggestion you need to decide that only in premake5.lua.

baconpaul:
Oh I get it. Yeah that’s a fine pattern too. I can shift it to that easy. Will do so Thursday and push 

And you are right we don’t need a windows to Mac runtime switch ever!

Thanks

baconpaul:
Oh and “what would we call it”. It’s more than an error reporter. You ok with Surge::USerInteraction or some such

jsakkine:
Right now it is solely used for reporting errors. Can you describe what  else it would be used for? If it is used for something else, I don't think it should take SurgeError as parameter.

baconpaul:
Well except it isn’t! Right now it has 3 features. It reports an error, it asks a user an ok/cancel choice, and it opens a web page. I have not used the last in this sweep but will very shortly. 

Basically when you want to tell or ask or show a user something outside the surge GUI the function would be here, was my thought. Once we have it stable I will do a separate pr to walk through the code for the various message box and fprintfs to use this. 

I will probably also move the mini edit here to consolidate 

So I think of this as “tell ask or show the user something outside the surge GUI”

baconpaul:
And the report error function takes a surge error, but the promptokcancel function doesn’t! I think that’s right don’t you?

baconpaul:
In windows they are both implemented by ::messagebox

On Mac they are both implemented by cfuseralertwhatzit

So the os has chosen to make them route through a pipeline

This is basically a wrapper on those two functions if you look

Also every function is different and we group them into classes :) The rationale here is that these are a set of functions which tell ask or show a user something. Which is a common pattern. 

See what I mean?

baconpaul:
I suppose now we are killing an instance we could also just make them free functions in a namespace. Same as class at this point

jsakkine:
@baconpaul, I think this is enough feedback for one review round :) Better to do now one iteration (once you have time, no rush), and then look at what we have.

baconpaul:
In fact putting these as free functions in a surge::userinteraction namespace is totally fine with me if you prefer. Breaks the bad habit of using classes as a namespace. 

Just let me know surge or Surge and userinteraction userInteraction or UserInteraxtion? My default would be surge::userInteraction sinde it seems odd to capitalize a namespace

baconpaul:
Yeah agree. I rather like the namespaces free functions actually. So I’ll do that with your other diffs and force push and ask for a second look

Appreciate the care and input here. Code reviews are so important. Thanks!

jsakkine:
I think you should also split the changes to the error class to its own commit.

jsakkine:
I don't know how experienced git user you, are but an easy way to split is to:

```
git reset --soft HEAD~1 && git reset
git add -i
git commit # no -a here, in interactive add you select stuff to the staging area with option 5
git commit -a # rest of the stuff
```

jsakkine:
@baconpaul, yeah probably won't have time to review anyway before the weekend and I think we should not rush with this change.

baconpaul:
OK @jsakkine your comments made this code way way better.

I did everything we discussed, except split out the commit for Surge::Error. That's just too gnarly a git untangling at this point. I think it is fine, although agree in the future we could do that as a one-two

If you take a look at the diffs now you will see

1. No more class or inheritance
2. Everything is in a Surge::UserInteractions namespace
3. There's a separate compile-time-selected version for Mac Win and Lin
4. All of them compile and link - I tested all 3 platforms. Mac and Win work also (I still am not sure how to test if Lin works).

When you have a moment, if you could take a look. Having this feature would help me close a bunch of tiny little issues kicking around in a way that we don't splat the code with OS-dependent stuff.

Thanks again for the excellent review and comments. Appreciated!

jsakkine:
Can you squash the commits (and update the commit message if there is need). Cannot really go much into detail before that.

kzantow:
@jsakkine we can just enable force-squash for PRs. makes this stuff lots easier

baconpaul:
Anyway I rebased it. So now you can't see that I took all your advice; but you can see the result of the advice in a single commit!

Let me know what you think of the diffs. I'd love to sweep this before the weekend.

baconpaul:
@kzantow I squash merge some things; and don't squash merge others. Sort of depends on the commit stream.

baconpaul:
Thanks

I have some time to squash a few other bugs tomorrow that need this so I will rebase to it in those branches and adjust if you and up wanting an api change. Take your time of course!

jsakkine:
Now my Windows environment is working. I will have to do my paid work now and might not have time to look into Surge before tomorrow but this pull request is next on my priority queue.

baconpaul:
Cool! And sorry for applying every possible thumbs up smiley frowny thumbs down to that. I meant thumbs up and it is early :)

PR #307, opened 2019-01-15
Premake fixes

jsakkine:
This to help with the error reporting effort.


jsakkine:
Generally explicit listing is better as we can see what gets build and nothing gets pulled into compilation by mistake. And if a file is missing, premake5 will report. This was already seen useful with VST2.

baconpaul:
Ok I will take a look on Thursday 

I also agree explicit is safer, philosophically. You want your source enumerated explicitly and once I think!

jsakkine:
Yeah, and it gives nice overview what we have :) You can peek a the build file.

baconpaul:
Heya! I think you may have made these changes off of an old master. I just did a clean build with them on mac and got a linker error on CScalableBitmap (which is a file we swept in over the weekend)

Here's the diff I needed to build:

```
paul:~/dev/music/surge$ git diff premake5.lua 
diff --git a/premake5.lua b/premake5.lua
index 43b55dc..c619ac0 100644
--- a/premake5.lua
+++ b/premake5.lua
@@ -171,6 +171,7 @@ function plugincommon()
         "src/common/gui/CNumberField.cpp",
         "src/common/gui/COscillatorDisplay.cpp",
         "src/common/gui/CPatchBrowser.cpp",
+       "src/common/gui/CScalableBitmap.cpp",
         "src/common/gui/CSnapshotMenu.cpp",
         "src/common/gui/CSurgeSlider.cpp",
         "src/common/gui/CSurgeVuMeter.cpp",
```

I don't think I have write-back to your branch.

Thanks!

jsakkine:
Thanks! WIll fix.

jsakkine:
Updated.

jsakkine:
Can this be merged?

baconpaul:
It’s on my list for today to look at!

baconpaul:
OK checked it out. builds mac and win and looks correct.  Will merge right now.

At some point in the future we should also get the globs for mac, win, lin and the hosts. But that's trickier since those lists are re-used so we need to make a premake variable.

Issue #305, opened 2019-01-15
Zoom Revisited


baconpaul:
So now we are getting bugs squashed and linux building and scalable assets, the most requested feature is zoom. I did a sorta-hack version of zoom in the audio unit and that's in the builds now, but we want a zoom which works all hosts all platforms so I plan to rip out and re-implement the audio unit zoom. I'm going to use this ticket as a place to store notes to myself. I'm also going to tag this for the 1.6 milestone. And, when the time comes, I think I will need help on windows.

Right now we have

* Zoom working in fixed increments in the AU
* On the mac all hosts we use 2x assets if they are available to give us a retina display and scale

So we know that we can resize (in some cases) and we can scale properly in vstgui. Good.

So my plan of attach is

* Move the about button over and put a little widget on the bottom for drag
* when people drag it, rescale the window appropriately
* inform the host that the window has resized
* set the scale

If the drag-to-resize proves to hard I may also just do fixed sizes. I note that Roli Equator has drag to resize but the Arturia V6 collection instead has click scale to resize. So step one above might be "replace the about button with a menu button which contains about and zoom". I'll see as I go.

As far as informing hosts of resize here's my prelim notes to self.

* AU doesn't seem to have an API but all the hosts seem perfectly fine if you send a clean NSWindow setSize message
* VST2 has a member `audioeffectx::sizeWindow` which requests the host resize. I think to use it you have to respond positively to the canDo message "sizeWindow"
* VST3 has a method "sizeWindow" which similarly responds to the canDo message "sizeWindow" but also seems to have a method resizeView; not sure which is from the host and do the host.

Also [this article](http://www.willpirkle.com/forum/advanced-gui-api-programming-in-rackafx-v6-6/how-do-you-resize-vst-window/) seems to have loads of useful information.

I'll probably take a swing at some code on Thursday. If any of you read this and have things to add about how to code it up, please add them to this issue.

Finally the way I code it should work on windows if we can get the resources; but how to get this plethora of bitmap resources into the windows build is far trickier. Once I get it working reliably on mac I will probably need some help there.

kurasu:
VST3 has a IPlugViewContentScaleSupport interface with the following method

typedef float ScaleFactor;
virtual tresult PLUGIN_API setContentScaleFactor (ScaleFactor factor) = 0;

For VST2, Presonus has added some extensions which are also used by some other hosts, which you can get [http://www.presonussoftware.com/en_US/index.php?id=developer](here).

[...]

PR #310, opened 2019-01-15
Original Skin: Adding Menu button (new assets)

itsmedavep:
I have added a "Menu" button at the request of @baconpaul .

This button has been added as bmp00164.png to persist the naming convetion that is already established.


baconpaul:
Thanks @itsmedavep - I will either use this in #305 or I'll delete it. For now I will sweep it in so I have the option. Appreciate the quick render!

PR #311, opened 2019-01-15
tell user _what_ they are actually overwriting.

esaruoho:
this does not make any sense:
image 1

"Overwrite a patch with the same name?"
"A patch by this name already exists, do you wish to overwrite it?"
however does kinda make more sense. maybe.

### DIFF
--- a/src/common/SurgeSynthesizerIO.cpp
+++ b/src/common/SurgeSynthesizerIO.cpp
@@ -192,15 +192,15 @@ bool askIfShouldOverwrite()
 #if MAC
    CFOptionFlags responseFlags;
    CFUserNotificationDisplayAlert(0, kCFUserNotificationPlainAlertLevel, 0, 0, 0,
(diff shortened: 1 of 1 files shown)


jsakkine:
LGTM

itsmedavep:
More broadly, isn't this the difference between save and save as? Save would overwrite. Perhaps you can warn in that case but I certainly wouldn't expect it. 

Save as, allows for rename or in the case of no name change prompt the user about what is being overwritten. 

Regardless, the point you raise is a good one. We should be explicit about what is going to get overwritten.

baconpaul:
I have this fixed in my big user feedback patch. Mind if we close it u til I sweep that to avoid merge nastiness?

esaruoho:
aww. well, okay

baconpaul:
My version actually gives you the patch and category name!! Coming soon I promise

PR #312, opened 2019-01-16
Fix wavetable browsing

jsakkine:
I'll do similar stateless functions for patches and categories to SurgeStorage. Makes sense to keep this logic there IMHO.

### DIFF
--- a/src/common/SurgeStorage.cpp
+++ b/src/common/SurgeStorage.cpp
@@ -393,18 +393,6 @@ void SurgeStorage::refresh_wtlist()
       errorbox("File IO Error: Couldn't locate wavetables on disk!\n\nPlease reinstall..");
    }
 
-   wtOrdering = std::vector<int>(wt_list.size());
-   std::iota(wtOrdering.begin(), wtOrdering.end(), 0);
-
-   auto wtCompare =
-      [this](const int &i1, const int &i2) -> bool
-      {
-         return _stricmp(wt_list[i1].name.c_str(),
-                         wt_list[i2].name.c_str()) < 0;
-      };
-
-   std::sort(wtOrdering.begin(), wtOrdering.end(), wtCompare);
-
    wtCategoryOrdering = std::vector<int>(wt_category.size());
    std::iota(wtCategoryOrdering.begin(), wtCategoryOrdering.end(), 0);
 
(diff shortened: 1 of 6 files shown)


jsakkine:
In the second patch I should probably check whether getNextWaveTable return -1 before assigning it. Otherwise, I think this should be OK.

jsakkine:
Updated.

baconpaul:
OK!

While you are in the code I just built it. Builds and runs fine but I noticed a behavior I'm not sure is right.

If you are the first waveform of a subcategory (say "sampled/banjo1") and you press "down" do you want to go to the last waveform of the preceding category (say "rhythmic/woolform") or do you want to stay in the category?

With patch you stay in category since category and patch have separate buttons but on wavetable there's only one button. So I think the "wavetable category wt id == 0" decrement operation should go to "prior category wt id == last". So the buttons really are running you through the entire wt library.

See what I mean?

I'm happy to make this change subsequent to yours if you want but I think it is just a tiny little more complexity in getNextWavetable.

baconpaul:
I built it before your pushes that is.

baconpaul:
Oh and I reviewed all the code and it looks great to me! Just the question about that behavior being slightly different for WT and Cat.

jsakkine:
@baconpaul, it is a regression. Should work in the same way. Also I think better name for stateless browsing functions in SurgeStorage would be getAdjacentPatch(), getAdjacentPatchCategory() and getAdjacentWavetable().

jsakkine:
@esaruoho, @baconpaul: so should wavetable browsing stay within a category or go through all categories? I'm not sure which one is right.

esaruoho:
@jsakkine my 2c: go through all categories
preferably both presets and wavetables would go through all categories

jsakkine:
OK, then I think it would make sense to tweak a bit how stuff gets ordered:

1. Sort by category first.
2. Then create order for wavetables such that wavetables in each category is sorted separately.

This kind of rework would simplify and speed up also patch browsing code (no need to have that do while loop that I have now). @baconpaul, what do you think? Not hard to implement. Just have to tweak precalculations for order. Would make most of things O(1) instead of O(n).

baconpaul:
Yeah I'm fine with that.

My answer was going to be

1. @esaruoho answer or, alternate
2. Add new UI widgets like << and >> next to the current < > and to browse category separately from wavetable and then show the category name (so show sampled/sine rather than sine) in the UI

Then I was going to say "but 2 is a new feature so the correct answer for now and for v1.6 is #1 but maybe we should consider a feature tagged ticket for #2"

So sounds like we agree! 👍 

As to the sort, I think that's - you are just changing the .order and the sort index though right not the underlying ID? I don't know if the IDs are streamed.

And yeah the 'Adjacent' name is much better. I like that improvement.

Thanks!

jsakkine:
@baconpaul, the same idea as how I sort category groups. I sort wavetables per category.

baconpaul:
Yeah got it that will be perfect then just change category 0 to global 0 in your jog function and it will be great. Thanks!

jsakkine:
Amazes me how tricky issue this sorting has been in the end :)

jsakkine:
Untested drafted code (even uncompiled) but this how I plan to lay out wavetables:

```
   // Sort wavetables per category in the category order.
   for (auto c : wtCategoryOrdering) {
      int start = wtOrdering.size();

      for (int i = 0; i < wt_list.size(); i++)
         if (wt_list[i].category == c) {
            wtOrdering.push_back(i);

      int end = wtOrdering.size();

      std::sort(std::next(wtOrdering.begin(), start),
                std::next(wtOrdering.begin(), end), wtCompare);
   }
```

We can  consider similar approach later on for patches and categories but since there the UI at least works properly I don't rush to break working code yet. I'll focus this PR purely on wavetables so that we can finally close the issue.

jsakkine:
@esaruoho, okay, what about this? Now it goes through the whole list.

esaruoho:
checking

jsakkine:
OK, so almost there. Almost patched should be browsed in the same way? Now it works for wavetables, it is easy fix but just sanity checking.

jsakkine:
Can we anyway close this PR since this is for wavetable browsing if the current works and I will do after that a PR for patch browsing that makes similar change there?

esaruoho:
LGTM, just tested with Renoise VST2, bonus was that you can just hold down a bunch of notes and keep cycling through wavetables.
Logic AU works too.

jsakkine:
Similar fix for patches is now merely copy-pasting.

baconpaul:
I don't think we should do the change for patches. See #314

But this one is correct and I tested it just now on windows, so I'm merging.

Issue #319, opened 2019-01-17
Gui Crashes from "init" patch when pressing "category" - or +


sense-amr:
Gui Crashes from "init" patch when pressing "category" - or +
in latest build 

![image](https://user-images.githubusercontent.com/1504019/51318137-92702c00-1aad-11e9-9632-47f9abb5c4d9.png)

[Surge_categoryCrashGui.zip](https://github.com/surge-synthesizer/surge/files/2768540/Surge_categoryCrashGui.zip)

1. Load plugin on midi channel
2. Open Gui 
3. init is selected ..
4. Press category - or + 
5. Crash gui/Plugin Crash in host

 - OS: windows 10
 - Host Ableton Live, Bitwig,Fruity Loops
 
Video provided of crash in bitwig .. 

![image 2](https://user-images.githubusercontent.com/1504019/51317784-98194200-1aac-11e9-8695-b74991a810ee.png)

Tested vst2 in Bitwig 1.3.16 /Ableton 9.7.4
Tested vst3 in Fruity Loops 20.0.1 Build 455

Note: In Bitwig 1.3.16 the plugin window and gui crashes only.. and can be re enabled from the plugin container.. 

In Ableton Live 9.7.4 the host crashes entirely.
In Fruity Loops 20.0.1 build 455 the host crashes entirely.

baconpaul:
Found the problem and have a fix coming.

sense-amr:
Excellent @baconpaul i'm glad to know it IS a legit problem .. :) cheers..

baconpaul:
Ok I found the edge case my fix didn’t fix and have a fix for that. But am away from computer now. Will do a pr and test this afternoon

PR #321, opened 2019-01-17
Handle out-of-range Patch.order

baconpaul:
One patch - init - has a .order which isn't in range. Others may.
This results in +/_ browser sometimes going out of range on
an array and segfaulting. So be defensive in our beyond-the-edge
checkes. Addresses #319


baconpaul:
this is incomplete - closing

PR #323, opened 2019-01-17
Handle out-of-range Patch.order

baconpaul:
One patch - init - has a category and id  which isn't in range. Others may.
This results in +/_ browser sometimes going out of range on
an array and segfaulting. So be defensive in our beyond-the-edge
checkes. Addresses #319


itsmedavep:
I took a look at this and it seems to be OK now.

Steps to attempt repro:

Load up AU
Right click patchbrowser<--Init <--Cycled through all patches in the Init category using  -/+ buttons
No crash



crash-fix

PR #325, opened 2019-01-17
Cleanup platform specific user prompts

baconpaul:
Now we have Surge::UserInteractions, deploy it to a few places
where we can use it to clean up some old pllatform specific
messages.

Pretty innocuous. Would appreciate a quick eyeball if someone wants to.

### DIFF
--- a/src/common/SurgeStorage.cpp
+++ b/src/common/SurgeStorage.cpp
@@ -376,7 +376,9 @@ void SurgeStorage::refresh_wtlist()
 
    if (wt_category.size() < 1 && wt_list.size() < 1)
    {
-      errorbox("File IO Error: Couldn't locate wavetables on disk!\n\nPlease reinstall..");
+       Surge::UserInteractions::promptError(std::string("Surge was unable to load wavetables from disk. ") +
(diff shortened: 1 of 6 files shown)


src/common/SurgeStorage.cpp
```diff
@@ -376,7 +376,9 @@ void SurgeStorage::refresh_wtlist()
 
    if (wt_category.size() < 1 && wt_list.size() < 1)
    {
-      errorbox("File IO Error: Couldn't locate wavetables on disk!\n\nPlease reinstall..");
+       Surge::UserInteractions::promptError(std::string("Surge was unable to load wavetables from disk. ") +
```
jsakkine: "the wavetables"?

jsakkine: OK,  I guess you're right.

baconpaul: Well English is so weird and ambiguous it’s not right or wrong, just a matter of style.  But the article certainly seems to add an emphasis which sounds a bit over specific, 

Anyway I’ll merge this since this was your only comment. Thanks for looking!

Issue #282, opened 2019-01-12
User can save an unnamed (empty) preset and it will be saved to Category/.fxp


esaruoho:
currently user is able to save an unnamed preset and if it is saved into a new category, the category-folder will get created, and then there'll be a `.fxp` file in there.

user can then load the preset, too.
<img width="268" alt="fullscreen_12_01_2019__18_43" src="https://user-images.githubusercontent.com/4966687/51075860-fce23000-1699-11e9-986a-2098417737d1.png">

baconpaul:
Well at least you can load it!

baconpaul:
This one needs #217 resolved first. My plan is to pop an error box saying "patch not saved since name was blank" or some such. But I would love that API to show an error box to user modally in code. (This is not a case where we can throw).

@jsakkine just flagging your attention to this also since I know you are working on the show-message stuff on linux.

Leaving this sit for a bit.

baconpaul:
This is now fixed.

PR #326, opened 2019-01-17
Dont allow blank patch names or categories

baconpaul:
Prompt a user error in this case rather than making blank named files
Fixes the error reported in #282 by not allowing the user to enter that
code path.

Eyeball or test appreciated. Thanks!

### DIFF
--- a/src/common/gui/SurgeGUIEditor.cpp
+++ b/src/common/gui/SurgeGUIEditor.cpp
@@ -19,6 +19,7 @@
 #include "vstcontrols.h"
 #include "SurgeBitmaps.h"
 #include "CNumberField.h"
+#include "UserInteractions.h"
 
 #if TARGET_AUDIOUNIT
 #include "aulayer.h"
@@ -2127,11 +2128,35 @@ void SurgeGUIEditor::valueChanged(CControl* control)
       // frame->setModalView(nullptr);
       frame->setDirty();
 
-      synth->storage.getPatch().name = patchName->getText();
-      synth->storage.getPatch().author = patchCreator->getText();
-      synth->storage.getPatch().category = patchCategory->getText();
-      synth->storage.getPatch().comment = patchComment->getText();
-      synth->savePatch();
+      /*
+      ** Don't allow a blank patch
+      */
+      std::string whatIsBlank = "";
+      bool haveBlanks = false;
+
+      if (! Surge::Storage::isValidName(patchName->getText().getString()))
+      {
+          whatIsBlank = "name"; haveBlanks = true;
+      }
+      if (! Surge::Storage::isValidName(patchCategory->getText().getString()))
+      {
+          whatIsBlank = whatIsBlank + (haveBlanks? " and category" : "category"); haveBlanks = true;
+      }
+      if (haveBlanks)
+      {
+          Surge::UserInteractions::promptError(std::string("Unable to store a patch due to invalid ") +
+                                               whatIsBlank + ". Please save again and provide a complete " +
+                                               whatIsBlank + ".",
+                                               "Error saving patch");
+      }
+      else
+      {
+          synth->storage.getPatch().name = patchName->getText();
+          synth->storage.getPatch().author = patchCreator->getText();
+          synth->storage.getPatch().category = patchCategory->getText();
+          synth->storage.getPatch().comment = patchComment->getText();
+          synth->savePatch();
+      }
    }
    break;
    default:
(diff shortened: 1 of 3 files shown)


src/common/gui/SurgeGUIEditor.cpp
```diff
@@ -2127,11 +2128,35 @@ void SurgeGUIEditor::valueChanged(CControl* control)
...
-      synth->storage.getPatch().category = patchCategory->getText();
-      synth->storage.getPatch().comment = patchComment->getText();
-      synth->savePatch();
+      /*
+      ** Don't allow a blank patch
+      */
+      std::string whatIsBlank = "";
+      bool haveBlanks = false;
+
+      if (std::string(patchName->getText()).find_first_not_of( ' ' ) == std::string::npos)
```
jsakkine: What about a tab character or something even weirder (from the ASCII range)? I think you should validate that a patch

- the only allowed non-displayable character is a space bar.
- has at least one displayable character.
- is at least one character length.

I think these are fair assumptions for a patch name.

src/common/gui/SurgeGUIEditor.cpp
```diff
@@ -2127,11 +2128,35 @@ void SurgeGUIEditor::valueChanged(CControl* control)
...
+      {
+          whatIsBlank = "name"; haveBlanks = true;
+      }
+      if (std::string(patchCategory->getText()).find_first_not_of( ' ' ) == std::string::npos)
+      {
+          whatIsBlank = whatIsBlank + (haveBlanks? " and category" : "category"); haveBlanks = true;
+      }
+      if (haveBlanks)
+      {
+          Surge::UserInteractions::promptError(std::string("Unable to store a patch with a blank ") +
```
jsakkine: Personal preference is `std::string() +` but not big issue for me.

src/common/gui/SurgeGUIEditor.cpp
jsakkine: I think doing this C-style instead of string class magic functions documents actually best how a patch name is validated and probably performs best too.

src/common/gui/SurgeGUIEditor.cpp
```diff
@@ -2127,11 +2128,35 @@ void SurgeGUIEditor::valueChanged(CControl* control)
...
+      ** Don't allow a blank patch
+      */
+      std::string whatIsBlank = "";
+      bool haveBlanks = false;
+
+      if (std::string(patchName->getText()).find_first_not_of( ' ' ) == std::string::npos)
+      {
+          whatIsBlank = "name"; haveBlanks = true;
+      }
+      if (std::string(patchCategory->getText()).find_first_not_of( ' ' ) == std::string::npos)
```
jsakkine: `validatePatchName()` could be called in both of these call sites. 

I would also use it when we scanning all the patches in SurgeStorage, but that is up to you. I would just skip the patches that have invalid filenames, possibly giving one promptError() after scanning them that reports the number of patches that were skip because of this reason. Allowing to load such patches brings it own set of corner cases, as you can probably understand.

jsakkine:
@baconpaul, glad to help :)

baconpaul:
So @jsakkine I really I want to make a function with name Surge::Storage::isValidPatchOrCategoryName since SurgeStorage is the representation of a patch used in all parts of the codebase.

So do I 

1. Put this namespace at the bottom of SurgeStorage.h
    * pros - it is code all in the same place
    * cons - it is just a namespace at the end
2. Put this namespace in a new header
    * pros - namespace stuff is distinct
    * cons - functions on storage are spread all around
3. Don't namespace this and make it a member of SurgeStorage
    * pros - the most like current code
    * cons - that's probably not true in the future and the future starts today

Your opinion welcome. I'm going to do 1 now but not force it until I hear from you

jsakkine:
I would name the function as isValidEntity() and put it to Surge::Storage namespace. I would also move the class to that namespace and rename it as Storage.

baconpaul:
Ha ha! IsValidThingWhichWeCanUseAsAStringForManyPurposesInSurge

I will do IsValidStorageName

I won't move the SurgeStorage class in this PR, but agree we should do it shortly.

jsakkine:
Why not just isValidName()? It is anyway in Storage namespace.

baconpaul:
Yeah OK. It feels weird but not enough that I can push back so I'll pick that!

baconpaul:
Thank you automated builds! Checking windows now.

Issue #181, opened 2018-12-29
Waveshaper asymmetric/ASSYM-mode glitches with (at least) two patches


esaruoho:
`Pluck (fast) -> Molusk`
`Polysynth -> FM Poly 1`

both use ASSYM -mode, which causes glitches. 

`FM Poly 1`: just hold down one note, then play another note. audible click/pop.
`Molusk`: Either hold down a note and play other notes quickly, or just play notes quickly, you'll hear an audible click/pop or resonance glitch or wotever.

original description for `Molusk`:
```
Getting some glitching: patch "Molusk" in "Pluck (fast)". 
...Playing a few notes fairly rapidly (8th notes) glitches out 
after about 9 notes (though the count varies).
It's the waveshaper. If you turn it off, no more glitching.
(The patch has it on Asymmetric, FWIW)
And, yup, if I turn other patches on ASSYM, causes glitching.
(The level of the waveshaper drive doesn't seem to make any difference.)
All of the other modes seem to behave just fine.
```

and for `FM Poly 1`
```
Select preset "Polysynth - FM Poly 1" and play just two notes legato - 
second one will glitch.
```

esaruoho:
another `ASSYM` bug:

Load `Polysynth - FM Poly 1` and play and hold down one note, then play another note. the other note will click/pop.

baconpaul:
These are great. Do we know platform version build number daw etc for the user? 

Clearly that “build info in about” will matter as people start using this.

esaruoho:
@baconpaul @getdunne reported the FM Poly 1 bug on Win32. I have seen both patches bug on macOS 64-bit AudioUnit. so this isn't a platform version or 32bit/64bit version.

esaruoho:
newest build from `master` shows these behaviours, best fastest example is `Polysynth>FM Poly 1`

VST2: Renoise
AU: Ableton Live
AU: Logic Pro X

kzantow:
Thanks for tracking this down to Surge-wide @esaruoho I'll see if I can scrape some time to figure out what's problematic with this stuff after getting all the installer (#180) stuff worked out

baconpaul:
OK I have a partial fix for this which also explains why VST2 and AU are broken and VST3 isn't.  Hold on for me to assemble it all together for at least AU and then you can do same with VST2, even though it's not a complete fix.

baconpaul:
Alright so there's a compound problem here. 

[...]

PR #330, opened 2019-01-18
Consolidate CPU detection to SurgeSynthesizer constructor.

jsakkine:
The only thing that initDllGlobals() does in addition to what
initCpuArchitecture() does, is to print an error message when either SSE2
or CMOV is missing. This commit rips of initDllGlobals() and moves the user
interaction inside initCpuArchitecture().


baconpaul:
So. When I made it so this was consistently called I for a while went this route (called initCpuArchitecture in the plugins rather than initDllGlobals) and then before I did the PR went back to initDllGlobals.

The reason I did that is, for another patch I was working on (which I since abandoned) I also had a function I wanted called cross host at startup reliably. Although as you'll see from the comments I added I hated the name. I think (but am not 100% sure) we will need that call point in the future.

So imagine if we had a second initCpuArchitecture type thing we need to call; or imagine if we had 2 or 3 more host styles (which in standalone modes we would).

So sure lets kill abstract synthesizer but I would really rather have a function called something like Surge::Globals::initialize() which for now only calls initiCpuArchitecture and then reports on the result.

Whether or not we should do user feedback on the architecture state when we set it or outside is a matter of taste. I sort of like it outside - especially if we have a Surge::Globals::initialize

But I consciously choose to not go straight-to-raw and I think that's a choice we will be happy with in the future.

Welcome your thoughts!

jsakkine:
It should be in the SurgeSynthesizers constructor. That is what requires these features. I don't really see why it should be global.

jsakkine:
This is in-between PR going to that direction. After this I'll add similar robustness to other plug-ins (can deal with the init failure). After all that is completed, I'm planning to move the detection code to the synth's constructor. I've written a few lines about this to Slack.

jsakkine:
Cannot be done with a single PR.

jsakkine:
Proposal for getting this PR to .somewhere:

- I will move the CPU detection code to SurgeSynthesizer constructor.
- Not going to throw an exception there, just print the error like initDllGobals does now.

This would not change anything for plugins but things would consolidated where they belong. Once  all plugins have proper exception handling we can make it throw as an exception but right now the UserInteraction would be done in the constructor without exception thrown.

jsakkine:
I will close this now and create a fresh one later.

baconpaul:
All sounds good. Like I said on slack test case is the patches in #181 still work on Mac all hosts. If they do you have it correctly ported. Thanks!!

PR #334, opened 2019-01-18
Capitalize max_voices

jsakkine:
Makes code hard to read when you don't know whether max_voices is a
constant or not, especially since it is used in the audio code, which
is complex to begin with.


baconpaul:
Great

PR #335, opened 2019-01-18
Migrate window size constants to globals.h

jsakkine:
Window size constants are essentially global constants. Thus, it makes
sense to place them in globals.h.


baconpaul:
This one is fine for 1.6; in >1.6 this will come from an SVG file I bet so we need to change it anyway. But happy to merge this now.

Issue #199, opened 2019-01-04
Drag n Drop a Wav into Oscillator


baconpaul:
@esaruoho had asked if you can drag and drop a sample into an oscillator window.

COscillatorDisplay.cpp has this code

```
bool COscillatorDisplay::onDrop(IDataPackage* drag, const CPoint& where)
{
   /*long size,type;
   char *filename = (char*)drag->first(size,type);
   //while(filename)
   if(filename)
   {
           if(type == CDragContainer::kFile)
           {
                   strncpy(oscdata->wt.queue_filename, filename, 255);
           }
//		filename = (char*)drag->next(size,type);
   }*/
   return true;
}
```

which looks like there was a start at an attempt at that. Don't think this is a release critical issue but a bit of work to pick up I stumbled across so wanted to drop it in an issue.

baconpaul:
So I got the VSTGUI 4 drag n drop API in place and on my mac at least, I now get a filename printed to stderr when you drop any file onto the oscillator.

baconpaul branch drag-n-drop-199

Issue #198, opened 2019-01-04
wav file loading on mac and linux


baconpaul:
Right now, .wav file (in SurgeStorageLoadWav.cpp) is implemented for windows only. We should fix that I guess!

esaruoho:
Aha!  So this is what the KVR ppl were talking about. When i asked for Details all i got was silence. Incredibly frustrating.

baconpaul:
Yeah; this is separate from the drag-and-drop stuff (which is cross-platform disabled) but basically when you load_wt, you check if it is a .wt file, and then if you aren't mac you and you are a .wav you call load_wt_wav.

so this means behavior is

* windows: a .wav file loads
* mac: a .wav file doesn't load and nothing happens because you never call load_wt_wav in SurgeStorage
* linux: a .wav file doesn't load but you do get an fprintf to stderr, because you do call load_wt_wav in SurgeStorage, but you don't meet the `#if WINDOWS` gate.

I guess if someone were to attack this issue you'd just sort of need to either (1) like some external library for reading wav files or (2) replicate the stuff in the windows block to not use the windows specific mmioOpen API and use instead uhh I guess fread? or some mem mapped file read?

Looks like "actual unfun work" tbh.

baconpaul:
For the person who does it on a mac, even if it ends up being me, I would use https://developer.apple.com/documentation/audiotoolbox/audio_file_services and then you can read basically anything and don't have to write all that code.

On linux I guess you could use libsndfile

baconpaul:
baconpaul branch mac-wav-file-198 is a healthy start; it loads a wav file and reads the bits. It just can't seem to push the bits into the wavetable data structure yet. So plugging away here.

baconpaul:
<img width="474" alt="screen shot 2019-01-05 at 11 40 51 am" src="https://user-images.githubusercontent.com/13387561/50726699-e1e24f80-10de-11e9-8c46-f2541c2ee591.png">

OK so I can read a wav file and create wavetables. I am not quite sure what the expected behavior is though so need to read some more code.

@esaruoho the users who complained about not being able to load wav? Did they provide a few sample wav files for us to try?

esaruoho:
@baconpaul they just wrote that they were not able to drag a WAV in. Does the WAV (in your screenshot) that you dragged in play in the plugin? Can i try it?-)

baconpaul:
[...]

PR #336, opened 2019-01-18
Implement .wt drag-n-drop

baconpaul:
Update the drag-n-drop API to be VST4 compatible for a single
file dropped on the Oscillator.

Support drag and drop of .wt files properly

Support error messages for non .wt files and non file drops.


baconpaul:
This will close #199 when swept (for a .wt; wav files are a separate PR on #198)

baconpaul:
Oh and I confirmed it works Mac and windows and @esaruoho confirmed Mac works also.

Issue #214, opened 2019-01-05
Use a custom and uniform font


baconpaul:
The design folks would love to use the open font @cyanit @itsmedavep 
https://fonts.google.com/specimen/Lato

To use it reliably in draws we would want to include it in the resources so we can load it in app without requiring it installed on the underlying system. That’s platform dependent.

It also may or may be something we get done. But lets put an issue in place since it would make painted labels more consistent with vector graphics and stuff if we use a scalable vector thingy.

On windows and linux I haven’t googled yet. On Mac I found this:

ATSApplicationFontsPath (String - macOS) identifies the location of a font file or directory of fonts in the bundle’s Resources directory. If present, macOS activates the fonts at the specified path for use by the bundled app. The fonts are activated only for the bundled app and not for the system as a whole. The path itself should be specified as a relative directory of the bundle’s Resources directory. For example, if a directory of fonts was at the path /Applications/MyApp.app/Contents/Resources/Stuff/MyFonts/, you should specify the string Stuff/MyFonts/ for the value of this key.

jsakkine:
Just a question (haven't studied sources) but how does the text rendering work at the moment? VSTGUI stuff uses its own path and stuff on screen other one?

kurasu:
Lato is perfect, its very pretty and and renders digits at constant size so its very well suited to numerical diplays. 

@jsakkine 
I'd assume VSTGUI just uses the platform font rendering, nothing custom. We don't want a fixed-size font for stylistic reasons, and any optimizations needed can be done at a string-level instead of character level. But since we're hardly doing any text rendering at real-time intervals anyway (apart form the text-popup on mouse-drag) I don't see how this sholud ever be an issue.

baconpaul:
yeah I think there's really no cpu benefit from a fixed width font nowadays and the UI is all double buffered properly anyway. 

@kurasu yup the vstgui uses the installed platform font tools in the PlatformFont class underneath CFont it seems. I took a quick tour after I opened this issue. Right now the 3 fonts we use in surge seem to be Lucida Grande (mac), MS Sans Serif and Arial (windows); and the "system small font" about screen.

But this issue is really about "how do we include a ttf file with our bundle which is loadable".

jsakkine:
OK, cool. Where is the text/number rendering code located for on-screen display?

baconpaul:
It’s all in the vstgui lib. Look in src for setFont and you will find the draw command. Pull from there

[...]

PR #338, opened 2019-01-18
Align text in orange boxes

baconpaul:
This is not a pretty diff. But it is in the middle of some code which
is very explicit about layout and the layout was wrong. As shown in
issue #295, the combination of graphics, box placement, and label on the
lfo selector were just a bit off. This makes it almost not a bit off.
Ideally #214 will save the day in the future by letting us have a uniform
font across all platforms.


esaruoho:
testing this with @itsmedavep 's vector surge look:
renoiseVST2
fullscreen_19_01_2019__20_31
LogicAU
logicorangebox

Issue #350, opened 2019-01-20
CpuArchitecture problems


jarkkojs:
Only QuadFilterUnit and WindowOscillator check CaSSE2. None of the code base uses CaCMOV to check anything. We can conclude from this that the flags are useless because in practice they are not used at all.

CMOV instructions were introduced in 1995 with P6 microarchitecture (from Pentium Pro to Pentium III) and SSE2 was introduced in 2000 with P68 microarchitecture (Pentium 4). I think the right thing to do is to remove the checks. Even the most conservative studio environments have newer hardware. Most of the operating systems would not even boot properly with hardware older than Pentium 4.

PR #351, opened 2019-01-20
Remove CpuArchitecture

jarkkojs:
Only QuadFilterUnit and WindowOscillator check CaSSE2. None of the code
base uses CaCMOV to check anything. We can conclude from this that the
flags are useless because in practice they are not used at all.

CMOV instructions were introduced in 1995 with P6 microarchitecture (from
Pentium Pro to Pentium III) and SSE2 was introduced in 2000 with P68
microarchitecture (Pentium 4). I think the right thing to do is to remove
the checks. Even the most conservative studio environments have newer
hardware. Most of the operating systems would not even boot properly with
hardware older than Pentium 4.

Fixes: #350 ("CpuArchitecture problems")


jsakkine:
It would fair, tho, to add to the documentation that Surge requires at least Pentium 4.

jsakkine:
@kurasu, Do you see this as a fair conclusion?

jsakkine:
I did not remove all the SSE1 in this PR because wanted to keep the diff as small as possible for easy review.

jsakkine:
Oops, forgot to `git commit --amend` part of the changes. Now should be good.

jsakkine:
Tested on macOS (used WindowsOscillator):

- VST2: Ableton 10
- AU: Ableton 10
- VST3: Reaper. I could load the plugin and switch to window oscillator  but I don't know how to play synth with Reaper. Maybe I'll try out Bitwig demo next time if the demo version allows to load VST3.

On Windows 10, I tested the VST2 plugin with OpenMPT.

Finally have quite decent coverage when it comes to testing.

jsakkine:
With 64-bit OS, you cannot have x86 processor that does not have CMOV and SSE2. If you are running 64-bit OS, you are guaranteed to have those features. And yeah, AMD has also supported SSE2 since 2003.

jsakkine:
Tested also with Bitwig Studio 2.4 demo the VST3 plugin by creating a MIDI pattern with Surge using Window oscillator.

baconpaul:
There was that poly patch which broke if CPU architecture was wrong, in #181 - did you test that particular patch doesn't glitch on second key? Super useful to make sure you got it all set properly.

baconpaul:
I like this change, by the way; if we can just confirm #181 patch works AU mac I'd be happy to sweep it.

Ok back to zoom!

jsakkine:
That should be OK since now ASYM_SSE2 is unconditionally returned in the aformentioned case.

jsakkine:
But I'll also sanity check with that patch!

baconpaul:
Yes from reading the code I am almost 100% sure it will work correctly

baconpaul:
Hey @jsakkine I had a minute free so built your patch on mac and can confirm the #181 patches all work.

Will just take a quick look on windows then sweep this in.

With the free 10 minutes you now have, though, if you could review my zoom PR, I would love it!

Thanks!

baconpaul:
OK builds windows and plays PolySynth / FM Poly 1 no problem there too.

And looks great. Thank you. Merging.

Issue #348, opened 2019-01-19
Re-implement zoom on AU using menu, better stops, retina


baconpaul:
Next step on my path to zoom is to take the current AU zoom +/- and replace it with a zoom menu. At the same time, stage the vector assets fully and use different size bitmaps based on different zoom levels.

Again all AU only for now. And the setting won't be remembered session to session. Those are future issues on my path to zoom.

baconpaul:
Oh and if along the way I can figure out how logic pro shows those sizes in the standard host control I will do that too. But no promises on that one.

baconpaul:
Tagging this to #305

PR #355, opened 2019-01-20
Smart bitmap selection on Zoom

baconpaul:
This change basically redoes zoom in the AU host to be way better in
two key ways

1: Moves it to a menu with more options than a simple +/-
2: Makes CScalableBitmap chose the best size bitmap for our display
setting

This is one diff on the path to zoom. Things like windows, non au hosts,
and detection of actualy screen resolution are still undone here. But
it is one I would like to sweep.

All these features remain off unless you consciously build mac with
vector assets.

### DIFF
--- a/src/common/gui/CScalableBitmap.cpp
+++ b/src/common/gui/CScalableBitmap.cpp
@@ -3,6 +3,13 @@
 #include <CoreFoundation/CoreFoundation.h>
 #endif
 
+// Remember this is user zoom * display zoom. See comment in CScalableBitmap.h
+int  CScalableBitmap::currentPhysicalZoomFactor = 100;
+void CScalableBitmap::setPhysicalZoomFactor(int zoomFactor)
+{
+    currentPhysicalZoomFactor = zoomFactor;
+}
+
 CScalableBitmap::CScalableBitmap(CResourceDescription desc) : CBitmap(desc)
 {
     int id = 0;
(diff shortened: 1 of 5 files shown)


baconpaul:
Oh and sweeping this would close #348

src/common/gui/CScalableBitmap.cpp
```diff
@@ -3,6 +3,15 @@
 #include <CoreFoundation/CoreFoundation.h>
 #endif
 
+/*
+** This is not the user zoom factor; this is the all in user + display zoom factor
+*/
```
jsakkine: Not sure if I understand this comment and maybe it should be with `//`? A very basic question: what are user and display zoom factors and what is this global variable?

src/common/gui/CScalableBitmap.cpp
```diff
@@ -40,17 +52,36 @@ CScalableBitmap::CScalableBitmap(CResourceDescription desc) : CBitmap(desc)
...
+            ns = scales.back();
+        }
+        bestFitScaleGroup = ns;
+        lastSeenZoom = currentPhysicalZoomFactor;
+    }
     /*
     ** For now this is a fixed 'retina' style draw of using 2x bitmaps at 0.5 scale if they exist
     */
-    if (scaledBitmaps[ 200 ] != NULL)
+    if (scaledBitmaps[ bestFitScaleGroup ] != NULL)
```
jsakkine: A nitpick: `scaledBitmaps[ bestFitScaleGroup ]`

jsakkine:
Kind of generic question but can we get to the point where you can scale the Window freely (in the correct w/h ratio) and zoom factor is set accordinly?

jsakkine:
That is how TAL's plugins, Serum and whatnot works, and I believe that is what people would want to see.

src/common/gui/CScalableBitmap.cpp
baconpaul: User zoom factors are things like "1 want this twice as large"
Display zoom factors are things like "My display has 4 physical pixels per logical pixel" (so like retina displays)

This is a static because I don't have access to every instance of a bitmap; it's a global state of the world. But it is a global because I was dumb and should have made it a private static member of CScalableBitmap which I'll do.

src/common/gui/CScalableBitmap.cpp
baconpaul: Isn't that what it says?

src/common/gui/CScalableBitmap.cpp
```diff
@@ -40,17 +52,36 @@ CScalableBitmap::CScalableBitmap(CResourceDescription desc) : CBitmap(desc)
...
 
+
 void CScalableBitmap::draw (CDrawContext* context, const CRect& rect, const CPoint& offset, float alpha )
 {
+    if (lastSeenZoom != currentPhysicalZoomFactor)
+    {
+        int ns = -1;
+        for (auto s : scales)
+        {
+            if (s>=currentPhysicalZoomFactor && ns<0)
```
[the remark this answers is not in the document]
baconpaul: Oh yeah I'll fix that when I'm in there for your comment above. Thanks!

baconpaul:
I think we can get to free scale but it's quite a lot harder. If you look at #305 I put quantized scale as a step along the way.

FreeScale is harder mostly because I need to support that window drag event. The code can support any integer scale; just making the UI gesture for it is tricky.

So I sort of decided on this path (which is the path that Arturia took, and yeah it frustrates me that I can't just drag that) to get it working then if it works try to add the drag vector all platforms all hosts.

I also agree it is what folks would want to see, but I think there's both precedent and utility in this halfway step.

src/common/gui/CScalableBitmap.cpp
baconpaul: Or do you mean kill the != NULL? Yeah you can go both ways.

Issue #360, opened 2019-01-20
./build-linux.sh


jarkkojs:
Implement ./build-linux.sh that compiles Surge and copies configuration.xml, presets and wavetables to their locations.

baconpaul:
Oh hey @jsakkine when you do the final sweep with your last step, can you also update README.md. @esaruoho did a first pass but getting it all ship shape for linux would be lovely. Thanks!

jsakkine:
@baconpaul, sure

falkTX:
For the linux builds, I really recommend updating vstgui to the latest version (which means your own vst3sdk repo, because vstgui is a submodule of this submodule..)

Current "stable" vstgui uses gtk3 for window and event management on Linux, which is pretty terrible.
It brings an entire big system toolkit for a plugin UI, which is supposed to be self-contained in the first place. (there are far too many reasons why gtk, or similar system/generic toolkits are bad ideas for plugins)
As it is, the plugin will fail to load (I mean, crash the host) when loaded in a few Linux hosts.

vstgui on its develop branch is going in the right direction, now using just cairo and xcb.

I got it built myself locally by using a slightly older git commit of vstgui (absolute latest has some API changes regarding menus that I did not look into).

jsakkine:
Switching to a newer VSTGUI is not an issue associated now we do Linux builds, or how we package. Lets not associate orthogonal problems with each other. I agree that getting rid of GTK would be nice, though.

falkTX:
I understand that this specific ticket is about the packaging.
It is still related a bit, because we will have to switch libraries used for linking.
I created the other ticket so we can leave updated-vstgui stuff there anyway.

Please note, if we do linux builds with the current state, it will be dangerous for users that try it.
Packaging as-is will result in a broken plugin.

jsakkine:
This not really a packaging issue, just a build script, and missing still presets and wavetables. Can you define what you mean by user? There is nothing too dangerous in that script, and it is not meant for end users.

esaruoho:
is the packaging finished or still under way? just wondering if this ticket can be closed or still under works? @jsakkine

PR #361, opened 2019-01-20
Initial version of build-linux.sh

jsakkine:
Add a similar build script as build-macosx.sh. Copies plugin DLL's and
configuration.xml to local directories. Does not yet copy presets and
wavetables, but it is a starting point.

### DIFF
--- a/build-linux.sh
+++ b/build-linux.sh
@@ -1,0 +1,242 @@
+#!/bin/bash
+#
+# build-linux.sh is the master script we use to control the multi-step build
+# processes.
+#
+# To understand what it does, run
+#
+# ./build-linux.sh --help
+
+help_message()
+{
+    cat << EOHELP
+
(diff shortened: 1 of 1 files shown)


jsakkine:
First PR for #360. Will need a 2nd one for presets and wavetables.

baconpaul:
Looks familiar! Chuckle

Will merge

I will also fnish the widows powershell version shortly mostly works now

jsakkine:
Yeah, there might be some similarities to build-osx.sh :)

baconpaul:
Anyway I just pushed the rough equivalent for windows

```
scripts\win\build-win.ps1
```

you run it in powershell and it will do everything you need without needing to start visual studio. You probably need to do a visual studio build once to do the nuget (I should add that) and it's not as good as the osx and lin versions but good enough that I use it now.

PR #364, opened 2019-01-20
Use over-aligned allocations for the class Effect

jsakkine:
Use C++17 over-aligned allocations for the class Effect.


jsakkine:
This is identical change what I already did for Oscillators. With the voice change, also macOS version of Bitwig won't load. It is not a Windows specific issue.

baconpaul:
OK so you've confirmed it loads bitwig windows?

baconpaul:
@jsakkine confirmed on slack that it does indeed load vst2 and 3 in bitwig. Looks good to me. Merging.

Issue #369, opened 2019-01-21
Don't resize larger than screen


esaruoho:
path to zoom #305 
currently if i have a smaller screen and set to 300%, i cannot re-resize it to lower size, instead i'm stuck and need to delete the plugin and re-open it.

baconpaul:
@esaruoho I have a preliminary fix for this on branch bacon paul / screensize-369

I want to do a few things still (like do a self-code-review, test win and lin still work, and squash) but if you want to try that on your mac and provide any feedback it would be great.

I have a 2 monitor setup and it worked fine detecting the right monitor and stuff but these graphics APIs can be fiddly so another test would be lovely.

Thank you!

baconpaul:
OK this is all swept into master. Thanks for the eagle eyes, @esaruoho! Hope you like the error message when you try to over-grow :)

PR #370, opened 2019-01-21
Limit zoom by screen

baconpaul:
The current implementation will let you zoom to a size much larger than your
screen, a change which you can't undo. Reported in #369.

To resolve this, query the screen size at zoom time. Querying the screen
size means I added (in this diff) DisplayInfo.h which is an implementation
of functions in Surge::GUI namespace to get backing scale and screen size
for the screen in which your current frame is resident (or if you are split
screen, the screen the OS determines is the actually 'screen' for your
window). Implemented in same cross platform style as UserInteractions.

We then call that with a check and clear error message in SurgeGUIEditor

Testing strategy: Loads VST2/3 bitwig mac, Hosting AU + Logic mac, VST2/3 bitwig windows, links linux.

Ran AU zoom on my two monitor different resolution setup. On big monitor can zoom  to 300% on small monitor I cannot.


jsakkine:
You should put the platform as postfix, not prefix. They order now badly when you list the files alphabetically.

jsakkine:
DisplayInfoWin.cpp, DisplayInfoMac.cpp and DisplayInfoLinux.cpp.

jsakkine:
I don't see any changes to premake5.lua.

src/mac/MacDisplayInfo.mm
```diff
@@ -0,0 +1,60 @@
...
+        
+        CRect getScreenDimensions(CFrame *f)
+        {
+            CRect defaultDimension(CPoint(0,0),CPoint(1440,900)); // This default is a 15" macbooc pro with scaling at half size
+
+            if (!f) return defaultDimension;
+
+            NSView *v = getViewFromFrame(f);
+            if( !v || ![v window] || ![[v window] screen])
+            {
```
jsakkine: Not sure why you want to have curly brackets here.

src/mac/MacDisplayInfo.mm
```diff
@@ -0,0 +1,60 @@
...
+        float getDisplayBackingScaleFactor(CFrame *f)
+        {
+            float defaultBackingScale = 2.0; // If I don't know, guess retina.
+            if (!f) return defaultBackingScale;
+            
+            NSView *v = getViewFromFrame(f);
+
+            // So objective c would run these but lets be explicit
+            if( !v || ![v window] || ![[v window] screen])
+            {
```
jsakkine: Not sure why you want to have curly brackets here.

src/mac/MacDisplayInfo.mm
```diff
@@ -0,0 +1,60 @@
...
+                Surge::UserInteractions::promptError("Platform frame is not an NSViewFrame. Please report this issue to developers.",
+                                                     "Software Error");
+                return NULL;
+            }
+            return nf->getNSView();
+        }
+        float getDisplayBackingScaleFactor(CFrame *f)
+        {
+            float defaultBackingScale = 2.0; // If I don't know, guess retina.
+            if (!f) return defaultBackingScale;
```
jsakkine: I would put the return statement to the following line.

src/mac/MacDisplayInfo.mm
```diff
@@ -0,0 +1,60 @@
...
+            
+            float res = [[[v window] screen] backingScaleFactor];
+            return res;
+        }
+        
+        CRect getScreenDimensions(CFrame *f)
+        {
+            CRect defaultDimension(CPoint(0,0),CPoint(1440,900)); // This default is a 15" macbooc pro with scaling at half size
+
+            if (!f) return defaultDimension;
```
jsakkine: I would put the return statement to the following line.

src/mac/MacDisplayInfo.mm
```diff
@@ -0,0 +1,60 @@
...
+            VSTGUI::NSViewFrame *nf = dynamic_cast<VSTGUI::NSViewFrame*>(pf);
+            if (!nf)
+            {
+                Surge::UserInteractions::promptError("Platform frame is not an NSViewFrame. Please report this issue to developers.",
+                                                     "Software Error");
+                return NULL;
+            }
+            return nf->getNSView();
+        }
+        float getDisplayBackingScaleFactor(CFrame *f)
```
jsakkine: Maybe one empty line between function declarations?

src/mac/MacDisplayInfo.mm
```diff
@@ -0,0 +1,60 @@
+#include "DisplayInfo.h"
+#include "UserInteractions.h"
+#include <CoreFoundation/CoreFoundation.h>
+#include <Cocoa/Cocoa.h>
+#include "vstgui/lib/platform/mac/cocoa/nsviewframe.h"
+
+namespace Surge
+{
+    namespace GUI
+    {
```
jsakkine: Do we want to indent namespaces in the cpp-file at least? Wastes screen space more than brings value. I would just not indent these i.e.

```
namespace Surge
{

namespace GUI
{

NSView *getViewFromFrame(CFrame *f)
```

src/mac/MacDisplayInfo.mm
```diff
@@ -0,0 +1,60 @@
...
+        }
+        float getDisplayBackingScaleFactor(CFrame *f)
+        {
+            float defaultBackingScale = 2.0; // If I don't know, guess retina.
+            if (!f) return defaultBackingScale;
+            
+            NSView *v = getViewFromFrame(f);
+
+            // So objective c would run these but lets be explicit
+            if( !v || ![v window] || ![[v window] screen])
```
jsakkine: Spacing.

src/mac/MacDisplayInfo.mm
```diff
@@ -0,0 +1,60 @@
...
+                return defaultBackingScale;
+            }
+            
+            float res = [[[v window] screen] backingScaleFactor];
+            return res;
+        }
+        
+        CRect getScreenDimensions(CFrame *f)
+        {
+            CRect defaultDimension(CPoint(0,0),CPoint(1440,900)); // This default is a 15" macbooc pro with scaling at half size
```
jsakkine: I would prefer to put long comment like to the line before. Probably will never read it in this form because I don't like to scroll that much horizontally in the text editor and absolutely hate wrapping.

src/mac/MacDisplayInfo.mm
```diff
@@ -0,0 +1,60 @@
...
+            if( !v || ![v window] || ![[v window] screen])
+            {
+                return defaultDimension;
+            }
+
+            NSRect screenRect = [[[v window] screen] visibleFrame];
+
+            CRect res(CPoint(0,0), CPoint(screenRect.size.width,screenRect.size.height));
+            
+            return res;
```
jsakkine: `return CRect(CPoint(0,0), CPoint(screenRect.size.width,screenRect.size.height));`

src/common/gui/DisplayInfo.h
```diff
@@ -0,0 +1,22 @@
...
+#pragma once
+
+#include "vstcontrols.h"
+
+namespace Surge
+{
+namespace GUI
+{
+    float  getDisplayBackingScaleFactor(CFrame *); // this is in units of 1.0 == 1.0
+    CRect  getScreenDimensions(CFrame *);
```
jsakkine: See my comment below about namespaces and indentation. I don't think this a good idea to indent these. I don't understand the comment at all and would put it on top (1.0 == 1.0??).

src/windows/WinDisplayInfo.cpp
```diff
@@ -0,0 +1,23 @@
...
+                                                 "Software Error");
+            return 1.0;
+        }
+        CRect getScreenDimensions(CFrame *)
+        {
+            CRect r(CPoint(0,0),CPoint(1024,768));
+            Surge::UserInteractions::promptError("getScreenDimensions not implemented yet on windows.",
+                                                 "Software Error");
+            return r;
+        }
```
jsakkine: `return CRect(CPoint(0,0),CPoint(1024,768));`

src/mac/MacDisplayInfo.mm
baconpaul: Thanks. I indent headers that way but generally not source. Happy to adapt.

src/mac/MacDisplayInfo.mm
baconpaul: I try to always use curly braces when I press return after an if.

baconpaul:
> You should put the platform as postfix, not prefix. They order now badly when you list the files alphabetically.

But they are in distinct directories.

baconpaul:
> > You should put the platform as postfix, not prefix. They order now badly when you list the files alphabetically.
> 
> But they are in distinct directories.

Also: This is the naming convention you suggested when I did exactly the same thing with UserInteractions!

src/common/gui/SurgeGUIEditor.cpp
```diff
@@ -2405,17 +2406,46 @@ long SurgeGUIEditor::applyParameterOffset(long id)
 
 void SurgeGUIEditor::setZoomFactor(int zf)
 {
-   int dbs = getDisplayBackingScale();
-   zoomFactor = zf;
+#if SUPPORTS_ZOOM
```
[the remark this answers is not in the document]
baconpaul: Yeah SUPPORTS_ is specific. I want a variable which indicates to compile code if and only if the particular type of configuration supports zoom.

I suppose CONFIG_SUPPORTS_ZOOM is just as good. Not sure it is any better. But happy to change it.

baconpaul:
> I don't see any changes to premake5.lua.

Yes there aren't any.

src/mac/MacDisplayInfo.mm
baconpaul: OK. I'll do that and add curly braces too! :)

src/common/gui/DisplayInfo.h
baconpaul: For your edification, here's the comment I replaced it with

```
/*
** Return the backing scale factor. This is the scale factor which maps a phyiscal
** pixel to a logical pixel. It is in units that a high density display (so 4 physical
** pixels in a single pixel square) would have a backing scale factor of 2.0.
**
** We retain this value as a float and do not scale it by 100, like we do with
** user specified scales, to better match the OS API
*/    
```

which is way better

baconpaul:
> > > You should put the platform as postfix, not prefix. They order now badly when you list the files alphabetically.
> > 
> > 
> > But they are in distinct directories.
> 
> Also: This is the naming convention you suggested when I did exactly the same thing with UserInteractions!

But the name is better with OS at the end; so I will rename consistently in the subduers for this fix.

src/common/gui/SurgeGUIEditor.cpp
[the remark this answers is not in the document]
jsakkine: Just thinking about future config flags, which I'm sure there will be. CONFIG_HAS_ZOOM might be short and clear.

baconpaul:
Yeah I agree. I think I mentioned that when we swept it. Since they are used multiple places it's a bit of premake refactoring so I've left it be. Another thing for the todo list...

baconpaul:
Also: Thanks as always for the diligent review!

PR #372, opened 2019-01-22
chore: add linux build to pipeline

kzantow:


kzantow:
@baconpaul here's an added linux build it currently only builds vst 3 of course; please have a quick look and you can see the output following the check links. this should fail the build just fine as long as `./build-linux.sh --build` returns nonzero on error.

Issue #356, opened 2019-01-20
Implement Zoom on Macintosh and Windows / VST2


baconpaul:
With #348 in place allowing full scalable zoom on AU driven by menu with proper retina assets, we need to implement Macintosh / VST2 zoom. All part of the #305 family of issues.

baconpaul:
<img width="1921" alt="screen shot 2019-01-20 at 8 47 58 pm" src="https://user-images.githubusercontent.com/13387561/51448450-ca1ce500-1cf4-11e9-8e47-01c5d5a34f87.png">

So the VST2 API partially supports zooming but has quite a few bugs which are quite hard to work around. Still not perfect but getting closer this evening.

If you want to try it bugs and all checkout branch 'zoom-vst2-356' but it is very flaky now. Especially menu positions pop up in the wrong place when you are zoomed which is tough.

baconpaul:
OK so here's the state of play on this

On the branch `baconpaul` `zoom-vst2-356` is a (not ready for merge yet) branch which implements zoom on both mac and windows using only the vstgui and vst2 apis. That is, it calls ::setSize in the vst2plugin and frame->setZoom() in the SurgeGUIeditor. On windows, of course, you don't get scalable assets so big zooms look cruddy.

But there are two bugs on vstgui, only one of which I can get around

1. The frame setZoom does not apply the zoom to the frame background image so I have to do a gross workaround specifically for the background in collaboration between SurgeGUIEditor and CScalableBitmap. This is barely commented now also but you'll figure it out
2. Way more importantly, with setZoom set COptionMenu popup in the wrong place both mac and windows. I do not know a workaround for this one at all. Probably something to do with a transform forgotten.

Both these bugs, but especially the second which I can't hack around, is what inspired #373 last night.

But good news: Using the portable vst api to implement zoom in vst did in fact implement zoom.

kzantow:
@baconpaul so an interesting thing about the menu offset: **both the patch drop-down and the effects drop-down are positioned properly**, however all the right-click menus and the 'menu' button are not. hope that helps to diagnose/fix the issue!

baconpaul:
So useful. Thank you.

baconpaul:
OK just as notes-to-self

The patch browser pops up a menu in the onMouseDown handler where it uses the coordinates handed to it by the event system

The slider and menu button use the alternate pattern of sending a message back to the parent class and then the parent class uses frame->getCurrentMouseLocation and then frame->localToFrame

[...]

PR #380, opened 2019-01-22
Implement zoom for VST2 Mac and Windows

baconpaul:
This diff implements zoom for VST2 Mac and Windows using the
VSTGUI scale APIs in conjunction with the existing ability
of surge to register a zoom callback.

This would have been a relatively prosaic change were it not for
two bugs in VSTGUI

1: When you setzoom on a frame, it does not setzoom on the frame->getBackground
and so we need to explicitly resize the background image in VST2 as we zoom.
See the instance issues and additional scale in CScalableBitmap and SurgeBitmaps

2: the frame->getCurrentMouseLocation method is mistakenly scaled up twice
(as opposed to, one supposes, scaling up and down). So we need to, for vst2,
scale down twice.

This is all discussed in github issue #356.

With these diffs, though, both mac and windows vst2 plugins support
zoom (even though, ad this time, windows is using original not vector
bitmaps).

My testing strategy was

1. Hosting AU and Logic on Mac
2. Bitwig on mac VST2/3
3. Bitwig on windows VST2/3
4. VST3 linux links 

@jsakkine or @kurasu or @kzantow I would love a review if you have time.

### DIFF
--- a/build-osx.sh
+++ b/build-osx.sh
@@ -36,6 +36,7 @@ Commands are:
         --install-local          Once assets are built, install them locally
         --build-and-install      Build and install the assets
         --build-validate-au      Build and install the audio unit then validate it
+        --build-install-vst2     Build and install only the VST2
 
         --package                Creates a .pkg file from current built state in products
         --clean-and-package      Cleans everything; runs all the builds; makes an installer; drops it in products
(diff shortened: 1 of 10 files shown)


baconpaul:
Oh for reviewers: There are a couple of things I had to do to work around bugs which are unpalatable. But I documented them exhaustively at least once place, and I think the most natural place, in the code.

baconpaul:
I'm going to close this PR and do a single PR with VST2 zoom and windows scalable assets

PR #383, opened 2019-01-23
Pipefail in ./build-unixen

baconpaul:
Both ./build-osx and ./build-linux when teeing output
supressed the return code of build. In osx this was only
in verbose mode; in linux it was in both. This meant a failed
build in these settings returned success

Since these are bash scripts, fix it with set pipefail

### DIFF
--- a/build-linux.sh
+++ b/build-linux.sh
@@ -107,6 +107,8 @@ run_build()
 
     echo
     echo Building surge-${flavor} with output in build_logs/build_${flavor}.log
+    # Since these are piped we lose status from the tee and get wrong return code so
+    set -o pipefail
     if [[ -z "$OPTION_verbose" ]]; then
         make config=release_x64 2>&1 | tee build_logs/build_${flavor}.log
     else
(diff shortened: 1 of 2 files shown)


baconpaul:
@kzantow or @jsakkine - a 10 second peek at this diff appreciated. Fixes the problem reported.

build-linux.sh
```diff
@@ -107,20 +107,24 @@ run_build()
 
     echo
     echo Building surge-${flavor} with output in build_logs/build_${flavor}.log
+    # Since these are piped we lose status from the tee and get wrong return code so
+    set -o pipefail
```
kzantow: Nice one! Should this be set for the duration of the script though? (same for mac)

baconpaul:
Yeah I was thinking of a big old set -euxo pipeline at the top of both or some such - basically the use strict of bash. But wanted to make smallest change since I’m not sure what else is going on in Linux land 

Fine either way honestly.

baconpaul:
Jjs was happy so sweeping

PR #385, opened 2019-01-24
Windows vector 358

baconpaul:
This pull request implements two key features

1: Zoom in VST2 on Mac and Windows

2: Scalable bitmaps on Windows

Implementing this required working around several vstgui bugs; and also deciding to eject a very large resource file using a script (and checking in the result) rather than typing the tedious file by hand.

The comments in the code explain the decisions but notably

1. The comment in emit-vector-resources.py explains why I decided to check in rather than generate at build time the rc files
2. The comments in the vst2 zoom handler detail the two bugs in vstgui which were a problem
      1. double transform of mouse coordinates
      2. setting zoom on a window not setting zoom on its background

This is a pretty big diff but it all hangs together and I commented the parts which were unexpected. With it, VST2 is at par with AU on mac and windows as far as zoom and vector graphics.

VST3 is proving trickier. But would love to get this version in folks hands.

Thank you for a careful review.


jsakkine:
Reviewing tonight!

src/common/gui/SurgeBitmaps.cpp
```diff
@@ -89,7 +90,34 @@ void SurgeBitmaps::addEntry(int id)
...
+#ifdef USE_SCALABLE_BITMAPS
+      /*
+      ** Background images are specially handled by the frame object with scaling and so each needs
+      ** to maintain an independent 'additional zoom'. For now handle that by allowing the construction of
+      ** a distinct bitmap. 
+      **
+      ** When the arity issues with scalable bitmaps get solved (github issue #356) this code will 
+      ** not be needed any more. Until then it is, but just for the BG image.
+      */
+      if( id != IDB_BG )
```
jsakkine: Maybe would be better to stick to C++ commenting style.

src/common/gui/CScalableBitmap.h
```diff
@@ -28,11 +28,19 @@ class CScalableBitmap : public VSTGUI::CBitmap
...
+    ** which we apply to the BG Bitmap in Vst2PluginInstance::handleZoom.
+    */
+    void setAdditionalZoom (int a) { additionalZoom = a; }
+    
 private:
     std::vector< int > scales;  // 100, 150, 200, 300 etc... - int percentages
     std::map< int, VSTGUI::CBitmap * > scaledBitmaps;
     std::map< int, std::string > scaleFilePostfixes;
-    int lastSeenZoom, bestFitScaleGroup;
+    int lastSeenZoom, bestFitScaleGroup, additionalZoom;
```
jsakkine: Maybe then name it as backgroundZoom? The current name is a bit confusing.

src/common/gui/CScalableBitmap.cpp
```diff
@@ -76,7 +77,7 @@ void CScalableBitmap::draw (CDrawContext* context, const CRect& rect, const CPoi
         // Seems like you would do this backwards; but the TF shrinks and the invtf regrows for positioning
         // but it is easier to calculate the grow one since it is at our scale
         CGraphicsTransform invtf = CGraphicsTransform().scale( bestFitScaleGroup / 100.0, bestFitScaleGroup / 100.0 );
-        CGraphicsTransform tf = invtf.inverse();
+        CGraphicsTransform tf = invtf.inverse().scale( additionalZoom / 100.0, additionalZoom / 100.0 );
```
jsakkine: The usual spacing complains :)

src/common/gui/CScalableBitmap.h
```diff
@@ -28,11 +28,19 @@ class CScalableBitmap : public VSTGUI::CBitmap
     */
     static void setPhysicalZoomFactor (int zoomFactor); // See comment on zoom factor units in SurgeGUIEditor.h
 
+    /*
+    ** VST2 has an error where the background doesn't zoom with the frame. Everything else
+    ** does though. So we need to hand scale the background and only the background image
+    ** when we draw it. We do that, simply, by having this class have an additionalZoom
+    ** which we apply to the BG Bitmap in Vst2PluginInstance::handleZoom.
+    */
+    void setAdditionalZoom (int a) { additionalZoom = a; }
```
jsakkine: C++ comments...

src/common/gui/SurgeBitmaps.cpp
jsakkine: Not going to mention about spacing in every place :)

src/common/gui/SurgeBitmaps.h
```diff
@@ -13,4 +13,4 @@ class SurgeBitmaps
    void addEntry(int id);
 };
 
-VSTGUI::CBitmap* getSurgeBitmap(int id);
\ No newline at end of file
+VSTGUI::CBitmap* getSurgeBitmap(int id, bool newInstance = false);
```
jsakkine: We should start documenting these starting from this commit. Lets us Doxygen: http://www.doxygen.nl/manual/docblocks.html. And I would recommend using the javadoc style e.g.

```
/**
 * @param id          <description>
 * @param newInstance <description>
 * @return description
 * Long description
 */
```

Lets not worry about generation at this point. Even having a structured style documented classes in the source code will help to understand.

jsakkine: Yes, and this is an exception where using C-style comment is OK, because the tool requires it.

src/common/gui/SurgeGUIEditor.cpp
```diff
@@ -2407,17 +2409,24 @@ long SurgeGUIEditor::applyParameterOffset(long id)
 void SurgeGUIEditor::setZoomFactor(int zf)
 {
 #if HOST_SUPPORTS_ZOOM    
+
+#if TARGET_VST2 || TARGET_VST3
+   // vstgui zoom is very unreliable shrinking with backgrounds.
+   if( zf < 100 )
+       zf = 100;
+#endif
```
jsakkine: Not sure if I understand what the comment is trying to say about what you are doing here.

src/common/gui/SurgeGUIEditor.cpp
```diff
@@ -2524,4 +2534,29 @@ void SurgeGUIEditor::showSettingsMenu(CRect &menuRect)
 
 }
 
+CPoint SurgeGUIEditor::getCorrectlyScaledMouseLocation()
+{
+    /*
+    ** Please see github issue 356 for discussion on this.
+    **
```
jsakkine: You could document properly what this function does with javadoc style in the class definition.

jsakkine:
OK, the only problem with Windows assets that I have is that you are pushing automatically generated file to the repository, which means that your build process is somehow broken.

Can I also review the updated PR. And can we move (iteratively commit by commit) to more structured commenting style? Doxygen support multiple but at least for my eye the javadoc style is most pleasing (maybe because kernel uses that one).

baconpaul:
OK thank you; I will enhance the comments and fix the spacing and make the other adjustments you suggest. Very useful. 

On spacing I thought I got it all. Like I said, decades of `if( foo )` will be a hard habit to break.

As to the build process being broken, I agree that's the conclusion. Here's my thinking as to my options

1. I could have (easily) added a requirement for python on build hosts, a step in premake to build the file, and a call to the script. It would have ejected the exact same output on every build unless we add an asset, which we almost definitely won't again in 1.6. I would also have updated the azure pipelines to add python as a dependency and added python 3 as a dependency in the README. And then I could have spent a few hundred hours on slack reminding windows users to install python and debugging that install or having them not build. 
2. I could have written the script in some scripting language which is pre-installed in windows10. That would basically be PowerShell I think. Pass.
3. I could have written the script in some language I could compile, but writing this in C++ rather than python and adding a two phase compile is, I think, hard for premake to do and also really rather dumb
4. I could have pretended i typed it in and not checked in my script. Too late for that strategy now of course, but in the sake of a complete enumeration wanted to add it :)
5. I could do the gross thing and check in the output, but not require all our users who want to build to have python

Now if the file changed every build, or every 3 PRs, or anything like that, I would have done the first. And the first is better in every way. Except those few hundred hours of slack support. So I chose the check in option.

As you can see even I'm rather critical of that approach. (Heck almost this exact same criticism is in the comments for emit-vector-rc.py). But I think it is the right tactic.

At some point we will snap and either the about screen or asset build or something else will make us require python in our build path.

So with that context, what do you think?

src/common/gui/SurgeGUIEditor.cpp
baconpaul: I think we could document what it does in the class definition; but should document how it does it here.

What it does is get the correctly scaled mouse location. I've always been skeptical of

```
  /**
   * getCorrectlyScaledMouseLocation
   * @decscription gets the correctly scaled mouse location, overcoming some vstgui bugs if needed
   * @returns the correctly scaled mouse location
   */
   CPoint getCorrectlyScaledMouseLocation()
```

as being useful (which is why i generally think the oxygen comments are tough for protected and internal methods which are well named).

So I guess: how would you separate out the comment between intent and implementation?

src/common/gui/SurgeBitmaps.h
jsakkine: Sure.

src/common/gui/SurgeGUIEditor.cpp
jsakkine: The correct format would be:

```
 /**
  * <short summary>
  * @return <what gives back>
  *
  * <long description what this does and maybe some implementation details>
  */
 CPoint getCorrectlyScaledMouseLocation()
```

In the long description you would explain what you mean by correctly scaled mouse location. It is far from obvious.

jsakkine:
I thought that in C++ you would always have to use `//` but as I look at http://www.doxygen.nl/manual/docblocks.html, it is really not mandatory :)

For stuff that is inside function I would use just

```
// We do it this way because blaa blaa ...
// bllaa
```
For function, class etc. descriptions I would use

```
/**
 * Foo
 */
```

That is one of the styles accepted by most documentation generators.

In general, at least non-trivial public functions ought to be documented, and class descriptions.

jsakkine:
I don't think that it makes sense to use comment block for comments inside a function but it should be uses for bigger constructs.

jsakkine:
My key remarks are:

- We need a standard for commenting functions.
- We need a standard for commenting classes.
- It would better be one that can be used for generation even if we don't do it now.
- Documenting classes at top level is not enough. At least non-trivial functions, like the one you mentioned, ought to be documented. It is not obvious what correctly scaled is unless you have implemented the function.

You can get ideas how to create long descriptions for functions from code reviews. 

Look at the Doxygen manual. It provides multiple options for styles. Unfortunately the one that you are used to use is not the one, but maybe something will work out for you that you can live with.

jsakkine:
And yeah my "C++ style" was invalid deduction. I think `//` would be good for comments that are inside the function implementation (that will not be generated in any way).

src/common/gui/SurgeGUIEditor.cpp
jsakkine: Maybe the function name is not good because correctly, in fact, does not mean anything. Dunno.

jsakkine:
If you are the person who originally did this, of course you know what correctly scaled means. If we want to port this to lets say Linux this to be properly documented. I kind of have now grip of everything right now, but probably will forget it within couple of weeks, since I didn't write the code.

src/common/gui/CScalableBitmap.h
baconpaul: I'll comment it and think of a better name, but won't be 'backgroundZoom' since we may need to apply this to other images; the reason I chose 'additionalZoom' is because it sort of is that; add this additional zoom.

baconpaul:
Yeah OK fair enough. I'll change the name of the function and add a class function comment which is a bit more explanatory.

You OK with the checkin-built reason? I think it's the right call for now.

src/common/gui/SurgeGUIEditor.cpp
baconpaul: I changed the function name to

```
   /**
    * getCurrentMouseLocationCorrectedForVSTGUIBugs
    *
    * This function gets the current mouse location for the frame
    * but adds, in necessary cases, workarounds for bugs in the
    * vstgui framework. Use it in place of frame->getCurrentMouseLocation
    */
   CPoint getCurrentMouseLocationCorrectedForVSTGUIBugs();
```

since I think that is the most correct name (since it replaces getCurrentMouseLocation)

baconpaul:
OK I just pushed a commit to this branch which addresses your comments. I will squash all the three commits into a single commit before merge, but left my changes separate to make sure you think they address your comments.

Thanks for the comments; it helps to see what parts of the code aren't clear.

jsakkine:
Looks good to me!

baconpaul:
Great thanks!

jsakkine:
Maybe we could start by little with coding conventions. Make a section to README.md that gives guidelines for comments. That's all for now. We can refine it whenever we end up into an argument. And also add a note that always link issue to the commit messaage.

baconpaul:
Agree! I'll merge this and lets write that doc.

Issue #288, opened 2019-01-12
~/Documents/Surge could have a wavetables folder for user-generated wavetables?


esaruoho:
Since presets are saved to `~/Documents/Surge/subfolders` - it would only make sense to have the wavetables be saved into the same place (like subfolders of `~/Documents/Surge/wavetables` maybe?

Otherwise the user needs to go to `~/Library/Application Support/Surge/wavetables` and copy them there. not very user-friendly

baconpaul:
The only trick is that we make categories based on parent directory and don’t recurse. We could easily have a special folder or we could just load looking for wt files anywhere also and you make too level dirs 

What do we want to directory layout to be in documents/surge basically?

esaruoho:
wavetables/subfolders/cool.wt

Is that possible?

baconpaul:
Well it’s code, it can do (almost) anything.

How about this

1. On startup if it doesn’t exist create doc/surge/user_wavetsbles
2. Scan that for directories containing wt
3. Modify save so a user can’t save a patch in user_wavetables

And be done

I’ll need to do #217 before that though to get the user feedback. @jsakkine here’s another spot where we need modal non throwing feedback. I will update 217 with some thoughts.

I don’t have a bunch of surge time until about Thursday is my guess. I’ll tuen to it the

baconpaul:
Yeah that’s a smart solution but I don’t know how to get there from current layout. Here’s an alternate idea

We just walk every subdir in documents / surge. If it contains a wt that becomes a wt cat. If it contains an fxp that becomes a patch cat.

That way third parties can make a single directory containing patches and wave tables and you just drop them into documents/surge and it works

Kinda like when someone shares a Dropbox folder with PowerPoint and excel both (if that analogy resonates with you)

See what I mean?

baconpaul:
Oh and obvious if doc/surge/baconfun contains a patch and a wavetable it shows up in the patch menus and the wavetable menus

baconpaul:
Ok so here is the code I plan to write to both add wavetables and subdir (so collapsing #283)

In documents surge do a full tree walk
If you find a .fxp in directory a/b/c the you get a patch with category name “a::b::c” where “::” is a separator defined in a constant. That way it all sorts still and all the category jogs work

At menu construction time show a::b:;c as submenus not as a flat hierarchy. In the display show a::b::c with an appropriate visual separator.

Finally if you hit a .wt do the same and show it in the wt menu

[...]

PR #386, opened 2019-01-25
Documents Patches with SubFolders

baconpaul:
If documents have subfolders they show up as submenus.
If you save a patch with category name "foo/bar/thisthat"
then you create directory structure foo/bar/thisthat and
associated nested menus. Addresses the patch part of #288 

This is lower priority to review than my zoom PR.

### DIFF
--- a/src/common/SurgeStorage.cpp
+++ b/src/common/SurgeStorage.cpp
@@ -8,6 +8,8 @@
 #include <set>
 #include <numeric>
 #include <cctype>
+#include <map>
+#include <queue>
 #include <vt_dsp/vt_dsp_endian.h>
 #if MAC
 #include <cstdlib>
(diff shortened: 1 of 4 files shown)


src/common/SurgeStorage.cpp
```diff
@@ -300,13 +302,43 @@ void SurgeStorage::refreshPatchlistAddDir(bool userDir, string subdir)
...
    }
 
-   for (auto& p : fs::directory_iterator(patchpath))
+   /*
+   ** std::filesystem has a recursive_directory_iterator, but between the 
+   ** hand rolled ipmmlementation on mac, expermiental on windows, and 
+   ** ostensibly standard on linux it isn't consistent enough to warrant
+   ** using yet, so build my own recursive directory traversal with a simple
+   ** stack
+   */
```
jsakkine: I would for inline comments inside function just use `//`.

```
// std::filesystem has a recursive_directory_iterator, but between the 
// hand rolled ipmmlementation on mac, expermiental on windows, and 
// ostensibly standard on linux it isn't consistent enough to warrant
// using yet, so build my own recursive directory traversal with a simple
// stack
```

src/common/SurgeStorage.cpp
```diff
@@ -300,13 +302,43 @@ void SurgeStorage::refreshPatchlistAddDir(bool userDir, string subdir)
...
+       {
+           if (fs::is_directory(d))
+           {
+               alldirs.push_back(d);
+               workStack.push_back(d);
+           }
+       }
+   }
+
+   // We want to knock out /user/foo or c:\\users\\bar\\ so get the length once
```
jsakkine: And here are using that style (and different one above).

jsakkine: If this was by any means consistent, this should be:

``
/*
** We want to knock out /user/foo or c:\\users\\bar\\ so get the length once.
*/
```
To follow the style inline comment above.

**Comment by jsakkine** [comment]
Referring to issues, especially as time goes on, is a really bad idea, except in the sense that this fixes this and this issue. But in the sense, that you should jump to some old closed issue to get idea what the commit does, is just stupid. Takes ages to look through the whole discussion to get the grip what is going on. I think we should take the time and summarise stuff in the commit message. It is also good exercise, because you have to actually think how you did the code change.

And it makes the commit log itself documentation. With this kind of commit messages it is just a dumpster without any real use.

**Comment by baconpaul** [comment]
Calling aspects of this PR "just stupid" or "a dumpster" is a bit strong isn't it? 

Some projects I've participated in, for instance, *require* that commits have reference to a permanent discussion of every design decision, so require commits to have a link to an issue otherwise they can't be merged. And not small projects. You may also have noticed that my branches have names referring to issues and my commit messages include issue numbers regularly.

In fact, I've actually considered asking you to change your commits to include links to issues (but wouldn't have said that your decision to not do so was "just stupid"). 

But you are correct, this commit message could be more fulsome and it's a good comment to expand it, even thought from the commit message it is easy to get to the answer to your question as to what the problem is and how to solve it. You should have a good first order understanding from the commit log.

I think your comments on comments are more s style thing. Again, in other large projects I've seen the standard has been "never use // comments except for one line comments"  and a few have "never use // at all" (since "//" makes it a pain to expand to a more complete multiline comment). 

```
   int foo(); // do the bar

   /**
    * Do the bar because jimbob
    */
   int foo();

   int foo()
   {
     // don't forget we need every mbar
     for( auto gylpyinothar = mbar )
     {
       /*
       ** You have a lot to say and have something trick
       */
     }
   }
```

Or of course there's the standard in our current code base. Which seems to be "not many comments".

The /** at a function decl is a good idea though and we can switch to that. And C++ doesn't have a standard, just projects do. (Other than // can't be multiline)

You won't have to do this twice on linux only once. This is actually done 3 times now on mac and windows since the recursion is in a function called with different configurations, if you look at the parent calls.

**Comment by baconpaul** [comment]
Anyway I pushed a branch which expanded comments, dropped the // single line, and also made a more expansive commit message. Also fixed the vestigial is_directory check you found (thanks!)

File: src/common/SurgeStorage.cpp (hunk 1 above)
Comment (baconpaul) [reply src/common/SurgeStorage.cpp]: Nah never use // for multi-line comments.

File: src/common/SurgeStorage.cpp (hunk 2 above)
Comment (baconpaul) [reply src/common/SurgeStorage.cpp]: See discussion below. But anyway I expanded this to be a multiline comment so it's now in /* style.

**Comment by jsakkine** [comment]
I can take the habbit of linking the issues to my commits. I've done to some, but for some I think I've forgotten/ignored that. Still I think nice summarisation of discussion would be nice.

**Comment by baconpaul** [comment]
Ok works for me. I always used the double star but can shift to single star.

Look ok otherwise?

**Comment by jsakkine** [comment]
Still one more question so that I know that I understood the code change right before your commit: are categories still meant to be flat and not hierarchical? It is just that you can organize a category into subdirectories?

**Comment by baconpaul** [comment]
> Still I think nice summarisation of discussion would be nice.

I agree. I rewrote the commit message here

**Comment by jsakkine** [comment]
And yeah I'm fine with having this as convention for multi-inline stuff
```
/**
 **
 */
```
As long as we use it consistently. It differs from what we use in Linux, but for me consistency is more important than personal preferences.

**Comment by baconpaul** [comment]
OK cool.

Not to be a pain but the format I naturally use is

```
 /*
 ** One star on the first and last line
 ** means that each comment line is preceeded
 ** by 2 characters consistently, either two stars or a star and slash
 */
```

and would be thrilled for that to be the standard

PR #399, opened 2019-01-27
removed IRC Channel, re-wrote slack text

esaruoho:


falkTX:
Why was this removed? IRC channel is still there, and people chat just fine.
It is a bit sad to be using proprietary ways of communication for an opensource project...

baconpaul:
I didn’t know it was even active with folks developing surge. Folks who put in prs recently have all been on slack. I don’t use irc myself but have no dog in a chat tool fight!

If it is used we should put it back no prob. I don’t mind!

Issue #373, opened 2019-01-22
Time to update vst3sdk pointer?


baconpaul:
Our vst3sdk pointer is very old. We are using the Jun 11 pointer and there have been 5 updates since, including ones from @whydoubt and others who have worked on surge.

I'm running into a vstgui problem now (when you setScale on a frame, opition menus appear at the wrong spot; they double-scale the mouse coordinate). Would love to know if it is fixed.

So I could, I suppose, pull vst3sdk to a latest (to hash 82380a8faa33824123104e005bcffb0451151d88) and see what happens

But before I do that work: Anyone object?

baconpaul:
Hmm well I still think it might be time to update our vst3sdk pointer
but I was able to work around the vst2 bug a bit grossly
so I'm not going to do this work anytime soon unless I run into another unsquashable bug

kurasu:
no objections here.. if it breaks everything we can always switch back

baconpaul:
From looking the problem seems to be of two types

1. Things like CPoint and CColor are now in the VSTGUI namespace so either we need a lot of using namespace VSTGUI or a lot of VSTGUI:
2. enums are scoped in classes. So kNoDrawStyle -> COptionMenu::kNoDrawStyle

tedious.

whydoubt:
I have a branch containing the needed changes (namespace and scoping), waiting for when they were useful. I've been busy elsewhere the last couple of weeks though, and hadn't been keeping up with the state of the project.

baconpaul:
Can you attach a .patch file here at least? Or if you branch is rebaseable, toss in a PR? Would be good to know about how much work it is. Happy to work starting at your branch also if you want. Thank you.

baconpaul:
OK I have this all building mac now with the updated pointer. Will begin some testing and seeing if it builds elsewhere.

`baconpaul/vst3sdk-update-373` is my branch

esaruoho:
hmm.. i wonder what kinds of gems we would find in https://github.com/whydoubt/surge/branches/all and how much of it is outdated (already implemented) and how much is yet to be merged in to the main repo.. would be interesting to know.

baconpaul:
Dunno. I do know my vst3sdk update didn't build on windows when I tried, and I'm not going to fix it tonight. But we can get this pointer moved this week (which will also help @sagantech).

baconpaul:
[...]

PR #401, opened 2019-01-28
Update to Jan 2019 vst3sdk

baconpaul:
This substantial diff undertakes the mechanical changes to update
to the Jan 2019 master version of vst3sdk (hash 82380a8f). This
move consuming nearly 7 months of changes introduced several breaking
changes in our code which required modification, and introduced a couple of
new dependencies and links.

The primary thing breaking the builds was that most classes, including
common ones like CRect and CPoint, are now in the VSTGUI namespace.
This diff follows the convention of never using a namespace in a header
file, and thus has quite a few changes which are namespace additions in
headers and using statements in cpp files. But ironically, since my first
attempt was to just put using in a header and break the rule for once,
if you put using in the header, the code doesn't compile on windows
since VSTGUI::IDropTarget and ::IDropTarget collide. So has to be this way.

On linux several xcb and xkb dependencies emerge with this vst, so they
were added to the azure pipeline. Also the vstgui_linux.cpp is no longer
needed in the way we link, and if linked, doesn't work.

On windows, an ambiguity in std::numeric_limits<T>::max() requires us to
set -DNOMINMAX=1 for, alas.

On mac, we expand the number of errors reported to 'as many as you generate'.

Commands used to update the module were

 git checkout master
 git checkout -b vst3sdk-update-373
 git submodule update --remote --merge --recursive
 git add vst3sdk/
 git commit
 git submodule update --init --recursive

This commit addresses github issue #373.

### DIFF
--- a/src/common/gui/CLFOGui.cpp
+++ b/src/common/gui/CLFOGui.cpp
@@ -4,6 +4,8 @@
 #include "CLFOGui.h"
 #include "LfoModulationSource.h"
 
+using namespace VSTGUI;
+
 extern CFontRef surge_minifont;
 extern CFontRef surge_patchfont;
 
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -91,6 +91,12 @@ jobs:
       popd
 
       sudo apt-get install -y libgtkmm-3.0-dev
+      sudo apt-get install -y xcb
+      sudo apt-get install -y libxcb-util-dev
+      sudo apt-get install -y libxcb-cursor-dev
+      sudo apt-get install -y libxcb-keysyms1-dev
+      sudo apt-get install -y libxkbcommon-dev
+      sudo apt-get install -y libxkbcommon-x11-dev
 
       ./build-linux.sh --build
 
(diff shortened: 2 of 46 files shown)


src/common/gui/CLFOGui.cpp
```diff
@@ -4,6 +4,8 @@
 #include "CLFOGui.h"
 #include "LfoModulationSource.h"
 
+using namespace VSTGUI;
```
kzantow: why `using` here and not other places?

baconpaul:
> I don't know better, so this looks like a "GREAT and we'll fix anything later" type of PR. My only question is why some `using` and not just more targeted `using` statements to make this all work?

Using is safe in cpp files but not headers basically.

kzantow:
Tested this and VST2 and 3 both work fine in Bitwig, no change in the behavior in REAPER (VST3 works fine). I think these changes are 👍

baconpaul:
OK I'm going to merge this so we can move forward. Will make some noise on slack also about the submodule pull. Thanks @kzantow and @esaruoho

PR #402, opened 2019-01-28
Restore patch checkmarks

baconpaul:
Restore functions from Surge 1.5.2.
- Add patch checkmarks next to category and patch.
- Restore checkmarks again and when patch is reloaded.
- Also expose patchname to host via a single program name.
- Integrate code review comments from @baconpaul and @jsakkine

The commit history here also got conflated with the vst3sdk update
and so is a mix of @sagantech and @baconpaul; but the code author
is @sagantech

### DIFF
--- a/src/common/SurgeSynthesizerIO.cpp
+++ b/src/common/SurgeSynthesizerIO.cpp
@@ -171,6 +171,10 @@ void SurgeSynthesizer::loadPatch(int id)
       kCFStringEncodingUTF8);
            ((aulayer*)parent)->SetAFactoryPresetAsCurrent(preset);*/
 #endif
+   /*
+   ** Notify the host display that the patch name has changed
+   */
+   updateDisplay();
 }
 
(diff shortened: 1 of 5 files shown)


src/common/SurgeSynthesizerIO.cpp
```diff
@@ -171,6 +171,11 @@ void SurgeSynthesizer::loadPatch(int id)
       kCFStringEncodingUTF8);
            ((aulayer*)parent)->SetAFactoryPresetAsCurrent(preset);*/
 #endif
+   /*
+   ** patch loading can be internal (menu) or external (host). In the event
+   ** it is from the host we need to update the display so do so here
+   */
+   updateDisplay();
 }
```
sagantech: Comment is incorrect. UpdateDisplay updates the host display not the plug-ins.

jsakkine:
Looks great now! Only thing would have been:

```
Fixes: #397 ("No checkmarks in preset browser")
```

After description, but not a biggie...

Issue #357, opened 2019-01-20
Implement zoom on Macintosh and Windows / VST3


baconpaul:
With #348 in place allowing full retina zoom on mac au, we need to do the same for mac vst3. Part of the path of issues required for #305

Issue #164, opened 2018-12-27
VST3: Split Processor vs Controller?


whydoubt:
SurgeVst3EditController is unused.  Either the processor and controller should be split properly into separate components, or this vestigial controller should go away.  Knowing the history of the VST3 code could be of value here.  FWIW, Steinberg discourages using the single-component pattern for VST3 plugins.

From messing with the code a bit, the main issue I run into with trying to split the components is that methods from both components access surgeInstance, and I don't know how to handle that.

baconpaul:
I'm struggling with VST3 zoom #305 #357 and I think it's because of this split also. The API to indicate resize to a host is not clearly reachable from our single class to me. Not sure I am correct, but just linking the issues.

baconpaul:
So I'm going to close this issue because I just opened #1208 which will subsume this. We need to do this for sure, but it's a big change which #1208 covers for all the host flavors.

PR #406, opened 2019-01-28
Implement VST3 zoom

baconpaul:
VST3 zoom is very similar to VST2 zoom since both rely on
VSTGUI (and the associated bugs therein), but differ on the approach
to notify the host. The primary difference in notifying the host
is requiring a reference to the Steinberg::IPlugFrame frame class,
which has to be publically exposed to allow the processor to see it.
(This may be indicative of issues still with us choosing to not have
a formal processor/editor split, as mentioned in #164). Regardless,
once the IPlugFrame is exposed, we can use the resizeView() API
in vst3 to inform hosts in our zoom callback, as implemented here.

Adresses issue #357

### DIFF
--- a/build-osx.sh
+++ b/build-osx.sh
@@ -37,6 +37,7 @@ Commands are:
         --build-and-install      Build and install the assets
         --build-validate-au      Build and install the audio unit then validate it
         --build-install-vst2     Build and install only the VST2
+        --build-install-vst3     Build and install only the VST3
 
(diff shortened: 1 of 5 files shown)


baconpaul:
Thanks @kzantow. I’ll merge this in now and then we will have zoom on all our working combos!! Yay.

And then to fix the rest of the zoom related issues...

PR #407, opened 2019-01-29
Add checks to wavetable popup menu

sagantech:
I looked into adding support for adding check marks to the wavetable menu after a preset or a host document is loaded but I don't think it is worth the effort if it is even possible as the data for the wavetable is stored inline in the preset's data and not from an external file.


baconpaul:
This looks good to me. Generally we refer to the issue number in the commit message but I can do that for you when I merge it.

I'm also going to add an issue to remind us to undo that .gitignore once your changes to vst3sdk are merged and we move the pointer again.

Thanks!

PR #412, opened 2019-01-30
Skin Source: Font layers back in source. Lato is used now.

itsmedavep:
There is Lato everywhere now. Basically, had to render out all the layers again though.


baconpaul:
These look great, Dave. Merging.

PR #418, opened 2019-01-30
Set Init patch HP to lowest frequency (6.875Hz)

### LINKED ISSUE #414: Set Init-Highpass setting to "zero" instead of 50hz
Currently the Highpass setting in Surge is 50Hz.

Proposal: Set the Highpass to minimum the slider will go (which is something like 6Hz).

The parameter in question is here:
vember_audio__surge__ 64bit__and_renoise__intel_64bit_

esaruoho:
Fixes #414 

This changes the Generic HighPass Filter (that is applied whether F1 or F2 are on or off) from 50Hz to minimum, which is 6.875Hz.

Before: HP is at 50Hz
After:
fullscreen_30_01_2019__18_59


baconpaul:
Yes that's a smarter default agreed. Happy to merge.

PR #420, opened 2019-01-30
No more using namespace; win -> c++17

baconpaul:
using namespace in header files is bad for a variety of
well documented reasons. This change removes them in favor
of (1) full scoping in header files and (2) moving the
using namespace std to cpp files after the last include

The reason to do this change is so we can compile windows
with /std:c++17 which this change also does. Without this
change, the using namespace std would conflict between
byte and std::byte in system headers. However, with this
change, combined with the NOMINMAX requires by the vst3sdk
upgrade, min( and max( in our copy of wtl are ambiguous. As
such I took the distasteful route of modifying lib/wtl to
add `#include <algorithm>` and use std::min and std::max.

With this diff, both mac and windows compile and run and
windows compiles with C++ 17

A prior change has resulted in linux linking problems which
I'll resolve separately, but this code all compiles to objects
on linux.

### DIFF
--- a/build-osx.sh
+++ b/build-osx.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 #
 # build-osx.sh is the master script we use to control the multi-step build processes
 #
--- a/libs/wtl/atlapp.h
+++ b/libs/wtl/atlapp.h
@@ -1638,7 +1638,7 @@ public:
 	}
 
 // COM Server methods
-	LONG Unlock()
+	LONG Unlock() noexcept // MODIFICATION added noexcept to compile c++17
 	{
 		LONG lRet = CComModule::Unlock();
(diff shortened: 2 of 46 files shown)


jsakkine:
This is a change we want anyway. Doing `using namespace std` is evil. I think it is evil for any namespace. If you need want to use using, you should do it for specific symbols. For std, I would never do it, as it is so small prefix anyway...  Makes code easier understand to have it.

baconpaul:
I don't mind using namespace in cpp program units that much (but would prefer we didn't with std indeed). At least that won't break includes. But I agree I would not use it in new code. At the port time, both here and with vstgui, I did add it to some cpp units.

The auto builds are broken but I'll go ahead and sweep this. I manually confirmed that it doesn't change build status on the platforms we have.

PR #421, opened 2019-01-30
Include libpng for windows in the repository instead of using nuget

kurasu:
Nuget turned out to be very unreliable, as it isn't included by default by visual studio 2017 and although it and MSBuild understands the nuget packages, it can't be coaersed into resolving the packages as part of the build process.

I also changed the build.cmd script to call 'git submodule update', so that a user doesn't have to do that manually to make sure the submodules are in sync.

Enabled /bigobj compile option on windows, as I could no longer build the debug x64 build without it.


baconpaul:
Any thoughts, rather than copying it all into lib. of adding https://github.com/glennrp/libpng as a submodule inside lib? Then we have an easier update path.

kurasu:
Well then we'd need to fork it as that one doesn't build out-of the box for visual studio 2017, and doesn't even have the x64 target in it afaics. And you need zlib too. That would be even more work than nuget

baconpaul:
Well those are all pretty good reasons yeah!

Let me fix the line xlink error I just found from the vstgui upgrade and then I'll unconflict this and merge it in.

baconpaul:
OK I resolved the conflict but looking at it, I think we may need to do something semi-clever to make 32 bit work with this diff. I'm going to leave it sit here for a bit until I have time to noodle since I'd sooner not have 32 bit work for one and only one commit :)

kurasu:
whats the issue with 32-bit that needs to be clever? if premake is your concern you can do this to filter on multiple conditions like this:
configuration { "x64", "Debug" }

baconpaul:
Clever is probably strong! We need to link the non x64 lib on non x64 builds don’t we?

kurasu:
Yes, I added those in the second commit

baconpaul:
Oh cool!

PR #438, opened 2019-02-01
Linux fixes

falkTX:
Ugly but working VST2 GUI under Linux.
Since we have no native event loop to attach to, we use host idle for it.
The logic for it needs work, right now for this PR I just got things working so we can get an UI that does something..


falkTX:
Progress so far:

image

baconpaul:
Thank you very much for these changes!

PR #441, opened 2019-02-01
Load bitmaps on linux from /usr/share

falkTX:
Experimenting with loading bitmaps on linux.
Obviously only works if "resources" dir is available, but allows Linux to test the full GUI for now.

We need to convert these bitmaps into object code, or embed them in the binary.
I believe premake has built-in utilities for that.
If not, we leverage some tool.
When done, we replace `createFromPath` with `createFromMemory`.

### DIFF
--- a/src/common/gui/SurgeBitmaps.cpp
+++ b/src/common/gui/SurgeBitmaps.cpp
@@ -87,8 +87,9 @@ void SurgeBitmaps::addEntry(int id)
 #ifdef USE_SCALABLE_BITMAPS
    VSTGUI::CBitmap *bitmap = new CScalableBitmap(CResourceDescription(id));
 #elif __linux__
+   // FIXME use scalable bitmaps, and load stuff from memory
    char filename [1024];
-   snprintf (filename, 1024, "./resources/bitmaps/bmp%05d.png", id);
+   snprintf (filename, 1024, "/usr/share/surge/bitmaps/bmp%05d.png", id);
    filename[1023] = '\0';
    VSTGUI::CBitmap* bitmap = new VSTGUI::CBitmap(IPlatformBitmap::createFromPath( filename ));
 #else
--- a/src/common/gui/SurgeBitmaps.cpp
+++ b/src/common/gui/SurgeBitmaps.cpp
@@ -86,6 +86,11 @@ void SurgeBitmaps::addEntry(int id)
 
 #ifdef USE_SCALABLE_BITMAPS
    VSTGUI::CBitmap *bitmap = new CScalableBitmap(CResourceDescription(id));
+#elif __linux__
+   char filename [1024];
+   snprintf (filename, 1024, "./resources/bitmaps/bmp%05d.png", id);
+   filename[1023] = '\0';
+   VSTGUI::CBitmap* bitmap = new VSTGUI::CBitmap(IPlatformBitmap::createFromPath( filename ));
 #else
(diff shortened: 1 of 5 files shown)


baconpaul:
But cool!

falkTX:
This is how it looks so far:

image

there are some oddities happening (kinda flickering) when changing values.
but after a few fixes, it runs quite nicely. :)

jsakkine:
@baconpaul, @falkTX, maybe we could merge these with a `FIXME` comment. It is a start.

jsakkine:
`// FIXME: Use scalable bitmaps after the resource manager has been landed`?

jsakkine:
I see developer value in this albeit it is not production quality e.g. others can continue while I work on the RM. That means also less pressure for me from other developers, which often implies better quality code.

jsakkine:
But definitely **must have** the FIXME comment.

falkTX:
I added a FIXME comment.
Also resources loading from /usr/share I guess makes more sense that way.

Note that PR #442 needs to be merged before this.
This PR has it included, because I am using this one as the working branch.

src/common/gui/SurgeBitmaps.cpp
```diff
@@ -86,6 +86,11 @@ void SurgeBitmaps::addEntry(int id)
 
 #ifdef USE_SCALABLE_BITMAPS
    VSTGUI::CBitmap *bitmap = new CScalableBitmap(CResourceDescription(id));
+#elif __linux__
+   char filename [1024];
+   snprintf (filename, 1024, "./resources/bitmaps/bmp%05d.png", id);
+   filename[1023] = '\0';
+   VSTGUI::CBitmap* bitmap = new VSTGUI::CBitmap(IPlatformBitmap::createFromPath( filename ));
```
jsakkine: std::string() and c_str().

falkTX: how do you string format it? we need the `%05d` bit..

jsakkine: Did some experimenting and a tes program:

```
#include <iomanip>
#include <iostream>
#include <sstream>

int main(int argc, char **argv)
{
    if (argc < 2)
    {
        std::cout << "Usage: " << argv[0] << "<number 1> ... <number N>"
                  << std::endl;
    }

    for (int i = 1; i < argc; i++)
    {
        int num = std::stoi(argv[i]);
        if (num >= 0 && num < 99999) {
            std::stringstream ss;
            ss << std::setw(5) << std::setfill('0') << num;
            std::cout << ss.str() << std::endl;
        }
    }

    return 0;
}
```

Sample output:

```
$ ./test 123 456 78901 23
00123
00456
78901
00023
```

falkTX: neat, though a bit overkill.
this is temporary code that is going to be replaced soon, once we load resources from memory...

baconpaul:
No problem keeping it open

I may move it to c scalable bitmap and make a function to get thenplatform bitmap. For now use your file load. Then there’s a place for @jsakkine to put his memory loader. Will post here if I do.

baconpaul:
I am pretty sure #443 does what this does, except for the /usr/share bit. I suggest we close this and sweep that, but would appreciate your view so we can form consensus before I press the buttons! :)

falkTX:
fine to close this one by me. I will go ahead and do it

Issue #454, opened 2019-02-02
W32 builds need a distinct target


baconpaul:
Right now W32 and X64 both create target/vst2/Release/Surge.dll

Make it so that W32 changes the name so we can build all 4 dlls at once.

esaruoho:
sounds awesome! nice :)

esaruoho:
yeah, for sure, now that people are downloading automated builds that are actually 32-bit versions.. here's an interesting thing for ya @kzantow :)

baconpaul:
Fixed with that PR

PR #463, opened 2019-02-02
Use a different name for 32 bit dll

baconpaul:
The target name for 64 and 32 was identical so the nightlies
were accidentally distroing the 32 bit plugin. This fixes
that by making the 32 bit dll have the name Surge32.dll.

This means in bitwig I can have all 4 flavors installed and pick
and choose between them!


baconpaul:
Addresses #454

PR #466, opened 2019-02-02
Move scalableresource.h

### LINKED ISSUE #453: Move scalable resource header
When @jsakkine did linux resource loader we realized I put scalaberesources.h in the wrong directory. It should be in src/windows. We should correct the script that makes it and move the file.

baconpaul:
When @jsakkine was doing linux resources, he and I realized
that I had put this file in the wrong directory. It should be
in windows and conditionally included. Closes #453


jsakkine:
Yeah, definitely in.

PR #468, opened 2019-02-02
Replace __linux__ with LINUX

jsakkine:
Make the Linux compilation flag more inline how the flags for macOS and
Windows have been named. Someone might say that, why don't you pick UNIX to
make code more compatible with CMake, which we plan to eventually to
migrate. There is a predefined flag called APPLE on CMake for macOS but
also UNIX is true on macOS compilation.

What people tend to do in their CMakeLists.txt usually is:

if(UNIX AND NOT APPLE)
    set(LINUX TRUE)
endif()

So, in a distant way, this commit makes also the code more ready for the
eventual CMake migration.

### DIFF
--- a/premake5.lua
+++ b/premake5.lua
@@ -77,7 +77,10 @@ elseif (os.istarget("linux")) then
     {
     }
 
-    defines { "WINDOWS=0" }
+    defines
+    {
+        "LINUX=1",
+    }
 
     buildoptions { "-std=c++17" }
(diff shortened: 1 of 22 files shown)


kzantow:
@jsakkine LGTM, and linux building fine note: I squashed it so rebase to `master`

jsakkine:
@kzantow, Thank you. :)

PR #470, opened 2019-02-02
Change #ifndef WINDOWS to #if WINDOWS in unitconversion.h

jsakkine:
Caused a compilation error when "WINDOWS=0" define was remove from
premake5.lua for Linux. This commit replaces it with a more proper
"#if WINDOWS" as it should be.

### DIFF
--- a/premake5.lua
+++ b/premake5.lua
@@ -46,7 +46,7 @@ if (os.istarget("macosx")) then
     {
     }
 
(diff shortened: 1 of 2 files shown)


baconpaul:
In fact: Wasn't windows=0 or windows=1 on all platforms prior, so ifndef windows was never true?

jsakkine:
OK cool :) even better

Issue #469, opened 2019-02-02
Release version of the Linux plugin crashes


jsakkine:
Debug version works but the release version crashes somewhere in the function load_mid_controllers().

jsakkine:
Noticed while working on my resource manager. The issue has predated that one.

baconpaul:
Yeah I saw the same when I tried the release version; agree this is not your changes.

baconpaul:
OK I know why this is happening and it is because of a very evil non standard thing in the xml parser. I actually trued to fix this when I did the mac port and it broke a lot of stuff.

load_mid_controllrs (and elsewhere) has code which looks like this

```
   TiXmlElement* mc = getSnapshotSection("midictrl");
   assert(mc);
   
   TiXmlElement* entry = mc->FirstChild("entry")->ToElement():
```

Now you may look at that and say "hold on; how can mc->FirstChild("entry")->ToElement be safe if mc->FirstChild("entry") is null"?

And you would be right.

But the people who wrote tinyxml thought about solving (or should I say "solving") that in a nasty way

```
	TiXmlElement*  ToElement()			{ return ( this && type == ELEMENT  ) ? (TiXmlElement*)  this : 0; } ///< Cast to a more defined type. Will return null not of the requested type.
```

That's right. ToElement actually checks that 'this' is never null.

Now on mac when you compile you get a bunch of warnings saying "this can never be null and it may be ignored". Because, hey, this can never be null and it may be ignored.

So when gcc -O3 runs, it knows this can never be null and branch jumps directly to side 2 of the &&

and then dumps core

because that code is wrong

but is used everywhere

So fun. Our choices are

```
TiXmlElement* entry = (mc->FirstChild("entry")?mc->FirstChild("entry")->ToElement():NULL);
```

lines like that through SurgeStorage (which I've done everywhere) or alternately do something clever with a return type other than NULL which is == NULL and replace the this check with it.

But that's your problem.

It is blind luck, by the way, that llvm on mac and cl on windows don't optimize that away. One day a compiler upgrade could just as easily make mac or windows core dump.

Ugh.

baconpaul:
The easiest (but least pleasant) solution is to introduce a macro TO_ELEMENT which SurgeStorage, CSnapshotMenu and SurgePatch all have which is

```
#define TO_ELEMENT(expr) (expr?expr->ToElement():NULL)
```

and then change the ->ToElement calls in the code.

[...]

PR #471, opened 2019-02-02
Replace ->ToElement with dynamic cast

baconpaul:
The TiXmlDocument thing does a check in a member function of
"if(this &&" which is non standard. Windows is fine, mac warns.
But gcc -O3 correctly optimizes it away. That optimization means
that linux release throws SEGV all over the place. Since the function
is replicating a dynamic_cast, replace it with that. See issue #469
for a more fulsome conversation of this change.

I want to test this a bit more on windows and mac before we sweep it, but am putting it up here so @falkTX and @jsakkine can (1) offer review comments and (2) sweep it into their builds so they can continue progress

### DIFF
--- a/src/common/SurgePatch.cpp
+++ b/src/common/SurgePatch.cpp
@@ -892,14 +892,14 @@ void SurgePatch::load_xml(const void* data, int datasize, bool is_preset)
    for (auto& i : fx)
       i.type.val.i = fxt_off;
 
-   TiXmlElement* patch = doc.FirstChild("patch")->ToElement();
+   TiXmlElement* patch = TINYXML_SAFE_TO_ELEMENT(doc.FirstChild("patch"));
    if (!patch)
       return;
 
    int revision = 0;
    patch->QueryIntAttribute("revision", &revision);
 
-   TiXmlElement* meta = patch->FirstChild("meta")->ToElement();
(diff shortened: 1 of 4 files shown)


baconpaul:
Confirmed this works just fine on mac and windows.

Your guys option. I think it's the best fix we have. But glad to take a more drastic approach if one of you wants to do that.

jsakkine:
Is this same regression in newer tinyxml versions?

jsakkine:
Bringing RTTI is more heavyweight than updating a library.

falkTX:
Yeah, best we can do considering circumstances.

jsakkine:
Hold on. Can you just check if FirstChild returns null? Remove the chaining.

jsakkine:
RTTI brings is like full stack of unnecessary bloat. Lets reconsider options before taking action.

baconpaul:
OK here's the macro version. Comments welcome. Still fixes release (along with new emit and log2 removed).

baconpaul:
and this PR *doesn't* contain the new emit or the log2 gone.

jsakkine:
OK, back at "work" (or should . I say hobby). LGTM.

baconpaul:
Merged

PR #478, opened 2019-02-03
build-linux.sh: Do not use the option syntax for commands

jsakkine:
My plan is to parse options with getopt, and differentiate them from
these command, e.g. to select debug/release and the type of plugin.

### DIFF
--- a/build-linux.sh
+++ b/build-linux.sh
@@ -214,22 +214,22 @@ fi
 prerequisite_check
 
 case $command in
-    --premake)
+    premake)
         run_premake
         ;;
-    --build)
+    build)
         run_all_builds
         ;;
-    --install-local)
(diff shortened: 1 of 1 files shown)


jsakkine:
Force-push just to fix typos in the commit message.

baconpaul:
Does this change how you call it? If so does the azure pipeline need an update?

jsakkine:
Does Azure pipeline use it already? If so, yes.

Anyway, I will include this PR to my PR that updates premake5.lua to create makefiles that do code generation for every build. I'll close this one.

baconpaul:
Yup and I see you did too. Cool!

baconpaul:
Just flagging @kzantow so he knows

PR #485, opened 2019-02-03
linux: Read the default configuration from memory

jsakkine:
Bundle configuration.xml with the DSO binary and parse it directly from the
memory. Saving should be postponed to the stage when we have plugin
settings (e.g. save setting buttons there).

Fixes: 465 ("Have a default configuration.xml bundled for Linux")

### DIFF
--- a/premake5.lua
+++ b/premake5.lua
@@ -80,7 +80,8 @@ elseif (os.istarget("linux")) then
 
     files
     {
-        "src/linux/ScalablePiggy.S"
+        "src/linux/ConfigurationXml.S",
+        "src/linux/ScalablePiggy.S",
     }
 
     includedirs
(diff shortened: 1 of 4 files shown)


baconpaul:
Looks good. Only comment is “configuration.h” is not a great file name. Mind changing it to something like “configurationxmlmemory” capitalizes as you see fit? Then I will happily merge

jsakkine:
@baconpaul, mind if I name it ConfigurationXml.h/S.

baconpaul:
Perfext

baconpaul:
Indeed it has! Great progress.

baconpaul:
Oh shoot. Not perfect!

This change means that data path isn't set in linux any more. This means that no patches or wavetables are loaded.

So the synth runs but it doesn't have anything else available even if you have them properly installed in the directory location we had chosen.

PR #489, opened 2019-02-03
Original Skin: Fixing FX diagram

itsmedavep:
I had missed the arrow going into the "+" in the signal path. Updated now.

screen shot 2019-02-03 at 12 16 57 pm


baconpaul:
Thanks!!

Issue #487, opened 2019-02-03
Linux Paths


baconpaul:
@jsakkine change https://github.com/surge-synthesizer/surge/pull/485 accidentally removed the initialization of datapath and userdatapath on linux

This means that, even though configuration.xml comes from memory, no patches or wavetables are loaded

To reproduce build the release version and click in the patch browser or switch to wavetables in the oscillator viewer. Nothing.

Also you can't save a patch.

The fix is to restore this bit of code

```
    string home = string(getenv("HOME"));
	   char* data_home_cstr = getenv("XDG_DATA_HOME");
	   string data_home = (data_home_cstr) ? string(data_home_cstr)
	                                       : home + "/.local/share";
	
	   datapath = data_home + "/Surge/";
	
	   userDataPath = home + "/Documents/Surge";
```

or an equivalent.

baconpaul:
Thanks!

baconpaul:
I would have restored but I still am not entire sure which paths you guys want as final for those! :)

PR #491, opened 2019-02-04
linux: Restore datapath and userDataPath calculations

jsakkine:
In a95e810921a3, removed datapath and userDataPath calculations by
mistake. This commit restores them. In addition, check on macOS and
Linux that the environment HOME, because at least artificially you
can create an environment where it doesn't.

Fixes: #487 ("Linux Paths")


jsakkine:
Following the campground rule i.e. the HOME existence check.

PR #498, opened 2019-02-04
Keycodes restored

### LINKED ISSUE #427: Keyboard hook not unregistered because it is called on every patch change
**Describe the bug**
Leaking hooks is not a good idea

**To Reproduce**
1. Go to 'Surge'
2. Click on preset and select one
3. Repeat
4. Keyboard hooks are leaked, one for each patch change

**Expected behavior**
No Leaks. (BTW There are many memory leaks as well). Depending on the implementation, no crashing later when hook is called.

**Additional context**
In  SurgeGUIEditor::openOrRecreateEditor This function is called:

 frame->registerKeyboardHook(this);

Simple fix is to move to SurgeGUIEditor::open and not call this function from SurgeGUIEditor::openOrRecreateEditor.

### LINKED ISSUE #479: Alt or Ctrl +- to toggle zoom up and down
Do this along with the #433 and default zoom and the max zoom change

### LINKED ISSUE #496: Keyboard Hook doesn't work MACOS (probably elsewhere)
In addition to #427 the keyboard hook generally doesn't do what you'd want. I think it is just coded incorrectly. Right now it is looking at KeyCode.character for things like LEFT and RIGHT, when those control keys send a character of 0.

Came across this doing #479 so will hold that one too, to do some general keyboard hook cleanups including fixing the leak referenced above, making existing functions work, and adding keyboard zoom.

baconpaul:
The onKeyDown handler was reading the wrong part of the data
structure, so keypresses didn't navigate patches. Moreover, the
handler was over-registered. Finally, once those problems were
fixes, make + and - zoom and unzoom the UI.

Closes #496 Keyboard Hook doesn't work
Closes #479 +/- for zoom
Closes #427 Keyboard hook over-registered

### DIFF
--- a/src/common/gui/SurgeGUIEditor.cpp
+++ b/src/common/gui/SurgeGUIEditor.cpp
@@ -419,44 +419,44 @@ void SurgeGUIEditor::refresh_mod()
 
 int32_t SurgeGUIEditor::onKeyDown(const VstKeyCode& code, CFrame* frame)
 {
-   switch (code.character)
-   {
-   case VKEY_ALT:
-      // modsource = modsource_editor;
-      // queue_refresh = true;
-      mod_editor = true;
-      refresh_mod();
-      return 1;
-   case VKEY_TAB:
-      toggle_mod_editing();
-      return 1;
-   case VKEY_LEFT:
-      synth->incrementCategory(false);
-      return 1;
-   case VKEY_RIGHT:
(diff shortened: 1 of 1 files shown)


baconpaul:
Huh the zoom + doesn't work on windows reliably I just realized.

Will close this PR and come back to it tonight.

PR #508, opened 2019-02-05
some minor tweaks to readme.md

esaruoho:
one more `--` removal

@baconpaul i think i guess this is alright to do?

### DIFF
--- a/README.md
+++ b/README.md
@@ -248,7 +248,7 @@ if you desire, but the build script will also do it.
 ./build-linux.sh install-local
 ```
 
-For other options, you can do `./build-linux.sh --help`.
+For other options, you can do `./build-linux.sh help`.
 
 ## References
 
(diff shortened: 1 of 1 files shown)


baconpaul:
Nope - help actually needs the dashes!

esaruoho:
ok, i'll fix it :D

baconpaul:
But we should fix the script so it also works. Happy to sweep the rest - just lemme know

esaruoho:
> But we should fix the script so it also works. Happy to sweep the rest - just lemme know

well i put it back to `--help`, hope that's fine. it'd be nice if the script was fixed eventually, but overall feels like maybe these tweaks + tiny fixes can be swept in.

baconpaul:
Yeah cool. Thanks!

PR #519, opened 2019-02-06
Initalize member variables

baconpaul:
Valgrnd on the VST2 shows some switch paths without an initialized
switch. This cleans up those warnings by making sure things are
initialized to reasonable values.

### DIFF
--- a/src/common/Parameter.cpp
+++ b/src/common/Parameter.cpp
@@ -527,6 +527,8 @@ void Parameter::set_type(int ctrltype)
       sprintf(dispname, "-");
       valtype = vt_int;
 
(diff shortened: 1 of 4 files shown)


baconpaul:
@jsakkine @falkTX this cleans up the valgrnd uninitialized switch errors on startup. A peek would be appreciated. I tested it works mac win and linux unchanged and used reasonable defaults (and definitely better defaults than "uninitialized memory!")

Thanks

baconpaul:
Thank you. And yeah I think there’s quite a few still. But will merge this one! Appreciate the test

PR #529, opened 2019-02-07
Fix Linux Drawing Errors

### LINKED ISSUE #513: Linux Drawing Errors
The CHSwitches are very eager to disappear. This is code in src/common/gui/CHSwitch2.cpp and .h mostly. Haven't debugged at all yet.

linux_draw_one

OS:
```
paul@ubuntu:~/surge$ uname -a
Linux ubuntu 4.15.0-45-generic #48-Ubuntu SMP Tue Jan 29 16:28:13 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux
```

Host:
```
 ~/Carla_2.0-RC3-linux64/carla/carla-bridge-native vst2 ./target/vst2/Release/Surge.so '(none)' '0'
```

baconpaul:
Linux drawing was mis-rendering terribly. Turns out this was
because, for some reason, the inbound alpha to bitmap::draw
was 255, not 1.0f, and cairocontext.cpp has a branch decision
based on equality with 1.0f. This indicates there may still be
a problem in vstgui (why was the 255 inbound?) but an interim
fix on linux is to reset alpha=255 to alpha=1.0 which completely
fixes the painting issues reported in #513

Closes #513 - Linux Painting Issues

### DIFF
--- a/src/common/gui/CScalableBitmap.cpp
+++ b/src/common/gui/CScalableBitmap.cpp
@@ -159,6 +159,19 @@ void CScalableBitmap::draw (CDrawContext* context, const CRect& rect, const CPoi
         CPoint ncoff = offset;
         CPoint no = invtf.transform(ncoff);
         
+#if LINUX
+        /*
+        ** For some reason alpha = 255 inbound to this function on linux (and maybe on
+        ** mac and windows) but unlike mac and windows that misscaling actually causes
+        ** draw errors on linux because of the comparison in cairocontext.cpp at line
+        ** 396 (the (if (alpha != 1.f )) so hand-rescale it here for now
+        **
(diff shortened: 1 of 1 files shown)


falkTX:
Very interesting, if it is this simple, then very nice!

baconpaul:
Well it is that simple to work around the bug. We should really find the actual bug that sends 255 in but if it's fine with you (which it seems it is) let me merge this one. Yes?

Have one coming for menu positioning in zoom also in a minute.

Thanks for looking.

falkTX:
I can confirm this fixes the random disappearing elements in the Linux build :+1:

baconpaul:
Thank you! Merging now. Yay!

Issue #525, opened 2019-02-06
Linux: When mis-installed crashes on drop down menu


tank-trax:
When the left hand menu below the Oscillator is selected a drop down menu appears.

If the VST is in its initial state,
If clicking on either Wavetable or Window the VST and DAW/Host will crash
The other selections will not crash the app.

![vlcsnap-2019-02-06-15h25m57s880](https://user-images.githubusercontent.com/41337033/52372772-fa6bbf80-2a26-11e9-9b1d-62e15ba37875.png)

However if a Preset is loaded clicking on Waveform will not crash the app however clicking on Window will still cause a crash.

![vlcsnap-2019-02-06-15h26m40s276](https://user-images.githubusercontent.com/41337033/52372821-1b341500-2a27-11e9-82b3-edc3db6b8ad6.png)

Using Debian Stretch KDE plasma and Fast Track Pro USB audio sound card

baconpaul:
Thank you again. These are really clear and useful bug reports!

baconpaul:
Just quickly: Are these 32 or 64 bit builds? Can you screen shot the Menu/About screen for us?

I poked at it a bit tonight and didn't get exactly your crashes so want to make sure I'm doing apples to apples.

Thank you.

tank-trax:
I only made 64 bit builds.

I was about to ask if you could explain where or what is the Menu/About screen, then I saw the Menu link at the bottom right hand corner. Funny thing is I never saw it until now... because it stays hidden in Linux until there is a Mouse Down/Click on that area. It does appear in the screen shot. 

Bad news is resizing does not work in Reaper. Good news is it works in Carla. However once resized the Menu option is dead. I noticed why. If made smaller the window that opens is misaligned towards the centre. I am thinking when enlarged the menu is flying off the grid.

FYI. The About and Manual links are dead.

Would it be possible to get involved on the design part and if so who would I contact to work on the GUI? 

I am seeing some other issues with the menu that I would post soon.

tank-trax:
Here is an example of the menu displacement after resizing...

![surge resize 01 gif](https://user-images.githubusercontent.com/41337033/52389571-0f197900-2a62-11e9-9da8-10feca69767e.gif)

baconpaul:
Yeah that menu repaint is part of Linux paint generally

[...]

PR #530, opened 2019-02-07
More Zoom VSTGUI Workarounds

baconpaul:
Linux apparently has even one more extra transform of
the frame position when scale != 1; so unwind one
extra with a comment. We should fix whatever code paths
are making this wrong once we have a vstgui fork, obviously.

Fixes a problem reported in #525 by @tank-trax

### DIFF
--- a/src/common/gui/SurgeGUIEditor.cpp
+++ b/src/common/gui/SurgeGUIEditor.cpp
@@ -2595,7 +2595,20 @@ CPoint SurgeGUIEditor::getCurrentMouseLocationCorrectedForVSTGUIBugs()
     CGraphicsTransform vstfix = frame->getTransform().inverse();
     vstfix.transform(where);
     vstfix.transform(where);
+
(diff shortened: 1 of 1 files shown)


baconpaul:
Heya @falkTX since you are online could you confirm this one too? Basically with the other one applied, zoom to 150%, right mouse on a slider, and see if the menu appears in the correct place?

Thank you very much.

falkTX:
I can confirm the issue on the original code, and the change that fixes it.
Also that menus still appear out of bounds and unclickable too :P
Another thing I found is that the size of the menu does not scale, so the words inside can sometimes be cropped off.

baconpaul:
Yeah menus under scale are not great - I think the fix to make this series of transforms go away has to include that too. But that’s a lot of vstgui debugging...

baconpaul:
And thank you for testing! I will merge this one too!

PR #546, opened 2019-02-09
Use Lato on MacOS

baconpaul:
As described in #214, we want to use the OFL-licensed font
Lato for all components in Surge with the constraint that we
want to ship the font with the DLL and not require it to be installed
on an end users system.

This change implements that in macOS by doing the following

1: Introducing the font to resources/fonts
2: Moving the fonts to the bundle with scripts/macOS/package*
3: Moving the font initiation for minifont and patchfont to
   runtime rather than surge dll load time (but only do this
   for fonts which USE_RUNTIME_LOADED_FONTS, which right now
   is only MAC. Once all platforms are done, that switch will
   disappear)
4: Write the MacOS specific CoreText code to load a font from
   memory into the system


baconpaul:
Heya @jsakkine if you get a moment to look this weekend would appreciate your careful eyes as always. It's actually not much code but since we are staging it over several builds as we do different platforms, want to make sure you are happy with how I started on this as mac-only.

baconpaul:
After review I will debase this just go check #549 works still!! Before push
Review still appreciated at this stage. Rebase won't change be the code
Thanks

baconpaul:
OK just rebased to make sure @rghvdberg linux change is OK with this mac change and it is.

baconpaul:
Ever have one of those moments where you think about a bit of code and realize you could have done it different ly and more cleverly 24 hours after you grapple with it?

Well I just had one of those. Closing this now and will be back with another version end of weekend or early next week

PR #549, opened 2019-02-09
use sans-serif on linux

rghvdberg:
makes the gui load on ubuntu 16.04 / mint 18


baconpaul:
Happy to merge this
Will be more happy to remove it when #214 wraps up
This may conflict out my #546 but I'll resolve it if so

Issue #486, opened 2019-02-03
Piano keys to the bottom of Surge


jsakkine:
Would be really nice to have for testing (especially on Linux where we still mostly use Carla) piano keys in the bottom of the UI, maybe with a show/hide button. Does VSTGUI have something of-the-shelf for this?

baconpaul:
I almost did this on Mac also but then just made midi routing work in the hosts and uses someone else’s keyboard

Afaics there is no such widget

A much easier thing is to add some special keystrokes which put you in midi keyboard mode. So press CTRL alt x r 17 or something and then a sends an a note

But Carla should have a routing screen keyboard available no? May be easier!

jsakkine:
Ping @falkTX (again) :)

jsakkine:
@baconpaul, I believe you, but feels that such isn't in the default stock of widgets.

falkTX:
I have almost no experience with vstgui, cant help much here.
you can use carla on mac and windows though, which provides the keyboard widget thing playable with pc keys :)

baconpaul:
Yup; bitwig does it with caps lock too on all platforms; and hosting AU has keyboard input as does logic on mac. So I think you and I are saying the same thing, @falkTX which is it is better to get running in one host and then use the host to route midi than to kill ourselves adding a keyboard.

baconpaul:
@falkTX does “Carla-single” have a way to generate midi into the plugin? That would completely solve the problem for @jsakkine. (That’s basically what I use hosting AU for on mac; a super lightweight AU host which can route audio and so on and starts in a second, not in 14 seconds like full blown logic)

falkTX:
> @falkTX does “Carla-single” have a way to generate midi into the plugin? That would completely solve the problem for @jsakkine.

carla-single plugs into JACK in a modular way, exposing midi and audio ports. so you can use then `jack-keyboard` to feed midi into it.

regarding juce: since this is already cross-platform, there is no need for bloating the code with juce stuff.
also, I got surge to build with my own tiny framework (DPF), this is how I made the first advances regarding the event loop. but for now getting the VST2 version out is already awesome.

jsakkine:
@baconpaul, Yes, this was not really proposal to use Qt right now. This was more like explanation why JUCE would be a bad choice at any point :) I could do a PR and derive something based on that text directly to our developer documentation. It is a design choice that should be documented.

If someone puts issue suggesting JUCE we can point them to the documentation.

[...]

PR #564, opened 2019-02-10
Add a note about JUCE and Qt to the documentation

jsakkine:
Summarizes the discussion in #486.

Closes: #486

### DIFF
--- a/doc/Architecture.md
+++ b/doc/Architecture.md
@@ -1,0 +1,13 @@
+## 3rd Party Frameworks
+
+### JUCE
+
+JUCE makes adapting to different audio APIs easy, which we don't need because we
+have our own code for that. It is plug-in framework that makes many things easy
+at the price of flexibility.
+
+### Qt
+
+Qt might be in a distant future an option worth of considering because we would
+get rid alot of Cocoa and Windows specific code. In addition the flexibility
+how things can be done would be compared to VSTGUI from a different dimension.


jsakkine:
Seems that Github recognizes that Closes tag, which I did not know before (shows a tooltip).

baconpaul:
Yeah looks good.

Redoing zoom without cocoa in AU will be a pain. But otherwise it looks like a fine summary of our discussions.

Wonder if dpf may also be an option?

But I'm happy with your doc!

jsakkine:
@baconpaul, maybe so but we still would reduce the amount specialized code. I don't know what dpf is but I know that Qt is the best application framework available for C++. And the subsection does not dictate to use it ever. It is just the current analysis of it. It is mentioned as a viable option.

jsakkine:
@baconpaul, good to here. Lets not **create** an architecture specifications. Lets instead **evolve** one over time. Otherwise we end up with a document that is both useless and unreadable.

jsakkine:
Can we close this now?

baconpaul:
Yeah sure! Fine with me.

falkTX:
Huh, Qt for GUIs is a really terrible idea.
It is a huge codebase which we would have to build statically for the plugin, just for a few OS independent things...

jsakkine:
@falkTX, there is no roadmap ATM to use Qt. And there is no requirement to link Qt statically even if we end up using it.

jsakkine:
Actually Windows would be a problem without static linking , so we would require static linking. However, static linking is not by itself any kind of problem.

- It does not really matter how fat the resulting binary is. Kernel mm will fault only the pages that it actually uses in Windows, macOS and Linux.
- Huge codebase is not a problem as it is a mature toolkit. We don't care all that much about Qt's source code.
- We are compatible in terms of licensing for static linkage. Qt is licensed with LGPL3 (among some other licensing options).

Only thing that it would affect is that the Qt SDK (which is free) would be required to compile Surge. Not a big deal.

PR #565, opened 2019-02-10
Context Menu: Set Controller To... "sort subfolders into groups of 20", "show CC127"

### LINKED ISSUE #558: Sub-folder says 120-127, but the last bindable CC is 126.
I'm experimenting with modifying the "subfolder split code" from "split by 10" to "split by 20". looks good so far, but raises weird questions:
Where's CC127? last CC is CC126.
image

esaruoho:
this is without all the spacing fixes. easier to review.

this: sorts subfolders by 20 instead of by 10 (less subfolders!)

also makes CC#127 bindable.

52533832-6198b500-2d42-11e9-9eec-e1c38956bfbb

fixes #558

### DIFF
--- a/src/common/gui/SurgeGUIEditor.cpp
+++ b/src/common/gui/SurgeGUIEditor.cpp
@@ -1561,7 +1561,7 @@ int32_t SurgeGUIEditor::controlModifierClicked(CControl* control, CButtonState b
              
              contextMenu->addEntry("-", eid++);
              
-             // Construct submenus for expicit controller mapping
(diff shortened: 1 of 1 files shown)


baconpaul:
Lets not merge this and instead just do #566 properly. That ok @esaruoho? The bad UI (which is entirely my fault) isn't much better with 20 terrible things in a group rather than 10!

If you feel strongly happy to merge also.

esaruoho:
Being able to select like with the dropdownmenu is good stuff but if there is better in the horizon then sure! Nvm the PR then

PR #573, opened 2019-02-10
Include Steinberg PR responses

baconpaul:
We sent 3 pull requests to steinberg for vstgui and they correctly
responded that 2/3 were indicative of us not using their API correctly

As such I've removed those commits from vstgui/surge, applied the
changes to our code, and have confirmed that the two issues
(capture on mouse click and diacriticals and emojis in patch
and category names) work with the vstgui without the associated
incorrect patches.

The final patch for checmark submenus is still in vstgui.surge in
this commit

### DIFF
--- a/src/common/SurgeStorage.cpp
+++ b/src/common/SurgeStorage.cpp
@@ -341,8 +341,26 @@ void SurgeStorage::refreshPatchlistAddDir(bool userDir, string subdir)
    for (auto &p : alldirs )
    {
       PatchCategory c;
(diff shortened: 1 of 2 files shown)


baconpaul:
@jsakkine would appreciate a quick glance. Didn't see a better way to do that UTF conversion in SurgeStorage than an #if WINDOWS.

jsakkine:
Looks good to me! Please merge.

baconpaul:
Thank you!

PR #577, opened 2019-02-10
doc/Architecture.md: Add a note about Qt and static linkage

jsakkine:
Added a comment about Qt and static linkage based on feedback from @falkTX.


falkTX:
If you link against shared library on Linux (or any other package-based OS), you end up with the very same issue I was warning about with Gtk and vstgui!
Please, STOP USING BIG UI FRAMEWORKS TO MAKE PLUGINS. (specially those that ship shared libraries by default)
This has been repeated for years now, and feels a bit useless...

jsakkine:
1. Static linkage is the route anyway since Windows sets a hard requirement for it.
2. According to Qt Company Qt 5.9 takes from 5.4MB or 13.4MB depending on whether you use the standard or the lite version [1]. To get some comparison figures, Massive.vst takes 75MB on my MacBook Pro.
3. Static linkage prevents causing the same problem as there was with GTK2.

I think neither shouting nor or using authorities (e.g. this has been discussed however many years) will take far.

[1] https://blog.qt.io/blog/2017/05/31/qt-lite-qt-5-9-lts/

jsakkine:
And this not the same as decision to use Qt. It doesn't hurt to have analysis about it in the architecture documentation.

jsakkine:
Updated with a note about Qt's size metrics.

jsakkine:
Professionally I do a lot of security related work for Linux and in that context I do understand the goal of reducing the size of the text segment. In that context the reason is obviously to keep the attack surface low for return-oriented-programming type of attacks.

In the context of the software synths, pragmatically speaking, I don't think there is any reason to care about the size that much as long as it isn't like hundreds of megabytes. They do not need to meet extreme security requirements.

jsakkine:
Not to mention that Qt would somewhat leverage the amount of work when we do an iOS version of Surge in distant future.

jsakkine:
Updated the commit again to have a note about the mobile versions.

jsakkine:
@thiagomacieira, is my analysis about the constraints to use Qt about right?

baconpaul:
Whoa guys this is a lot of argument about something we aren’t doing! We all ok? 

(Also iOS requires auv3 to work properly in GarageBand iPad and stuff which is a lot of work)

jsakkine:
Not in 1.6 timeline but definitely mobile port is a feasible goal in distant future. Landing ARM DSP within next few days. My thinking is that already 1.6 scales internally to both mobile to desktop, it is almost done. Please read my analysis carefully and point out the wrong deductions. How the visible interface is actually implemented is an orthogonal concern.

jsakkine:
@paynebc, Could you also take a quick peak what I wrote about Qt? Remember, I was the guy who bothered you with questions about how you implemented TuneFish. Sorry to bother again :)

thiagomacieira:
Nothing wrong with your analysis. LGPLv3 does allow for static linking, especially if the application in question is open source. YMMV for the size numbers, of course, so the reduction to 5.4 MB may not match your results, especially since those came from professional people who knew what they could disable. But in general, static linking *does* result in shrunken binaries, compared to a full-size DLL/shared library.

jsakkine:
@thiagomacieira, Thank you for your feedback!
@kzantow, Any clues how this can cause the cloud build fail? Updates only a .md file.

jsakkine:
@baconpaul, Can be merge this soonish? Facts should be correct. This is just making notes instead of having to dig them from closed issues and PR's. I cherry-pick this to bigger "doc update" if that is what you prefer.

baconpaul:
Uhh I’ll be honest I haven’t had time to read this whole thread. I can but not today. Is it still controlvrrsial?

jsakkine:
@baconpaul, there has never been anything controversial in this one. Just writing down what we have learned from issues and pull requests and email discussions.

baconpaul:
Ok! Let me take some time to look. Busy few days sorry!

jsakkine:
No rush with this!

kzantow:
@jsakkine odd about the failure. it's complaining about something to do with `vstgui.surge` maybe just update to the latest `master`?

falkTX:
Sorry for shouting, but it pains me to see the same mistakes being repeated over and over again.
On Linux we have been struggling with projects that decide to use big toolkits for their plugin UIs, just for convenience I guess, but ends up being a mess to solve later on.

If it comes to the point where surge really needs something more than vstgui, Juce should be the way to go. It already supports AUv3 and iOS, as it seems to be the point here for considering Qt.

jsakkine:
Yet I'm not sure where does the documentation update go wrong. If there is a mistake, it should be easy to pinpoint that. If not, I don't see any reason this could not be merged. For JUCE the section can be extended with new remarks as they come.

baconpaul:
Tell you what guys I’m a bit further away from this than you both are and I know you are both reasonable and accurate devs. So let’s put this convo on pause until I get a chance to read the thread and the doc and see if I can understand what’s causing the angst here. I just haven’t had time to do that yet because of a busy non surge week. Sound ok?

baconpaul:
OK I read this. I think the problem is that (1) @falkTX thinks your analysis is correct but also (2) @falkTX thinks we should not couple to a big framework like QT, GTK, WX, TK/TCL or any such for our gui since the complexity at link, integrate, and run time is high. See, for instance, the GTK problems we dodged.

So everything you say about QT is correct, @jsakkine, but it seems to not be the stuff which @falkTX thinks is a relevant reason to not use QT.

I don't think it makes a lot of sense to (1) argue about stuff we aren't doing or (2) throw away analysis. I propose we to the top of this something like

------

## Frameworks

We often get asked about why we chose VSTGUI (answer: @kurasu open sourced it after porting to VSTGUI) and why we don't choose framework A, B, C or UI Library X, Y, or Z. Our current stance in Surge dev is to stick with VSTGUI for the foreseeable future. 

There's a long history of plugins binding to large frameworks and getting good results - Juce being a prime example - and a long history of plugins binding to large frameworks and getting bad results - complexities with GTK versions on linux, for instance.

But today for Surge our approach of VSTGUI plus hand implemented wrappers to plugin and headless styles works well enough. So at this point we are sticking with our current course at least until an imagined version 2.0. 

Along the way we recorded quite a bit of analysis of each choice, and record that here along in pull requests #xxx and #xxx, even though our collective decision was to not use any of these frameworks

--------

Then keep your analysis of the issues, which seems correct.

How does that make folks feel? It seems to me to be what we've decided, and so in that regard, I like writing it down.

baconpaul:
Oh and @falkTX reasons for not linking to QT make sense to me also - didn't mean to imply otherwise. Seems like a big compatibility layer with no direct benefit for our windows or mac ports from our current state.

Cool?

jsakkine:
Actually, there is no debate here whether to use QT or not. It is just a PR documenting some facts about QT.

baconpaul:
Yeah my prolog was meant to say that!

jsakkine:
@baconpaul,

Well the alternative proposal is not very accurate. We know exactly the reason why GTK2 and GTK3 cannot run in the same process. There are no random complexities when it comes to software.

I think it is already a failure in the documentation if it has to cross-reference Github pull requests.

I think that text you wrote could included to that section as the **Conclusion** for the time being but that does not give any rationale to hide any facts about other frameworks. JUCE and Qt are probably both kind of popular frameworks that might be proposed many times. It is good that the documentation shows that they are considered but this is our conclusion for now.

jsakkine:
IMHO, bringing WX or TK to this discussion is a bit archaic :) Better to have some analysis about stuff that is popular and widely used today...

jsakkine:
@baconpaul, I'd propose that based on your text I'll edit Conclusion subsection and amend that to the current commit. Then it is explicit that we stick to VSTGUI.

baconpaul:
Perfect!! And then let’s close this and fix menus instead :)

jsakkine:
@baconpaul, I'll update this tomorrow or day after tomorrow. Right now preparing PR for headless linux build so that I can finally close ARM support soonish.

PR #580, opened 2019-02-11
Use Lato on MacOS and Windows

baconpaul:
As described in #214, we want to use the OFL-licensed font
Lato for all components in Surge with the constraint that we
want to ship the font with the DLL and not require it to be installed
on an end users system.

To do this cross platforms, introduce a new C++ file
"RuntimeFont.h" which defines Surge::GUI::initializeRuntimeFontFromDistribution
and implement that in src/mac/RuntimeFontMac.cpp, Windows, etc.
This PR implements for Mac and Win and neither implements
nor references the symbol in Linux.

This change implements that in macOS by doing the following

1: Introducing the font to resources/fonts
2: Moving the fonts to the bundle with scripts/macOS/package*
3: Moving the font initiation for minifont and patchfont to
   runtime rather than surge dll load time (but only do this
   for fonts which USE_RUNTIME_LOADED_FONTS, which right now
   is only MAC. Once all platforms are done, that switch will
   disappear)
4: Write the MacOS specific CoreText code to load a font from
   memory into the system

This change implements that in Windows by doing the following

1: Including the font in the .rc file and subsequent res bundle
as a binary
2: Load the font and place it in a well known temp file
3: Load that temp file into the font registry using AddFontResource
4: Following the subsequent standard cfont implementation.


baconpaul:
@jsakkine a review greatly appreciated

@itsmedavep or @esaruoho if you want to build and test this too that would be lovely

No rush. Nothing here should conflict with anything.

Oh and: Linux was too hard. I gave up. vstgui makes the API too far away.

src/common/gui/SurgeGUIEditor.cpp
```diff
@@ -30,19 +30,28 @@
...
-/*
-#elif TARGET_VST3
-#include "surgeprocessor.h"
-#elif TARGET_VST2
-#include "vstlayer.h"
-#endif*/
-
-//#include <commctrl.h>
+
+#if MAC || WINDOWS
```
jsakkine: Should be done in premake5.lua, not here. Is use the standard for us for feature flags?

src/common/gui/RuntimeFont.h
```diff
@@ -0,0 +1,34 @@
...
+#include "vstgui/lib/cfont.h"
+#include "vstgui/lib/platform/iplatformfont.h"
+
+namespace Surge
+{
+namespace GUI
+{
+
+/**
+ * initializeRuntimeFontFromDistribution
```
jsakkine: `initializeRuntimeFont() - load the default font we use for VSTGUI`

src/common/gui/RuntimeFont.h
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * The role of this function is to load a font from the bundle, dll, so whatever
```
jsakkine: `a bundle, DLL or some other source`

src/common/gui/RuntimeFont.h
```diff
@@ -0,0 +1,34 @@
...
+ * and plumb through the VSTGUI to create the 9 and 14 point fonts onto the two
```
jsakkine: I don't understand what kind of act it is when you "plumb through".  I would start a fresh new sentence instead of a clause "The loaded font is ..." and explain preferably roughly in one or two sentences what is done with VSTGUI.

src/common/gui/RuntimeFont.h
```diff
@@ -0,0 +1,34 @@
...
+ * global variables "surge_minifont" and "surge_patchfont" if they are NULL
+ * at calltime.
+ *
+ * This is a remarkably OS-specific thing to do, so it is implemented in
+ * mac/win/linux RunTimeFontOS.cpp. (As of this writing it is neither implemented
+ * nor referenced on linux)
```
jsakkine: Just one sentence would be enough: "The implementation is OS specific.". That's it. We don't want over explain this in every function will be like this. And we do not want to keep backlog of the implementation status here.

src/common/gui/RuntimeFont.h
```diff
@@ -0,0 +1,34 @@
...
+namespace GUI
+{
+
+/**
+ * initializeRuntimeFontFromDistribution
+ *
+ * The role of this function is to load a font from the bundle, dll, so whatever
+ * and plumb through the VSTGUI to create the 9 and 14 point fonts onto the two
+ * global variables "surge_minifont" and "surge_patchfont" if they are NULL
+ * at calltime.
```
jsakkine: I'm missing the detail why Surge needs two different fonts.

src/common/gui/RuntimeFont.h
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * This is a remarkably OS-specific thing to do, so it is implemented in
+ * mac/win/linux RunTimeFontOS.cpp. (As of this writing it is neither implemented
+ * nor referenced on linux)
+ */
+void initializeRuntimeFontFromDistribution();
```
jsakkine: Could be just initializeRuntimeFont() IMHO.

src/common/gui/RuntimeFont.h
```diff
@@ -0,0 +1,34 @@
...
+
+}
+}
+
+/*
+** The two extern globals we need to initialize, which are defined/created
+** in SurgeGuieditor
+*/
+extern VSTGUI::CFontRef surge_minifont;
+extern VSTGUI::CFontRef surge_patchfont;
```
jsakkine: I think we could rename stuff every once in a while when we do something else related. You could do a prepending commit for this that renames these as:

- displayFont (maybe?)
- patchFont

No reason to use surge prefix.

With Git renames are a breeze:

```
git grep -l get_modulation_f01|xargs sed -i -e 's/get_modulation_f01/modToFactor/g'
```

(Ignore the specific example. One day I was just thinking that clearer names for those would be just modToFactor and factorToMod, not part of any PR ATM)

src/common/gui/SurgeGUIEditor.cpp
```diff
@@ -30,19 +30,28 @@
...
+#define USE_RUNTIME_LOADED_FONTS 1
+#else
+#define USE_RUNTIME_LOADED_FONTS 0
+#endif
```
jsakkine: I mean this :)

jsakkine:
OK, I'm ready with my review.

baconpaul:
Thank you. My guess is I will look at this tomorrow maybe Wednesday.

Sorry for using the idiomatic language “plumb through”! Kinda means attach from one end to another like when you are running pipes through a house. I’ll clean it all up.

Appreciated!

src/common/gui/SurgeGUIEditor.cpp
baconpaul: This is the way we staged zoom also. Have a HOST_SUPPORTS_ZOOM (I could do HOST_SUPPORTS_RUNTIME_FONT) and put it in the code; when we got the last host, remove the #ifdef from the code. I could move it to premake but that seems both more icky and less clear to the reader of the code. Mind if I leave it? (Happy to change the name of the variable. USES_RUNTIME_FONTS? HAS_RUNTIME_FONT_SUPPORT? What do you think?)

src/common/gui/RuntimeFont.h
baconpaul: I am happy to do this but would sooner do it as a separate PR and commit. I prefer text-only changes to be commits on their own not intermixed with function changes (and I know you do too!) So once this is swept I'll do another PR renaming the variables.

src/common/gui/SurgeGUIEditor.cpp
baconpaul: See comment above. I'm basically following the same way we rolled out zoom. If you feel strongly I can move it to premake, but we've used this convention as we partially roll out OS specific things before, and I think it makes the code nicely readable to have this inline.

baconpaul:
HI @jsakkine 

I rewrote the comments based on your feedback and renamed the function

I didn't change the feature guard - that's the same pattern we used in zoom and I'd sooner stick with it. And I am happy to rename those weirdly named global variables but would sooner that was not intermixed with this function change. Once this is approved I can make that mechanical change, test the builds, and the push it at the same time I push this, but with a separate commit and PR.

Please let me know what you think. Thanks!

PR #581, opened 2019-02-11
Remove unused getSelfLocation

baconpaul:
At the very early days of the mac port, we thought we needed
getSelfLocation in SurgeStorage, but then realized we didn't
so took out that code path. But we didn't take out the implementation.
Do so.


baconpaul:
Thanks for the look! I'll merge it in.

baconpaul:
Oh and for the future, VSTGUI::getBundleRef() does basically the same thing on macOS for us. #include "vstgui/lib/platform/mac/macglobals.h" in case you want it.

PR #598, opened 2019-02-14
DisplayInfoLinux improved

### LINKED ISSUE #594: Linux:  Zoom and resize issues
When attempting to Zoom to a size larger than the screen in Windows it will give a warning. Here an attempt at 150% was made and it warned that it was selecting 135% and refit.

resize warning win

In Linux if selecting 150% it will not warn that the monitor/screen limits will be exceeded and will in fact exceed them.

noresize warning lin 150

baconpaul:
1: getBackingScaleFactor doesn't need to alarmingly report an error
   when it returns 1.0, but rather just spool a log message to stdout

2: getScreenDimensions constrains zoom limits (or fails to) on large
   (or small) screens, so do a simple implementation of returning the
   size of the primary screen

Closes #594: Zoom limit issues linux vs windows

@falkTX or @jsakkine you know these APIs better than I do, but this seems like a positive step forward and at least will help @tank-trax continue to debug problems for us without overblowing his surge. Comments welcome!

### DIFF
--- a/premake5.lua
+++ b/premake5.lua
@@ -296,7 +296,7 @@ function plugincommon()
 
         buildoptions {
             "-Wno-unused-variable",
-            "`pkg-config --cflags cairo fontconfig freetype2 xkbcommon-x11 xcb-cursor xcb-keysyms xcb-xkb xcb-util`",
+            "`pkg-config --cflags cairo fontconfig freetype2 xkbcommon-x11 xcb-cursor xcb-keysyms xcb-xkb xcb-util x11`",
             "-std=c++14"
         }
 
@@ -333,7 +333,7 @@ function plugincommon()
         }
 
         linkoptions {
-            "`pkg-config --libs cairo fontconfig freetype2 xkbcommon-x11 xcb-cursor xcb-keysyms xcb-xkb xcb-util`",
+            "`pkg-config --libs cairo fontconfig freetype2 xkbcommon-x11 xcb-cursor xcb-keysyms xcb-xkb xcb-util x11`",
             "-Wl,--no-undefined",
         }
     elseif (os.istarget("windows")) then
--- a/src/linux/DisplayInfoLinux.cpp
+++ b/src/linux/DisplayInfoLinux.cpp
@@ -1,6 +1,10 @@
 #include "DisplayInfo.h"
 #include "UserInteractions.h"
 
+#include <X11/Xlib.h>
+#include <iostream>
+#include <iomanip>
+
 namespace Surge
 {
 namespace GUI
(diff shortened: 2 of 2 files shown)


premake5.lua
```diff
@@ -330,6 +330,7 @@ function plugincommon()
             "gcc_s",
             "gcc",
             "dl",
+            "X11",
```
falkTX: This should be in the pkg-config lines, not as library to link against directly.
Put it as lowercase "x11" at the end of the --cflags and --libs

src/linux/DisplayInfoLinux.cpp
```diff
@@ -10,16 +14,21 @@ using namespace VSTGUI;
...
-    Surge::UserInteractions::promptError("getScreenDimensions not implemented yet on linux. Returning 1400x1050",
-                                         "Software Error");
-    return CRect(CPoint(0,0), CPoint(1400,1050)); // a typical mid-range 15" laptop
+    /*
+    ** For now make the assumption of single screen to at least avoid
+    ** clip and bound errors like shown in #594
+    **
+    ** FIXME: Deal with multiple screens using the appropriate API
+    */
+    Display* d = XOpenDisplay(NULL);
```
falkTX: This is leaking memory/resources everytime it is called, it needs the cleanup methods

baconpaul: Thanks. That’s just xclosedisplay? Or do I need to dispose of the screen also

baconpaul: Googling shows me xclosedisplay is the call I need. I’ll do that. But also I will cache this value since we don’t support multi screen anyway in this revision and will update the fixme

premake5.lua
baconpaul: Thanks will adjust.

baconpaul:
If you have another minute, @falkTX, I integrated your comments and it all works fine. A quick peek is appreciated. Thanks!

src/linux/DisplayInfoLinux.cpp
```diff
@@ -10,16 +14,36 @@ using namespace VSTGUI;
...
-    return CRect(CPoint(0,0), CPoint(1400,1050)); // a typical mid-range 15" laptop
+    if(dispinfoScreenW < 0 || dispinfoScreenH < 0 )
+    {
+        /*
+        ** For now make the assumption of single screen to at least avoid
+        ** clip and bound errors like shown in #594
+        **
+        ** FIXME: Deal with multi-display linux
+        */
+        Display* d = XOpenDisplay(NULL);
```
falkTX: looks good, though I would add a verification here on the `d` pointer for nullness.
Also, since we are using C++14/17 after all, no need to use NULL, use nullptr instead :)
Copy&pasted from somewhere I guess :P

baconpaul:
Oh definitely copied and pasted! My Linux ui skills amount to carefully navigating google searches! Lol

Thanks v much

PR #601, opened 2019-02-14
UTF8 for Wavetables; Noisier Mis-installation; Win Categories

### LINKED ISSUE #543: Surge when mis-installed should be much noisier
@tank-trax found a lot of problems on linux around crashes and non working elements which were all because his `~/.local/share/Surge` was not correctly populated.

Now of course Surge shouldn't crash in those situations and so on but also: Surge should report that it can't find information.

So I propose a simple check, at the end of SurgeStorage, that if wavetable count is 0 or factory patch count is 0, we show a user error clearly reporting the problem. If we follow good install procedures this should "never" be needed. But it could have saved some headaches for @tank-trax and will for other users in the future I bet.

### LINKED ISSUE #584: UTF8 paths for wavetables
In surge storage we convert patch names and category names to utf 8 from a wstring on windows. We need to do this for wavetables also even though wavetables don’t have any charset exercising names today (so we don’t experience the bug which the lack implies)

baconpaul:
UTF8 filename conversion is required on windows for vstgui so factor
the UTF8 string conversion into a function and adjust the callpoints.

Those callpoints intersect with places where we load patches and
wavetables, so at the same time, make the system report user errors
if wavetables or factory patches are non-loadable

In the course of testing this, I also realized that nested directories
didn't work properly on windows due to path separator being "\" rather
than "/", which this commit also addresses.

Closes #584 UTF8 for wavetables
Closes #543 Alert user to mis-installation

### DIFF
--- a/src/common/SurgeStorage.cpp
+++ b/src/common/SurgeStorage.cpp
@@ -27,6 +27,8 @@
 #include <Shlobj.h>
 #endif
 
+#include <sstream>
+
 float sinctable alignas(16)[(FIRipol_M + 1) * FIRipol_N * 2];
 float sinctable1X alignas(16)[(FIRipol_M + 1) * FIRipol_N];
 short sinctableI16 alignas(16)[(FIRipol_M + 1) * FIRipolI16_N];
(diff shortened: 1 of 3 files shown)


baconpaul:
@kzantow or @jsakkine appreciate a once over of the code. @esaruoho if you could take a look at the diffs also to make sure you think the user feedback is appropriate. Will probably sweep this tommorrow night NYC time if I don't hear otherwise. Thanks!

PR #604, opened 2019-02-14
Sub-menus from the snapshot configuration

### LINKED ISSUE #539: Sub-folders for EFX presets too
currently configuration.xml is where the efx are stored. so you gotta kinda carve it up a bit to add more stuff.

Would be great to support sub-folders for EFX presets, so that we can have an `Arty` folder for Reverb + Distortion instead of dumping the Arty EFX parameters (for Reverb,Distortion, still trying to find the Delay ones) into the selector itself, like in the screenshots at #518

baconpaul:
As @esaruoho is finding people with preset settings, but
before we can save presets, we want to have user presets in
sub-folders rendered from the XML. So modify the snapshot
to have types-inside-types so we can imply recursive folders.

Closes #539: SubFolders for EFX


esaruoho:
logic-au works, renoise-vst2 works. as expected. get it in! :)

Issue #607, opened 2019-02-14
Mouse Wheel controls Sliders


baconpaul:
@rghvdberg wants to do this and I think it is cool

You select a slider and scroll mouse wheel. Slider scrolls.

control has a `virtual bool onWheel()` method which we can override to do the appropriate setValue for the slider akin to the onMouseMoveDelta etc... stuff already in the codebase. (May need a bit of a refactor to avoid copy and paste).

rghvdberg:
I've tried
```
bool CSurgeSlider::onWheel(CPoint& where, const float& distance, const CButtonState& buttons)
{
printf("CSurgeSlider::onWheel distance %f\n",distance);
return true;
}
```
nothing showing up in ..
however
CMouseEventResult CSurgeSlider::onMouseEntered(CPoint& where, const CButtonState& buttons)
{
printf("Mouse entered\n");
 return kMouseEventNotImplemented;
}
works
so ....

baconpaul:
Two thoughts 

1: virtual and override done properly in the h file? 
2: move the print statement to the vstgui lib platform code which calls onwheel if it exists
2a: if it doesn’t exist there’s the problem

rghvdberg:
I did it like this
https://github.com/rghvdberg/surge/commit/26e6910ff26751a385e3df3f5e23745ca630d7fe
trying to override
https://github.com/surge-synthesizer/vstgui/blob/1dced2b2d82d78173dcaddfff998c0d21f199871/vstgui/lib/cview.h#L109

baconpaul:
Right
So that looks correct
I wonder if the implementation of cview on Linux just doesn’t send the message
I would do a grep -r of onwheel in vstgui/lib/platform/Linux as a next step and stare at whatever code you find maybe?

rghvdberg:
https://github.com/rghvdberg/surge/commit/d3f29c97fab238b5fddbceca8cca394e2aa470af
https://github.com/rghvdberg/surge/commit/72e88b626a9958088cbc0a8802e16d4f5d203a84
I haven't figured out yet to determine the type of a control value (int/float/bool)
so any advice there is welcome

mkruselj:
Would be cool to have Shift+mousewheel doing fine movement just like when you drag via mouse cursor itself. :)

rghvdberg:
and we probably want the popup with values , just like when moving the slider with the mouse

baconpaul:
Absolutely want the popup
[...]

Issue #605, opened 2019-02-14
holding shift-down while hovering on macvst2/au will shoot forth tons of popups/tooltips that take ages to vanish


esaruoho:
<img width="345" alt="fullscreen_14_02_2019__21_07" src="https://user-images.githubusercontent.com/4966687/52811355-8f149400-309d-11e9-8064-4ff4990ce808.png">
<img width="1440" alt="fullscreen_14_02_2019__21_08" src="https://user-images.githubusercontent.com/4966687/52811360-90de5780-309d-11e9-806c-b6012758abca.png">

video posted at
https://surgesynth.slack.com/files/UF7BKN3D2/FG79RPW3V/hold-shift-down-and-move.mov

baconpaul:
The problem here is that the drag implementation doesn't implement beginEdit/endEdit the same way the click implementation does.

baconpaul:
And there's some awful state dependent behavior which makes VST2 and AU different

in AU shift-moving over the control doesn't even show the popup. In VST2 it goes away eventually - probably due to the dismiss timer loop - but doesn't go away as quickly as on mouse click.

I think I need to review the whole info window open close state. That's a big annoying problem.

But this is nothing to do with the shift change. The shift change just lessens the footprint of the problem.

baconpaul:
So here's the problem (adding @rghvdberg )

if you call beginEdit and not endEdit, a timer is supposed to kill the popup. Line 431 of SurgeGUIEditor

but if you have popped it up somewhere else and left it open and then haven't hit the timeout and popup a second place, voila

So the trick is before you go moving the parameter tooltip for the next show you have to check if it is showing and if so hide it.

baconpaul:
So that's around line 2491 of SurgeGUIEditor 

```
 if (buttons || forceMB)
   {
/*
   add here if info window is visible then hide it and invalidate before you move it
 */
      ((CParameterTooltip*)infowindow)->setViewSize(r);
      ((CParameterTooltip*)infowindow)->Show();
      infowindow->invalid();
      clear_infoview_countdown = 40;
   }
   else
```

baconpaul:
(I'm in a totally different branch now so can't fix this right now but wanted to put this in here)

rghvdberg:
like this ?
```
 if (buttons || forceMB)
   {
      if (((CParameterTooltip*)infowindow)->isVisible())
      {
         ((CParameterTooltip*)infowindow)->Hide();
[...]

PR #619, opened 2019-02-17
Scrollwheel 607

rghvdberg:
add scrollwheel support

Sliders and horizontal switches can be edited with scrollwheel.
No need to select the control, hovering works.
Holding shift enables fine control on sliders.
When "`modmode`" is enabled the modulation parameter is changed.
I've added some changes in SurgeGUIEditor, but I'm not 100% sure this is needed for my patch.


baconpaul:
Hi @rghvdberg and thanks for doing this.

Couple of code reviews

1. Your onWheel has signature bool but doesn't return a value. This means windows doesn't build. Can you add a 'return true' in both your implementations
2. Your edit to surgeguieditor "double copies" code inside an if. I'll tag the diff inline but you can see it looks rather odd.
3. Does the "/10" make chswitch2 act naturally? Like does the mouse wheel do the right thing in those cases?

I'll add these inline on the diff too.

src/common/gui/CHSwitch2.cpp
```diff
@@ -107,3 +107,11 @@ CMouseEventResult CHSwitch2::onMouseMoved(CPoint& where, const CButtonState& but
    }
    return kMouseEventNotHandled;
 }
+bool CHSwitch2::onWheel(const CPoint& where, const float& distance, const CButtonState& buttons)
+{
+   float newVal=value;
+   newVal += (getRange() / 10) * distance;
```
baconpaul: Why "/10" here? Can you at least add a comment explaining that constant

src/common/gui/CHSwitch2.cpp
```diff
@@ -107,3 +107,11 @@ CMouseEventResult CHSwitch2::onMouseMoved(CPoint& where, const CButtonState& but
...
+   value = newVal;
+   bounceValue();
+   setValue(value);
```
baconpaul: Needs to return a value

src/common/gui/CSurgeSlider.cpp
```diff
@@ -435,3 +435,32 @@ void CSurgeSlider::setBipolar(bool b)
...
+   }
+   else
+   {
+      setValue(value);
+   }
+   setDirty();
+   if (isDirty() && listener)
+         listener->valueChanged(this);
+   //endEdit();
+   edit_value = nullptr;
```
baconpaul: Needs to return a value

src/common/gui/CSurgeSlider.cpp
```diff
@@ -435,3 +435,32 @@ void CSurgeSlider::setBipolar(bool b)
 
    setDirty();
 }
+
+bool CSurgeSlider::onWheel(const VSTGUI::CPoint& where, const float &distance, const VSTGUI::CButtonState& buttons)
+{
+   double rate = 0.1 * moverate;
+   if (buttons & kRButton)
+      rate *= 0.05;
```
baconpaul: Is it even possible to mouse wheel with the right button down? More curious than anything else

src/common/gui/CSurgeSlider.cpp
```diff
@@ -435,3 +435,32 @@ void CSurgeSlider::setBipolar(bool b)
...
+      setModValue(*edit_value);
+   }
+   else
+   {
+      setValue(value);
+   }
+   setDirty();
+   if (isDirty() && listener)
+         listener->valueChanged(this);
+   //endEdit();
```
baconpaul: Can you add a comment here like 

```
/*
** No need to call endEdit since the timer in SurgeGUIEditor will close the
** info window, and  SurgeGUIEditor will make sure a window
** doesn't appear twice
*/

File: src/common/gui/SurgeGUIEditor.cpp [hunk 6]
```diff
@@ -2478,6 +2478,21 @@ void SurgeGUIEditor::draw_infowindow(int ptag, CControl* control, bool modulate,
...
    if (buttons || forceMB)
    {
+      if (buttons || forceMB)
+      {
+         if (((CParameterTooltip*)infowindow)->isVisible())
+         {
+            ((CParameterTooltip*)infowindow)->Hide();
+            ((CParameterTooltip*)infowindow)->invalid();
+         }
+
```
baconpaul: You only need this block (if ->invisible) not the copy of the if and the inside. Your current code is structured as

```
if a
   if a
      if b
         x
      y
   else
      y
```

which is equivalent to

if a
  if b
     x
  y
```

see what I mean?

baconpaul:
Alright I added comments on the sections I flagged when I read this. Let me know when you get a chance to resolve. You can just push another commit to this branch and I'll squash it all when I merge. Don't worry about flattening or anything.

I will also test on Win and Mac for you when the change is settled.

Oh and this will close #607 and might also close #605 (I'll tell when I test it).

Best

Issue #586, opened 2019-02-12
Zoom state vs View state on open / close


baconpaul:
Reported by @sense-amr 
vst2 bitwig

* Open bitwig
* Add an instrument and pick surge
* open the editor
* shrink by 25%
* close the editor
* open the editor - it appears to be 100%
* shrink by 25% it probably goes to 50% view
* close the editor
* open the editor - it appears to be 100%
* shrink by 25% and you get the "no teensy surge for you warning" even though it looks like 100%
* and then it goes to 50%

What is happening is when you open the editor the display is ignoring the in memory zoom value. So re-open is displaying at 100% but "shrink by 25%" is reading the prior zoom value.

Another perverse version of the bug is

* add a surge
* zoom to 200%
* close surge
* open surge and see 100%
* shrink by 10% and it gets bigger (goes to 190)

The fix is in the ::open to make sure you setZoom properly. Just find the right codepath.

baconpaul:
A simple fix (setZoomFactor of getZoomFactor in ::open) doesn’t work since the window isn’t shown yet and so the panel is blown up larger but the window frame isn't. Same reason that I can't set initial zoom to something other than 100. So related to implementing #456 properly also.

Probably need to find some frame onOpen handler thingy or something in CFrame land.

baconpaul:
Yup same problem as #456. This obvious change doesn't resize the window at the right time.

```
--- a/src/common/gui/SurgeGUIEditor.cpp
+++ b/src/common/gui/SurgeGUIEditor.cpp
@@ -1266,6 +1266,11 @@ bool PLUGIN_API SurgeGUIEditor::open(void* parent, const PlatformType& platformT
    frame->registerKeyboardHook(this);
    openOrRecreateEditor();
 
+   setZoomFactor( getZoomFactor() );
+   // Now it's too early for the vstgui to work so
+   frame->setSize( WINDOW_SIZE_X * getZoomFactor() / 100.0,
+                   WINDOW_SIZE_Y * getZoomFactor() / 100.0 );
+   
    return true;
 }
```

so we need to be 'later' to resize it in vst2 land.

baconpaul:
OK i have figured out a way to do the zoom call at the right time with one minor side effect. I've only tested it mac vst2 and it is late but I will walk it through the tests this week and see if I can get this fixed.

baconpaul:
branch `baconpaul/zoomstate-586` has my WIP just in case I forget.

geoslake:
[...]

PR #620, opened 2019-02-17
UserDefaults; Zoom Defaulting

### LINKED ISSUE #374: retain zoom settings in plugin configuration across multiple plugin instances
Suggestion:
When you have set your plugin to 200% and are happy with that zoom, the next time you load an instance of Surge, it could also default to 200%.

So some sort of cookie or a cache or a configuration flag that defines it to be  zoom type X

path to #305

### LINKED ISSUE #375: Have UserDefaults
Right now, Surge makes a lot of choices for you. patches are in ~/Documents/Surge (or whatever) your default zoom is 100% your default patch is the built in init. If you imagine a world with skins (which is post 1.6) etc...

But most proximate is that default zoom so on path to zoom I was planning on adding a user defaults mechanism.

Here's a rough design of how I would do it

* Implement in the Surge::Storage namespace functions like `loadUserDefaults` 'getUserDefault(string key)`, `setUserDefault(string key, string value)`. 
* Have `getUserDefault` be implemented (in pseudocode) as

```
if( loaded_defaults.contains[ key ] ) return loaded_defaults[ key ]
else if( system_wide_initi_defaults.contains[ key ] ) return the
else promptError
```

* have `loadUserDefaults` and `setUserDefaults` write an XML file of the implicit key value map
* *Here's the one bit I haven't figured out*. Store that XML file somewhere. But where
     * An obvious choice is `~/Documents/Surge` and the mac and lin equivalent. But then we can't change that directory with defaults
    * Another obvious choice is `~/.surge_defaults` on mac and linux and the registry on windows. But registry. Ugh.
    * So this one gets a 'design required' flag

I would like to resolve this on the path to zoom #305 so we can resolve #374 in a more general way.

Welcome any thoughts. No way I'm working on this anytime soon.

### LINKED ISSUE #456: Change default zoom to NOT 100%
I believe great majority of people are not running 1024x768 or similar monitor resolutions, so it might be very safe to assume that the default zoom level of 100% should not be considered the default anymore.

Over here, at 1920x1200, 150% zoom is perfection. 125% still looks a bit smallish, particularly some fonts are tiny and harder to read (but at least things are sharp!).

So my suggestion is twofold: first renaming the zoom levels. Maybe 125% could be the "new" 100%, 100% would be 75%, and then go from there. Second, detect monitor resolution when instantiating the plugin and pick the most comfortable zoom level for that resolution - there's already overzooming detection so that sounds to me like it would go hand in hand :)

Thanks!

baconpaul:
Introduce a UserDefaults mechanism for surge, where users can set
a persistent value for some configuration element of surge and
store it between sessions. This is accomplished using the API in
src/common/UserDefaults.h which is pretty simple key/value store
and retrieve of strings and ints (now).

Use UserDefaults to create a default zoom option. This option
stores the defaultZoom and restores it across host and session
when you restart surge.

Closes #375: Have UserDefaults
Closes #374: Retain Zoom Settings
Closes #456: Allows a default zoom setting not 100%

### DIFF
--- a/premake5.lua
+++ b/premake5.lua
@@ -227,6 +227,7 @@ function plugincommon()
         "src/common/SurgeStorageLoadWavetable.cpp",
         "src/common/SurgeSynthesizer.cpp",
         "src/common/SurgeSynthesizerIO.cpp",
+        "src/common/UserDefaults.cpp",
         "libs/xml/tinyxml.cpp",
         "libs/xml/tinyxmlerror.cpp",
         "libs/xml/tinyxmlparser.cpp",
--- a/src/common/UserDefaults.cpp
+++ b/src/common/UserDefaults.cpp
@@ -1,0 +1,190 @@
+#include "UserDefaults.h"
+#include "UserInteractions.h"
+
+
+#include <string>
+#include <iostream>
+#include <iomanip>
+#include <map>
+#include <sstream>
+#include <fstream>
+
+#if LINUX
+#include <experimental/filesystem>
+#elif MAC
+#include <filesystem.h>
+#else
+#include <filesystem>
+#endif
+
+namespace fs = std::experimental::filesystem;
+
+namespace Surge
+{
+namespace Storage
+{
+
+/*
+** These functions and variables are not exported in the header 
+*/
+struct UserDefaultValue
+{
+    typedef enum ValueType
+    {
+        ud_string = 1,
+        ud_int = 2
+    } ValueType;
+
+    std::string key; 
+    std::string value;
+    ValueType   type;
+};
+
+std::map<std::string, UserDefaultValue> defaultsFileContents;
+bool haveReadDefaultsFile = false;
+
+std::string defaultsFileName(SurgeStorage *storage)
+{
+    std::string fn = storage->userDataPath + "/SurgeUserDefaults.xml";
+    return fn;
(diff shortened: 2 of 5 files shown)


baconpaul:
@jsakkine or @kzantow - I would very much appreciate a review of this code to make sure I didn't do anything stupid.

@esaruoho - if you would like to test to see if it works for you (save zoom state; restart; open and get zoom state) I would appreciate it

I tested 

Mac: AU in Hosting AU and Logic Pro
         VST2 and VST3 in Bitwig
Windows: VST2 and VST3 in Bitwig
Linux: VST2 in Carla

and all behaved properly over the save / restart / show cycle.

Thanks folks!

baconpaul:
Also will close #586 when swept

jsakkine:
Yeah, looks good to me except if you have energy check the spacing...

baconpaul:
Thanks appreciated.

I wrote a script to pre-check my PRs for spacing, and got many of the spacing things. I'll eyeball it once more and merge.

baconpaul:
Ahh added one more case to my spacing quick check and fixed that too. Thanks!

I wonder if I should commit my little set of util scripts. I'll ponder that separately.

Will merge this once the automated builds complete. Thanks again for looking!

PR #630, opened 2019-02-19
Refactor Menus

### LINKED ISSUE #536: Refactor Surge Menu to be callback only since Linux  COptionMenu::popup() is non blocking
Linux Menu / About does nothing.

First reported by @tank-trax

baconpaul:
This change is a large code change for zero functionality change,
required because of the semantics of menus and ::popup on Linux.
On Linux, because of various event loop handling, making
menu::popup() block is very difficult. The same idle loop which
pops up the UI also paints the on-screen menus, unlike in mac and
windows where they are separate elements.

So we have to change all of our menus from return-a-value-after-block
to bind-a-lambda-and-call fashion. That is, from

  id = 3;
  menu->add("do this", id );
  menu->popup();
  r = menu->getLastResult();
  if (r == 3)
    doThis()

we have

   menu>-add("do this", []() { doThis(); } )
   menu->popup()

which is cleaner but requires us to bind the correct state into lambdas
and reorder the code. Thus this commit is a massive code shuffle to
do exactly that, along with a helper function which actually adds a
no arg lambda to a menu in SurgeGUIEditor.

All of the menus were accordingly adjusted for copy, paste, rebind,
about, and so on.

Closes #536 popup menus on Linux

### DIFF
--- a/src/common/gui/CSnapshotMenu.cpp
+++ b/src/common/gui/CSnapshotMenu.cpp
@@ -125,11 +125,12 @@ void CSnapshotMenu::populateSubmenuFromTypeElement(TiXmlElement *type, VSTGUI::C
       addEntry("Store Current");
       }*/
     /*
-      int32_t sel_id = contextMenu->getLastResult();
+      int32_t sel_id = contextMenu->get Last Result();
+      this is commented so put in spaces so my grep to kill all these doesn't flag it PW
       int32_t sub_id, main_id;
       COptionMenu *b = contextMenu->getLastItemMenu(sub_id);
       if (b) main_id = b->getTag();
-      
+
       if ((sel_id >= 0) && (sel_id < contextMenu->getNbEntries()))
       {
       if (sel_id < main)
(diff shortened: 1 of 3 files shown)


baconpaul:
OK folks this is a big change. @jsakkine I used clang-format on my diffs so the spacing and commas are all OK! But I think this is one we just need to test the heck out of. I've tested every menu I can find on mac and linux but would really appreciate more eye either on the code or the result before i merge.

esaruoho:
i tried this with renoise/vst2 and found no issues opening any context menu (wavetables, set controller to, fx presets, patches..) looks good!

jsakkine:
I don't have any objections merging this.

baconpaul:
Thanks. Let me beat it up a little on windows and then if it all works there I will merge

baconpaul:
Alright I spent some time with this on Windows and it seems to work as well as linux and mac. Thanks for the tests and looks @esaruoho and @jsakkine - I'm gonna merge this one.

Issue #556, opened 2019-02-10
Performance Profiling Surge


esaruoho:
report from KVR: ( 
https://www.kvraudio.com/forum/viewtopic.php?p=7310445&sid=74fe7926c1b5f832ffcbcbb4cd8ed8af#p731044 )5
 
-----
i recently try official v152 build from friend, for compare, so, my assumptions were confirmed - by some reason new version load cpu already in initial (idle) stage, just after load plugin (it's a pretty small cpu load, but strange it is). While in v152 absolutely zero cpu usage in the same idle condition. Hmm.
I dont know, maybe i miss something in settings, would be good if you check this too. Or maybe reason is known ?
-----

baconpaul:
Hmm what a good and worrying report. I wonder what it's doing. I do notice some non-zero cpu clip at idle also.

baconpaul:
OK I took a peek at this using the macOS sampling profiler on the release

```
instruments -l 30000 -t Time\ Profiler -p 5773
```

where 5773 is the pid of the host. Found the time used was about half drawing and half in the processing loop

Didn't dig into the processing loop but the drawing time came a lot from something called 'background refresh' type stuff in NSView. (This was all mac).

And then I noticed something which seems to be worrying. The background image draws on every idle.

So if I place a print in CScalableBitmap::draw it gets called every idle loop.

More to investigate here, but that's definitely an issue.

baconpaul:
The stack that gets us to the background draw is

```
0   Surge                               0x000000010aabb7c5 _ZN15CScalableBitmap4drawEPN6VSTGUI12CDrawContextERKNS0_5CRectERKNS0_6CPointEf + 261
1   Surge                               0x000000010aaf69fc _ZN6VSTGUI14CViewContainer18drawBackgroundRectEPNS_12CDrawContextERKNS_5CRectE + 780
2   Surge                               0x000000010aae6762 _ZN6VSTGUI14CViewContainer8drawRectEPNS_12CDrawContextERKNS_5CRectE + 706
3   Surge                               0x000000010aae6402 _ZN6VSTGUI6CFrame8drawRectEPNS_12CDrawContextERKNS_5CRectE + 306
4   Surge                               0x000000010aaedebd _ZThn32_N6VSTGUI6CFrame16platformDrawRectEPNS_12CDrawContextERKNS_5CRectE + 13
[...]

Issue #446, opened 2019-02-02
Some drawn elements are still pixel-y


baconpaul:
Some drawn elements are still pixel-y

<img width="196" alt="screen shot 2019-02-01 at 8 28 20 pm" src="https://user-images.githubusercontent.com/13387561/52157781-ebb88d80-265f-11e9-9ca6-7d6070237699.png">

arrows in the wavetable browser

<img width="421" alt="screen shot 2019-02-01 at 8 28 40 pm" src="https://user-images.githubusercontent.com/13387561/52157787-f83ce600-265f-11e9-8eea-74fcb98443a6.png">

edges of VU meter

These are all things drawn by code not by bitmaps so need C++ changes.

(If we find more add them to this issue)

tagging @itsmedavep and @esaruoho

baconpaul:
And connecting with #305

baconpaul:
<img width="469" alt="screen shot 2019-02-01 at 8 55 24 pm" src="https://user-images.githubusercontent.com/13387561/52158172-b3b34980-2663-11e9-9055-895bfaced6ba.png">

Got one of them.

esaruoho:
imo those arrows could be at least one pixel higher.. and .. sharper? :)

itsmedavep:
Nice! 

@esaruoho 
`//// Drawing Methods

        public static void DrawTriangle()
        {
            DrawTriangle(new CGRect(0.0f, 0.0f, 3.0f, 5.0f), ResizingBehavior.AspectFit);
        }

        public static void DrawTriangle(CGRect targetFrame, ResizingBehavior resizing)
        {
            //// General Declarations
            var context = UIGraphics.GetCurrentContext();
            
            //// Resize to Target Frame
            context.SaveState();
            var resizedFrame = ResizingBehaviorApply(resizing, new CGRect(0, 0, 3.0f, 5.0f), targetFrame);
            context.TranslateCTM(resizedFrame.X, resizedFrame.Y);
            context.ScaleCTM(resizedFrame.Width / 3.0f, resizedFrame.Height / 5.0f);


            //// Color Declarations
            var color = UIColor.FromRGBA(0.000f, 0.000f, 0.000f, 1.000f);

            //// Polygon Drawing
            context.SaveState();
            context.TranslateCTM(3.0f, -0.0f);
            context.RotateCTM(90.0f * NMath.PI / 180.0f);

            var polygonPath = new UIBezierPath();
[...]

PR #640, opened 2019-02-21
Optimize VU Meter Draw

### LINKED ISSUE #639: Re-implement the VU meter for two reasons
There are two reasons I would like to re-implement the VU meter

1: It invalidates *and invalidates the background of its parent* on every call. This means we are doing lots of redraw calls to the background bitmap. Some careful subclassing to suppress invalidation to the background is probably the way to go

2: It looks like pixel bits at high zoom since it is using pixels to draw rather than paths.

So add this to the list.

Couples to #446

baconpaul:
The VU Meter was a cause of substantial idle load. In a 30 second
timer sample, a full half second was spent drawing background
images - as much time as was spent filling up the audio buffer
with no sound. There were a couple of key reasons for this:

1. The VU Meter invalidates every idle even if value hasn't changed
2. When the VU meter invalidates even legitimately it forces a
   full repaint of the background image into a clip rect because
   that's what VSTGUI::CViewContainer does

So - somewhat unsatisfactorily in the second case - this diff
fixes those by only invalidating when value changes (which is
great for idle) and supressing the background image repaint
using knowledge of the placement of the VU meter (which is
fragile and unsatisfying but the only option we have given
the protection of VSTGUI::CFrame's final status and so on).

While at it, replace the pixel painted edges so the VU meeter
has a better boundary when zoomed.

Closes #639 - Re-Implement VU Meter
Addresses parts of #446 and #556; performance and pixely bits

### DIFF
--- a/src/common/gui/CScalableBitmap.cpp
+++ b/src/common/gui/CScalableBitmap.cpp
@@ -115,6 +115,32 @@ CScalableBitmap::CScalableBitmap(CResourceDescription desc)
 
 void CScalableBitmap::draw (CDrawContext* context, const CRect& rect, const CPoint& offset, float alpha )
 {
+    /*
+    ** CViewContainer, in the ::drawRect method, no matter what invalidates, calls a drawBackground
+    ** on the entire background with a clip rectangle applied. This is not normally a problem when
+    ** you invalidate infrequently or just have a constant color background. But in Surge we have
+    ** a drawn background on our frame and we invalidate part of our UI every frame because of the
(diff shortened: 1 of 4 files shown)


baconpaul:
OK this is a fairly important little tweak here both for idle CPU (which this improves by 2x on my mac) and for soon when we have SVG backgrounds; but I had to do it in a somewhat gross way. So @jsakkine or @kzantow could you do me the favor of reading the comment in CScalableBitmap change at least to see if it makes sense to you? I had to do an if based on actual position of a parent element. It's not great, but it is an important improvement. Literally a 2x CPU improvement on my mac when idle.

PR #641, opened 2019-02-21
Tab key corrected in Save dialog

### LINKED ISSUE #633: Pressing Tab while in Store-preset-mode no longer does tab in the dialog, instead toggles modulation binding on/off
here's me pressing Tab while in the Store preset dialog window. tab doesn't actually happen, cos something else grabs it first.

presstab

baconpaul:
With the restoration of key bindings, the Tab toggled
modulation. That interfered with the save dialog.
So have the GUI editor not bind and eat the tab key
if the saveDialog is visible.

Closes #633 - tab key interfering with store

### DIFF
--- a/src/common/gui/SurgeGUIEditor.cpp
+++ b/src/common/gui/SurgeGUIEditor.cpp
@@ -495,6 +495,13 @@ int32_t SurgeGUIEditor::onKeyDown(const VstKeyCode& code, CFrame* frame)
         switch (code.virt)
         {
         case VKEY_TAB:
+            if (saveDialog && saveDialog->isVisible())
+            {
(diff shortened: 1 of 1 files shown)


baconpaul:
Hi @esaruoho - can you test this tiny change which fixes the tab problem you reported WYHAS? Thanks!

baconpaul:
yeah it’s the exact same diff. Just move it from the tab block up to the entire keyhandling block.
I can get this tonight. Or if you just move that block to above the switch statement rather than after the tab case it wil work

esaruoho:
above the switch statement rather than after the tab case. will try that too.

Issue #514, opened 2019-02-06
VST3 Linux Inconsistently Working


baconpaul:
Just here at the top of the thread

As of sep 2019

- builds and runs in Steinberg editor host and reaper on linux
- ui shows up as black in bitwig - the idle loop isn’t getting to the draw loop

baconpaul:
Forever ago in #128 @asimilon told us the bundle needs to look like vst3sdk/doc/vstinterfaces/vst3loc.html on Linux so that's probably the issue here.

baconpaul:
Hey OK that @asimilon document is right; we need to package it up like a mac bundle. When I do that I can get bitwig to recognize a VST3 then crash when it loads the plugin. But that's OK. Basically we can make a post-build script to build the .vst bundle.

baconpaul:
OK great so if you build the steinberg editor host it can also load the .vst3 and gets a core dump in (of all places) xcb_generate_id from VSTGUI::ChildWindow. So clearly something amiss but also clearly our VST3 is now starting to work.

If you want to try this here's a set of commands

```
# first build and localy install surge vst3 using build-linux

mkdir VS
cd VS
git clone --recursive https://github.com/steinbergmedia/vst3sdk.git
mkdir build
cd build/
sudo apt-get install libsqlite3-dev
cmake ../vst3sdk/
cd public.sdk/samples/vst-hosting/editorhost
make
../../../../bin/Debug/editorhost ~/.vst3/Surge.vst3
```

and you'll get the loading starting, a window almost popping up, and this core trace:

```
#0  0x00007ffff6ace736 in xcb_generate_id () from /usr/lib/x86_64-linux-gnu/libxcb.so.1
#1  0x00007ffff54a094e in VSTGUI::X11::ChildWindow::ChildWindow(unsigned long, VSTGUI::CPoint) ()
   from /home/paul/.vst3/Surge.vst3/Contents/x86_64-linux/Surge.so
#2  0x00007ffff549c249 in VSTGUI::X11::Frame::Frame(VSTGUI::IPlatformFrameCallback*, VSTGUI::CRect const&, unsigned int, VSTGUI::IPlatformFrameConfig*) () from /home/paul/.vst3/Surge.vst3/Contents/x86_64-linux/Surge.so
[...]

PR #646, opened 2019-02-22
VST3 on Linux packages and installs

baconpaul:
VST3 on Linux requires a mac-like bundle, as described in
the VST3 documentation vst3sdk/doc/vstinterfaces/vst3loc.html.
These changes move VST3 from a raw dll to a bundle using the
same techniques we use on macOS. Since we bake the images
and so on into the bundle, the bundle contains only the DLL.

With this change, bitwig studio will load our VST3, but it
crashes if you open the UI and gives corrupted parameter names.
But we are able to begin to debug.

Addresses part of #514

### DIFF
--- a/build-linux.sh
+++ b/build-linux.sh
@@ -141,8 +141,9 @@ run_install()
 
     if [ ! -z "$OPTION_vst3" ]; then
         echo "Installing VST3"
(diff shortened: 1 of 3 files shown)


baconpaul:
Thanks @kzantow - I'll merge this once the auto checks complete.

PR #649, opened 2019-02-23
script debian package creation

rghvdberg:
run `make_deb.sh` from the installer_linux dir to get a deb package
I've renamed some paths to make them more 'linuxy'
Only the vst plugins are packaged.


rghvdberg:
Some explanation:
The `make_deb.sh` creates a directory structure for making the Debian deb package.
I looked at http://www.hackgnar.com/2016/01/simple-deb-package-creation.html and the package provided by @falkTX.

This script assumes `dpkg-deb` (part of `dpkg`) is installed and that surge has been compiled successfully. The script needs the surge version as argument (`make_deb.sh 1.6b` for example), as this will be used in the `control` file for package creation.

I've mostly replaced `Surge` with `surge-synthesizer` in the paths and files, for the reason that there's already a plugin called Surge by Airwindows and I like file names to be descriptive and lower case.
Exception being the user data location $HOME/Documents/Surge.

falkTX:
This is really not the proper way to build a deb package.
You should use the "debuild" script instead, so that it catches dependencies and other things properly.
Manually specifying dependencies is nasty, it is the reason why bitwig continues to be broken to this day (it requires jack2 instead of the usual jack2|jack1 optional dependency that debian does in their scripts)

falkTX:
The name we use for the linux binary should be consistent with the official non-deb binaries.
Otherwise there will be conflicts with sharing sessions with other people that install in different ways

baconpaul:
I agree we shouldn’t rename. Or should rename all use cases.

Do you have a script which uses debuild to do this? I’m sort of hoping to cut a stable beta and would love a Linux distro to be part of it; having the ci pipeline is really the only missing link. Also happy to chat on irc if that’s easier

baconpaul:
Rob: Mmost of this has gone over to another branch per our convo on slack. Closing this and reopening another PR in a minute.

PR #650, opened 2019-02-23
Scrollwheel 607 - Add scrollwheel support to various controls

rghvdberg:
Add scrollwheel support to 
- `CHSwitch2`
- `CSurgeSlider`
- `CNumberfield`

Minor edits to `SurgeGUIEditor`
- slow  down the close time of the `infowindow` on **Linux**
- prevent `infowindow` from appearing twice

I added this support to speed up editing. 
I didn't add support for `CSwitchControl` because that wouldn't speed up editing and I think might lead to accidental flipping of switches which aren't easily noticed.
I didn't add support to the patch browser because that wouldn't speed up editing either.


baconpaul:
Thanks Rob! As I mentioned on Slack, I'm going to hold on review and test of this until after we print a beta. Not ignoring it but just want a 16beta5 out and then we can get this (and also I hope SVG) into the nightlies

baconpaul:
OK I just reviewed the code. It all looks good. Let me test it on Mac and Windows in a couple of ways and if that works I'll squash and merge.

baconpaul:
Confirmed it functions macOS AU in HostingAU and Logic ProX, and mac VST2 and VST3 Bitwig. Trying windows next.

baconpaul:
OK I tried it in windows in parallels. Seems everything is backwards but I'm not sure if that's my trackpad mapping into my VM or not. I'm tempted to sweep this to nightlies and find a windows physical desktop user to check.

baconpaul:
Oh yup this is all just parallels mis-interpreting "scroll direction natural" setting on macOS in the VM. If I disable that everything works. (but my trackpad is a nightmare to use because the scroll direction for mac thigns is no longer natural).

So gonna merge this. Thanks Rob!

PR #653, opened 2019-02-23
Visibility Flags for Linux

### LINKED ISSUE #648: VST2 on Bitwig Linux
We have the general #492 issue for host issues on linux, but I wanted to create a special one for VST2 Bitwig Linux, since this is @kurasu 's synth and I'd be bummed if our first beta release didn't actually work with bitwig.

But right now, surge linux vst2 doesn't work with bitwig! So here's what I've found

If you build the dll and run it in the bitwig opt host you get this error

```
./build-linux.sh build --project=vst2 && ./build-linux.sh install --local --project=vst2
/opt/bitwig-studio/bin/BitwigPluginHost64 test-local ./target/vst2/Release/Surge.so

(... output redacted ...)

Failed plugins:

    /home/paul/surge/target/vst2/Release/Surge.so:      createAEffectInstance returned NULL.
```

So what is createAEffectInstance? `public.sdk/source/vstplugmain.cpp` basically constructs the plugin and then calls `getAeffect` on it. 

```
	virtual AEffect* getAeffect ()   { return &cEffect; }		///< Returns the #AEffect structure
```

so what the heck. How does returning a reference to a member return NULL ever? And why only in bitwig. I don't know

But what I do know is if at the end of the constructor for our `VstPluginInstance` I put this

```
   std::cout << "/TOR P=" << this << " C=" << &cEffect << " G=" << getAeffect() << std::endl;
```

and then change our createEffectInstance to look like this

```
   auto res = new Vst2PluginInstance(audioMaster);
   std::cout << "crEI P=" << res << " C=" << "--" << " G=" << res->getAeffect() << std::endl;
   return res;
```

I see this

```
/TOR P=0x7f2d04022cf0 C=0x7f2d04022d20 G=0x7f2d04022d20
crEI P=0x7f2d04022cf0 C=-- G=0 
```

so somewhere between the very last line of the constructor and the print immediately after the constructor is done, something makes the object `getAeffect()` return 0.

Even more weirdly if I add this

```
   virtual AEffect *getAeffectFF() // override
   {
      return &cEffect;
   }
```

and then after my cout above add

```
   std::cout << "FF " << res->getAeffectFF() << std::endl;
``` 

I see 

```
/TOR P=0x7f7d98022cf0 C=0x7f7d98022d20 G=0x7f7d98022d20
crEI P=0x7f7d98022cf0 C=-- G=0
FF 0x4e82d1
Segmentation Fault (core dumped)
```

so (1) why the heck is that a short pointer for FF; (2) why does it core and (3) why is the result different.

So I have no idea what is going on. But it's clear I'm not finding this problem so wanted to write some exhaustive notes either to myself or maybe to @kurasu if he gets a chance to look; or @falkTX or @jsakkine if they want to take it on.

baconpaul:
Bitwig 2 with Surge consistently crashed as described in issue
number #648. After banging at it for a while, I gave up, but
@falkTX found the answer, basically resolving a symbol clash which
corrupted the base class with -fvisilbity flags. This change turns
those on on linux for the releas build only.

Closes #648 Linux

Co-authored-by: Filipe Coelho <falktx@falktx.com>

### DIFF
--- a/premake5.lua
+++ b/premake5.lua
@@ -302,6 +302,15 @@ function plugincommon()
             "-std=c++14"
         }
 
+        configuration { "Release*" }
+        buildoptions {
+            "-fvisibility=hidden",
+            "-fvisibility-inlines-hidden",
+            "-fdata-sections",
+            "-ffunction-sections",
+        }
+        configuration {}
+
         files
         {
             "src/linux/*.mm",
@@ -338,6 +347,17 @@ function plugincommon()
             "`pkg-config --libs cairo fontconfig freetype2 xkbcommon-x11 xcb-cursor xcb-keysyms xcb-xkb xcb-util x11`",
             "-Wl,--no-undefined",
         }
+
+        configuration ("Release")
+        linkoptions {
+            "-fdata-sections",
+            "-ffunction-sections",
+            "-Wl,-O1",
+            "-Wl,--as-needed",
+            "-Wl,--gc-sections",
+            "-Wl,--strip-all",
+        }
+        configuration()
     elseif (os.istarget("windows")) then
 
         pchheader "precompiled.h"


baconpaul:
Oh thanks. This will just apply the flag twice. I'll knock it out.

PR #659, opened 2019-02-25
slow down blinking of modulation control on linux 

rghvdberg:


baconpaul:
We have to clean up your master branch so you don’t generate all these spurious commits - makes merging a bit of a pain. 

The change looks fine and I can squash merge this one and rewrite the commit message later this morning but let’s get your master back to upstream!

Issue #662, opened 2019-02-25
AU Mac Cracks and Pops if you change sample rate


baconpaul:
Investigating #660 I played around with AU Mac in Logic after @davephillips reported cracking in bitwig linux at 48k jack.

If I change audio sample rate back and forth between 44.1 and 48 in a session enough times, eventually surge will get wedged into a state where it pops only, or pops plus plays.

My original debugging thought it was the timepos hardcoded to 44.1 in the ::Render of aulayer but fixing that doesn't help. (PR to fix that inbound though).

So something in the internal state of the oscillators, envelopes, or wavetables is wrong when we reset sample rate on mac. Sometimes. Not always though. Which is super fun.

My notes

1. All the sample rates are globals in SurgeStorage.cpp externed in SurgeStorage.h
2. setSampleRate sets them all then inits the wavetables. A printf shows that the at least the env_linear wave table does init correctly under this change
3. The 44100 hard coded in unitcoversion.h is not it; that's just used for display in the cm_lforate section of CNumberField where it is matched by another hardcoded 44100.
4. Not clear this is daves issue, but it seems likely. But lets keep a separate issue for mac debug and test of the switch open also.

Issue #663, opened 2019-02-25
Code-reformat the AU layer


baconpaul:
The AU layer is a mix of hard tabs, weird spacing, and so on. Plus a bunch of ancient non running carbon code. So remove the dead code and run clang format on it one day.

PR #664, opened 2019-02-25
Don't hardcode timepos sample rate to 44100 in AU

baconpaul:
AU had a hard coded 44100. Rather than use that, cache the last sample
rate we saw when reset and use it to advance timepos in the sample block.
Related to #662.


baconpaul:
Oh the code formatting here is a mess. #663 will have me clean that up in a separate commit.

PR #668, opened 2019-02-26
Plus key for zoom on Linux

### LINKED ISSUE #600: Linux: +/- Resize issue
the minus key to the left of Backspace if pressed will Shrink the app however the plus key will not Zoom in

the - key is active whether shift is pressed or not
the + key is inactive whether shift is pressed or not

baconpaul:
Unsatisfyingly, the plus key (shift-= on US keyboards) on linux
doesn't report the character '+' to onKeyDown in VSTGUI but rather
reports '=' with shift depressed. Ideally the keyboard layer would
handle keyboard mapping of course (it does on mac) but this is a
useful enough feature and common enough keyboard mapping that I
choose to make the ugly choice to also bind shift-= to zoom increase
which defacto fixes the +/- zoom on linux.

Closes #600: +/- key on linux zoom


baconpaul:
tested and confirmed by @tank-trax

PR #669, opened 2019-02-26
frame->close in ~SurgeGUIEditor

### LINKED ISSUE #503: SurgeGUIEditor::~SurgeGUIEditor() should call frame->close
Right now it calls removeall and forget(). I believe this is incorrect as it misses implementation specific stuff and other cleanup.

baconpaul:
Call frame->close which does a removeAll and a forget, like we
used to, but also does a bunch of other internal implementations.
Reported by @sagantech. Reading the vstgui code for close this
is obviously correct.

Closes #503 call close


baconpaul:
I tested this by repeatedly opening and closing the app in host and flavor combinations on all platforms in my vm as well as reading the cframe::close implementation. Gonna sweep this one once the auto builds work.

PR #670, opened 2019-02-26
cmake: Headless version for macOS

jsakkine:
Add CMakeLists.txt that is able to build the headless version for macOS:

mkdir build
cmake -GXcode -Bbuild
cd build
xcodebuild
cd ..

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,118 @@
+cmake_minimum_required(VERSION 3.5)
+project(Surge VERSION 1.0.0 LANGUAGES CXX)
+
+set(SURGE_COMMON_SOURCES
+    src/common/dsp/effect/ConditionerEffect.cpp
+    src/common/dsp/effect/DistortionEffect.cpp
+    src/common/dsp/effect/DualDelayEffect.cpp
+    src/common/dsp/effect/Effect.cpp
+    src/common/dsp/effect/FreqshiftEffect.cpp
+    src/common/dsp/effect/PhaserEffect.cpp
+    src/common/dsp/effect/Reverb1Effect.cpp
+    src/common/dsp/effect/Reverb2Effect.cpp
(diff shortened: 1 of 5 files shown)


CMakeLists.txt
```diff
@@ -0,0 +1,118 @@
...
+)
+
+set(SURGE_COMMON_INCLUDES
+    src/common
+    src/common/dsp
+    src/common/gui
+    libs/xml
+    src/common/vt_dsp
+    src/common/thread
+    vst3sdk/vstgui4
```
baconpaul: This should definitely be vstgui.surge not vst3sdk/vstgui4 so we get our overridden vstgui

jsakkine:
@baconpaul, can you give it a shot?  my cmake build file ever

baconpaul:
Yup. Will give it a shot tonight

baconpaul:
Oh did you see my comment about your using the wrong vstgui? You should use vstgui.surge not vst3sdk/vstgui4

Won’t matter here but will for proper plugins of course

baconpaul:
Sorry @jsakkine I won't get to this tonight. Will look tomorrow in a break though.

baconpaul:
Actually had 30 seconds so checked out 

```
paul:~/dev/music/surge$ cmake -GXcode -Bbuild
CMake Error: The source directory "" does not exist.
```

turns out my cmake is 3.1.something but the minimum version check didn't fire. I did a brew upgrade and it resolved.

baconpaul:
Once I upgraded though I was able to build headless following your directions yes.

I'd be happy to merge this if we remove headless from premake and update the build-linux and build-osx to have commands to do this. Also change the vstgui location of course!

jsakkine:
@baconpaul, OK whatya think?

jsakkine:
I'll do Linux as a separate PR.

PR #671, opened 2019-02-26
disable mousewheel on category and patch buttons

rghvdberg:
added bool and a setter function to enable/disable use of mouse-wheel for editing `CHSwitch2`


baconpaul:
Thanks looks good. Let me just test it in a couple of other oses tonight and then if it is ok I will merge

rghvdberg:
:+1:

Op di 26 feb. 2019 19:50 schreef Paul <notifications@github.com>:

PR #676, opened 2019-02-27
Separate Chorus and Delay, Pingpong -> Ping-Pong

### LINKED ISSUE #675: FX section Delay/Chorus = Delay only
After looking through the FX section scrutinizing various aspects of it with @itsmedavep and @oddy.o.trax
I discovered something which i had noted before but not thought much of .. 

The Delay/Chorus section.. actually NEVER shows Chorus and only ever delay .. regardless of any preset that is selected .. 
delaychorusneverchorus

in the above picture i selected the wide chorus preset effect.. as you can see no indication of it being selected is present in the UI...and to the contrary the UI displays that i have selected a DELAY not a CHORUS.. effect..

Whilst understanding clearly that technically a "Chorus" is just a delay with modulation.. we consider that it might be useful to have it explicitly separated for the user .. 
 
We propose as a fix.. that the Chorus related FX be moved to a separate Chorus related dropdown .. which will populate in the dropdown .. as Chorus .. and  "CHOR or CHRS" in the box area.. 

thoughts?

esaruoho:
- Moves all Choruses from Delay/Chorus to Chorus-subfolder
- Renames Delay/Chorus to Delay
- Renames Pingpong FX to Ping-Pong

fixes #675


esaruoho:
not what's wanted since delay/chorus fx are specifically different from chorus fx, not the same parameters at all

PR #677, opened 2019-02-27
FX renaming: Pingpong -> Ping-Pong

esaruoho:
Pingpong -> Ping-Pong


esaruoho:
appears to work, so merging

PR #682, opened 2019-02-27
Check for LATO built in

### LINKED ISSUE #673: Windows Lato-Semi-Bold is too bold
I picked a slightly bolder font on windows than mac due to render differences when I was working. Seems those render differences are a function of parallels VM and the font looks a bit too bold on windows. Back it down from lato-semi-bold to lato-regular in time for 1.6. beta6. Reported by @sense-amr

### LINKED ISSUE #678: Lato font locked access - plugin doesn't load on Windows (beta-5)
This one is for @baconpaul.

image

Long story short, seems there's an issue in loading Lato font on Windows, which makes the VST not loadable at all. Solution would be to install the font to the system, and this is how:

https://stackoverflow.com/questions/28186646/install-fonts-to-pc-using-inno-setup

baconpaul:
Rather than try and runtime load the Lato font assume it is installed
and react accordingly if it is not. No longer bundle it with the dll
and instead bundle it with the installer.

Closes #678 : Font causes Hang on some Windows
Closes #673 : De-Bold our Lato


baconpaul:
I tested this several ways, including checking if Lato installed when I ran the installer, running the installer twice to make sure the second install worked, etc...

Would really appreciate an eyeball from @kzantow on those iss changes

The rest of it makes it super simple. If lato is found use it; otherwise use the old defaults. So even if you never run the installer you should be OK.

baconpaul:
This is an obvious improvement - just gonna merge it

We should make a beta 6 - this bug is real

PR #685, opened 2019-02-28
Remove headless app from premake

baconpaul:
As @jsakkine moves headless to CMake as part of our CMake migraiton
we don't need to build it with premake. Also make clean-all grab
the new cmake directory


jsakkine:
Thanks!

Issue #316, opened 2019-01-17
LFO: Enabling synctempo on LFO rate yields several values that seem incorrect


itsmedavep:
When enabling synctempo on LFO rate, if i pull the rate slider all the way to the right it appears that I have many different 64th note positions. 

My expectation is that if I am temp syncing LFO rate to the host, I see all of the standard note durations up to 64th (IDK maybes somebody has use for 128th notes ). Additionally, I would expect to see dotted and triplet notes along with the standard note durations. 

Im not sure if the real issue here is a bug or feature improvement TBH. In the attached video you should see that it appears as though I can move the rate slider and there too many stops that are showing the label as 64th notes. 

This was using the AU in Logic ProX

![lfo-temposync](https://user-images.githubusercontent.com/11078351/51288417-cec65e00-19c1-11e9-9bc1-d2d98ad12f60.gif)

baconpaul:
Aren't the 1.5 and 1.33 along the way the dotted and triplet notes?
Is this a display thing (like you wish 0.6667 / 64th actually said 1.333 / 128th)?
Or is the tempo set incorrectly?
Thanks!

itsmedavep:
I believe you're correct. They are stops along the way and *to me* this seems like its just a label issue. The tempo is set correctly. 

Using your example, my expectation would be to see the 1.333/128th as opposed to .6667/64th. 

Additionally, as Ive been playing around with this some demarkation of dotted or triplets would be a welcome addition instead of the ratio only. 

As an example, If I hit the 1.5/64th ratio the label would be "1.5/64th d" or similar. Unfortunately my brain doesn't work out those ratios at a glance.

baconpaul:
Not a fix, but some comments

The code that does the tempo sync to string conversion is in Parameter.cpp. It is in two places
 
1. The ct_delaymodtime has one implementation which is around line 690
2. The ct_lforate (the one you screen shot here) has a second implementation around line 712

You can read the formulae there but basically it is the pow based approach you would suspect.

But the reason for the two implementations is the value (the 'f' in the code) has different meaning in LFO vs Delay. So those ranges need unpicking.

But I guess if you wanted a 128th display in the same flavor as today you would just ad an `if( f>5)` around line 719

I'm not going to work on this now. (I'm not 100% sure what the correct fix is). But am trying to put some hint notes on some issues in case other devs want to whack on them.

baconpaul:
[...]

PR #688, opened 2019-02-28
Lfo temposync label update

itsmedavep:
Changes to tempo sync in parameters.cpp 

Adding in 128th notes and 256th notes.


itsmedavep:
This does not fix the 1.5, 1.333333, etc by making them dotted or triplets respectively.  This just adds in 128th and 256th notes. So only partially addresses #316

baconpaul:
Looks good to me, Dave, and works fine too when I test it. Gonna squash and merge.

PR #690, opened 2019-02-28
update dependencies for Debian package

rghvdberg:
- Updated dependencies per recommendation of @falkTX
- add `fonts-lato` as dependency as discussed in #687


baconpaul:
Yeah that’s way better. Thanks guys!

PR #700, opened 2019-03-02
A Proposal for organizing Headless

baconpaul:
This PR expands the capability of the headless app to start adding
functions and helpers we can use to debug. Specifically this adds
HeadlessApps.h and HeadlessTools.h which have various top level
activities and tools you can use, and replaces the main (which is
still available as Surge::Headless::runSimpleSynthToStdout) with
a program which loads all the presets into a headless surge.

Soon I'll make it so it can play them too, probably by adding
a "HeadlessPlayer.h/cpp" which plays various note sequences
into various synths.

### DIFF
--- a/src/headless/HeadlessTools.cpp
+++ b/src/headless/HeadlessTools.cpp
@@ -1,0 +1,29 @@
+#include "HeadlessApps.h"
+#include "HeadlessPluginLayerProxy.h"
+
+namespace Surge
+{
+namespace Headless
+{
+
+SurgeSynthesizer* createSurge()
+{
+   HeadlessPluginLayerProxy* parent = new HeadlessPluginLayerProxy();
+   SurgeSynthesizer* surge = new SurgeSynthesizer(parent);
+   surge->setSamplerate(44100);
+   return surge;
+}
+
+bool loadPatchByIndex(SurgeSynthesizer* s, int index)
+{
+   if (index > s->storage.patch_list.size() || index < 0)
+      return false;
+
+   s->patchid_queue = index;
+   s->processThreadunsafeOperations();
+
+   return true;
+}
+
+} // namespace Headless
+} // namespace Surge
--- a/src/headless/HeadlessTools.h
+++ b/src/headless/HeadlessTools.h
@@ -1,0 +1,25 @@
+#include "SurgeSynthesizer.h"
+
+namespace Surge
+{
+namespace Headless
+{
+
+/**
+ * createSurge
+ *
+ * Create an instance of a SurgeSynthesizer object correctly configured
+ * to run in headless mode
+ */
+SurgeSynthesizer* createSurge();
+
+/**
+ * loadPatchByIndex
+ *
+ * Appropriately loads the surge patch at "index", returning false if
+ * no patch is available
+ */
+bool loadPatchByIndex(SurgeSynthesizer* s, int index);
+
+} // namespace Headless
+} // namespace Surge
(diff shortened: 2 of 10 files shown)


baconpaul:
@jsakkine I am not committed to merging this code, first of all, but it is turning out to be useful in a couple of ways, so wanted to let you see it in case you agree or disagree with the direction. Totally happy to close this PR, but like I mentioned on slack would be great to get a way to try and debug clicks and pops with headless.

Anyway this diff adds helper functions and starts reorganizing headless so that on macOS it loads every patch once in category order. Kinda nice. Pretty easy to make it start playing sound into a buffer also using the pattern implied here.

Curious as to your thoughts. Like I said this PR is as much a design proposal for us to discuss as it is a request to merge!

src/headless/HeadlessTools.cpp
```diff
@@ -0,0 +1,29 @@
+#include "HeadlessApps.h"
+#include "HeadlessPluginLayerProxy.h"
+
+namespace Surge
+{
+namespace Headless
+{
+
+SurgeSynthesizer* createSurge()
```
jsakkine: Maybe a singleton and `getSurgeInstance()`?

baconpaul: I’d thought about that but I wanted to be able to simulate a host which was running 5 surge instances in one process. But some management utilities for the pointer would be useful I agree

baconpaul:
Obviously if this ends up merging I would fix whatever is breaking windows first :)

src/headless/HeadlessTools.cpp
```diff
@@ -0,0 +1,29 @@
...
+   return surge;
+}
+
+bool loadPatchByIndex(SurgeSynthesizer* s, int index)
+{
+   if (index > s->storage.patch_list.size() || index < 0)
+      return false;
+
+   s->patchid_queue = index;
+   s->processThreadunsafeOperations();
```
jsakkine: I looked in call sites and I think we could reorganize the code in a way that there would instead a function:

``
void SurgeSynthesizer::loadPatch(int id)
{
   s->patchid_queue = index;
   s->processThreadunsafeOperations(); 
}
```

I think it would be a good idea to abstract away the exact load mechanism.

Comment (baconpaul) [reply src/headless/HeadlessTools.cpp]: Yeah I agree. We can’t call it loadPatch I don’t think since SurgeSynthesizer already has one of those which does the innards, but we can call it something reasonable. “setPatchIndex” or something

Comment (jsakkine) [reply src/headless/HeadlessTools.cpp]: How do oadPatchInBackgroundThread and processThreadunsafeOperations() collaborate? The existing loadPatch() could loadPatchPrivate().

File: src/headless/HeadlessTools.h [hunk 3]
```diff
@@ -0,0 +1,25 @@
...
+ */
+SurgeSynthesizer* createSurge();
+
+/**
+ * loadPatchByIndex
+ *
+ * Appropriately loads the surge patch at "index", returning false if
+ * no patch is available
+ */
+bool loadPatchByIndex(SurgeSynthesizer* s, int index);
```
jsakkine: For consistency sake I would name this as loadPatch().

src/headless/HeadlessTools.cpp
baconpaul: yeah that’s the right thing to do. I haven’t figured out how the UI thread and audio thread and headless all collaborate in detail so I may be able to call the direct method here also. This needs resolving for sure.

jsakkine:
I would like to focus on the first commit first and then you could do a follow-up PR of the rest.

src/headless/HeadlessTools.cpp
baconpaul: Turns out we don't need this at all; since there is no "UI Thread" in headless we can just call loadPatch directly.

PR #704, opened 2019-03-02
A Proposal for organizing Headless

baconpaul:
This PR expands the capability of the headless app to start adding
functions and helpers we can use to debug. Specifically this adds
HeadlessApps.h and HeadlessTools.h which have various top level
activities and tools you can use, and replaces the main (which is
still available as Surge::Headless::runSimpleSynthToStdout) with
a program which loads all the presets into a headless surge.

### DIFF
--- a/src/headless/HeadlessApps.cpp
+++ b/src/headless/HeadlessApps.cpp
@@ -1,0 +1,105 @@
+#include "HeadlessApps.h"
+#include "HeadlessTools.h"
+#include <iostream>
+#include <iomanip>
+
+namespace Surge
+{
+namespace Headless
+{
+
+void runSimpleSynthToStdout(SurgeSynthesizer* surge)
+{
+   /*
+   ** Change a parameter in the scene. Do this by traversing the
+   ** graph in the current patch (which is in surge->storage).
+   **
+   ** Clearly a more fulsome headless API would provide wrappers around
+   ** this for common activities. This sets up a pair of detuned saw waves
+   ** both active.
+   */
+   surge->storage.getPatch().scene[0].osc[0].pitch.set_value_f01(4);
+   surge->storage.getPatch().scene[0].mute_o2.set_value_f01(0, true);
+   surge->storage.getPatch().scene[0].osc[1].pitch.set_value_f01(1);
+
+   /*
+   ** Play a note. channel, note, velocity, detune
+   */
+   surge->playNote((char)0, (char)60, (char)100, 0);
+
+   /*
+   ** Strip off some processing first to avoid the attach transient
+   */
+   for (auto i = 0; i < 20; ++i)
+      surge->process();
+
+   /*
+   ** Then run the sampler
+   */
+   int blockCount = 30;
+   int overSample = 8; // we want to include n samples per printed row.
+   float overS = 0;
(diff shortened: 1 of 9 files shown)


baconpaul:
OK @jjs this is the first-commit-only with your comments and the loadPatch change in place.

If you like this, lets sweep it and I will add the playNote stuff as a second PR, perhaps along with mac-only wav file rendering.

baconpaul:
Oh I meant to tag @jsakkine not @JJS

src/headless/HeadlessTools.h
```diff
@@ -0,0 +1,25 @@
...
+ */
+SurgeSynthesizer* createSurge();
+
+/**
+ * loadPatchByIndex
+ *
+ * Appropriately loads the surge patch at "index", returning false if
+ * no patch is available
+ */
+bool loadPatchByIndex(SurgeSynthesizer* s, int index);
```
jsakkine: This now unused.

src/headless/HeadlessApps.cpp
```diff
@@ -0,0 +1,105 @@
...
+
+            std::cout << "idx= " << std::setw(4) << i << "; cat = '" << pc.name << "'; patch = '"
+                      << p.name << "'" << std::endl;
+
+            surge->loadPatch(i);
+
+            /*
+            ** So lets play some notes and gather some waveforms
+            ** FIXME
+            */
```
jsakkine: Maybe you could encapsulate your old note playing code to a function and call it here? Instead of  runSimpleSynthToStdout() I would create playNote(int id) and do not worry about sink at this point. I would rename this function as playNoteOnAllPresets().

src/headless/HeadlessApps.h
```diff
@@ -0,0 +1,23 @@
...
+ * Shows how to create and run a synth putting output to stdout
+ */
+void runSimpleSynthToStdout(SurgeSynthesizer* surge);
+
+/**
+ * scanAllPresets
+ *
+ * Loads each preset, scans for volume characteristics under a few playing modes
+ */
+void scanAllPresets(SurgeSynthesizer* surge);
```
jsakkine:
```
void playNote(SurgeSynthesize* surge);

void playNoteOnAllPresets(SurgeSynthesize* surge);
```

We can always refactor the API later on more generic but these exactly describe what is going on right now.

src/headless/HeadlessApps.cpp
baconpaul: That’s what happens in the extra commit yeah. Your other comment made me think I should factor this properly now though. So pass in collectors and generators cleanly. That’s sort of in my other commit but let me go back to the drawing board and try something a bit more expansive and send another commit over which does that. It will be quite different

But thank you. Your comments have made me realize what I was doing and how I want to factor!

src/headless/HeadlessApps.h
baconpaul: Yeah I will do a bit more expansive a factor now that I see what you see in the current code. Very very useful as always thank you!

src/headless/HeadlessApps.cpp
jsakkine: Great, NP :)

PR #707, opened 2019-03-02
A More Expansive Headless Proposal

baconpaul:
This commit suggests a way to implement Headless which is as a
player of internal events to buffers. For now the buffers are
vectors as are the events. Right now, this implements the
play and monitor of every patch with a simple C major scale,
as well as a quick test of a direct patch configuration.

APIs for Wav and Gnuplot are declared but not implemented.

### DIFF
--- a/src/headless/HeadlessPluginLayerProxy.h
+++ b/src/headless/HeadlessPluginLayerProxy.h
@@ -8,6 +8,5 @@ class HeadlessPluginLayerProxy
 public:
     void updateDisplay()
     {
-        std::cerr << "HeadlessPluginLayerProxy::updateDisplay" << std::endl;
     }
 };
--- a/src/headless/HeadlessUtils.h
+++ b/src/headless/HeadlessUtils.h
@@ -1,0 +1,27 @@
+/*
+** HeadlessUtils provides a collection of utility functions to manage and output
+** data and construct and manipulate surge
+*/
+#pragma once
+
+#include "SurgeSynthesizer.h"
+
+namespace Surge
+{
+namespace Headless
+{
+
+SurgeSynthesizer* createSurge(int sr);
+
+void writeToStream(const float* data, int nSamples, int nChannels, std::ostream& str);
+
+/*
+** One imagines expansions along these lines:
+
+void writeWavFile(const float * data, int nSamples, int nChannels, std::string fname);
+void writeGnuplotFile(const float *data, int nSamples, int nChannels, std::string fname);
+
+*/
+
+} // namespace Headless
+} // namespace Surge
--- a/.gitignore
+++ b/.gitignore
@@ -67,3 +67,6 @@ premake-stamp
 /Debug
 src/linux/ScalablePiggy*
 *.deb
+
+# cmake
+build/
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -71,6 +71,8 @@ add_executable(surge-headless
    src/headless/DisplayInfoHeadless.cpp
    src/headless/UserInteractionsHeadless.cpp
    src/headless/LinkFixesHeadless.cpp
+   src/headless/HeadlessUtils.cpp
+   src/headless/Player.cpp
 )
 
 target_compile_features(surge-headless
--- a/build-osx.sh
+++ b/build-osx.sh
@@ -369,8 +369,11 @@ case $command in
         run_build_install_vst3
         ;;
     --build-headless)
-        run_premake_if
         run_build_headless
+        ;;
+    --run-headless)
+        run_build_headless
+        ./build/Release/surge-headless 
         ;;
     --clean)
         run_clean_builds
--- a/premake5.lua
+++ b/premake5.lua
@@ -673,7 +673,10 @@ if (os.istarget("windows")) then
         "src/headless/main.cpp",
         "src/headless/DisplayInfoHeadless.cpp",
         "src/headless/UserInteractionsHeadless.cpp",
-        "src/headless/LinkFixesHeadless.cpp"
+        "src/headless/LinkFixesHeadless.cpp",
+        "src/headless/HeadlessUtils.cpp",
+        "src/headless/Player.cpp"
+        
     }
 
     excludes
--- a/src/headless/HeadlessUtils.cpp
+++ b/src/headless/HeadlessUtils.cpp
@@ -1,0 +1,39 @@
+#include "HeadlessUtils.h"
+#include "HeadlessPluginLayerProxy.h"
+
+#include <iostream>
+#include <iomanip>
+
+namespace Surge
+{
+namespace Headless
+{
+SurgeSynthesizer* createSurge(int sr)
+{
+   HeadlessPluginLayerProxy* parent = new HeadlessPluginLayerProxy();
+   SurgeSynthesizer* surge = new SurgeSynthesizer(parent);
+   surge->setSamplerate(sr);
+   return surge;
+}
+
+void writeToStream(const float* data, int nSamples, int nChannels, std::ostream& str)
+{
+   int overSample = 8;
+   float avgOut = 0;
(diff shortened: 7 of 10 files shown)


baconpaul:
OK @jsakkine here you go. This is sort of what you were getting at and what I should have done first time around. It's a pretty big hunk of proposal but I look forward to your thoughts.

jsakkine:
Well what if you just cut off Application.h/c and the Apps-namespace and declare them as static functions in main.cpp? I don't see the point. It is over-engineered.

src/headless/Applications.h
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * statsFromPlayingEveryPatch
+ *
+ * Play a C Major scale on every patch; and see what happens
+ */
+void statsFromPlayingEveryPatch();
+
+} // namespace Apps
+} // namespace Headless
+} // namespace Surge
```
jsakkine: Seriously, just two simple local functions without any namespaces in main.cpp would be just fine:

```
static void testSimpleTone()
{
}

static void testScaleOnAllPatches()
{
}

int main(int argc, char **argv)
{
   testSimpleTone();
   testScaleOnAllPatches();
}

```

src/headless/HeadlessPluginLayerProxy.h
```diff
@@ -8,6 +8,6 @@ class HeadlessPluginLayerProxy
 public:
     void updateDisplay()
     {
-        std::cerr << "HeadlessPluginLayerProxy::updateDisplay" << std::endl;
+       // std::cerr << "HeadlessPluginLayerProxy::updateDisplay" << std::endl;
```
jsakkine: You should make a decision to leave it or keep it.

src/headless/HeadlessUtils.h
```diff
@@ -0,0 +1,25 @@
...
+{
+namespace Headless
+{
+
+SurgeSynthesizer* createSurge(int sr);
+
+typedef std::vector<float> audioData_t;
+
+void writeWavFile(const audioData_t& data, int nChannels, std::string fname);
+void writeGnuplotFile(const audioData_t& data, int nChannels, std::string fname);
```
jsakkine: Even if we know that these would be needed there is no really any good reason to add any dummy functions. When you don't have to keep a stable API/ABI things can be always refactored if they turn out to be not a good fit.

src/headless/HeadlessUtils.h
```diff
@@ -0,0 +1,25 @@
...
+
+void renderDataTo(const audioData_t& data, int nChannels, std::ostream& str);
```
jsakkine: `void writeToString(const void *data,  size_t size, int nChannels,  std::ostream& str);`

The typedef does make the code any cleaner. It only causes a distraction when you have to cross-reference it.  Here the raw data is probably the best option.

src/headless/HeadlessUtils.h
baconpaul: Agree completely; in this case these were part of an ‘implied api’ for your consideration. If these functions existed would you use these signatures? Again this PR is more a proposal than a bit of code. (Although I have a draft of writeWavFile on mac kicking around)

src/headless/HeadlessUtils.h
baconpaul: Ooh `const void*` vs a type? I actually don’t agree with that choice! Why would you do that?

src/headless/HeadlessPluginLayerProxy.h
baconpaul: Yeah indeed! Sorry about that. It should go

src/headless/Applications.h
baconpaul: OK I can do that. But main is gonna get messy as heck when we have 19 flavors which is why I started splitting them out. I can go either way now of course. They are simple and tiny functions.

jsakkine:
Looks going somewhere. Just don't like when code does "proactive" things when you don't have any hard reasons to do so. The only hard reason that exist in this universe to do so is when you freeze an API/ABI of some kind. Only freezed APIs that we have are the ones that come from outside (e.g. VST2 and VST3).

baconpaul:
> Well what if you just cut off Application.h/c and the Apps-namespace and declare them as static functions in main.cpp? I don't see the point. It is over-engineered.

Yeah I was thinking of when we had more flavors of activities this could take but at this point I can’t just kill the application file.

jsakkine:
J

> > Well what if you just cut off Application.h/c and the Apps-namespace and declare them as static functions in main.cpp? I don't see the point. It is over-engineered.
> 
> Yeah I was thinking of when we had more flavors of activities this could take but at this point I can’t just kill the application file.

Not sure exactly. Not really a massive change to what it is right now.

baconpaul:
I meant “I can just kill” not “I can’t just kill”. Will nuke the application file remove the unused api points and send another version across the transom tomorrow! Thanks as always for a careful review

src/headless/HeadlessPluginLayerProxy.h
jsakkine: Nothing to be sorry about, just pinpointing what I see :)

src/headless/Applications.h
jsakkine: This commit introduces already much new structure. Maybe the best route would be to put them right now to main and then look into this after this has been landed?

jsakkine: In the scope of just commit this like the amount of complexity that makes sense.

src/headless/HeadlessUtils.h
jsakkine: Probably yes. But we have another review process lets say for my gnuplot commits :) We will then look at the best alternative, which is probably the above.

src/headless/HeadlessUtils.h
jsakkine: It was a typo!

baconpaul: Gotcha. Yeah so `const float* data, size_t size` vs `std:;vector<float>` - I prefer the vector if you don't particularly mind. Easier to do memory management and the like. Happy to change to that.

src/headless/HeadlessUtils.h
baconpaul: Yup cool so I will remove it. It was definitely one of those "could I imagine the API" things not "this is finished" things :)

src/headless/HeadlessUtils.h
jsakkine: I'm just a big fan of loose binding as far as it possible makes sense.  For example, you could use the same function to process only portion of the data without having to use iterators and such things. With either signature obviously the function does not do any memory management on the parameter.

jsakkine: - Standard call would something like `writeToStream(result.data(), result.size(), 2, std::cout);`.
- Lets say you want to render `maxSamples` amount of samples is where you want to cut. Then you can just `writeToStream(result.data(), result.data() + maxSamples, 2, std::cout);`.

There is a lot robustness value to do this as you can probably see...

baconpaul: OK you convinced me. But the accumulators still need a place to put the data; I'll keep a vector& there rather than a float** int* if that's OK with you.

The STL idiom of iterator from and to are a bit clunky. but we can insert the thousand year history of C and STL here.

baconpaul: Actually you know what I'll just change it everywher

jsakkine:
OK, I'll just hold for the next review! Enough of comments :)

baconpaul:
Just force pushed a version with your review comments in. I think we are converging. Let me know what you think.

I have some dev time tomorrow but I have plenty to do in other issues if you don't get a chance to look.

jsakkine:
Don't see anything over the top! Please merge this. Thanks for the patience.

baconpaul:
Great thanks!

PR #714, opened 2019-03-04
Python based .wt file manipulator

### LINKED ISSUE #522: kvr-ideas: a program that takes 16bit mono single-cycle wavfiles and assembles them into 1024 length .wt files for surge
from https://www.kvraudio.com/forum/viewtopic.php?p=7306639

----
i'm not a programmer or i would do this: a program that takes a bunch of mono 16bit single-cycle wav files in a folder and randomly assembles them in your desired waveform length .wt files for surge, then dump the assembled .wt's into another folder. would be nice, huh?
----

baconpaul:
This adds a script which gathers info on, explodes to constituents
and creates from constituents a surge .wt file. This command line
gives is a lot mroe option to create and modify .wt files.

Closes #522 program to make wt from wav

### DIFF
--- a/doc/wavetables.md
+++ b/doc/wavetables.md
@@ -1,0 +1,64 @@
+# Surge WaveTables
+
+Surge has a wavetable oscillator which plays a wavetable file. A wavetable is a collection 
+of short (128-1024 or so) sample waves which have multiple waves in a single file. The 'shape'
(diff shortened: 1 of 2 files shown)


baconpaul:
This is purely additive extra so I'm just going to merge it and let others improve as they see fit.

PR #728, opened 2019-03-05
disable cursors in save dialog from changing presets

### LINKED ISSUE #727: Critical: Disable Cursor keys from bleeding to Surge when Patch save dialog is visible
as reported by @sense-amr 

let's disable cursor keys from save dialog so that they do not bleed to Surge and change presets. 

You want to be able to go left and right with cursor keys while saving preset name, without losing the preset.

esaruoho:
fixes #727 but in a blunt / heavyhanded way.

### DIFF
--- a/src/common/gui/SurgeGUIEditor.cpp
+++ b/src/common/gui/SurgeGUIEditor.cpp
@@ -506,15 +506,43 @@ int32_t SurgeGUIEditor::onKeyDown(const VstKeyCode& code, CFrame* frame)
             ** event loop)
             */
         case VKEY_LEFT:
+            if (saveDialog && saveDialog->isVisible())
+            {
+               /* 
+               ** SaveDialog gets access to the cursor keys if it is open
+               */
+               return -1;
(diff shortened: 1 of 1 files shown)


esaruoho:
tested with RenoiseVST2 and LogicAU. cursors work in save dialog and don't change the presets while in the save dialog.

esaruoho:
here's what baconpaul said

yeah it’s the exact same diff. Just move it from the tab block up to the entire keyhandling block.
I can get this tonight. Or if you just move that block to above the switch statement rather than after the tab case it wil work

_Originally posted by @baconpaul in https://github.com/surge-synthesizer/surge/pull/641#issuecomment-469686433_



-- i think i might be able to figure out how to do it maybe, if i read it very carefully.

"move the block, above the switch statement instead of after tab case". let's see.

baconpaul:
This is ok for now but we should clean it up to do that check once and set a bool. If you push this and make a new ticket I can do that or you can

Issue #720, opened 2019-03-04
Linux: error compiling headless


tank-trax:
@jsakkine @baconpaul 

when building VST2, VST3 and headless on Linux

uname -a => Linux LinuxDAW 4.9.0-8-rt-amd64 #1 SMP PREEMPT RT Debian 4.9.144-3.1 (2019-02-19) x86_64 GNU/Linux

cmake --version => cmake version 3.7.2

gcc --version => gcc (Debian 6.3.0-18+deb9u1) 6.3.0 20170516

the VST2 and VST3 build but headless has errors

```
build of surge-vst3 succeeded                                                                                                                                                                                                               

Building surge-headless with output in build_logs/build_headless.log                                                                                                                                                                        
-- The CXX compiler identification is GNU 6.3.0                                                                                                                                                                                             
-- The ASM compiler identification is GNU                                                                                                                                                                                                   
-- Found assembler: /usr/bin/c++                                                                                                                                                                                                            
-- Check for working CXX compiler: /usr/bin/c++                                                                                                                                                                                             
-- Check for working CXX compiler: /usr/bin/c++ -- works                                                                                                                                                                                    
-- Detecting CXX compiler ABI info                                                                                                                                                                                                          
-- Detecting CXX compiler ABI info - done                                                                                                                                                                                                   
-- Detecting CXX compile features                                                                                                                                                                                                           
-- Detecting CXX compile features - done                                                                                                                                                                                                    
CMake Error at CMakeLists.txt:78 (target_compile_features):                                                                                                                                                                                 
  target_compile_features specified unknown feature "cxx_std_17" for target                                                                                                                                                                 
  "surge-headless".                                                                                                                                                                                                                         


-- Looking for C++ include pthread.h                                                                                                                                                                                                        
-- Looking for C++ include pthread.h - found                                                                                                                                                                                                
-- Looking for pthread_create                                                                                                                                                                                                               
-- Looking for pthread_create - not found                                                                                                                                                                                                   
-- Looking for pthread_create in pthreads                                                                                                                                                                                                   
-- Looking for pthread_create in pthreads - not found                                                                                                                                                                                       
-- Looking for pthread_create in pthread                                                                                                                                                                                                    
-- Looking for pthread_create in pthread - found                                                                                                                                                                                            
-- Found Threads: TRUE                                                                                                                                                                                                                      
-- Configuring incomplete, errors occurred!                                                                                                                                                                                                 
[...]

PR #732, opened 2019-03-06
Remove a C++-17ism

baconpaul:
the c++-17-ism auto [k,v] = (thing-returning-pair) doesn't work with
gcc6 it seems. So knock it out rather than requiring users to
upgrade their compiler.

### DIFF
--- a/src/headless/main.cpp
+++ b/src/headless/main.cpp
@@ -51,7 +51,9 @@ void statsFromPlayingEveryPatch()
 
       if (nSamples * nChannels > 0)
       {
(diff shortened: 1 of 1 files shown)


baconpaul:
Should end up closing #720 for real.

tank-trax:
Debian 9 is held at cmake 3.7 and ran into errors build surge headless, using a locally compiled and installed more recent version of CMAKE version 3.14

`Build of surge-headless succeeded`

tank-trax:
./surge-headless executes and enumerates all presets and wavetables in succession

baconpaul:
Thanks! This is a handy idiom but one I can easily skip

PR #735, opened 2019-03-06
Remove Arrow Keys altogether from changing category+presets since it was destructive

### LINKED ISSUE #729: check cursor+tab boolean only once
"This is ok for now but we should clean it up to do that check once and set a bool. If you push this and make a new ticket I can do that or you can" from @baconpaul 

i tried to make it like this:
image

not sure if that's valid though

### LINKED ISSUE #734: remove cursor keys (up,down,left,right) from doing anything at all in the code
as per discussion at #729 , related to making shift to #733 less drastic.

esaruoho:
This _should_ remove all cursor up down left right keys from working in Surge (no more category+preset changing). This saves the user from accidentally destroying their preset with an accidental cursor key press.

closes #734 
closes #729


baconpaul:
if not linux, that is

PR #736, opened 2019-03-06
Surge32.dll -> Surge_x86.dll = fixes Windows issues

### LINKED ISSUE #610: surge_32.dll apparently causes issues with windows users because of other plugins by that name
@baconpaul suggested `Surge32bit.dll` 

from KVR:
```
32bit Windows keeps giving:
”The 32-bit Windows build has the same file name (Surge32.dll) as 32-bit Airwindows Surge, http://www.airwindows.com/surge/
At least 32-bit REAPER finds/loads only one of these, might be a problem in some others too but I have only REAPER to test with (and Jeskola Buzz 32-bit as a sort of a relic, loads both plugins fine).
Both plugins load in the 64-bit version of REAPER as the file names are different (and 64-bit REAPER is what I use anyway).”
```

So looks like filenames could be tweaked.

esaruoho:
fixes #610 

where surge32.dll would actually accidentally open some completely different plugin. this way the names are completely separated for .dll and .vst3.

### DIFF
--- a/premake5.lua
+++ b/premake5.lua
@@ -132,7 +132,7 @@ elseif (os.istarget("windows")) then
     platforms { "x64", "x86" }
 
     filter "platforms:x86"
-        targetsuffix "_32bit_x86"
+        targetsuffix "_x86"
     filter "platforms:x64"
         targetsuffix ""
     filter {}
--- a/premake5.lua
+++ b/premake5.lua
@@ -132,7 +132,7 @@ elseif (os.istarget("windows")) then
     platforms { "x64", "x86" }
 
     filter "platforms:x86"
-        targetsuffix "32"
+        targetsuffix "_32bit_x86"
     filter "platforms:x64"
         targetsuffix ""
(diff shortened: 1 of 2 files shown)


premake5.lua
```diff
@@ -132,7 +132,7 @@ elseif (os.istarget("windows")) then
     platforms { "x64", "x86" }
 
     filter "platforms:x86"
-        targetsuffix "32"
+        targetsuffix "_32bit_x86"
```
kzantow: why not just `Surge_x86.dll`? since `x86` basically means `32-bit`

baconpaul:
Also what’s gonna happen to a user who installed the prior version? They have two surges? Curious how we deal with that

esaruoho:
@baconpaul it would appear that the user would need to remove the previous version..

premake5.lua
esaruoho: resolved

Issue #710, opened 2019-03-04
tixml locale / Presets don't parse in locales which don't use "." as a decimal separator


tavasti:
_Edit: this Original description and beginning if discussion is on false assumptions of reason. Read more at the end_

When running surge in xubuntu 18.04, bit part of the presets do not produce any sound. With failing presets, surges own output meter stays on zero. 

1. Download xubuntu install media https://xubuntu.org/download
2. Install to new virtual machine
3. On fresly installed VM, run 'sudo apt install ardour qjackctl'
4. Download surge deb, install it and dependencies
5. Start jack.
6. Run ardour, create project with jack, add midi track, add surge to track, set pinnings correct
7. With init-patch you should get sound. Start selecting another patches, and most of them do not produce any sound

Presets that work:
From beginning of Bass presets:
Bass 10, Deep end, Distorted MW, FM Bass 1
From beginning of pads: Ghost pad

Tested with 48k and 44.1k and different Frames/period (256,512,1024)
Have tested with 2 different USB audio interfaces, and also with virtual machines default audio interface.
Same virtual machine with ubuntu studio live dvd, all presets work, so it is verified that this is not hardware problem.

baconpaul:
To clarify your last comment: ubuntu studio live dvd, a different distribution, you can load surge and play every preset successfully? So these presets are only broken in your xubunutu install?

Any ideas on what's different between the two distributions?

Thank you for the clear report.

tavasti:
Yes, with ubuntustudio dvd, everything works. Don't have any idea what is difference. I have also one xubuntu machine where surge works, but at least all libraries which surge depends are identical. 

Tried also build surge myself, no difference

baconpaul:
OK so @jsakkine and I just pushed the first version of a headless which actually plays samples to an internal buffer. This will be useful to understand how it works on your system, especially since you can build the synth. So if you don't mind could you do us the following on the broken system

1. Get the latest master from the surge-synthesizer repo. 
2. in your surge directory with latest master do this:

```
./build-linux.sh clean-all
./build-linux.sh build --project=headless
./build/surge-headless
```

this will start a program running which plays every pad with a C major scale into a buffer and then prints the min and max sample. You should see output like this

```
cat = Monosynth (smooth)            ; patch = Glisslead                     ; data = [-0.830299 , 0.831995  ] in 175297 samples
[...]

PR #745, opened 2019-03-07
Fix locale problems with preset loading and saving

tavasti:
Preset loading and saving used sscanf and sprintf. This will fail
with locales which have some other decimal separator than '.'
Closes surge-synthesizer/surge#710

### DIFF
--- a/libs/xml/tinyxml.cpp
+++ b/libs/xml/tinyxml.cpp
@@ -23,11 +23,11 @@ distribution.
 */
 
 #include <ctype.h>
+#include <clocale>
+#include <iomanip>
+#include <sstream>
 #include "tinyxml.h"
 
-#ifdef TIXML_USE_STL
-#include <sstream>
-#endif
 
 
 bool TiXmlBase::condenseWhiteSpace = true;
@@ -722,9 +722,15 @@ void TiXmlElement::SetAttribute( const char * name, int val )
 
 void TiXmlElement::SetDoubleAttribute( const char * name, double val )
 {	
-	char buf[128];
-	sprintf( buf, "%f", val );
-	SetAttribute( name, buf );
+	char buf[64];
+	std::stringstream sst;
+	sst.imbue(std::locale::classic());
+	sst << std::fixed;
+	sst << std::showpoint;
+	sst << std::setprecision(6);
+	sst << val;
+	strncpy( buf, sst.str().c_str(), 63 );
+	SetAttribute( name, buf);
 }
 
 
@@ -1126,7 +1132,10 @@ int TiXmlAttribute::QueryIntValue( int* ival ) const
 
 int TiXmlAttribute::QueryDoubleValue( double* dval ) const
 {
-	if ( sscanf( value.c_str(), "%lf", dval ) == 1 )
+	std::stringstream sst;
+	sst.imbue(std::locale::classic());
+	sst << value.c_str();
+	if ( sst >> *dval )
 		return TIXML_SUCCESS;
 	return TIXML_WRONG_TYPE;
 }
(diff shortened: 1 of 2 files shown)


libs/xml/tinyxml.cpp
```diff
@@ -1141,7 +1152,13 @@ void TiXmlAttribute::SetIntValue( int _value )
 void TiXmlAttribute::SetDoubleValue( double _value )
 {
 	char buf [64];
-	sprintf (buf, "%lf", _value);
+	std::stringstream sst;
+	sst.imbue(std::locale::classic());
+	sst << std::fixed;
+	sst << std::showpoint;
+	sst << std::setprecision(6);
```
baconpaul: Why 6? Doubles have up to 13 points of precision. Perhaps removing the fixed, showpoint, and precision is best to match the original %lf behavior; but if you want to keep them at least setting the precision higher than 6 seems appropriate for a double.

tavasti:
Replacement of sprintf in tinyxml.cpp are not tested, because I don't know even if they are used.

Preset writing sprintf is in src/common/Parameter.cpp

baconpaul:
Thank you very much for doing this! I appreciate it.

I had a question about setpreicision(6) which seems too low. Doubles have up to 13 points of precision (floats have 6). 

I appreciate you finding and fixing the one in Parameter very much. Thanks!

And also (although I’m sure you tested this) with this change you can load all the patches properly right? So cool. Thank you so much for finding this.

libs/xml/tinyxml.cpp
tavasti: 6 is default value of sprintf, and presets this far have used that.

tavasti: Without fixed and showpoint instead of value="1.000000" we will get value="1"
Sure functionally that is same.

tavasti: And for higher precision, I would expect that more than 6 decimals would cause any audible difference in any parameter :-)

tavasti:
Yes, tested that loading factory presets work ok, and saving produces similar output as factory presets.

libs/xml/tinyxml.cpp
baconpaul: OK that’s a pretty good argument about keeping same behavior. Sure great.

If the automated builds work, I’ll merge this.

If you’d like, please add yourself as you would like to be attributed in the AUTHORS file and send in a pull request. We are keeping it alphabetical but happy to have you added!

Thanks again

baconpaul:
Uh oh. The automated builds failed. Let me take a look at that later tonight or tomorrow before I sweep, OK?

we will definitely get this in before the end of the weekend. Thanks again

libs/xml/tinyxml.cpp
```diff
@@ -23,6 +23,8 @@ distribution.
 */
 
 #include <ctype.h>
+#include <clocale>
+#include <iomanip>
```
baconpaul: The windows build problem is you need to `#include <sstream>` to get std::stringstream building. Want to add that and push a commit to your branch and we can see the auto-builds again?

tavasti: What is preference, fix this old commit and forced push, or new commit to fix?

baconpaul: If you’re comfy with a rebase and force push, that’s better yeah. (Some of our devs aren’t and so I just squash at merge so I’m fine either way)

baconpaul:
Sigh looks like the mac CI pipeline is acting up - don’t think that fail is your code. I’ll see if @kzantow has any ideas.

I have some time tomorrow to build on my mac to confirm. Looks like windows is progressing properly with your push. Should be able to merge this then.

baconpaul:
OK as well as the automated builds, I manually built and ran this branch on both mac and windows to make sure it could still load and save patches; and it can. So I'm going to merge this.

But lets all remember if anyone reports an odd patch load behavior that this change is in there. flagging to @esaruoho for that reason.

Issue #746, opened 2019-03-07
Selecting OSC type FM2 will be shown and FM3


tavasti:
When I select OSC type FM2, OSC type FM3 is shown. Parameter setting seem to be bit different on selecting FM2 and FM3, so I assume this is just copy-paste in string. 

**Desktop (please complete the following information):**
 - OS: Linux
 - Host carla

baconpaul:
That’s actually an error in the bitmaps! I noticed it when porting to SVG but forgot to open an issue.

@itsmedavep check this out! 

![](https://github.com/surge-synthesizer/surge/blob/master/assets/original-vector/exported/bmp00119%402x.png)

So good we have FM3 twice :)

Good eyes @tavasti thanks!

itsmedavep:
Done now for the 1.6 bitmaps here:

https://github.com/surge-synthesizer/surge/pull/748

itsmedavep:
> When I select OSC type FM2, OSC type FM3 is shown. Parameter setting seem to be bit different on selecting FM2 and FM3, so I assume this is just copy-paste in string.
> 
> **Desktop (please complete the following information):**
> 
> * OS: Linux
> * Host carla

Thanks for reporting this! Nice catch and sorry for any frustrations I may have caused.

PR #748, opened 2019-03-07
FIx for https://github.com/surge-synthesizer/surge/issues/746 Fixing …

itsmedavep:
…FM2 Label

Fixing FM2 Label


itsmedavep:
FWIW it is already fixed and checked in on the SVG branch.

PR #778, opened 2019-03-13
Headless can write wav files

baconpaul:
To debug click n pop I want headless to be able to make a file
I can hear. This change allows that by (1) making CMakeLists
check for libsndfile and (2) if it is there building a libsndfile
writer

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,8 @@
 cmake_minimum_required(VERSION 3.10)
 project(Surge VERSION 1.0.0 LANGUAGES CXX ASM)
 
+list(APPEND CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/cmake/")
+
 set(SURGE_COMMON_SOURCES
     src/common/dsp/effect/ConditionerEffect.cpp
     src/common/dsp/effect/DistortionEffect.cpp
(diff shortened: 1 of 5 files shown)


baconpaul:
@jsakkine hey. When debugging click-n-pop I realized I wanted headless to be able to write wav files. The change to debug click and pop is bigger than this one (including making headless no longer assume 44.1k as it does here). But along the way I had to figure out how to create a .wav

So this change adds the (incredibly small) libsndfile code to write a wav file. But it also does the cmake shenanigans to get libsndfile if it is installed and not if it isn't. I tested this by building on mac where i had libsndfile, then on linux where I didn't, then installing libsndfile and building again. It all worked as expected.

The packages are

```
brew install libsndfile
```

on mac and

```
sudo apt-get install libsndfile1-dev
```

on ubuntu

I am *not* going to sweep this for the beta7 tomorrow. But wanted to put it here for (1) you to review and (2) if you think it is good to go for you to merge.

I can get on with my debugging without this in master so no rush.

baconpaul:
Oh also I didn't use `find_package` because libsndfile has (apparently) an experimental CMake build only so I used the find_library / find_path combo.

baconpaul:
Actually I am going to force push this so sample rate is an argument. Hold on before you review.

baconpaul:
OK @jsakkine I just pushed a version which does exactly this. You are right - it is cleaner. Makes me think about how to factor the CMakeFile generally... but have to make headless work windows before we think about bigger things.

Anyway I'm not going to merge this before beta 7 tomorrow morning. But if you want to merge it while I'm away or fix it up and merge it, that would be super. I am sure it will help folks working on headless and c-n-p problems.

Cool stuff!

jsakkine:
Yeah, now it looks great!

baconpaul:
OK I'm going to merge it then, now that beta 7 is out!

Issue #783, opened 2019-03-16
Small GUI bug with FX Bypass buttons


mkruselj:
![fx byp offset](https://user-images.githubusercontent.com/2393720/54478626-8c2fcf00-4814-11e9-9b99-2c3a1f4f2487.gif)

That's at 125% and 150% zoom. Some other zoom levels just seem to offset Send+Master button. W10 64-bit.

itsmedavep:
I have a PR for the send+master button label. That is on this PR

https://github.com/surge-synthesizer/surge/pull/784

I cant repro your button frame jiggle issue on my mac unfortunately. So Im not sure if my PR will resolve that part of it. 

150%
![FX150](https://user-images.githubusercontent.com/11078351/54484893-1b50dd00-483d-11e9-915c-66f1d6152ebf.gif)

125%
![FX](https://user-images.githubusercontent.com/11078351/54484894-1b50dd00-483d-11e9-9e21-dd279e0e1696.gif)

baconpaul:
Dave: Rememberer on mac it uses 2x resolution since you have a retina display. I find that if I edit src/mac/DisplayInfo.cpp and make BackingStore return 1.0 I can often repro windows issues more easily (since the windows HDPI API is too painful for me to support).

itsmedavep:
Nice! TY for that bit of info. I’ll fold that into my testing regimen.

baconpaul:
Np. Oh and svg should fix most of this. Once I’m back from vaca! (To which I now return :))

itsmedavep:
![Testing@125x](https://user-images.githubusercontent.com/11078351/54570868-3f7ffb80-49ae-11e9-9d3f-236d1468bfc0.gif)


Ive tried what Paul suggested to get at 1x desktop usage. So the above gif is with returning 1 with BackingStore. Im unable to repro the button frame wiggle present in Evil's video.

kzantow:
Pretty sure @itsmedavep fixed this one; please reopen if the latest nightly still seems to have the issue.

mkruselj:
Unfortunately still happens in beta-8 over here :(

mkruselj:
Fixed in 165844b nightly!

baconpaul:
Awesome thanks! Closing.

PR #784, opened 2019-03-17
fixing placement of send + master labels

itsmedavep:
Somehow send+master label was slightly moved. I must have nudged it.

This is in issue https://github.com/surge-synthesizer/surge/issues/783

Im not sure if this entirely fixes that issue though. I only see the send+master label being off. I cant seem to see the button frames movement that is reported there. 

Screen Shot 2019-03-16 at 10 20 41 PM


kzantow:
Tested this; nice one @itsmedavep !

PR #792, opened 2019-03-30
Play MidiFiles in Headless mode

baconpaul:
To better debug click n pop I thought it woudl be useful if the
headless player could play midi files. Here's the first step
on that path to that, handling note on and note off events
and rendering a wav file, and showing a little example which renders
some bach.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -5,3 +5,6 @@
 	path = vstgui.surge
 	url = https://github.com/surge-synthesizer/vstgui.git
 	branch = surge
+[submodule "libs/midifile"]
+	path = libs/midifile
+	url = https://github.com/craigsapp/midifile.git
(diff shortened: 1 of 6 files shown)


baconpaul:
@jsakkine or @esaruoho any objections if I merge this? I wrote it on my first day of vaca so I could have a more robust test framework for click pop. Now I'm back figure I'd put it in as a pull request since it still looks reasonable 2 weeks later :)

Will require a git submodule update --init --recursive to work.

esaruoho:
i object nothing

baconpaul:
OK I'll let it sit for another day in case @jsakkine has a comment and then merge on Monday morning. Then see if I can get anywhere replicating click and pop. Thanks for the quick answer!

PR #802, opened 2019-04-01
UserInteractions: API update

jsakkine:
Add a new parameter guiEditor so that optionally implementations can use
VSTGUI to show dialogs. In particular, this will give a sane way for
Linux to display errors and prompts.


jsakkine:
Better to get this right first and deal with the GUI change separately. @baconpaul, this follows your advice.

baconpaul:
Yup that’s the right API. Fails to build mac and windows because of &error signature replacing &e signature variable name tho!

jsakkine:
Right, did not have my MacBook Pro available when doing this. I'll do testing on Windows and Mac tomorrow (also check that plug-in still works runtime).

jsakkine:
Hmm... Wonder what is causing the Windows compilation error.

baconpaul:
Just a guess: the cpp has only visibility on the class predefinition. Wonder if including surgeguieditor.h in the win cpp would fix it?

If it does that’s actually ok; as long as the userint header doesn’t link GUI we don’t gave a bad cycle.

Just a guess tho

jsakkine:
Aah! I'll try it out. BTW, just realized that nice thing in your design proposal is this: we can start using control in one place at a time so first version of the control does not have to do everything (since it depends on giving the parameter). Makes the whole effort manageable.

jsakkine:
Fixed the build issue.

baconpaul:
Oh thanks! Merging

PR #804, opened 2019-04-02
Move the QFC Init

baconpaul:
I had implmented initToZero as a mebmer of QuadFilterChain but
QuadFilterChain is align malloced in a wierd way. While this
hadn't caused a problem yet, it gave me concern that the additional
vtable element would misalign the FBQ _aligned_malloc array
and so I moved it to a standalone function.


baconpaul:
I hand confirmed that 32 bit windows surge works fine with this diff. (It may have worked fine before; this change is out of an abundance of caution). If the automated builds work properly I will merge this today and then stamp beta 8 later tonight.

PR #808, opened 2019-04-03
fonkle/linux-mousebehavior-793 adds optional settings for 'Mouse

fonkle:
Behavior' of the sliders to the 'Menu'.

One of 4 slider modes can be selected:

Classic: same as before (0.3 * moverate) = default
Slow:    (0.3)
Medium:  (0.7)
Exact:   (1.0)

On selecting either menu item, the new value get written to the system-
wide 'sliderMoveRateState' value in the defaults file
'SurgeUserDefaults.xml' which is respected when (re)opening Surge.

Note: I also added '*.txt.user' to .gitignore to ignore the
'CMakeLists.txt.user'-file that Qt Creator creates.

### DIFF
--- a/src/common/gui/CSurgeSlider.h
+++ b/src/common/gui/CSurgeSlider.h
@@ -16,15 +16,6 @@ enum CControlEnum_turbodeluxe
    kNoPopup = 1 << 22,
 };
 
-typedef enum MoveRateState
-{
-   kUnInitialized = 0,
-   kClassic,
-   kSlow,
-   kMedium,
-   kExact
-} MoveRateState;
-
 class CSurgeSlider : public CCursorHidingControl
 {
 public:
@@ -85,6 +76,15 @@ public:
 
    bool is_mod;
    bool disabled;
+
+   enum MoveRateState
+   {
+      kUnInitialized = 0,
+      kClassic,
+      kSlow,
+      kMedium,
+      kExact
+   };
 
    static MoveRateState sliderMoveRateState;
 
--- a/src/common/gui/CSurgeSlider.h
+++ b/src/common/gui/CSurgeSlider.h
@@ -16,6 +16,15 @@ enum CControlEnum_turbodeluxe
    kNoPopup = 1 << 22,
 };
 
+typedef enum MoveRateState
+{
+   kUnInitialized = 0,
+   kClassic,
+   kSlow,
+   kMedium,
+   kExact
+} MoveRateState;
+
 class CSurgeSlider : public CCursorHidingControl
 {
 public:
@@ -76,6 +85,8 @@ public:
 
    bool is_mod;
    bool disabled;
+
+   static MoveRateState sliderMoveRateState;
 
 private:
    VSTGUI::CBitmap *pHandle, *pTray, *pModHandle;
--- a/src/common/gui/SurgeGUIEditor.cpp
+++ b/src/common/gui/SurgeGUIEditor.cpp
@@ -1237,6 +1237,11 @@ void SurgeGUIEditor::openOrRecreateEditor()
    patchCreator = new CTextEdit(CRect(CPoint(96, 85), CPoint(340, 21)), this, tag_store_creator);
    patchComment = new CTextEdit(CRect(CPoint(96, 112), CPoint(340, 21)), this, tag_store_comments);
 
+   // Mouse behavior
+   if (CSurgeSlider::sliderMoveRateState == MoveRateState::kUnInitialized)
+      CSurgeSlider::sliderMoveRateState = (MoveRateState)Surge::Storage::getUserDefaultValue(
+          &(synth->storage), "sliderMoveRateState", (int)MoveRateState::kClassic);
+
    /*
     * There is, apparently, a bug in VSTGui that focus events don't fire reliably on some mac hosts.
     * This leads to the odd behaviour when you click out of a box that in some hosts - Logic Pro for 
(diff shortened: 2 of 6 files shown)


fonkle:
I hope all went well; I am rather new to this so please let me know what I can improve :).

baconpaul:
Hey thanks! I should have time to look tonight. Really appreciate it!

src/common/gui/CSurgeSlider.h
```diff
@@ -16,6 +16,15 @@ enum CControlEnum_turbodeluxe
    kNoPopup = 1 << 22,
 };
 
+typedef enum MoveRateState
```
baconpaul: I think this should be a member of CSurgeSlider not a standalone (which will mean you change the constants to CSurgeSlider::kUnInitialized and so on). My thinking is "it only makes sense in the context of the slider.

Alternately it could be a standalone named "SurgeSliderMoveRateState" I suppose but a member would be better if you don't mind!

src/common/gui/SurgeGUIEditor.cpp
```diff
@@ -2636,6 +2641,72 @@ void SurgeGUIEditor::showSettingsMenu(CRect &menuRect)
...
     mpeSubMenu->forget();
 
+    // Mouse behavior
+
+    int mid = 0;
+
+    COptionMenu* mouseSubMenu =
+        new COptionMenu(menuRect, 0, 0, 0, 0, VSTGUI::COptionMenu::kNoDrawStyle);
+
+    std::string mouseClassic = "Classic";
```
baconpaul: Hey cool! Exactly one part of the Surge code is ready for internationalization :) Good stuff here.

src/common/gui/SurgeGUIEditor.cpp
```diff
@@ -2636,6 +2641,72 @@ void SurgeGUIEditor::showSettingsMenu(CRect &menuRect)
     eid++;
     mpeSubMenu->forget();
 
+    // Mouse behavior
+
+    int mid = 0;
+
+    COptionMenu* mouseSubMenu =
+        new COptionMenu(menuRect, 0, 0, 0, 0, VSTGUI::COptionMenu::kNoDrawStyle);
```
baconpaul: As you'll see in a second probably want to make this `NoDrawStyle | MultiCheckStyle`

src/common/gui/SurgeGUIEditor.cpp
```diff
@@ -2636,6 +2641,72 @@ void SurgeGUIEditor::showSettingsMenu(CRect &menuRect)
...
+       break;
+    case MoveRateState::kMedium:
+       mouseMenuName.append("(Medium)");
+       break;
+    case MoveRateState::kExact:
+       mouseMenuName.append("(Exact)");
+       break;
+    default:
+       break;
+    }
```
baconpaul: This is a clever solution but the approach we've generally taken is to use menu checkmarks rather than menus this dynamic. I think the "Mouse Behavior (Slow)" menu would be a bit confusing compared to "Mouse Behavior >" and then "(check) Slow".  The only problem is addCallbackMenu doesn't let you set the check, but that's no problem. I'd suggest

1. Change the signature of addCallbackMenu to have a fourth argument `bool checked=false` and
2. Set that argument to the right boolean 

or

1. Change the signature of addCallbackMenu to return a CComandMenuItem *; and make it return menu then
2. Capture the return value in the parent

the VSTGUI API for checking a menu item is super simple. `menu->setChecked(true)`.

baconpaul:
OK this is super. Really nice clean correct code. 

I had two changes one of which was stylistic and one of which was functional but neither of which was big. Look forward to your feedback.

If you agree with the changes, feel free to make them, rebase, and force push your branch. It will just update this pull request and then run continuous integration again and then I can look.

When you do your push, though, could you also add yourself to the AUTHORS file in the second section at the appropriate alphabetical point in the manner you'd like to be recognized?

And thanks again! Super stuff! If you want I can look around for another set of issues for you to work in if interested. There's plenty out there :)

src/common/gui/CSurgeSlider.h
fonkle: I agree. I hadn't thought of that tbh.

src/common/gui/SurgeGUIEditor.cpp
fonkle: Yes I agree again. I will improve it.

fonkle: Actually I like the 2nd choice better since it interferes less in menus without `kMultipleCheckStyle`.

fonkle:
I think I have rebased and force pushed the branch. The code in this branch seems updated, so I guess it worked. It was the 1st time I did this, so please let me know if anything went wrong.

baconpaul:
Looks like the code came through. I had a couple of things cancel tomorrow morning so I’ll do a second look then! Thanks so much

baconpaul:
Hey this looks great and works perfectly. I'm going to merge it!

You could have rebased so the branch only had one commit - I'll update the git-howto.md with how to do that now - but I can do that when I merge also so it's no problem! (Just means I need to rewrite your commit message a little bit but that's no problem).

Thank you so much and welcome to the surge developer community!

baconpaul:
Oh and I added a bit of documentation at the end of https://github.com/surge-synthesizer/surge/blob/master/doc/git-howto.md on how to send a single commit up when you get code review feedback. Thanks again!

fonkle:
Thanks!

I must admit that I didn't understand the 'rebase' part completely (a lot of other commits showed up, probably due to my 'fetch upstream' inbetween. Sorry for that. I will use the _git-howto.md_ next time!

Currently I am finding out how to properly debug a VST within a host using Qt Creator and gdb. If I get my setup for that working, I would like to help with further improving Surge. Preferably on Linux and where needed on Windows.

I started using Surge recently, when the open port to Linux became available, so I am fairly new to both the user- and the developer aspects of it (as well as collaborative development in Github). While learning, I will no doubt run into either issues or new ideas. If there is anything particular I could help with in the meantime, please feel free to ask.

baconpaul:
Super

The developer guide in doc has comments on how to run and debug in various configurations. If you get your setup working please feel free to add to that doc. 

We’ve tagged Linux issues as Linux in the issue tracker pretty consistently. Some are more straight forward. And there’s some bigger standalone projects kicking around (like redo/undo) which could be developed on one platform only and would work well on the others with high certainty. 

Just if you are going to work on something please drop a comment on the issue when you start so other folks don’t. And have fun!

fonkle:
I have tested the nightly build with these code changes in Bitwig and Carla, and it still has the -minor- stability issue:

If I repeatedly modify the Mouse Behavior setting and adjust some sliders in turn while playing some MIDI notes, Surge seems to crash on choosing any of the new menu items.

I have tried to catch one of these crashes in the debugger. I can only get that to work attaching to Carla running Surge-Debug.so. Sometimes it stops on line 1072 of `/vstgui.surge/vstgui/lib/cframe.cpp`, and at other times the debugger does not pause at all yet the plugin gets disabled (red icon) in Carla.

I cannot determine the cause of this instability, but it seems not directly related to the committed changes. Possibly it's related to vstgui?

Although the instability is very small since most people likely set the Mouse Behavior just once per system, I think it should be fixed so Surge becomes rock solid. The improvement in usability justifies the release of the changes imo, but I thought you should know.

You mentioned creating a _stack trace_ previously, can you give me a hint how to do that?

baconpaul:
On let’s move this over to a fresh issue then rather than discuss it here. If you can crash it in Carla single you can get a stack by unlimitingncore or running Carla in gdb- not sure how in bitwig

PR #832, opened 2019-04-12
sweetspot-819: Fixed slider overshoot

fonkle:
This commit fixes the 'overshoot'-issue with the sliders.
CSurgeSlider::bounceValue() now remains a restvalue at the end of the value range, making the sliders more stable.

Classic mode hasn't been altered to keep the Windows version intact.

Fixed reset on 2nd slider action:

When ending a slider action outside of its range, its restvalue and restmodval were not reset.
Starting a 2nd slider action would use these -now obsolete- values, resulting in
unexpected behavior. This is now fixed; restvalue and restmodval are now reset (to 0.0f) in CSurgeSlider::onMouseDown().

Update: initialized restvalue and restmodval in CSurgeSlider constructor.

### DIFF
--- a/src/common/gui/CSurgeSlider.cpp
+++ b/src/common/gui/CSurgeSlider.cpp
@@ -48,6 +48,9 @@ CSurgeSlider::CSurgeSlider(
    has_modulation = false;
    has_modulation_current = false;
 
+   restvalue = 0.0f;
+   restmodval = 0.0f;
+
    CRect size;
 
    if (style & CSlider::kHorizontal)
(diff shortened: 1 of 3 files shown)


src/common/gui/CSurgeSlider.cpp
```diff
@@ -286,17 +289,36 @@ void CSurgeSlider::draw(CDrawContext* dc)
    setDirty(false);
 }
 
-void CSurgeSlider::bounceValue()
+void CSurgeSlider::bounceValue(const bool keeprest)
 {
-   if (value > vmax)
+    if (keeprest) {
+        if (restvalue) {
```
baconpaul: Oh I’m sorry to ask for one last change but: we really shouldn’t use floats as Boolean implicitly. Do you mind changing this to “if (restvalue != 0.0f)” and the same below and doing a force push? I had misread it before as a bool and it would be much clearer. (And since you are there, this looks 4 tabbed; did you run clang-format vs master? No big deal there tho).

Otherwise all LGTM! If you force push that change I’m happy to merge it.

baconpaul:
Looks great. I just had one comment which I didn’t pick up when I looked at the branch before sorry. If you don’t mind that minor change then I can sweep it this afternoon after you force push and the builds finish.

Thanks so much!

fonkle:
Stop the presses!

I had my doubts about that, but I let it go (shouldn't have). I'll fix it, force-push it and send anothe PR. Stay tuned!

baconpaul:
OK all looks great now! Once the automated builds are done I will merge it in. Thanks so much!

PR #833, opened 2019-04-12
fonkle/sweetspot-819: Fixed slider overshoot

fonkle:
This commit fixes the 'overshoot'-issue with the sliders.
CSurgeSlider::bounceValue() now remains a restvalue at the end of the value range, making the sliders more stable.

Classic mode hasn't been altered to keep the Windows version intact.

Fixed reset on 2nd slider action:

When ending a slider action outside of its range, its restvalue and restmodval were not reset.
Starting a 2nd slider action would use these -now obsolete- values, resulting in
unexpected behavior. This is now fixed; restvalue and restmodval are now reset (to 0.0f) in CSurgeSlider::onMouseDown().

Update: initialized restvalue and restmodval in CSurgeSlider constructor.
Update: changed if (float) into if (float!=0.0f).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -70,6 +70,7 @@ src/linux/ScalablePiggy*
 
 # Qt Creator
 *.txt.user
+*.txt.user.*
 
 # cmake
 build/
--- a/src/common/gui/CSurgeSlider.cpp
+++ b/src/common/gui/CSurgeSlider.cpp
@@ -48,6 +48,9 @@ CSurgeSlider::CSurgeSlider(
    has_modulation = false;
    has_modulation_current = false;
 
+   restvalue = 0.0f;
+   restmodval = 0.0f;
+
    CRect size;
 
    if (style & CSlider::kHorizontal)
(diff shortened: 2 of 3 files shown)


fonkle:
There ya go. Did that work? I closed the PR then made a new one...?

baconpaul:
Yup the old one would have worked too! Just the ci builds take a few minutes. Merging now

fonkle:
Does a PR include all commits since the PR was made or just the ones that are force-pushed upstream?

baconpaul:
The PR contains only the commits that were force-pushed.

Also i generally merge with a “squash” which just puts one commit into the master branch. You can see the single commit into master here https://github.com/surge-synthesizer/surge/commit/9bb612edf19feb6dec7040e2490e5c98c6bb21bf

(You can also use github where there’s N commits; each of the PR commits and a merge commit by the merger; that would mean we didn’t both appear on the commit here. I prefer to not do that because it means I can do `git cherry -v upstream/release/1.6.0-beta-8` and see all the things that changed one per line)

But basically: Think of git as a transaction manager. The transactions you put into the PR are the ones we merge or compress and merge

itsmedavep:
Man. This. Is. GREAT. Thank you for the improvement. 

@fonkle I really appreciate your efforts.

Issue #562, opened 2019-02-10
Make Linux UserInteractions use something other than fork or popen zenity


jarkk0sakkinen:

jsakkine:
I guess I'll start working on this now.

baconpaul:
Great!

We will almost definitely want to take the Surge::UserInteractions::promptError about unimplemented out of src/linux/DisplayInfoLinux when you do since that will over-pop error messages. We can just replace them with a silent default and a FIXME for now I think.

baconpaul:
Over in #749 @tank-trax and @jsakkine propose a solution to this where we fork either Zenity or kdialog for now. I think both issues can close at the same time.

baconpaul:
(@tavasti not @tank-trax sorry!)

baconpaul:
We had a long discussion over in #749 which amounted to "we should do this right and not fork a sub process (mostly)". I made a comment about "do this like the about box" but that made me realize there's a super quick path here. Which is actually USE the about box.

So in 

1. CAboutBox.h add an enum "AboutBoxMode" which has values "About", "Error", "OKCancel" and add a method "setAboutBoxMode(enum)"
2. in SurgeGUIEditor::showSettingsMenu where you do if(about box) aboutbox->show make it instead
```
  if(about box)
   {
     about box->setMode(CAboutBox::About)
    about box->show()
   }
```

(in pseudo code)

3. In UserInteractions promptError make "promptError" reach back to the GUIEditor and set the about box to error and show it. This may mean weaving an optional GUIEditor through the Surge::UserInteractions code paths, which is a bummer, since we don't always have an editor available (so there will be some errors Surge pops which still go to stderr if an editor hasn't started; and on mac and linux you will see them; the zenity solution would avoid this).

4. Modify CAboutBox::Draw to draw the right thing for prompt error

5. Repeat for OK Cancel just with the addition of a callback.

jsakkine:
@baconpaul, you are speaking of Zenity as it would be some kind of solution to something. It seriously isn't. For me it will break my whole development environment. I think I quite clearly showed the reasons why it should be avoided...

jsakkine:
Anyway, I've spoken.  Not going to comment about this as I've presented hard facts. Better to do some code now...

baconpaul:
Yeah @jsakkine you just aren't reading my ifs correctly is all! I was never suggesting requiring zenity. Just using it if we had it and solving the immediate problem for almost all of our users. 

[...]

PR #839, opened 2019-04-14
Implement FileOpen dialog in UserInteractions

### LINKED ISSUE #829: Add OpenFileDialog to UserInteractions
If UserInteractions had

```
void OpenFileDialog( std::string initialDirectory, std::string suffix, 
                 std::function<void(std::string)> callBackOnSelection);
```

which called you back when you selected, we could do things like open .scl files (#828) and also not worry so much about wt drag n drop (as I was discussing on KVR).

Mac and Windows is easy. Linux is hard. As always.

baconpaul:
There are increasingly cases where we want to prompt users for a file.
Rather than write this in VSTGUI I'm using the OS specific APIs behind
a UserInteraction common point. As of this writing Mac and Windows work but
ignore the options for starting directory and filter; and Linux returns a
UserError.

Since this move required me to move to cocoa for Mac UserInteractions,
also set up the .clang-format file to work properly with objective c

Closes #829 (but will require us to open another issue for linux and the
undone parts of the API). Pushing this now so I can use it for scale file
opening in the tuning issue.


baconpaul:
@jsakkine note the addition to the API here that you are working on in #562

PR #880, opened 2019-05-19
Linux/package/VST3/813: Fix package-vst3.sh

fonkle:
The 'package-vst3.sh' script tried to copy 'target/vst3/Release/Surge.so' disregarding which version (debug_x64 or release_x64) was being built.

It now copies either 'target/vst3/Release/Surge.so' or 'target/vst3/Debug/Surge-Debug.so' depending on $config.


baconpaul:
Thanks! Merged

Issue #828, opened 2019-04-10
Feature Request: .scl / .tun Retuning


bit-101010:
I would like to see alternate tuning options for oscillators, and perhaps everytihng that does key-tracking.

The most common standards for retuning are the [AnaMark (.tun)](https://www.mark-henning.de/files/am/Tuning_File_V2_Doc.pdf) and [Scala (.scl / .kbm)](http://huygens-fokker.org/scala/scl_format.html) formats. loading these should change the frequencies of each MIDI note.

There are some open tools for doing this already:
There is some C++ source code [provided by Mark Henning](https://www.mark-henning.de/am_downloads_eng.php#AnaMark_Misc) (creator of the AnaMark standard) for using both formats of tuning files, but it handles .scl files incorrectly.
ZynAddSubFx handles .scl files as expected, but doesn't load TUN files. [Microtonal.h](https://github.com/zynaddsubfx/zynaddsubfx/blob/4a740f070762a14ea46ce4ee85f2c90880603892/src/Misc/Microtonal.h)/.cpp handles this function.

In addition (this is a stretch goal, and maybe beyond the scope of this initial request) being able to load and blend between multiple scales would open up a world of harmonic possibilities.

baconpaul:
Oh that’s a really cool idea. The key to frequency code is fairly well isolated so it wouldn’t even be that tough

Question : surge modulates pitch by semitones like with an lfo or whatever. Wonder if other synths bind their modulation to the scale or use base note in scale space and frequency modulation in even temper space. 

Obviously base note in scale space and frequency modulation in even temper space is way way easier!

bit-101010:
The synths I've worked with keep modulations by semitones.
Even pitch bend stays in semitones for some synths. It really comes down to implementation, which varies greatly for this stuff. 
The main exception I've seen to the pitch bend staying in semitones is AnaMark, which gives you the _option_ of using the pitch bend wheel to morph between two scales.

baconpaul:
One other question: do you think we should have tuning per osc, per scene, or per patch? I can make all the arguments but interested how other synths do this.

bit-101010:
I'd go by scene or patch; loading somethings for each oscillator seems tedious.

baconpaul:
So OK here's some notes on this issue

[...]

PR #1032, opened 2019-08-16
Support DAW state in the patch

### LINKED ISSUE #890: MPE Status stored in patch
currently if MPE is enabled and saved in a project Surge will open with MPE disabled

would it be possible to have a setting in SurgeUserDefaults.xml

similar to

`<default key="mpeEnabled" value="1" or "0">` 

so that if MPE is set to enabled in a project it will open with the MPE feature enabled

### LINKED ISSUE #914: DAW does not save zoom state between sessions
We don't want to store zoom in the .fxp but we do want to store it in the DAW streaming state.  It never has because I set the goal of "change nothing with patch streaming until first stable release". Now we have 1.6.1.1 we can move ahead.

Per #912 tagging @mireiner

### LINKED ISSUE #915: A Host of things are not stored in patch master issue for 1.6.2
Some things should be stored in patch; some things should be stored in DAW state but not patch. This issue is a master issue for those as we change patch format

Stored in .fxp
* tuning file (#828)
* output volume (#475)

Stored in daw-only patch
* zoom state (see #912 comment and #914)
* MPE state (#890)
* snap state

baconpaul:
Some features of the synth - notably, zoom, MPE Enablement, and
Tuning - are features of the DAW environment you are working in and
were not persisted. This fixes that by adding a dawExtraState section
to the streaming protocol which is only populated and read at DAW time
not at general patch time.

Closes #890
Closes #914
CLoses #915
Mostly wraps up #828

### DIFF
--- a/src/au/aulayer.cpp
+++ b/src/au/aulayer.cpp
@@ -525,6 +525,10 @@ ComponentResult aulayer::RestoreState(CFPropertyListRef plist)
       p = CFDataGetBytePtr(data);
       size_t psize = CFDataGetLength(data);
       plugin_instance->loadRaw(p, psize, false);
+
+      plugin_instance->loadFromDawExtraState();
+      if( editor_instance )
+          editor_instance->loadFromDAWExtraState(plugin_instance);
    }
    return noErr;
 }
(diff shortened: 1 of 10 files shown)


gnips:
I don't think zoom should be part of the patch. It's part of the current UI.
If I load your patch in Cakewalk, how do I restore my zoom?
If you use 4k, and I'm on 1280*1024, how can I use this patch?

baconpaul:
It's not part of the patch; don't worry. For exactly that reason.

The zoom is part of the DAW state. So if you manage to set your zoom the DAW will persist it and restore it. But if you store the patch the contents won't be read. (And as of an upcoming push won't even be written).

VSTi use patches for two things. Storing presets and streaming to the DAW when you save your workspace. I segregated those behaviors more cleanly in this commit so we can choose which to persist when.

gnips:
Sorry, missunderstanding on my side, good to hear.

baconpaul:
No worries - thanks for the comment!

Issue #1063, opened 2019-08-20
Unison phase randomization sounds less effective than before


VincyZed:
**Bug Description**
The behavior of the phase randomization regarding Unison has changed and sounds weird in the latest nightly.

* Surge Version: Latest Nightly, VST3, 64 bits

**To Reproduce**
1. Open Surge (keep it on stock saw).
2. Set Uni Count to 16.
3. Set Uni Spread to 10.0 (those are just settings that I used to realize the difference).
4. Play C4 on the two version side by side.

**Expected behavior**
The attack should sound lush and less abrupt.

Issue #1060, opened 2019-08-20
Copying and pasting wavetable oscillators sometimes does not work properly


VincyZed:
**Bug Description**
When pasting certain wavetables in an oscillator after copying them from another oscillator, the result is not the same as the first one.

* Surge Version: Latest Nightly
* Plugin type: VST3
* 64 bits

**To Reproduce**

1. Open an instance of Surge
2. Set the type of the first OSC to Wavetable
3. For example, select the wavetable "bells" under the "rhythmic" category
4. Morph through the wavetable to see expected result
5. Copy the first oscillator (right-click on the "1" button at the top of the display)
6. Paste it on OSC 2 (right-click on the "2" button at the top of the display)
7. Go on the second OSC and morph through the wavetable. It is not the same content as the first OSC.

**Expected behavior**
It should behave like the first OSC we copied the data from.

**Screenshots**

OSC 1 (the one we copied from)

![image](https://user-images.githubusercontent.com/50145178/63315999-561edb80-c2db-11e9-80fe-95ec9e2d6135.png)

OSC 2 (the one we pasted data from 1)

![image](https://user-images.githubusercontent.com/50145178/63316008-5dde8000-c2db-11e9-87bd-451d28300202.png)

**Desktop config:**
 - OS: Windows 10 x64
 - Host FL Studio 20.5.1

baconpaul:
I've confirmed this is a bug in 1.6.1.1 as well; which is good. It means my .wav file changes didn't introduce it. But also is bad. It means I have more code to look at to figure it out :)

Thanks for the accurate bug report.

baconpaul:
Oh geez - it's super clear in the code

SurgeStorage::clipboard_paste in the cp_osc block just ignores the wavetable on the clipboard. Should be an easy fix.

baconpaul:
OK this bug was introduced between 1.6.0 beta-9 and 1.6.1.1 

Looking at the "copy" code I think it is that the "copy" code doesn't deal with dynamic wavetables properly.

VincyZed:
Confirmed, it works now! 

The only thing is that the name of the selected wavetable from the copied OSC does not appear after pasting, instead it says the generic "(Patch Wavetable)" name.

baconpaul:
Yeah that’s right; cloning makes an in memory copy and loses affiliation with the on-disk image. Just like if you save your state in the DAW and then re-open it. 

[...]

PR #1064, opened 2019-08-20
Unison Timing in SuperSaw division

baconpaul:
When fixing the unison for tuning, I divided where I should have
multiplied. This restores the untuned unison to the 1.6.1.1 behavior
and still works coherently with tuning.

All parth of #1060


baconpaul:
This will close #1063

PR #1092, opened 2019-09-02
Fix mismatched new[] and delete

jpcima:
Hi, I found this programming error detected by Valgrind memory checker.


baconpaul:
Oh thanks! I only ever had the patience to run valgrnd on the headless. Appreciate this fix. 

If you want to look at the Authors file and drop a note here about how you'd like to be credited, or add yourself alphabetically and add another pull request, that would be great.

Merging this change which is yeah obviously correct.

Issue #1105, opened 2019-09-04
Overview of bugs in VST/VST3/LV2 on Linux


mortfell:
Just pooling a few issues together since they are all done in the same setup.
I have not used Linux for a while but have used it a lot in the past, I'm fairly well versed in the way Linux does audio.

If it seems like these issues could be coming from my end, or if anyone looking into these bugs needs more info about my particular machine, please let me know if there is a particular command or anything that would give helpful info.

I'm testing Ardour, REAPER, Renoise and Carla
with LV2 VST VST3 where possible.

Overall I'm really surprised with the how functional it is so far!!! especially the LV2 one. Great work to everyone. In terms of it's actual audio functionality I haven't noticed a single bug yet!
There is a lot of bugs with the GUI in both versions though. 

VST2 and VST3 actually both work fine for me in REAPER from an audio perspective, but they both have the same graphical issues. All of my bugs below that mention VST are the same with VST3.

I've made a few bug reports for the biggest issues I've on VST and LV2 faced:
#1100 #1101 #1102 #1103 #1104 

The **biggest** issue is #1101 the Preset menu is pretty much non functional in both the LV2 and VST2/VST3 versions for me.

baconpaul:
Thank you! I tagged them all “linux” and added a few comments. Very useful!

mortfell:
Cool, so lots of improvements already:
- ~#1100~ was fixed by using a non "demo" version of Ardour (though it's unclear why that would have any bearing on things) I closed it because it wasn't an issue anywhere else as far as I could see.
- ~#1102~ was totally fixed by the update today (thank you @jpcima !!) closed it.
- #1104 was fixed (for me) by not using a desktop environment that relies on "openbox" window manager. This should still be considered a bug I think. So far it's unknown if this is a total incompatibility with that window manager or to do with something else in my system.

These next two are the big ones, if I wasn't experiencing these, Surge would feel as stable and quick on Linux as the other platforms, I'm happy to give any info at all that could resolve these:
- #1103 I am still experiencing this issue with VST and VST3, I'm interested if anyone else is also. Also sliders that adjust any graphics in the interface are VERY laggy on  VST2(and 3) and LV2 (for instance morph, LFO rate, anything like that)
- #1101 presets menu is almost unusable which is too bad.

jpcima:
Another small remark I will add: the general menu at bottom right, it tries to open itself on the right side instead of left, which is, while still usable, pretty unpractical.

[...]

PR #1106, opened 2019-09-04
Improving LV2

jpcima:
Hello, I add two things.

1. it's a mistake about taking the wrong variable in a function call, which is able to crash the host if reached.
2. this is window resizing of zoom.

About zoom, it's fine with Carla. With Jalv, the window accepts to make itself larger, but not to shrink back. I'm guessing the problem is probably Jalv's regarding this matter.

### DIFF
--- a/src/lv2/SurgeLv2Ui.cpp
+++ b/src/lv2/SurgeLv2Ui.cpp
@@ -5,6 +5,8 @@
 #include "SurgeGUIEditor.h"
 #include <cassert>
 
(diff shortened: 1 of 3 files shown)


baconpaul:
Awesome. Merging now. Did you also see #1105 which @mortfell put together?

jpcima:
Yes, I have seen it and subscribed the thread.
I was now building the Reaper configuration to attempt reproducing the problem.

PR #1108, opened 2019-09-04
Make the pthread routine return null

jpcima:
Since I've seen such things be at the origin of crashes before, it's better to be safe.


baconpaul:
Thanks!

Issue #1102, opened 2019-09-04
Can not select custom user folder on LV2 or VST in Linux


mortfell:
**Describe the bug**
The "zenity" dialogue which opens for selecting a custom user preset folder from the menu (if it opens at all) doesn't actually allow you to select a folder. If you press "ok" it just steps to the next directory level until it can't anymore.
![surgezenitybug](https://user-images.githubusercontent.com/4732794/64228283-efd5b380-ceab-11e9-8432-d72883676593.png)

This is on VST2/VST3/LV2 and all hosts. (REAPER, Ardour, Carla, Renoise)

SYSTEM INFO
The latest nightly, first version with LV2
I'm on the latest lubuntu 19.04 a lightweight Ubuntu distro.
UNAME printout:
5.0.0-13-generic #14-Ubuntu SMP Mon Apr 15 14:59:14 UTC 2019 x86_64 GNU/Linux
Latest stable Ardour, REAPER, RENOISE and Carla

Let me know if any other info would be helpful!

baconpaul:
zenity command line is constructed in src/linux/UserInteractionsLinux.cpp if that helps folks

mortfell:
This works for me now!
thanks

mortfell:
this worked for me yesterday! but something broke the mouse support since then

jpcima:
Can you explain in detail what is "broken mouse support in zenity" ?
Zenity is external software, it handles mouse in full independent way of what surge does, none of our code has to do with it.

baconpaul:
@tank-trax reported the same. The mouse can't click on any of the zenity windows for some reason. 

Oh and I bet I know what it is

Our menu change probably isn't releasing mouse focus properly.

And zenity is only launched from menus.

baconpaul:
I will hae to debug that for you. It's too tricky to debug if you don't know vstgui. I'll open a separate issue and look this weekend.

baconpaul:
Moved it over to #1124

mortfell:
ahh @jpcima  sorry for not being more specific.
The mouse just doesn't do anything at all in the zenity interface which Surge opens

baconpaul:
i have a fix in flight @mortfell

mortfell:
I will close this one since technically you can select it with the keyboard, and the other issue more accurately describes the problem

PR #1110, opened 2019-09-04
Add support of various file dialog options on Zenity

jpcima:
This adds better zenity file dialog support. #1102

This is supported:
- initial directory
- selecting directories
- file filters

Also, there is a quoting routine for passing safe arguments to popen.

### DIFF
--- a/src/linux/UserInteractionsLinux.cpp
+++ b/src/linux/UserInteractionsLinux.cpp
@@ -21,6 +21,33 @@ namespace Surge
 namespace UserInteractions
 {
 
+// @jpcima: string quoting for shell commands to use with popen.
+static std::string escapeForPosixShell(const std::string &str)
+{
+   std::string esc;
+   esc.reserve(2 * str.size());
+
+   // 1. if it's a newline, write it verbatim enclosed in single-quotes;
+   // 2. if it's a letter or digit, write it verbatim;
+   // 3. otherwise, prefix it with a backward slash.
+
+   for (char c : str)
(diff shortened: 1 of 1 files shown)


baconpaul:
Looks fantastic; thanks.

I would say the entire zenity implementation is something I did quickly because it was becoming untenable to have nothing. Happy for any other changes, corrections, or thoughts you have.

Will merge now.

baconpaul:
Also: It's wonderful to have another dev working on this stuff. Really glad for the efforts!

jpcima:
I noted 2 other defects, related:

- it invokes the callback even without selection. If the user clicks cancel, Linux user interations are invoking it with empty string.
One could imagine it may be interesting to act in response to a "cancel" click.
However it doesn't seem how Surge works: when you ask to open a tuning, and cancel, it will warn that it's not a valid SCL file.

- About SCL: open an invalid file, or non existing. If the tuning variable "count" is equal to 0, it attemps a divide by it, and raises FPE which crashes.

baconpaul:
On behaviors: The mac is really the definitive example. Since that's the one I use almost all the time. If you look at UserInteractiosnMac.mm you can see it only calls the callback if you click OK. Clearly linux should do the same.

On the second: non-existing gets fixed by above. Invalid is much harder. There's no great way to detect if an sci file is valid. But an FPE crash is dumb - we should clearly be more defensive in the parser. That's all in src/common/Tunings.cpp if you want to look!

PR #1111, opened 2019-09-04
Linux file dialog

jpcima:
Solving the few items noted in the previous comment.

https://github.com/surge-synthesizer/surge/pull/1110#issuecomment-528020293

### DIFF
--- a/src/common/gui/SurgeGUIEditor.cpp
+++ b/src/common/gui/SurgeGUIEditor.cpp
@@ -3066,6 +3066,13 @@ VSTGUI::COptionMenu *SurgeGUIEditor::makeTuningMenu(VSTGUI::CRect &menuRect)
                                 }
                             }
                             auto sc = Surge::Storage::readSCLFile(sf);
+
+                            if (sc.count <= 0) // invalid
+                            {
+                               Surge::UserInteractions::promptError( "This .scl file is not valid", "File format error" );
+                               return;
+                            }
+
                             this->synth->storage.retuneToScale(sc);
                         };
                         Surge::UserInteractions::promptFileOpenDialog(this->synth->storage.userDataPath,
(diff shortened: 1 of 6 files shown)


src/common/gui/SurgeGUIEditor.cpp
```diff
@@ -3066,6 +3066,13 @@ VSTGUI::COptionMenu *SurgeGUIEditor::makeTuningMenu(VSTGUI::CRect &menuRect)
                                 }
                             }
                             auto sc = Surge::Storage::readSCLFile(sf);
+
+                            if (sc.count <= 0) // invalid
```
baconpaul: This is a smart check!

But  SurgeStorage::retuneToScale is called several places. I would rather put this check inside the retuneToScale and add a Scale.isValid() method which for now just checks count > 0 but later could check more.

Do you mind doing that change? I'm happy to add it to my list if you aren't. But these are all pretty clean classes in src/common/Tuning

baconpaul: Oh and retuneToScale should just start with if( ! scale.isValid() ) { UserInteractions... return } and leave the state unchanged. Thanks!

jpcima: Ok I'll do this

baconpaul: Thank you very much!

baconpaul:
Looks fantastic

I'm in the middle of messing with the CI/Nightly pipeline right now so I want to finish that before I merge your change (which will kick off that pipeline - which is not what I want!). But I will merge it today for sure.

Thank you very much.

baconpaul:
Great. My pipeline changes are done so I merged this. Thanks again!

PR #1114, opened 2019-09-05
Do not regenerate source files which are identical

jpcima:
This doesn't rewrite piggy files when they don't have any change.
It skips recompiling `CScalableBitmap` every single build.
It makes the recompiling cycle a tiny bit faster.

### DIFF
--- a/scripts/linux/emit-vector-piggy.py
+++ b/scripts/linux/emit-vector-piggy.py
@@ -12,11 +12,18 @@
 import os
 import re
 import sys
+try:
+    from io import StringIO
+except ImportError:
+    from StringIO import StringIO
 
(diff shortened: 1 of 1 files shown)


baconpaul:
Oooh I’ve wanted to fix this forever

But did you see it broke in ci? Need to import the filenotfounderror I think

jpcima:
Seems FileNotFoundError is a builtin, but only available since some version of Python 2.
I have tried on my machine Python2 and 3.

baconpaul:
Yeah we build on an older version of Ubuntu (16.04 LTS) in the pipeline so sometimes these edge cases pop up. Fix looks great though and worked! So I'll merge.

Thanks

PR #1115, opened 2019-09-05
Add xdg-open and zenity commands as Debian dependencies

jpcima:
These packages have been reported to me as missing dependencies, in the Debian binary package.
It's the commands called by Linux User Interactions.
- https://packages.debian.org/stretch/xdg-utils
- https://packages.debian.org/stretch/zenity


baconpaul:
Thanks!

PR #1128, opened 2019-09-06
Implement OK/Cancel prompt for Linux

jpcima:
This is the OK/Cancel box.
The result is indicated by command's exit code: 0 for OK, 1 for cancel or other failures.

Regarding passing some window ID from the editor, it has more complication that I first thought, so I keep this PR to this for the moment.

(a remark: I was reported that, when building the software as LV2, the RDF dumper which is invoked by the python build script will popup a dialog if Surge is not set up in /usr/share. In case of a non-existing editor, I would be happy to bypass this)

### DIFF
--- a/src/linux/UserInteractionsLinux.cpp
+++ b/src/linux/UserInteractionsLinux.cpp
@@ -2,7 +2,9 @@
 #include <iostream>
 #include <iomanip>
 #include <sys/types.h>
+#include <sys/wait.h>
 #include <unistd.h>
+#include <string.h>
 #include <thread>
 
 /*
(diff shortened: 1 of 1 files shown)


baconpaul:
That /usr/share popup needs to happen even if the editor isn’t open since the mis-installed synth won’t start in a DAW without those files.

The correct solution for the LV2 - if you want it to run without the configuration installed - is to have your routine thread a global through that suppresses that message.

I am surprised you get that message at all on linux though; the configuration is built into the DLL.

But please don’t remove that popup for configuration.xml missing; rather suppress it from your call point if your callpoint is a non-standard synth entry point.

baconpaul:
Oh and this change looks great! Thanks. Merging.

jpcima:
I quote it to you from the linux-audio-dev IRC (#lad)

> (00:24:15) it showed me a popup during build!
> (00:24:18) "Surge was unable to load wavetables from '/usr/share/Surge/'. The directory contains no wavetables. Please reinstall using the Surge installer."
> (00:24:43) but it looks like it created the lv2 plugin

Ok I won't suppress; what I will do instead: ttl generator will pass an environment variable, and I'll check how to make it behave correct when detecting this variable.

baconpaul:
Yeah that's exactly right.

So how I would do it is

1. Add a static to SurgeStorage called loadWaveTables default to true
2. In your routine set SurgeStorage::loadWaveTables to false
3. at line 290 of SurgeStorage.cpp inside the TARGET_RACK do a if( loadWaveTables ) around both the patches and wavetables

That will suppress this message properly. And you don't need the wavetables to get your configuration for parameters

Thanks!

jpcima:
I cannot reproduce it unfortunaley.
I wonder if this user had a particular pre-existing configuration which caused this problem to occur.

baconpaul:
Yup if you have never installed surge you will see that; if you have a working install you won't. 

That makes it something we miss as developers since we usually have working installs :)

PR #1129, opened 2019-09-06
When exporting LV2 manifest, avoid error dialog situations

jpcima:
As discussed in the previous PR.
Don't load the WT, nor patches. Both are sources of error messages.

### DIFF
--- a/src/common/SurgeStorage.cpp
+++ b/src/common/SurgeStorage.cpp
@@ -287,8 +287,16 @@ SurgeStorage::SurgeStorage(std::string suppliedDataPath)
    load_midi_controllers();
 
 #if !TARGET_RACK   
-   refresh_wtlist();
-   refresh_patchlist();
+   bool loadWtAndPatch = true;
+#if TARGET_LV2
+   // skip loading during export, it pops up an irrelevant error dialog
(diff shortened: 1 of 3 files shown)


baconpaul:
Hi!

I’ll merge this if you really want, but can I offer one suggestion? The getenv in the SurgeStorage means that a user running the synth in a daw could set that environment variable and suppress their content loads which would be sub-optimal.

I was kind of hoping you could put the getenv in your export function in the LV2 to set a static variable in Storage; and then read that static variable. That way there is no way that accidentally having the variable set could lead to an improper load of the synth in normal course of action.

So proposed change is

1: Python is fine
2: Add a static var to SurgeStorage which is true 
3: In your LV2 custom exporter function read the environment and set the variable (or heck, don’t even read the environment; just set the variable)
4: Modify Surge Storage from getenv to if (static variable)

See what I mean?

Happy to chat if you disagree. But the idea of giving an option to not load in the normal synth for LV2 only seems like it could be an edge case which could lead to a user support question.

Thanks so much!

jpcima:
Ok I understand perfectly what you mean, and I will change it.

baconpaul:
Thank you very much!

baconpaul:
Great thank you. Merging.

PR #1130, opened 2019-09-07
Move Version Numbers to 1.6.2

### LINKED ISSUE #1116: Versions for 162
1. As @gnips reports there’s a version in the windows rc 
~2. Wanna change info.plist so parent display name is surge synth team (but keep Id at vm a) to match the effect bank~

baconpaul:
Ahead of the 1.6.2 release, move the DLL and Plugin version
nubmers from 1.6.0 (since we had never moved these before)
to 1.6.2. Also add a doc on where they all live.

Closes #1116

### DIFF
--- a/doc/versions.md
+++ b/doc/versions.md
@@ -1,0 +1,13 @@
+# Version Numbers
+
+This document is a reminder list of all the places you need to update when you have a version 
+number tick.
+
+So version numbers live in
+
+* src/common/surge_auversion.h
+* resource/osx-*/Info.plist
+* src/windows/surge.rc (in several places in for 1,6,0,0)
+* src/common/version.h
+* src/common/version.h also contains a copyright year as does the AU info.plist
+* src/lv2/SurgeLv2Export.cpp generates minor and micro version
--- a/resources/osx-au/Info.plist
+++ b/resources/osx-au/Info.plist
@@ -40,7 +40,7 @@
 	<key>CFBundleSignature</key>
 	<string>VmbA</string>
 	<key>CFBundleVersion</key>
-	<string>1.6.0</string>
+	<string>1.6.2</string>
 	<key>CSResourcesFileMapped</key>
 	<true/>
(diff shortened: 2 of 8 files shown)


jpcima:
There is actually more here.

https://github.com/surge-synthesizer/surge/blob/1cd009b58dd47819ccc78da10d406672717fdefc/src/lv2/SurgeLv2Export.cpp#L150-L152

I'll quote a relevant part of LV2 spec.

> The minor version of an LV2 Resource. This, along with lv2:microVersion, is used to distinguish between different versions of the same resource, e.g. to load only the bundle with the most recent version of a plugin. An LV2 version has a minor and micro number with the usual semantics:
> 
> The minor version MUST be incremented when backwards (but not forwards) compatible additions are made, e.g. the addition of a port to a plugin.
> The micro version is incremented for changes which do not affect compatibility at all, e.g. bug fixes or documentation updates.
> Note there is deliberately no major version; all versions with the same URI are compatible by definition. Replacing a resource with a newer version of that resource MUST NOT break anything. If a change violates this rule, then the URI of the resource (which serves as the major version) MUST be changed.

- http://lv2plug.in/ns/lv2core/lv2core.html#microVersion
- http://lv2plug.in/ns/lv2core/lv2core.html#minorVersion

baconpaul:
Oh yeah I saw that but also saw the TODO so left it alone and figured I would ask you. 

Shall we make that a "2" and a "6" respectively? I haven't swept this yet so happy to merge that in.

jpcima:
Yes, that would make sense.
Also, better to reorder minor as first, and micro as second, to save some potential confusion.

baconpaul:
OK I'll add that to this commit and push it for you to look at. Hold on.

baconpaul:
OK just force pushed. It build and loads in Carla for me no problem on linux. The change looks OK to you?

jpcima:
Looks fine, runs, passes the lv2lint. It's good :+1:

Issue #1131, opened 2019-09-07
Velocity 127 registers as 0 with VST3


mortfell:
I noticed there are no bugs related to Studio One anywere, and I also wasn't sure if anyone had tested Surge as an effect in Studio one (which I'm interested in for the wiki). Posted on the Studio One sub-reddit a few days ago, to see if anyone had tested Surge and got a useful bug report that it seems like no one has noticed yet!

from user @noisedroid on Reddit:
"Seems like there's a problem with MIDI notes with maximum velocity? With many of the presets a note at max velocity doesn't sound at all.

Or if I start with an init patch and route velocity to filter cutoff, the filter opens up as expected with increasing velocity but then is fully closed when velocity is 127.

It's as if 1 is being added to the 7-bit input velocity, but it's being processed internally as a byte with a 01111111 bit-mask, so at maximum velocity the high bit is simply discarded."

Upon further prompting about platform:
"VST3 on Studio One 4 / Windows 10"

I can confirm this bug exists on both Linux and MacOS VST3, in REAPER.
I tested some other midi inputs (channel pressure, Mod wheel) and none seem to have that problem.

AU,VST2 and LV2 don't have the issue. 
VST3 only.

jpcima:
Oh, I'm betting it's this computation. The result of real-valued operation Is overflowing the `char` type.

```
char cVel = std::min((char)(e.noteOn.velocity * 128.0), (char)127);
```

PR #1132, opened 2019-09-07
Fix the VST3 velocity overflow

jpcima:
#1131

Handle adequately the case velocity >= 1.0, by not overflowing the value to -128.

While at it, defend against values outside of [0.0,1.0] range.

### DIFF
--- a/src/vst3/SurgeVst3Processor.cpp
+++ b/src/vst3/SurgeVst3Processor.cpp
@@ -231,7 +231,8 @@ void SurgeVst3Processor::processEvent(const Event& e)
       }
       else
       {
-         char cVel = std::min((char)(e.noteOn.velocity * 128.0), (char)127); // Why oh why is this a float in VST3?
+         // Why oh why is this a float in VST3?
+         char cVel = (char)std::min(127, std::max(0, (int)(e.noteOn.velocity * 128.0)));
          getSurge()->playNote(e.noteOn.channel, e.noteOn.pitch, cVel, e.noteOn.tuning);
       }
       break;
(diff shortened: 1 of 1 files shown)


jpcima:
As a note, it has others as well. Maybe these would also benefit to be also range-checked.
```
surgeInstance->channelController(channel, cont, (int)(value * 128));
```

There seem to be some disagreements within the SurgeVst3Processor.
The MIDI controller is scaled by 128, whereas aftertouch is by 127.

mortfell:
awesome thanks @jpcima !

baconpaul:
Awesome! Vst3 denying midi exists is one of the worst features.

baconpaul:
And I think The right fix is multiply by 127 not 128 - in theory we wouldn’t need the max then but it is good defense

jpcima:
I am about confident regarding latest code, but I don't have experience with Reaper and I don't know how to interact my usual MIDI software with it, so couldn't really test.

mortfell:
[vst3 velocity test.zip](https://github.com/surge-synthesizer/surge/files/3587202/vst3.velocity.test.zip)
If it's helpful here is a reaper session with a quick test of velocity effecting filter cutoff.
notes 1 and 4 of that midi file do not produce sound with the current nightly.
note 4 SHOULD have filter cutoff full up in note 4

baconpaul:
Yeah ok lemme give it a while tomorrow or Monday - I’m out at dinner now. The 128 looks wrong - your fix looks correct but your fix with a 127 may be even more correct

baconpaul:
Oh and I’m pretty sure the 128 was my error so appreciate you guys chasing this!

baconpaul:
OK good morning! The change at the latest push looks great. I also tried the reaper file before and after and the problem is resolved. So I'll merge this. Thank you very much.!

Issue #1139, opened 2019-09-09
LV2 Saving in a DAW


baconpaul:
With the excitement today, @jpcima, I was trying to figure out how LV2 saves its state

Notably, src/lv2 doesn't call saveRaw anywhere

Right now Surge saves its state as a binary patch which is a serialization of the parameters plus the binaries of its wavetables. That's what saveRaw/loadRaw do

The lv2 doesn't call these; and I do notice that if I save a session in carla, my settings get lost.

So somehow we need to make sure the LV2 interacts properly with this serialization code. Surge's state is more than just the parameters (the wavetables are not a parameter for instance).

How exactly is it working now?

jpcima:
You raise a good point regarding this.

Regarding how LV2 works, it's by saving two things:
- a set of the values of parameters
- a key-value dictionary, where keys are string and values are of any type

The parameters are core LV2 and needs nothing special, the extra data is by requiring the "state" extension feature.

It's no problem to add this raw saving feature, except not storing parameters in the blob.
Hosts store parameters themselves according to their preferred format.

baconpaul:
Right. So just the parameters isn’t enough - we need those wavetables.

So I think the solution is to take the output of saveRaw and put it into the kv table, and when the kv table is presented to us, take that value and put it on loadRaw

Otherwise you’ll set up a session with a wavetable, save it, and won’t get the same sound when you restore.

(The XML also contains more than the parameters in some cases; so I think saving and restoring that, even though some of it is redundant, is the way I would go. Since the synth will be in a consistent state when you stream it doesn’t really matter which one you stream on first.)

jpcima:
Agreed on the solution, I'll give it an implementation as quick as possible.

The only thing worrying me, it's I'm not sure how safe it is to access into the active editor in there.

LV2 has preference for a purely separated DSP/UI model (which we explicitly bypass by instance-access extension, to follow the model how Surge operates).
I'm not sure, even under the instance-access, I can always count on editor not being destroyed while I access it to save.

baconpaul:
The load and save methods are all outside the editor; it’s all in src/common in SurgeSynthesizer, SurgeStorage, and SurgePatch. I use the same methods in the Rack version of the code without problem.

So while I also am unhappy with how closely the editor is coupled in memory, in this case I think we will be fine.

baconpaul:
[...]

PR #1141, opened 2019-09-09
Save the LV2 raw state

jpcima:
#1139

This is raw state saving using data chunks. I tried quickly in Jalv host if it was recognized.
It needs that I test it a bit to check.

### DIFF
--- a/src/lv2/AllLv2.h
+++ b/src/lv2/AllLv2.h
@@ -13,6 +13,6 @@
 #include <lv2/time/time.h>
 #include <lv2/state/state.h>
 
-#define SURGE_PLUGIN_URI "https://github.com/surge-synthesizer/surge"
-#define SURGE_UI_URI "https://github.com/surge-synthesizer/surge#UI"
-#define SURGE_PATCH_URI "https://github.com/surge-synthesizer/surge#patchData"
+#define SURGE_PLUGIN_URI "https://surge-synthesizer.github.io/lv2/surge"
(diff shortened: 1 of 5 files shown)


jpcima:
I introduced another LV2 URI as identifier.
It's a possibility to change them prior to release, if they are not satisfying. (they are not significant for functionality, but some dislike using github in URI)

baconpaul:
On the URIs for user exposed URLs we’ve used https://surge-synthesizer.github.io/ rather than github/surge-synt/surge. Don’t know if that’s appropriate here.

Will review this tomorrow at some point. Off for the day now. Thanks for putting it together!

baconpaul:
Oh now I see about the editor. That’s right I added an optional editor state in the recent change to the streaming format.

It’s OK to skip that. We will miss a few things like retaining zoom in the session but I think it’s OK.

I reviewed the rest and it looks correct to me, inasmuch as I understand LV2. I’m happy to merge it. Just let me know what you think about the URI

jpcima:
The choice is yours regarding the URI, they don't have to necessarily be accessible, they are identifiers which are to remain durably for this major version of the synth plugin.

I did a few simple edits.
- I used the website as URL base
- I inserted an imaginary subpath /lv2/
- also I have capitalized `PatchData`. (on second examination of the spec, I think a capital letter is more appropriate for types, whereas a small letter is used for the attributes)

baconpaul:
OK all looks great to me! Merging.

PR #1153, opened 2019-09-10
Prevent overflowing VST2 program name buffers

jpcima:
This truncates the name to the 24 byte limit.
I spend the last byte for the null-character because Qtractor requires it to be present.


jpcima:
It's annoying to restrict the length as such, even though many hosts are supporting it.
Afaik unfortunately, there's no way to know how much buffer the host has.
I posted an issue at Qtractor regarding this.

baconpaul:
Ha ha ha. I remember writing that comment (apparently at the end of January) when I moved that function around for some reason.

And yeah its a bit of a bummer to restrict it, but VST2 is old and crufty for reasons like this. Great find and great fix. Sweeping.

PR #1159, opened 2019-09-11
Ensure LV2 ports values and chunk data are matching

jpcima:
As noted from a previous IRC discussion with @falkTX, there is a potential issue when restoring parameters from the raw chunk.

It's expected to have parameters identical with the parameter ports, but that is potentially not the case, due to 2 reasons. I have noted a more lengthy description in the patch's comment.

The following code is a workaround; not a code I'm particularly proud of making.
The purpose is to force parameters and chunk to have matching values, while saving the chunk.

It's simpler than to edit into SurgeSynthesizer which has a more complex I/O logic, but may serve as a workable patch until implementing a better solution.


jpcima:
unfortunately, despite this working in hosts, it's not guaranteed to have ports connected at save-time.
I will post a better solution later.

PR #1192, opened 2019-09-20
List ui:idleInterface also as lv2:requiredFeature

ventosus:
See http://lv2plug.in/ns/extensions/ui#idleInterface


baconpaul:
@jpcima @falkTX you good to merge this?
@ventosus thank you! I don’t really understand the LV2 so flagging the folks who wrote it here for review.

Issue #1199, opened 2019-09-21
crash while storing preset


spurkopf:
When storing an existing user-preset host crashes. Storing an new preset works.

* Surge Version: Surge-linux-x64-NIGHTLY-2019-09-21
* Plugin type (VST2, 3 or AU) vst2 + lv2
* Bits (32/64) 64

**Desktop (please complete the following information):**
 - OS: [e.g. windows 10] Opensuse Tumbleweed
 - Host [e.g. bitwig, logic] Traction Waveform (freezes) Qtractor (crashes)

baconpaul:
By “existing user preset” you mean you save a preset, chose store, edit it, and chose store again with the same name and category? Or you mean you edit a factory preset and save it as a new preset?

Also did you happen to get a stack trace out of either of your hosts?

Thank you

spurkopf:
Yes, the first one. no need to edit it, just save ist again and it happens.
Sorry I couldn´t  produce now useful stack-trace till now.

baconpaul:
Ok I’ll take a peek when I’m next at my laptop. Thanks!

@tank-trax do you see this behavior in Linux?

spurkopf:
Tested with Surge-linux-x64-1.6.1.1 - there it works.

baconpaul:
Ahh I have a theory. Hold on. Booting up my linux VM.

baconpaul:
OK so between 1.6.1.1 and 1.6.2 we added a code path that prompts you before you overwrite on linux and that uses zenity to prompt for ok/cancel with `zenity --question`

Can you confirm for me that you do or don't have zenity installed? 

Let me tag @jpcima also who wrote the code but I bet I can fix it too just by simulating a missing zenity if JP's not online. But I can super easily reproduce your crash by renaming `zenity` to `qenity` in the C++. Then we get a core dump right away. 

Let me tag this one for 'immediate fix' in our 1.6.2.1 release. Thank you for the report.

baconpaul:
OK I've confirmed that if you don't have zenity the execlp falls through and then you get an immediate crash it seems in the child process trying to do some UI thing (my crash backtrace is that the platform xcb handle is bad; but parent has not it seems returned from waitpid so I think it is things going very wrong when excclp can't replace the child process)

Here's a couple of solutions

1. @jpcima or @falkTX can we do anything more defensive in the vfork/exec pattern
[...]

PR #1204, opened 2019-09-21
Exit with '_exit' when zenity is missing

baconpaul:
Per comment from @falktx on #1199 this will stop our missing-zenity-core

### DIFF
--- a/src/linux/UserInteractionsLinux.cpp
+++ b/src/linux/UserInteractionsLinux.cpp
@@ -61,7 +61,7 @@ void promptError(const std::string &message, const std::string &title,
                  "--title", title.c_str(),
                  (char*)nullptr) < 0)
       {
-         exit(0);
+         _exit(0);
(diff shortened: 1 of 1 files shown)


baconpaul:
Oh this one avoids the crash but there's one more change we need. This will default to "cancel" which is not what we want when zenity isn't there.

baconpaul:
Alright just did a force push of the 'fail open' for OK/Cancel. @falkTX would you mind another peek? I think this is the right solution. Thank you ver much!

baconpaul:
Actually i don't need to burn your time look at this @falkTX - thanks for the comment but I'm sure my additional "if" statement is fine now I look at it. I'll merge and then ask original user to test the nightly.

Really appreciate that _exit tip. I would have never found it and had a rube goldberg solution with making sure zenity was in path starting to get ready for me to show you :)

PR #1225, opened 2019-09-30
Allow to build the Linux x86 configuration

jpcima:
Hi, this allows to package Surge for Linux in x86 configuration.
The current problem is, under 32bit, running premake5 does not create any targets.

cc @trebmuh
Surge x86 is buildable by `config=release_x86` on the `make` line.

### DIFF
--- a/premake5.lua
+++ b/premake5.lua
@@ -83,7 +83,7 @@ elseif (os.istarget("linux")) then
 
     buildoptions { "-std=c++17" }
 
-    platforms { "x64" }
+    platforms { "x64", "x86" }
 
     files
     {


baconpaul:
I want to make sure both the CI and Release pipelines work as expected before I merge this which I should be able to do in the next couple of days. Just need to find 20 minutes to take a look at them and be sure they run the right version and the default hasn’t changed. So gonna pause on merging until I can check that if that’s OK. Might have time tomorrow - if not will on Weds.

trebmuh:
Just letting you know that with this fix, current git master 3677dda builds fine on a Debian Buster. Note that I didn't test and then don't know yet if the soft actually works.

baconpaul:
Cool. The pipelines call build-linux which sets the config to release_x64 explicitly so no chance of accidentally choosing 32 bit there.

If you would like to, adding a -x86 option to build linux and adjusting those config lines may be nice. But also happy for you to not! gonna merge this.

tank-trax:
will there be a 32 bit version be available for release?

baconpaul:
If we want to do that we would need someone to make build-linux set up a 32 bit pipeline and have installer make a 32 bit deb. Then I can add another pipeline to the release to build it sure!

Basically right now the pipeline does 'build-linux' then 'installer_linux/make_installer' or whatever. What's the equivalent set of those things for a 32 bit release? Then it's just pipeline work.

tank-trax:
I tried to build it locally but was not able to on a multiarch system. There are certain libs like cairo that if installing the i386 version require something like 45 new packages to be installed with the all the equivalent x64 versions removed

PR #1261, opened 2019-11-01
Fixed internal link on line 163

kant:
* plus minor proposals


baconpaul:
Compressed into #1265

PR #1262, opened 2019-11-01
Minor formatting proposal on line 05

kant:


baconpaul:
Thanks very much for these!

Do you mind resubmitting them as a single pull request rather than on a file-by-file basis? (Or if you can't do that, I'll go ahead and make a single PR later so we don't end up with lots of log entries in the master commit log).

Best

baconpaul:
Compressed into #1265

Issue #14, opened 2018-09-26
Win32 build possible, but has bugs


getdunne:
I was able to build the VST2 for Win32/x86 in VS2017, but it exhibits glitches with some presets, most notably "Polysynth - FM Poly 1".

1. Use VS2017 Configuration Manager to add the Win32 platform, then build the 32-bit VST2.
2. Open 32-bit VST2 in a 32-bit VST host (I used Hermann Seib's VSTHost).
3. Select preset "Polysynth - Embrass" and play lots of notes
4. Select preset "Polysynth - FM Poly 1" and play just two notes legato - second one will glitch.

This does NOT happen with the 64-bit build. There may be an underlying bug which mainly manifests in the 32-bit version for some reason.

synthtester:
Can you share it? I'm here after following link from Vember that says binaries are available but the only Windows installer I found is x64.

getdunne:
> Can you share it? I'm here after following link from Vember that says binaries are available but the only Windows installer I found is x64.

@kurasu can you share this? If I do it, it will violate my license from Steinberg; I'm not prepared to risk that.

esaruoho:
@getdunne hold on, what? Is this because of Steinberg's VST2 license or something, or why can't a regular VST be made available for free without running into Steinberg issues? can ya somehow explain this further?

esaruoho:
@getdunne is there any chance you could document the VS2017 - settings and building further so someone could make a PR / Branch out of that and maybe update the `README.md` file so we'll eventually be able to bake VSTs. This way nobody needs to create a `.dll` with their own VST license and share it somewhere and get into trouble with steinborg.

esaruoho:
can the vs2017 configuration manager be used to build a 64-bit VST2 for Windows?

getdunne:
[surge-vst32proj.zip](https://github.com/kurasu/surge/files/2696528/surge-vst32proj.zip)
I don't have time to go through the pull-request process, but here is my .vcxproj for the VST2 build and instructions for how to use it, assuming you have access to the necessary files from the old VST2 SDK.

jsakkine:
Going through old issues for fun to find the dead leaves. Maybe you could try with a newer build? @kzantow was working on automatic Windows builds so maybe you could grab on of those. Dunno, what is the status of that effort, though.

getdunne:
[...]

PR #1264, opened 2019-11-01
Minor proposal on paragraph #14

kant:


baconpaul:
Compressed into #1265

PR #1286, opened 2019-11-13
Update README.md

adiastra:
Remove IRC

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ This README serves as the root of developer documentation for the project.
 # Developing Surge
 
 We welcome developers. Our workflow revolves around github issues in this github repo
-and conversations in our slack and IRC chatrooms. You can read our developer guidelines
+and conversations in our slack chatroom. You can read our developer guidelines
 in [our developer guide doc](doc/Developer%20Guide.md).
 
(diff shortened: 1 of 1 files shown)


trebmuh:
Why removing it?

esaruoho:
@adiastra let's not have personal preference dictate what should be in the surge readme.md or not.

we had a ton of Linux users on the surge IRC channel helping with the Linux implementation. Let's not just randomly remove IRC "just because"...?

baconpaul:
Yeah @adiastra was missing a bit of historical context; and also the channel has been dead for a while. We chatted about it some on slack and I’m not going to merge the change. The irc channel is still alive and Bridged to slack

trebmuh:
Maybe it's dead because everything runs smoothly so far... for now.
Just as a bit of contextual, it's not because I'm not talking in there right now that I'll never ever do it again. I'm working on other stuff right now. And the LV2 version seems to work quite well. But if I found something strange or if I've got a Feature Request to do, I'll probably use the IRC channel to talk about it.

baconpaul:
That’s exCtky what I explained on slack

trebmuh:
:sunglasses: then

PR #1298, opened 2019-11-15
Move dem sliders

adiastra:
moved cutoff and resonance


baconpaul:
I think this isn’t quite right; there should be two frequency and resonance sliders - one for each filter - I’ll explain over on slack.

PR #1309, opened 2019-11-17
Keep cmd open when something goes wrong

cyanit:
keep cmd open in case of an error, so in case something fails you can read that error comfortably.

### DIFF
--- a/build.cmd
+++ b/build.cmd
@@ -14,4 +14,5 @@ exit /b 0
 
 :error
 echo build failed!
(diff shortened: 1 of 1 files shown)


baconpaul:
Just as a note to self: This .cmd is not used in the azure pipelines so this is user change only and is fine. As long as the CI build finishes successfully I’ll merge it. Thanks!

baconpaul:
Squashed into #1319

PR #1312, opened 2019-11-17
moved some sliders

adiastra:


baconpaul:
I notice you moved the feedback slider but didn’t add it to the xml. Also I think the feedback is probably filters.feedback as the right name since it is feedback in the filter bank. Mind fixing those before I merge?

baconpaul:
Looks great now. Thanks. Merging.

Issue #1337, opened 2019-11-22
Surge++ streaming with n_osc != 3


baconpaul:
Old patches have wavetable offset issues with n_osc != 3. For now I put an n_osc = 3 override in a couple of key spots but we need to fix this. Places are in the associated commit.

baconpaul:
Re-tooling the S++ plan so closing these issues.

PR #1340, opened 2019-11-23
More streaming n_osc vs 3 problems

baconpaul:
I had fixed load_patch but, of course, that broke save_patch.
So here's another thing I need to clean up when I fix


baconpaul:
Attach to #1337

Issue #1316, opened 2019-11-18
Surge++ CHSwitch2 to CGlyphMulti moves


baconpaul:
So once the sliders are moved the next thing is the multi switches.  Controls like this.

<img width="47" alt="Screen Shot 2019-11-18 at 5 48 09 PM" src="https://user-images.githubusercontent.com/13387561/69100640-9bae8a00-0a2b-11ea-992c-2aa5414fd747.png">

These have a similar but slightly different recipe from the sliders.

These start the same way. Namely edit SurgePatch to change the signature of the param. So in this case

https://github.com/surge-synthesizer/surge/blob/a554049beb6bfa1bf860b1f7a3b9a69f8b8ee643/src/common/SurgePatch.cpp#L348

that line would add a type, probably "global.waveshaper_type" just like for a slider.

But now things get different. The control types for these switches are individual case statements in SurgeGuiEditor.cpp which need to be adjusted. Each type has an individual case statement in SurgeGUIEditor::openOrRecreateEditor. This needs to be removed. So find the block

https://github.com/surge-synthesizer/surge/blob/a554049beb6bfa1bf860b1f7a3b9a69f8b8ee643/src/common/gui/SurgeGUIEditor.cpp#L1093

and remove everything up to the next case statement, including the break.

Then add the type to the list of types which have an explicit layout load

https://github.com/surge-synthesizer/surge/blob/a554049beb6bfa1bf860b1f7a3b9a69f8b8ee643/src/common/gui/SurgeGUIEditor.cpp#L999

by adding the `case ct_wstype:` or equivalent to that list.

At this point you will now be using the layout manager for the component. We could use the CHSiwtch2 component again but that has problems. So for now we can just use one of the multi switches with a choice.

Here's the node statement which makes the "poly|single|etc.." switch which we could copy and paste for this one now

https://github.com/surge-synthesizer/surge/blob/a554049beb6bfa1bf860b1f7a3b9a69f8b8ee643/resources/data/layouts/original.layout/layout.xml#L375

So there you go. Good luck!

baconpaul:
Re-tooling the S++ plan so closing these issues.

PR #1356, opened 2019-12-01
Move multi switches to XML

apalm:
This PR addresses https://github.com/surge-synthesizer/surge/issues/1316.

It moves the following to XML:

* waveshaper type
* filter envelope mode
* amp envelope mode
* LFO shape
* LFO trigger mode

Screen Shot 2019-12-01 at 10 35 19

One issue is that the `Filter EG` section doesn't show the envelope mode multi switch; I'm not sure why.

Since https://github.com/surge-synthesizer/surge/issues/1316 states "once the sliders are moved", I also moved a few more sliders:

* waveshaper drive
* filter balance

However, if you prefer that the slider moves aren't part of this PR, I can back them out.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -14,6 +14,7 @@ The Emu (J Riley Hill)
 EvilDragon 
 Erik-Jan Maalderink <fonkle@gmx.com>
 Kjetil Matheussen <k.s.matheussen@notam02.no>
+Andrew Palm <andrewtpalm@gmail.com>
 Dave Palmer <itsmedavep@gmail.com>
 Adam Raisbeck <sense@i2pi.com>
 Esa Juhani Ruoho <esaruoho@gmail.com>
--- a/resources/data/layouts/original.layout/layout.xml
+++ b/resources/data/layouts/original.layout/layout.xml
@@ -688,7 +688,7 @@
 
       </layout>
       <layout height="auto" width="auto" style="roundedlabel" bgcolor="$mod.bg" fgcolor="$mod.fg" label="Modulator">
-        <node component="60x12 Multi" guiid="filter.env.mode" xoff="170" yoff="20" width="60" height="36">
+        <node component="60x12 Multi" guiid="lfo.trigmode" xoff="170" yoff="20" width="60" height="36">
           <nodeproperties rows="3" cols="1" choices="$freerun|$keytrigger|$random"/>
         </node>
         <node component="60x18 Switch" xoff="170" yoff="65" guiid="lfo.unipolar">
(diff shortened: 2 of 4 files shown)


baconpaul:
Wow thank you! And welcome to surge! You may want to also join our slack by the way. We end up chatting about the tool quite a lot over there. But GitHub is great too.

Let me take a few minutes and do a quick review and also see if I can figure out why the filter env isn’t showing up. From your code changes it looks like they should so will pull it into my copy and run it.

Oh and you probably also want to add yourself to the AUTHORS file if you want to push that into your branch.

baconpaul:
OK here's why it didn't show the filter and why it didn't work.

You called your 'LFOs trigger' thing 'filters.env.mode' 

```
diff --git a/resources/data/layouts/original.layout/layout.xml b/resources/data/layouts/original.layout/layout.xml
index 1ab9f68..1f5e50d 100644
--- a/resources/data/layouts/original.layout/layout.xml
+++ b/resources/data/layouts/original.layout/layout.xml
@@ -688,7 +688,7 @@
 
       </layout>
       <layout height="auto" width="auto" style="roundedlabel" bgcolor="$mod.bg" fgcolor="$mod.fg" label="Modulator">
-        <node component="60x12 Multi" guiid="filter.env.mode" xoff="170" yoff="20" width="60" height="36">
+        <node component="60x12 Multi" guiid="lfo.trigmode" xoff="170" yoff="20" width="60" height="36">
           <nodeproperties rows="3" cols="1" choices="$freerun|$keytrigger|$random"/>
         </node>
         <node component="60x18 Switch" xoff="170" yoff="65" guiid="lfo.unipolar">
```

this diff makes it work.

I found this as follows

1. I ran it with stdout attached
2. On stdout I saw the error

```
ERR :  Failed to construct component 'lfo.trigmode' with failmsg 'No Component for guiid '
```

3. I then realized you hadn't laid it out in the layout but it was there so went and looked
4. et voila

We should really make it so that if you have two nodes with the same guild we pop a clearer error.  I meant to do that when I added readonly nodes last week but didn't.

Let me do a quick review of the rest of it now, but if you want to also push up the fix to the XML I think we will be in good shape.

baconpaul:
Yup, just did a quick once over and other than that layout issue above, it's all good.

If you do another push fixing the XML I will squash and merge.

Welcome to team surge synth!!

apalm:
Oops 🤦‍♂️. I pushed the change. Thanks for digging in, the detailed explanation, and the welcome :)

baconpaul:
Great, I merged it in. Super stuff! You just took my Monday Morning dev list and cleared it :)

Issue #1413, opened 2019-12-10
Add a “Channel Split” mode


baconpaul:
Right now surge has a scene “split” mode where notes below the split key are sent to one scene and above to another.

I propose to add a second split mode “channel split mode” where notes below some channel split are sent to one and above to a second. This, combined with MPE and a device like a Linnstrument, will allow a single surge patch to work with a split mode on that instrument without an artificial note split (the linnstrument split because of the geometry has notes which are in both split region and it track maps).

Do this by having the split UI go from

SINGLE
SPLIT
DUAL

To

SINGLE
DUAL
KEYSPLIT
CHSPLIT

And if you select CHSPLIT the splitkey transmogrifies into a split channel control. Deal with all the streaming consequences, implement the routing, and voila.

When doing this also, I should add perhaps two patches to the MPE folder which demonstrate this capability using the default LS split.

baconpaul:
Branch `baconpaul:channel-split-1413` contains my progress.  That branch has this feature working properly using a hard coded split-at-8/9.

This is the checklist of things I have to do still

- [ ] make it so you can set split channel
- [ ] modify the position of the "mode" label in the background
- [ ] Make the "split key" label change to "split ch" when you are in channel mode
- [ ] quadruple check streaming and so on
- [ ] that chswitch2 click still seems a bit off to me.

baconpaul:
OK this is wrapped up. Let me push where I am now and then do some more testing before I close this.

baconpaul:
As @VincyZed points out the gray on the selector is the darker not lighter gray; and should be the lighter. So fix that up.

baconpaul:
Oh @VincyZed thanks for fixing up the color over the holidays! Closing this issue since the feature is in place now!

PR #1423, opened 2019-12-16
Start to Implement Channel Split mode

baconpaul:
Addresses #1413

- Update graphics with correct states
- Unstream handling to make states map to graphics order
- Route in sm_chsplit around channel 8
- Begin state management tricks on the parameter
- Update the bakground graphics

This is a partial implementation of this feature. Items still
to be done are in the issue tagged above.


VincyZed:
This happened in the latest nightly:

image

PR #1437, opened 2019-12-20
Fixed the background color of the Scene Mode selector.

VincyZed:
Set the background color of the Scene Mode selector to the same as it was before adding Channel Split option (lighter gray, same color as A|B selector).


fluffy-uni163-the-best:
that sounds a good idea.

baconpaul:
@fluffy-uni163-the-best thanks for looking and welcome!

PR #1441, opened 2019-12-22
Fixed new parameter messages sent to host, Fixed scrolling on buttons.

### LINKED ISSUE #1440: Scrolling through buttons is not consistent
**Description:**
Scrolling through orange buttons such as **Scene Mode, FX Bypass/Character, OSC FM Routing, Filter Configuration, Octave switches, Filter algorithms, Waveshaper Shape,** etc. is inconsistent.
Sometimes it takes **two scroll** steps for it to switch to the next / previous option, when it should only take one.

**Version:**
Latest nightly, VST3, 64 bits.

**Desktop (please complete the following information):**
 - Windows 10 x64
 - FL Studio 20.6

Seems like a small bug, probably valid for #1390 ?

VincyZed:
- Fixed the parameter message that is sent to the host for the new Channel Split scene mode. (it sent **"Scene Mode: 1.00"** before, it now sends **"Scene Mode: Channel Split"**).
- Fixed inconsistent scrolling behavior on buttons. It now works correctly every time.

Closes #1440.

### DIFF
--- a/src/common/SurgeStorage.h
+++ b/src/common/SurgeStorage.h
@@ -85,7 +85,7 @@ enum sub3_scenemode
    n_scenemodes,
 };
 
-const char scenemode_abberations[n_scenemodes][16] = {"Single", "Split", "Dual"};
+const char scenemode_abberations[n_scenemodes][16] = {"Single", "Key Split", "Dual", "Channel Split"};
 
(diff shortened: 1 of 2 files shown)


baconpaul:
Awesome!

PR #1448, opened 2020-01-02
Fix the shift without mouse down behavior on buttons

### LINKED ISSUE #1447: Holding down shift and sliding above buttons with mouse pointer activates them without left-clicking
**Description:**
Holding down shift and sliding across buttons with the mouse pointer activates them without left-clicking, which is probably not intended behavior.

**Version:**
Latest nightly, VST3 64 bits.

**Expected behavior:**
Pressing "Shift" should not affect buttons whatsoever.

**Additional context:**
It does this with Scene mode, FX Bypass / Character, Polymode, Filter type selectors, octave buttons, Waveshaper shape, the EG stage curve controls (orange rectangles), and maybe more. It **DOES NOT** do it with a row or column with 2 buttons or less (A|B, Digital/Analog mode for EGs, etc.) It also does not do it with OSC 1-2-3 buttons. After a quick peak in the code I haven't found anything yet. I might take a deeper look eventually, but this is (I hope) probably a very easy fix nonetheless.

VincyZed:
This fixes a bug where you could just slide across buttons while holding down "Shift" without left-clicking.

Closes #1447

### DIFF
--- a/src/common/gui/CHSwitch2.cpp
+++ b/src/common/gui/CHSwitch2.cpp
@@ -76,7 +76,7 @@ CMouseEventResult CHSwitch2::onMouseUp(CPoint& where, const CButtonState& button
 }
 CMouseEventResult CHSwitch2::onMouseMoved(CPoint& where, const CButtonState& buttons)
 {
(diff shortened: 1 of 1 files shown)


baconpaul:
Cool!

PR #1461, opened 2020-01-09
Windows touch

nathankopp:
For windows, handle touch events differently from mouse events.  Note this includes/requires an update to the vstgui subproject.


baconpaul:
Hey I reviewed and merged. Looks great. Can you fire in a separate PR to add yourself to the AUTHORS file? Format should be obvious. Thank you!

VincyZed:
Great stuff !  :-)  Mouse behavior has changed since this update though. The sliders move faster than before, and pressing Shift for more precision doesn't work at all..

VincyZed:
Looks like it thinks I'm using a touch screen when I'm dragging with the mouse?

PR #1478, opened 2020-01-14
More slider color and contrast tweaks

VincyZed:
More slider color and contrast tweaks.
Before VS After:
image


baconpaul:
Re-running the pipeline; seems azure didn’t give us a macOS instance that ran so lets try again!

PR #1558, opened 2020-02-06
Renaming spree

mkruselj:
Wanted to bring some more consistency into parameter naming of Surge. So there we go...

### DIFF
--- a/src/common/dsp/effect/DualDelayEffect.cpp
+++ b/src/common/dsp/effect/DualDelayEffect.cpp
@@ -224,13 +224,13 @@ const char* DualDelayEffect::group_label(int id)
    case 0:
       return "Input";
    case 1:
-      return "Delay time";
+      return "Delay Time";
    case 2:
-      return "Feedback/EQ";
+      return "Feedback";
    case 3:
       return "Modulation";
    case 4:
-      return "Mix";
+      return "Output";
    }
    return 0;
 }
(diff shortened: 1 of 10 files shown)


src/common/dsp/effect/DualDelayEffect.cpp
```diff
@@ -260,7 +260,7 @@ void DualDelayEffect::init_ctrltypes()
    fxdata->p[0].set_type(ct_envtime);
    fxdata->p[1].set_name("Right");
    fxdata->p[1].set_type(ct_envtime);
-   fxdata->p[2].set_name("Feedback");
```
baconpaul: this one seems odd change; it actually is the feedback you are controlling there.

mkruselj: Sure, the amount of signal being fed back, no? 

I renamed these while considering eventual implementation of #1559 as well. If automation names are constructed from section names + parameter names, then this works out nicely. Otherwise we'd get Feedback Feedback for automation name if it was kept how it was :)

baconpaul:
Hey so alas the grouping-included-in-vst change in #1559 isn't going to happen so we need to revisit these for things like "Feedback" staying "Feedback" or the VST names will be borked.

mkruselj:
I will close this PR since a better one is incoming. :)

Issue #1579, opened 2020-02-15
deb file has wrong ownership of files


baconpaul:
@bgins points out the deb file has some weird user id for the files it installs

![image](https://user-images.githubusercontent.com/13387561/74593242-41f03880-4ff7-11ea-9119-54d8d871dae1.png)

According to https://askubuntu.com/questions/735883/what-user-file-owner-to-use-when-creating-debian-packages a reasonable answer is to modify installer_linux/make-deb.sh to add the option `--root-owner-group` to deb-dpkg at line 93.

bgins:
I added the `-root-owner-group` flag and that worked out fine. When I tested it before the change, it was setting me as the user and group, so the user and group from the screenshot are probably from the build pipeline.

baconpaul:
So --root-owner-group doesn't exist on the deb in the azure pipeline. Back to drawing board. Probably need to investigate fake root (and apt-getting that in the pipeline).

Reverted the change for now.

bgins:
The recent changes to the pipeline in https://github.com/surge-synthesizer/releases/commit/d3e2a513a826069454f3e6ae850eebe4c4ffdef3 have this one resolved.

![perms](https://user-images.githubusercontent.com/7957636/74597660-a3c09a80-5017-11ea-98e2-f13740746a60.png)

baconpaul:
Great!

PR #1580, opened 2020-02-15
Update deb file ownerships

bgins:
This PR adds the `--root-owner-group` flag to make root the owner and group on installed files on Linux. Addresses surge-synthesizer/surge#1579.


baconpaul:
Thanks! Will merge now. If you watch the GitHub repo channel on slack you can see when the new nightly ships and we can test that deb

Issue #1577, opened 2020-02-15
LV2 patch corruption when re-opening GUI 


unfa:
I've been trying to get into Surge for a good while now.

My platform of choice is Linux, and until recently that's the only OS group I've been running Surge on.
However recently I had something similar happen to me under Windows 10.

If I close the Surge window (GUI) and later open it up again in my DAW - at that moment the patch gets corrupted, usually in very specific ways.

Most often the oscillators all get reset to the default Sawtooth. Often LFO settings are corrupted too. Sometimes the patch is completely reset to Init state.

If I store my patch first before closing the GUI I can just reload it immediately after it gets corrupted, but it'll break again after I reopen the Surge window for another time.

As you can expect this is a big problem and prevents me from being able to do anything more than fiddling with Surge, and it's a big loss for me, as I see a great potential for amazing sounds in it.

I've had this issue for about a year now (basically since the first time I tried Surge on Linux).
This happens to me no matter the host: Ardour, Carla or Zrythm.

However this week I had a similar problem happened on a Windows 10 system where I so far was able to use Sure with success with Ardour 5.12.

one of my patches got reset to Init when loading a project created a few days earlier. I have saved and loaded it before without problems, and other patches were fine. I don't know if this could be related.

Please let me know what can I do to help you fix this problem. I know I'm not the only Linux user suffering from it.

baconpaul:
Hi!

OK you are the first report we have to it. Can I ask a couple of rudimentary questions?

Are you using the VST2, VST3 or LV2?
Are you using Surge 1.6.5?
Do you have the ability to build surge from source if we have a debugging build?
Do you have a DAW file for some easily available DAW which shows this?

We have recently been adding automated tests and stream/unstream invariance is one of the things we partly test. But some more details could help us loo for sure.

Thanks

baconpaul:
Oh also: I’ve used surge for a year now (mac AU and VST3) as havre many of our users and I haven’t seen this behavior once. So step one is pinning down what code path you are on which we are not so we can look there,hence my block and tackle questions above.

Thanks!

baconpaul:
One last thought: The activity you describe seems like some sort of race condition between the DAW restoring the state and the internal state of the synth. In the VST2/3/AU implementations we don’t have this problem because the synth is the sole master of the values, but I understand in the LV2 there is a question about whether the patch is mastered by the LV2 or by the engine. 

[...]

PR #1583, opened 2020-02-16
lv2: reload parameters in host after editor loads a patch

jpcima:
#1577 

This is a better fix for updating patch after reloading, although still hackish.
In particular, this one is thread-safe and will possibly help with some crash reports I received. (@alcomposer)

```
   // if a patch was loaded from storage, the plugin atomically sets a bit to
   // indicate it. In this case, write the control value again using UI's write
   // interface, in order to update the host side.
```

From what I can tell, this works in Ardour, Jalv.

However it still fails in Carla, for some reasons I can't explain. On saving, it doesn't pick up the parameters which have been set by UI's write function. (cc @falkTX)

Note: Carla was `1:2.1.b1.test1.r210.gd17c94b3f-1`
I've checked control ports values in and out, after saving closing carla, and reloading. It was different values.


baconpaul:
Great thanks! Don’t worry about that failing test it’s my fault not yours. Merging this

PR #1595, opened 2020-02-20
LV2 State Restore and Port Consistency

baconpaul:
When we restore the state form XML, we have to set it onto
the ports also, otherwise the value caches won't work.

Addresses #1577

### DIFF
--- a/src/lv2/SurgeLv2Wrapper.cpp
+++ b/src/lv2/SurgeLv2Wrapper.cpp
@@ -10,6 +10,9 @@ SurgeLv2Wrapper::SurgeLv2Wrapper(double sampleRate)
 {
    // needed?
    _synthesizer->time_data.ppqPos = 0;
+
+   for( int i=0; i<NumPorts; ++i )
+      _dataLocation[i] = nullptr;
 }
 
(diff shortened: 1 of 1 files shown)


baconpaul:
@falkTX @jpcima pragmatically, this makes Carla save and load state from the LV2 with no problem at all. Can you review and comment?

I would like to release 1.6.6 on Saturday, and would like to include this diff. Appreciate your looking.

@unfa if you know how to build from a branch in my fork, you can also build this branch and test, but if not that's cool and I will just merge it after review. (or won't and then will release broken still)

Thanks!

falkTX:
Thanks for the patch, but this breaks LV2 spec.
Plugins can only read from ports during run()

falkTX:
Also, plugins are not allowed to change their own input ports.

baconpaul:
Ok
So when I unstream the xml how do I let the ports know they are stale?

baconpaul:
Oh geez your comment is super helpful not o look again. I shouldn’t set the port there I should set the old controller value structure. Ignore my prior question

baconpaul:
Huh yeah ok Carla is restoring port values which are defaulted not which are the saved values - I will close this and write up a clearer report tomorrow

PR #1596, opened 2020-02-21
LV2 Symbol Name Uniqueness

baconpaul:
The LV2 Symbols were not unique; leading to inconsistent
restore, automation, streaming, and many other problems.
This diff fixes that and makes sure that the XML and internal
state buffer are consistent.

This change will "break" saved sessions in that they would have
mostly not restored correctly before, and will not restore correctly
at all now. But sessions going forward will work.

Addresses #1577

### DIFF
--- a/src/lv2/SurgeLv2Wrapper.cpp
+++ b/src/lv2/SurgeLv2Wrapper.cpp
@@ -10,6 +10,9 @@ SurgeLv2Wrapper::SurgeLv2Wrapper(double sampleRate)
 {
    // needed?
    _synthesizer->time_data.ppqPos = 0;
+
+   for( int i=0; i<NumPorts; ++i )
+      _dataLocation[i] = nullptr;
 }
 
 SurgeLv2Wrapper::~SurgeLv2Wrapper()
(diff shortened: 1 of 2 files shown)


baconpaul:
@falkTX I force pushed my branch with the correct fix per our convo. Don't suppose you could take a peek this morning could you? Thank you!

src/lv2/SurgeLv2Wrapper.cpp
```diff
@@ -330,10 +351,25 @@ LV2_State_Status SurgeLv2Wrapper::restoreState(LV2_Handle instance, LV2_State_Re
...
 
    if (type != self->_uridAtomChunk)
+   {
       return LV2_STATE_ERR_BAD_TYPE;
+   }
 
    s->loadRaw(data, size, false);
    s->loadFromDawExtraState();
+
+   // Restore the ports with the loadRaw state since the synth is now in good shape
```
baconpaul: This comment is wrong but the code is right just FYI. I fill fix before I mergr

falkTX:
seems fine to me, though I would have to check if the resulting values are all correct. let me do that now..

falkTX:
Seems fine to me, only minor things:
1. license is `<GPL-3.0-only>` but this is not valid, as like that it ends up parsed as `file:///usr/lib/lv2/Surge.lv2/GPL-3.0-only`. It needs to be a URL, or perhaps an actual string (though not everyone agrees on the string case)
2. parameter values are floats, but they are in the ttl as integers (missing decimal point). though this is being picky really

falkTX:
Oh, the things I mentioned are not related to this PR, but the LV2 plugin in general.
The PR is fine

baconpaul:
Thanks! On the first https://spdx.org/licenses/ may be useful to people who are disagreeing? Using those keys if you aren’t a url may be a smart design choice? But I can change it to a url sure. And I actually know if I’m a float or an int at build time so can whack that code too. Will do and repush

baconpaul:
Oh we actually just put the 0/1 values in. Cool. So all fixed. I'll merge this when CI is done and then let unfa know over on the main issue.

Appreciate the help

Issue #413, opened 2019-01-30
Memory alignment breaks win32; move to C++17 on Windows


baconpaul:
@sagantech reports

Also the change that killed 32 bit is removing the "            _aligned_free(t_osc);" and
osc = (Oscillator*)_aligned_malloc(sizeof(SurgeSuperOscillator), 16);
      new (osc) SurgeSuperOscillator(storage, oscdata, localcopy);
with new they are allocated all over the place on 8 byte boundaries

On windows 7 32 bit build

baconpaul:
Oh sheesh. @jsakkine @kurasu you are not going to enjoy the content of this issue.

So @sagantech reported that the win32 build crashes reliably because SurgeSuperOscillator is not aligned. I went digging a bit. This is only a 32 bit windows problem.

So I wrote this code:

```
#include <iostream>
#include <iomanip>
#include <vector>
#include <string>

class unAl
{
public:
    char x;
};

class unAlSub : public unAl
{
};

class alignas(16) al
{
public:
    char x;
};


template< class T >
void testAlign( std::string name )
{
    std::cout << "\nRunning test for " << name << " alignof=" << alignof( T ) << std::endl;
    std::vector<T*> v;
    int n = 1000;
    for( auto i=0; i<n; ++i )
    {
        v.push_back( new T() );
    }
    int sixteens = 0;
    for( auto p : v )
    {
        if( ((size_t)p) % 16 == 0 )
        {
            sixteens ++;
        }
    }
    std::cout << "16 bit alignments is " << sixteens << "/" << n << std::endl;
    for( auto p : v ) delete( p );

}
  
int main( int argc, char **argv )
{
    std::cout << "SizeOf size_t is " << sizeof( size_t ) << std::endl;
    std::cout << "SizeOf void * is " << sizeof( void * ) << std::endl;

    testAlign<unAl>("unAl");
    testAlign<unAlSub>("unAlSub");
    testAlign<al>("al");
}
```

and on a 32 bit windows build it gives the following

```
c:\Users\paul\w32>testal
SizeOf size_t is 4
SizeOf void * is 4

Running test for unAl alignof=1
16 bit alignments is 601/1000

Running test for unAlSub alignof=1
16 bit alignments is 601/1000

Running test for al alignof=16
16 bit alignments is 499/1000
```

So then I went reading MSFT docs and found here: https://docs.microsoft.com/en-us/cpp/cpp/align-cpp?view=vs-2017

the following

```
[...]

PR #1606, opened 2020-02-27
SSE2 to NEON compatibility

falkTX:
Everything builds, but I did not test them yet. :black_joker: 
This patch-set is used in the KXStudio repositories armhf and arm64 builds.

### DIFF
--- a/src/common/dsp/QuadFilterUnit.cpp
+++ b/src/common/dsp/QuadFilterUnit.cpp
@@ -508,7 +508,7 @@ __m128 SNHquad(QuadFilterUnitState* __restrict f, __m128 in)
    return f->R[1];
 }
 
-#if !_M_X64
+#if !_M_X64 && !defined(NEON_SSE)
 __m128 COMBquad_SSE1(QuadFilterUnitState* __restrict f, __m128 in)
 {
    assert(FIRipol_M == 256); // changing the constant requires updating the code below
@@ -653,7 +653,7 @@ __m128 DIGI_SSE2(__m128 in, __m128 drive)
    return _mm_mul_ps(drive, _mm_mul_ps(m16inv, _mm_sub_ps(_mm_cvtepi32_ps(a), mofs)));
 }
 
-#if !_M_X64
+#if !_M_X64 && !defined(NEON_SSE)
 __m128 DIGI_SSE1(__m128 in, __m128 drive)
 {
    const __m128 mofs = _mm_set1_ps(0.5f);
@@ -691,7 +691,7 @@ __m128 TANH(__m128 in, __m128 drive)
    return _mm_max_ps(_mm_min_ps(y, y_max), y_min);
 }
 
-#if !_M_X64
+#if !_M_X64 && !defined(NEON_SSE)
 __m128 SINUS_SSE1(__m128 in, __m128 drive)
 {
    const __m128 one = _mm_set1_ps(1.f);
@@ -770,7 +770,7 @@ __m128 SINUS_SSE2(__m128 in, __m128 drive)
    return x;
 }
 
-#if !_M_X64
+#if !_M_X64 && !defined(NEON_SSE)
 __m128 ASYM_SSE1(__m128 in, __m128 drive)
 {
    const __m128 one = _mm_set1_ps(1.f);
--- a/src/linux/ConfigurationXml.S
+++ b/src/linux/ConfigurationXml.S
@@ -5,6 +5,8 @@ configurationXmlStart:
     .incbin "../../resources/data/configuration.xml"
 configurationXmlEnd:
     .global configurationXmlNullTerminator
+#ifndef __arm__
     .type configurationXmlNullTerminator, @object
+#endif
 configurationXmlNullTerminator:
     .int 0
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -120,7 +120,6 @@ endif()
 if (UNIX)
     target_compile_options(surge-headless
         PRIVATE
-            -msse2
             "-D_aligned_malloc(x,a)=malloc(x)"
             "-D_aligned_free(x)=free(x)"
     )
--- a/premake5.lua
+++ b/premake5.lua
@@ -85,7 +85,7 @@ elseif (os.istarget("linux")) then
 
     buildoptions { "-std=c++17" }
 
-    platforms { "x64", "x86" }
+    platforms { "x64", "x86", "armv7ve", "arm64" }
 
     files
     {
--- a/src/common/globals.h
+++ b/src/common/globals.h
@@ -12,10 +12,25 @@
 #if MAC
 #include "vt_dsp/macspecific.h"
 #endif
-#include <xmmintrin.h>
 
+#ifdef NEON_SSE
+#include "sse2neon.h"
+#define __m128d __m128
+#define _mm_mul_ss _mm_mul_ps
+#define _mm_sub_ss _mm_sub_ps
+#define _mm_set_ss _mm_set1_ps
+#define _mm_set_sd _mm_set_ss
+#define _mm_rcp_ss _mm_rcp_ps
+#define _mm_min_sd _mm_min_ss
+#define _mm_max_sd _mm_max_ss
+#define _mm_cmpgt_ss _mm_cmpgt_ps
+#define _mm_rsqrt_ss _mm_rsqrt_ps
+#define _mm_cvtsi32_ss(a,b) _mm_cvtepi32_ps(vdupq_n_s32(b)) /* this is wrong, but only used once, so whatever */
+#else
+#include <xmmintrin.h>
 #if LINUX
 #include <immintrin.h>
+#endif
 #endif
 
 #if MAC || LINUX
--- a/src/common/precompiled.h
+++ b/src/common/precompiled.h
@@ -1,4 +1,6 @@
+#ifndef NEON_SSE
 #include <emmintrin.h>
+#endif
 
 #include "globals.h"
 
--- a/src/common/sse2neon.h
+++ b/src/common/sse2neon.h
@@ -1,0 +1,2766 @@
+#ifndef SSE2NEON_H
+#define SSE2NEON_H
+
+// This header file provides a simple API translation layer
+// between SSE intrinsics to their corresponding ARM NEON versions
+//
+// This header file does not yet translate all of the SSE intrinsics.
+//
+// Contributors to this work are:
+//   John W. Ratcliff     : jratcliffscarab@gmail.com
+//   Brandon Rowlett      : browlett@nvidia.com
+//   Ken Fast             : kfast@gdeb.com
+//   Eric van Beurden     : evanbeurden@nvidia.com
+//   Alexander Potylitsin : apotylitsin@nvidia.com
+//   Hasindu Gamaarachchi : hasindu2008@gmail.com
+//   Jim Huang            : jserv@biilabs.io
+//
+// A struct is now defined in this header file called 'SIMDVec' which can be
+// used by applications which attempt to access the contents of an _m128 struct
+// directly.  It is important to note that accessing the __m128 struct directly
+// is bad coding practice by Microsoft: @see:
+// https://msdn.microsoft.com/en-us/library/ayeb3ayc.aspx
+//
+// However, some legacy source code may try to access the contents of an __m128
+// struct directly so the developer can use the SIMDVec as an alias for it.  Any
+// casting must be done manually by the developer, as you cannot cast or
+// otherwise alias the base NEON data type for intrinsic operations.
+//
+// A bug was found with the _mm_shuffle_ps intrinsic.  If the shuffle
+// permutation was not one of the ones with a custom/unique implementation
+// causing it to fall through to the default shuffle implementation it was
+// failing to return the correct value.  This is now fixed.
+//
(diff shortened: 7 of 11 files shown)


src/common/dsp/QuadFilterUnit.cpp
```diff
@@ -691,7 +691,7 @@ __m128 TANH(__m128 in, __m128 drive)
    return _mm_max_ps(_mm_min_ps(y, y_max), y_min);
 }
 
-#if !_M_X64
+#if !_M_X64 && !defined(NEON_SSE)
```
baconpaul: These SSE1 instructions aren't actually used anywhere now anyway - I should really remove them.

src/linux/ConfigurationXml.S
```diff
@@ -5,6 +5,8 @@ configurationXmlStart:
     .incbin "../../resources/data/configuration.xml"
 configurationXmlEnd:
     .global configurationXmlNullTerminator
+#ifndef __arm__
```
baconpaul: We will need this to be null terminated somehow otherwise the read will fail and surge can crash in some situations.

baconpaul:
Tossed in a few comments. All looks pretty reasonable. Thank you for pushing these back upstream!

One other one is: sse2neon.h should be placed in libs/sse2neon with a clear license and provenance file (like we did with catch2) and not into src/common.

falkTX:
No problem.
I did this as a test first to see if it was possible. So you think eventually it could be merged?
The xml assembly stuff I have no idea how to fix or handle.

baconpaul:
Yeah I would have no problem merging this. In fact I would like to. Probably needs a bit of focus but not impossible at all. 

And we are sort of increasing our risk with the head of the codebase for a bit now we have 1.6.6 shipped and are pushing towards 1.7 - so this is a good time to do it.

hexdump0815:
just for reference: i'm doing something similar (but more ugly i guess) on my armhf/aarch64 build of vcvrack - i use the more complete simde headers:
https://github.com/hexdump0815/vcvrack-dockerbuild-v1/blob/master/prepare.sh#L28
and then simply rename all the function call accordingly in a brute force way (the ugly part):
https://github.com/hexdump0815/vcvrack-dockerbuild-v1/blob/master/simde-ify.sh
is this maybe what you ment with xml stuff?
https://github.com/hexdump0815/vcvrack-dockerbuild-v1/blob/master/surge-rack-surge.armv7l.patch
and on armhf i had to disable unaligned access to avoid crashes:
https://github.com/hexdump0815/vcvrack-dockerbuild-v1/blob/master/surge-rack.armv7l.patch

best wishes - hexdump

falkTX:
That is very helpful, thanks!
Going to see if the `%` operator works for my builds, I assume yes.

Any special reason to pick `-mfpu=vfpv3-d16` over the usual `-mfpu=neon-vfpv4` ?

falkTX:
Sadly I still get a crash in the build server.
So I will have to run the armhf in actual hardware now for real testing :/

baconpaul:
So all that is doing is null terminating the configuration xml memory block. Any other way to stick a zero at the end would work too

falkTX:
I expected the `%object` to do the trick, since it seems to work for @hexdump0815 
But perhaps the issue is somewhere else, or the cloud builder environment has missing instructions.
I will have more details soon, need to find and install debian on my Pi. I am not in a rush though

baconpaul:
One other thought: linux and only linux reads that configuration from memory. on mac and windows we read it from resources. If you have a #if ARM you could modify SurgeStorage.cpp (you can clearly see where it gets it) and add an if dev to read from resources.

I think, in retrospect, it was a mistake to do linux this way but back when we all were doing it it seemed reasonable. Since we need so many other things from the shared resource directory though singling this one out on only linux is something we could reverse if it becomes a real problem

hexdump0815:
@falkTX - in case you would like to have a ready to go debian or ubuntu image for a raspberry-pi 3 - you might use those: https://github.com/hexdump0815/imagebuilder/releases/tag/200206-01 ... they are available in armhf and aarch64 versions :)

and here are some useful audio optimizaions for the rpi (they are for raspbian, but should be working on debian and ubuntu too): https://github.com/hexdump0815/vcvrack-dockerbuild-v1/blob/master/readme-raspbian.txt ... in that context maybe interesting too: https://github.com/hexdump0815/sonaremin/blob/master/files/extra-files/etc/rc.audio

good luck and best wishes - hexdump

baconpaul:
Is there a docker image or some such we can use to try a pi build?

hexdump0815:
@baconpaul - maybe this might help you to get going: https://github.com/hexdump0815/vcvrack-dockerbuild-v1/blob/master/readme-arm-build-on-intel.txt

hexdump0815:
@baconpaul @falkTX - if any of you should have any compiled builds for armhf and aarch64 ready and want me to test them, please let me know - i can easily do so on ubuntu 18.04 and debian buster on both armhf and aarch64 - just let me know ...

baconpaul:
Hi @hexdump0815 sorry I haven't actually been active on this - been working on skinning engine for a 1.7 release. Nothing coming from me.

hexdump0815:
@baconpaul - no problem, there is no hurry :)

hexdump0815:
ok - i gave the hacking approach i used for the surge-rack module for vcvrack a try to compile the surge plugin itself that way too and finally got it working on both armv7l and aarch64 ... this approach is really just a hack to get it compiled, but its at least a proof of concept that surge works on those platforms - i tested the vst's with reaper on both of them ... here are my notes and patches from this try: https://github.com/hexdump0815/surge-arm-build ... under releases one can find compiled versions ...

baconpaul:
Wow simdeify.sh has quite a regex!!

This is cool. It looks like mostly flag changes. I wonder if we could coax premake, if an ARM variable is set, to generate those diffs without a patch. So basically have premake make a makefile which runs simdyfy and outputs the changed flags.

Thank you for sharing that. Neat stuff.

hexdump0815:
@baconpaul - i played around a bit more with surge in reaper on armv7l and it looks like this architecture (i tried it on a rockchip rk3288, which already is one of the faster armv7l  cpus) is maybe a bit weak for surge - as soon as i use more complex patches (like "back to tyrell" from noisegeek) it quickly results in dropouts as soon as i play polyphone notes and somehow reaper/libjack and qjackctl both start to saturate a cpu each ... i only see this with surge there - helm or zynaddsubfx for instance do not show this phenomenon ... are there any tricks or changes (maybe even compile time) to make surge a bit less cpu hungry on such low end platforms (maybe increasing some internal buffers or such)?

a lot of thanks in advance and best wishes - hexdump

baconpaul:
So the trick is really “use the SSE2 instruction set!” When you substitute it out what are you replacing it with? The synth is super duper CPU efficient on x86/SSE2 architectures precisely because of all the hand coded SSE2 and if you replaced that with something less optimal it may destroy performance.

That said I have no idea how powerful armv71 is. Surge was originally written to run OK on 2005 era intel chips. How does the armv71 compare?

hexdump0815:
it should be in that cpu power range, maybe even a bit faster - its similar to the 2x1.6ghz sl7500 of the thinkpad x61s i'm writing this on right now - its a 4x1.8ghz 32bit cpu and it has something simd comparable to sse2 called neon and this simde stuff is trying to translate sse2 as good as automatically possible to neon which will not be as good as hand coded for sure (maybe with the compiler options i'm using to avoid the unaligned access crashes the neon instructions are not yet fully utilized even) ... at least on the 64bit aarch64 arm it seems to work pretty well - there it is pretty useable on an odroid c2 for instance (4x1.5ghz arm a53 64bit cores) ... all this is no reason to worry, its at least running to some degree even on armv7l and maybe someone with good arm know how will come around and has some good ideas to improve it further even on this dying species of older 32bit arm cpus :)

baconpaul:
Yeah so the last time I thought about a non-x86 instruction set was very very long ago so I am way out of my league. But your comment on alignment is one that perhaps matters. Surge makes an effort to align blocks on 16 bit boundaries (of course - SSE2 requires that) using the C++-11 and C++-14 alignas directive. Does your compiler toolchain obey that? On win 32 bit with vs2017 for instance to make it work I had to turn on the C++-17 flag (even though alignas is a C++-11 and 14 directive). (Win 64 bit, linux, and mac were fine with alignment directives without the 17 uptick).

I don’t know if any of this helps, but I would like to try and help.

Amazing that it is working on the other chipset. Do you think somehow that it is mergable? I would be thrilled to try and do so.

hexdump0815:
that c++17 seems to partially apply to the armv7l gcc as well: i recompiled with c++14 replaced with c++17 and with this the lv2 linking does not crash anymore and i can scan the vst3 plugin properly in reaper, but as soon as i send some midi to it it still crashes with a bus error due to unaligned access.

so we have the choice between "-march=armv7-a -mfpu=vfpv3-d16 -mno-unaligned-access" which does not seem to use neon simd and thus is slow  but does not crash or "-march=armv7-a -mfpu=neon -mno-unaligned-access" which would most probably be faster as it would use neon simd but crashes due to unaligned access - so we have an alignment problem somewhere on this platform: either in the surge code or in the simde code.

i tried to build the vst3 with debug info and run reaper from gdb, but this did not give any more details where it crashed. interesting was that i first replaced "-O3" with "-g" and then reaper crashed already on startup (most probably during plugin scan), putting back "-O3" as well made it start and crash only on midi input - so looks like all this is quite fragile. looks like here someone with deep arm and neon know how would be required.

baconpaul:
So I think the question is whether alginas works properly. Lemme see if I can dig up the teensy test program (or recreate it) I used when doing 32 bit windows

baconpaul:
Oh smart I put it all in a GitHub issue

https://github.com/surge-synthesizer/surge/issues/413

Can you take the very very last version of my test program in that issue and run it in your various flag environments with your compile tool chain? That issue explains the problem we faced basically.  Thanks!!

hexdump0815:
i can only run your first test program, as the second uses _aligned_malloc and _aligned_free which seems to be a microsoft only thing ... independent of compiler options i always get the following on armv7l:
```
SizeOf size_t is 4
SizeOf void * is 4

Running test for unAl alignof=1
16 bit alignments is 344/1000

Running test for unAlSub alignof=1
16 bit alignments is 280/1000

Running test for al alignof=16
16 bit alignments is 1000/1000
```
on aarch64 i get:
```
SizeOf size_t is 8
SizeOf void * is 8

Running test for unAl alignof=1
16 bit alignments is 1000/1000

Running test for unAlSub alignof=1
16 bit alignments is 1000/1000

Running test for al alignof=16
16 bit alignments is 1000/1000
```
does that tell you anything in any way?

baconpaul:
Right so that means that arm71 is ignoring the c++ alignas directive 
That’s why you get unaligned memory
And that’s why you crash without noalign
And that’s why you are slow without aligned neon

So cool now we know the problem! Now off to google...

(By the way one solution is to in an ifdef arm71 overload opnew perhaps so not all hope is lost but it may be real chunk of work)

baconpaul:
oh i'm sorry ignore me. i mis-read the program output and alignment *is* working on arm71
harumph
sorry lots going on and was trying to answer between other things.

schnitzeltony:
Have a question related the original PR:

Why do you suggest NEON_SSE (guess it is intended to be set by build-system) and don't use __ARM_NEON__ which is set automatically by compiler?

schnitzeltony:
Aargh - cmark magic - I meant

\_\_ARM_NEON__

baconpaul:
I just last week moved master from premake to cmake; I wonder if that would make some of these ports easier (for instance, we could avoid all the substitution steps and make a header which was included only in a neon build; merge that into master; etc...)

But again I don’t have the hardware or the know how to work on this. Happy to help though!

schnitzeltony:
> I just last week moved master from premake to cmake; I wonder if that would make some of these ports easier (for instance, we could avoid all the substitution steps and make a header which was included only in a neon build; merge that into master; etc...)

I saw that after writing here. FWIW this PR does not apply any more.
Am working with cross environment (Yocto) and dozens of Raspis to test. Hope I'll find some time to rebase+rework this to master. It is worth because main coding effort was done by @falkTX.

baconpaul:
Great! That’s cool.

And yeah also @hexdump0815 got it building in the rack-version (which contains all the dsp but none of the UI) and it worked fine in his arm for vcv rack. So definitely workable. Just would love to figure out what to merge back into the master branch and (aspirationally) if there’s a way I could have our pipelines test it still works somehow.

I have a mac and parallels. If yocto makes it easy for me to run your env under a docker container or in a linux vm happy to lend a hand.

But super happy to do a merge. We have continuous integration running to make sure mac win and linux keep working on any changes you suggest.

Thanks!

hexdump0815:
@baconpaul - yes i think that header solution would be the right one and it could then easily be used for both native surge and surge-rack ... its just a matter of time to build that header file, which will be quite a bit of hand work i guess ... i was thinking of starting with my replace script and then use the diff as a base for which functions need to be mapped in such a header file - this way it should at least be possible to half automate the process ...

baconpaul:
OK I want to at least do something with this PR before 1.7.0 so I am right now merging it into our cmake and testing it with the ubuntu cross compiler chain, using ARM_NEON and so on. Will update the readme and I hope merge. Probably will end up closing this PR and adding a new one - will update here when I do.

baconpaul:
Closing in favor of #2342.

PR #1611, opened 2020-03-05
Handle VST3 poly aftertouch as real value

### LINKED ISSUE #1610: VST3 polyphonic aftertouch incorrect
**Describe the bug**
VST3 polyphonic aftertouch incorrect

**Please let us know your surge version**
Surge Git 44e2fed

I find a programming problem which seems to me rather evident with a look at source code.

The poly pressure from VST3 is documented as 0-1 float, and passed into an int argument expecting 0-127. I'm guessing the aftertouch also required the `value01ToMidi7Bit` treatment.

https://github.com/surge-synthesizer/surge/blob/44e2fedc335cd32984395e8243c85fd6a7d7b562/src/vst3/SurgeVst3Processor.cpp#L260-L262

jpcima:
This is supposed to fix incorrect interpretation of the poly aftertouch.
It's just as is, I have not tried.
I have noticed this while studying how VST3 works.

fixes #1610

### DIFF
--- a/src/vst3/SurgeVst3Processor.cpp
+++ b/src/vst3/SurgeVst3Processor.cpp
@@ -258,9 +258,11 @@ void SurgeVst3Processor::processEvent(const Event& e)
    }
 
    case Event::kPolyPressureEvent:
-      getSurge()->polyAftertouch(e.polyPressure.channel, e.polyPressure.pitch,
-                                 e.polyPressure.pressure);
+   {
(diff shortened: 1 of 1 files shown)


baconpaul:
That CI failure is not due to you. I’ll look int he next day or so. Seems like a test booger - just can’t fix it from phone/ipad. Sorry bout that and thanks1

baconpaul:
Thanks @jpcima - all merged!

Issue #1724, opened 2020-04-15
cmake linux vst2, Surge.so  fails to appear in the products directory


mickski56:
As title really, the lib builds fine and is in the build directory as libsurge-vst2.so but doesn't appear in the products directory under Surge.vst2

One possible solution is to patch build-linux.sh and CMakeLists.txt as follows.
Plus the addition of one file.

```
--- surge/build-linux.sh-orig	2020-04-15 10:33:35.495314342 +0100
+++ surge/build-linux.sh	2020-04-15 13:55:16.275694773 +0100
@@ -95,7 +95,7 @@
 {
     if [ ! -z "$option_vst2" ]; then
         run_cmake_if
-        run_build "surge-vst2"
+        run_build "Surge.vst2"
     fi
 
     if [ ! -z "$option_vst3" ]; then
@@ -231,7 +231,7 @@
 
 if [ -z "$option_debug" ]; then
     config="config=release_x64"
-    vst2_src_path="build/libsurge-vst2.so"
+    vst2_src_path="products/Surge.vst2"
     vst3_src_path="products/Surge.vst3"
     lv2_bundle_name="Surge.lv2"
     lv2_src_path="products/$lv2_bundle_name"
     

+++ surge/CMakeLists.txt	2020-04-15 13:59:41.937709220 +0100
@@ -617,7 +617,18 @@
       COMMAND (cd .. && ./scripts/macOS/package-vst.sh )
       )
    endif()
-
+  if( UNIX AND NOT APPLE )
+    add_custom_target( Surge.vst2 ALL )
+    add_dependencies( Surge.vst2 surge-vst2)
+    add_custom_command(
+      TARGET Surge.vst2
+      POST_BUILD
+      WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
+      COMMAND echo "Packaging up VST2 component"
+      COMMAND pwd
+      COMMAND ./scripts/linux/package-vst2.sh
+      )
+   endif()
    if( WIN32 )
      add_custom_target( surgev2-dll ALL )
      add_dependencies( surgev2-dll surge-vst2 )

new file required at surge/scripts/linux/package-vst2.sh

#!/bin/sh

OUTPUT_DIR=products

BUNDLE_NAME="Surge.vst2"
BUNDLE_DIR="$OUTPUT_DIR/$BUNDLE_NAME"

echo "Creating Linux VST2 Bundle..."

# create basic bundle structure

[...]

PR #1725, opened 2020-04-15
Address #1724, cmake linux vst2, Surge.so fails to appear in the products directory #1724

mickski56:
adjust cmake build to address packaging issue on linux.


baconpaul:
Thank you!

baconpaul:
Oh shoot actually this breaks one or two things. I'll clean it up now. Sorry for that!

Issue #1146, opened 2019-09-10
LFO Magnitude bipolar, rename, reposition


mkruselj:
"Magnitude" is an extremely rarely used word when LFOs are concerned. Renaming this parameter to "Intensity" or "Amplitude" would be much better.

Also, after a bit of messing around, when you modulate LFO magnitude/amplitude/intensity with another LFO that is bipolar, Surge allows the modulation to overshoot the parameter range. So in case of bipolar LFO, it's going to eventually invert the amplitude of the modulated LFO. Which is COOL!

However, with that in mind, I think that LFO amplitude should be a bipolar slider, not a unipolar one like it is now. It would be more representative. Plus you would also be able to very easily create a reverse ramp LFO, for example.

Lastly, since this parameter is basically scaling the output of the LFO, I think it should swap places with the Deform slider, and be at the bottom there.

sense-amr:
i second the motion..

baconpaul:
OK these are all good ideas; but I’m going to leave them til post 1.6.2 - I’m in the mood to ship what we have!

mkruselj:
Sure I didn't even mean to imply that it should be in 1.6.2 :)

baconpaul:
Yeah but you know - I listen to your ideas man!

PR #1732, opened 2020-04-15
ED's renaming spree, part 2 take 1

mkruselj:
* forgot to rename FM3's M3 Freq to M3 Frequency
* renamed FM2's Mx parameters to M1/2 Offset and M1/2 Phase
* renamed send FX send/return level parameters
* renamed LFO Magnitude to Amplitude
* various other parameter renames visible in right-click menu only)


mkruselj:
This one partially implements #1146!

PR #1791, opened 2020-05-05
About box tweaks

mkruselj:
* link to Surge Github page is clickable now
* tweaked the capitalization of OS and plugin types
* separated factory and user data folders to separate lines
* used full legal names of Steinberg and Apple
* removed the gray rectangle from SVGs, generated in code instead


src/common/gui/CAboutBox.cpp
```diff
@@ -152,7 +160,19 @@ CAboutBox::onMouseDown(CPoint& where,
    if (!(button & kLButton))
       return kMouseEventNotHandled;
 
-   boxHide();
+   //char debug[128];
```
baconpaul, line 163: please don't leave this in the code. thanks!

src/common/gui/CAboutBox.cpp
```diff
@@ -2,9 +2,12 @@
 #include "globals.h"
 #include "resource.h"
 #include "RuntimeFont.h"
+#include <windows.h>
```
baconpaul, line 5: bet that won't work on linux :)

mkruselj: was just for debugging, oops!

src/common/gui/CAboutBox.cpp
mkruselj: oh drat, forgot to remove that.

src/common/gui/CAboutBox.cpp
mkruselj: updated in cf82b26

src/common/gui/CAboutBox.cpp
mkruselj: Updated in cf82b26

Issue #840, opened 2019-04-14
Implement directory and filter in fileOpenDialog in UserInteractions


baconpaul:
The first version of FileOPenDialog I just pushed (https://github.com/surge-synthesizer/surge/commit/fcc8c43bb315c7a86b91f6ffbdce3b75080b0416) ignores the filter and starting directory arguments.

This issue exists to nag me to not do that.

baconpaul:
@mkruselj if you want to here's the issue

mkruselj:
Managed to implement initial directory working just fine, but not the filter.

Also if the initial directory is not found (i.e. somebody renames it), it should fallback to the factory data folder, but for now I didn't manage to figure out how to do that.

mkruselj:
Additional nicety that could be implemented (again hand-holding welcome): maybe remember the last folder the user browsed to in configuration.xml (separately for SCL and KBM of course!), and then use that the next time loading dialog is called. If the folder is later not found, revert back to factory tuning folder for SCL or KBM, and again if THAT is not found, fallback to factory data root.

baconpaul:
the macOS does that automatically and i think zenity does too but I'm not sure.
there's some advanced windows ones which did too but they coupled to some newer sdks which we didn't use

mkruselj:
We have a separate issue for "additional niceties" (#1864) so closing this one!

PR #1836, opened 2020-05-08
SCL/KBM load dialogs open in respective factory folders

mkruselj:
* partial implementation of #840
* passing initialDirectory to promptFileOpenDialog will work now, but filterSuffix doesn't, for to me unknown reason
* TODO: fallback if those factory folders aren't found - open Surge factory data folder root (pointers welcome!)

### DIFF
--- a/src/common/gui/SurgeGUIEditor.cpp
+++ b/src/common/gui/SurgeGUIEditor.cpp
@@ -4129,7 +4129,11 @@ VSTGUI::COptionMenu *SurgeGUIEditor::makeTuningMenu(VSTGUI::CRect &menuRect)
                            scl_folder = "tuning-library/SCL";
                         #endif
                         Surge::UserInteractions::promptFileOpenDialog(this->synth->storage.datapath + scl_folder,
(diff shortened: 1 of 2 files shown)


src/common/gui/SurgeGUIEditor.cpp
```diff
@@ -4122,8 +4122,14 @@ VSTGUI::COptionMenu *SurgeGUIEditor::makeTuningMenu(VSTGUI::CRect &menuRect)
...
-                        Surge::UserInteractions::promptFileOpenDialog(this->synth->storage.userDataPath,
-                                                                      ".scl",
+                        std::string scl_folder;
+                        #if WINDOWS
+                           scl_folder = "tuning-library\\SCL";
+                        #else
+                           scl_folder = "tuning-library/SCL";
+                        #endif
+                        Surge::UserInteractions::promptFileOpenDialog(this->synth->storage.datapath + scl_folder,
+                                                                      "Scala microtuning files (*.scl)",
```
baconpaul, line 4132: Does this postfix work on mac and linux? It doesn't look like it will to me!

PR #1839, opened 2020-05-08
A couple of quick renames in SurgeStorage.h

mkruselj:
* add extra & for Windows in Sample & Hold filter type
* no shorthands in play modes


mkruselj:
Well I guess only the "no shorthands in play modes" should stay then, since the && thing impacts host automation label...

baconpaul:
correct

Issue #1846, opened 2020-05-09
Surge LV2 UI does not scale with the host


falkTX:
When an LV2 plugin host runs in a high-dpi screen or is set to render at a custom scale, Surge LV2 UI does not scale to match.
See https://github.com/falkTX/Carla/issues/1121

This happens in any surge version.
For LV2, there is a UI property to inform the plugin about the host GUI scale.
Surge should be reading this information and automatically set UI scale to match.

PS:
Not sure about VST2 or VST3, might be the same also happens there.
Focusing on LV2 first, then we can see the rest.
I will try to fix this myself, as a way to ensure Carla is implementing this properly too.
I have opened the ticket just to get some visibility on the topic, and I might need some help soon :)

baconpaul:
The HDPI support is mostly in DisplayInfo/getBackingStore which is what upscales us. Don't know anything about linux apis to find out our density but you can see that plus the zoom handler in SurgeGUIEditor is basically where it is. You can see there's handling of physicalzoomfactor as well in surgeguieditor around like 3730 to help compensate for some vstgui bugs. that's probably where you want to read your property to increase zoom.

Note the LV2 still has loads of bugs with parameters and basically doesn't work in ardour. Those are still open! So just when you test don't expect consistent behavior of the plugin.

falkTX:
Yeah I know of the other issues, sorry been a lot busy with work.
Today I thought to look into Carla UI issues, this one is relevant since I need an opensource plugin that can handle scaling and supports multiple formats. So surge is quite the nice fit.
I will look into the other LV2 issues another day, if that is okay.

Anyway, is there a way to force a specific UI scale before creating the UI? so it starts up with that?
I think if the host informs the plugin UI what scaling to use, the plugin should use that instead of trying to guess the desktop one (which sometimes is not correct, could be also that host is running at a custom scale set by the user)

baconpaul:
Yeah what I would do if I had to was (presuming you have a function called “getExtraScale()” which returns 1, 2, etc...)

You can probably do it all in the zoom callback per host

https://github.com/surge-synthesizer/surge/blob/1aa548947e441364448812e7da984b7b53eb9254/src/lv2/SurgeLv2Ui.cpp#L150

If you need to intervene earlier than that, In surgeguieditor.cpp SetZoom with an #ifdef TARGET_LV2 could do it also

But i think the lv2 local one will let you do it

baconpaul:
[...]

PR #1851, opened 2020-05-09
Implement lv2ui:scaleFactor

falkTX:
As discussed in #1846

Some screenshots to demonstrate (I do not manually set anything, just open the UI from Carla)

With 1.0 scale:
Screenshot_20200509_221228

With 1.5 scale:
Screenshot_20200509_221417

Cannot test with 2x scale since surge gives warning about window not fitting on screen, which is indeed the case :joy:

### DIFF
--- a/src/lv2/AllLv2.h
+++ b/src/lv2/AllLv2.h
@@ -13,6 +13,10 @@
 #include <lv2/time/time.h>
 #include <lv2/state/state.h>
 
+#ifndef LV2_UI__scaleFactor
+#define LV2_UI__scaleFactor LV2_UI_PREFIX "scaleFactor"
+#endif
+
 #define SURGE_PLUGIN_URI "https://surge-synthesizer.github.io/lv2/surge"
 #define SURGE_UI_URI     SURGE_PLUGIN_URI "#UI"
 #define SURGE_PATCH_URI  SURGE_PLUGIN_URI "#PatchData"
--- a/src/lv2/SurgeLv2Export.cpp
+++ b/src/lv2/SurgeLv2Export.cpp
@@ -154,6 +154,7 @@ void lv2_generate_ttl(const char* baseName)
               "                        ui:noUserResize ;\n"
               "    lv2:requiredFeature ui:idleInterface ,\n"
               "                        <" LV2_INSTANCE_ACCESS_URI "> ;\n"
+              "    opts:supportedOption ui:scaleFactor ;\n";
               "    lv2:extensionData ui:idleInterface .\n";
    }
 }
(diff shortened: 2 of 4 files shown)


baconpaul:
Hey neat. Merging!

falkTX:
You publish nightly builds of surge online right?
Can you provide me a link so I can pass it along as testing, thanks!

baconpaul:
sure. they are just at the bottom of https://surge-synthesizer.github.io/
contain the hash
usually take about 30 mins from merge

falkTX:
oh it is that simple, nice.
thanks again

agaskins:
Amazing work! :)
Screenshot from 2020-05-26 15-04-31

baconpaul:
Oh @falkTX you had asked about VST3. 

https://github.com/surge-synthesizer/surge/issues/1951

That has a link to the api. https://steinbergmedia.github.io/vst3_doc/base/classSteinberg_1_1IPlugViewContentScaleSupport.html

So basically if you have a VST3 you can check if it is cartable to IPlugViewContentScaleSupport and if it is pass it the scale factor.

Surge will support this api in our VST3 before 1.7 ships I hope.

falkTX:
Thanks, but right now I am just letting juce handle the vst3 support.
So this is already implemented there.

baconpaul:
Oh cool!

kybos:
I can confirm that the lv2 ui:scaleFactor option works nicely with HiDPI screens (it opens up with 200% percent zoom on my screen, which is the correct scaling). Tested with MusE (where I also implemented the support for this option).

However, it seems that another issue arises from that. If another zoom level is set as default in Surge (lv2) by the user, it is saved in the config file correctly, but when Surge lv2 is reopened, this setting is ignored and the zoom is again set to the "optimized" zoom derived from the scale factor.

I assume the zoom level set explicitly by the user should have priority over the scale factor, or is it intentional?

baconpaul:
Certainly looks like the diff above does what you say when it sets the zoom factor to ui scale * 100 but I’m not sure why that was the choice for the lv2

kybos:
@falkTX 

Could you please check if the custom zoom level set explicitly by the user could/should override the scaleFactor?

falkTX:
the scale factor from the host is just a hint, so of course you **can**. the **should** I dont know, it is more of a UX question.

kybos:
Yes, but currently if lv2ui:scaleFactor is passed by the host to Surge, the Default zoom setting in Surge is ignored, i.e. the user has no chance to change the default zoom level there any more because Surge is always opened in the zoom level given by lv2ui:scaleFactor. 
IMO if the user explicitly sets the default zoom level in Surge to another value, this should take priority. The default zoom level value is saved to the Surge config file if (and only if) the user has customized it.
That's at least my understanding so far. We already had a user complaint about it, which I was able to reproduce.

baconpaul:
1: the scale factor should change the backing store resolution not the zoom
2: but the 1.8 and 1.9 LV2 are buggy in innumerable other ways and you should use the VST3 instead

PR #1852, opened 2020-05-09
Finessed Ctrl+drag feature, param unit fixes for certain FX

### LINKED ISSUE #1593: Ctrl+drag on %-based parameters jumps too much
For any %-based parameters, the jump is from 0 to 100% or 0 to -100% if bipolar, rather than rounding to 1% steps, which would feel the most natural to me personally at least.

mkruselj:
* all (if I got them all!) parameters now have a lot more sensible Ctrl+drag behavior, usually rounded to an integer unit
* fixed db_to_amp() function in DspUtilities.h
* delay, chorus and frequency shifter Feedback parameters were in dB rather than % like other Feedback parameters - made more consistent
* this PR effectively closes #1593!

### DIFF
--- a/src/common/Parameter.cpp
+++ b/src/common/Parameter.cpp
@@ -791,7 +791,7 @@ void Parameter::bound_value(bool force_integer)
          break;
       case ct_chorusmodtime:
          val.f = limit_range(
-                    log2(round(powf(2.0f, val.f) * 100) / 100.f),
+                    (float) log2(round(powf(2.0f, val.f) * 100) / 100.f),
                             val_min.f, val_max.f);
          break;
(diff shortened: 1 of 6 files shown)


baconpaul:
Looks like you need explicit casts to float for your limit_range to compile on linux as an fyi. Once the ci passes I’ll do a review! Thanks for doing this!

baconpaul:
This all looks good to me other than the linux break

the linux fix is at the two lines where it didn't compile, explicitly cast your arguments to limit_range to float

So either limit_range( (float)( my math ), 1.f,. 2.f ) or change the constants from 1.0 to 1.f

PR #1861, opened 2020-05-11
Adjust text for MPE/tuning buttons

mkruselj:


baconpaul:
"Scala" seems like the wrong label.

mkruselj:
Changing it to "Tuning", then!

PR #1862, opened 2020-05-11
Don't reset to default value on Ctrl-click anymore

### LINKED ISSUE #1844: Ctrl+drag always starts from default value
Regardless of where the slider actually is, it seems that Ctrl+drag always first executes Ctrl+click, which defaults the control. This can be quite inconvenient, for example if the volume slider was at a low value, and its default value is 0 dB.

We already have defaulting controls on double-click, so Ctrl+click should just do nothing here.

mkruselj:
* closes #1844

### DIFF
--- a/src/common/gui/SurgeGUIEditor.cpp
+++ b/src/common/gui/SurgeGUIEditor.cpp
@@ -2460,7 +2460,7 @@ int32_t SurgeGUIEditor::controlModifierClicked(CControl* control, CButtonState b
 
    if (tag < start_paramtags)
       return 0;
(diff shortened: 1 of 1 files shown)


baconpaul:
just confirming: the double-click still works right?

baconpaul:
Oh also: Above this there's something that looks like

if( buttons & kDoubleClick )
  buttons |= kControl

which we can probably remove also. Can you just peek at that and see if removing it it all still works OK? Thanks!

mkruselj:
Double-click still works, yes. If I remove that part above that you mention, then it makes double-click not work, so let's keep it there :)

PR #1868, opened 2020-05-12
Path to typeins, Crossfeed in Delay now uses % unit

mkruselj:


mkruselj:
Edge cases I'd like to see somehow resolved (pointers welcome!):

- right now typing -inf into the `ct_amplitude` types does work, but it would be super-nice if it also recognized -oo as the same as -inf
- right now if you type a value lower than -192.00 dB for `ct_amplitude` it will clamp to the minimum value and show it in the context menu (instead of showing -inf). the main problem here is that it doesn't return false if you type a value below -192 dB, not sure why
- Send FX levels are also using `ct_amplitude` type but they have a larger range (multiplied by 1.5874 in SurgePatch.cpp), which created some problems, I had to special handle the case when you type 12 dB because it didn't work (I assume `float` inaccuracies). it should've been a separate ctype, but not sure if adding a new type now would mess stuff up
- it would be mega-awesome if we could parse `k` as a thousands multiplier when entering frequencies in Hz. i.e. `5.67k`, `0.35k`, `16k` and so on

PR #1880, opened 2020-05-13
Change Modulator names for Step/Env

### LINKED ISSUE #1201: Rename (S)LFO to ENV or SEQ when LFO mode is set to those
I believe the summary in the title is quite self-explanatory. All instances of (S)LFO in the plugin should be affected - both the GUI tabs, and various mentions in dropdown menus.

This should make things easier to discern when dissecting the patches.

baconpaul:
Closes #1201


baconpaul:
Don't merge this et. I gotta get the context menus.

PR #1883, opened 2020-05-14
Implemented blocking certain MIDI CCs from being MIDI learnable

### LINKED ISSUE #575: Disallow (at least) following MIDI CCs for MIDI learn: 0, 32, 120, 123
The above mentioned MIDI CCs are not supposed to be utilized for MIDI learn because they have reserved special functions as far as MIDI goes: CC#64 is sustain pedal, CC#120 is all-sound-off (kills all sound and releases all voices immediately) and CC#123 is all-notes-off (sends note off to all keys, all voices have their envelopes immediately proceeding to their release stages).

0 and 32 are used for Bank Select, which is important if Program Change support is going to be implemented to Surge (which I will open another ticket for).

mkruselj:
* also updates the "Set controller to..." menu by showing those entries as disabled rather than skipping them in the list
* closes #575


baconpaul:
I wanna pull this into Mac and check one thing. Code looks good but I think there may be a pre existing bug now I review it.

Thanks!

mkruselj:
Yeah looks like it doesn't wanna pass the unit tests on Mac :(

baconpaul:
oh no that's just noise. don't worry.

baconpaul:
Pushed two fixes. Can either merge this or wait for more from you. lemme know.

PR #1891, opened 2020-05-14
Alt+drag in step seq quantizes to number of degrees in the loaded microtuning

mkruselj:
Also Alt+Shift+drag quantizes to double the scale length (covering up to 4 octaves in the designated scale)

### DIFF
--- a/src/common/gui/CLFOGui.cpp
+++ b/src/common/gui/CLFOGui.cpp
@@ -974,10 +974,6 @@ CMouseEventResult CLFOGui::onMouseDown(CPoint& where, const CButtonState& button
    {
       if (ss && lfodata->shape.val.i == ls_stepseq)
       {
-         // find out if a microtuning is loaded and store the scale length for Alt-drag quantization to scale degrees
-         if (!storage->isStandardTuning && storage->currentScale.count > 1)
-            altQuantStep = storage->currentScale.count;
-
          if (rect_steps.pointInside(where))
          {
             controlstate = cs_steps;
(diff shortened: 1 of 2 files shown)


src/common/gui/CLFOGui.cpp
```diff
@@ -1180,16 +1184,34 @@ CMouseEventResult CLFOGui::onMouseMoved(CPoint& where, const CButtonState& butto
             float f = (float)(steprect[i].bottom - where.y) / steprect[i].getHeight();
             if (buttons & (kControl | kRButton))
                f = 0;
-            else if (lfodata->unipolar.val.b)
-               f = limit_range(f, 0.f, 1.f);
             else
-               f = limit_range(f * 2.f - 1.f, -1.f, 1.f);
-            if ( (buttons & kShift) )
+               if (lfodata->unipolar.val.b)
```
baconpaul, line 1188: why did this change from an else if to an if? Seems wrong no? (This is why I wish the codebase used curly braces everywhere btw :) )

src/common/gui/CLFOGui.cpp
```diff
@@ -974,6 +974,10 @@ CMouseEventResult CLFOGui::onMouseDown(CPoint& where, const CButtonState& button
    {
       if (ss && lfodata->shape.val.i == ls_stepseq)
       {
+         // find out if a microtuning is loaded and store the scale length for Alt-drag quantization to scale degrees
+         if (!storage->isStandardTuning && storage->currentScale.count > 1)
+            altQuantStep = storage->currentScale.count;
```
baconpaul, line 979: Oh another bug. You made altquantstep a member so it will stick around, therefore if you make a 15 note tuning it gets set to 15, then if you untune it doesn't get set back to 12.

the trick is to *not* put it in the h but make it a local here and initialize it to 12. So remove the altQuantStep in the header file below and add the line `int altQuantStep = 12` right above this if.

mkruselj: OK that's cool but then altQuantStep is not visible in ::onMouseDrag...

baconpaul: It’s not a query it’s just a memory access. Really about 3-10 instructions. But you are correct you could do an if. 

It’s sort of a trade off (and I would merge either) 

But: More bugs come from state needing to be coherent than come from doing an extra dereference so if I were coding it I would just look up the value in each onMouseDrag. They happen outside the UI thread and already access the storage object so it’s not a performance problem. But again it would be correct in either case this way and I’m happy for either change, just sharing a purely philosophical outlet.

(You do need to do one or the other though otherwise the state hangs around; if it was a local that wouldn’t be a bug available to us)

mkruselj: Alrighty! All done now :)

PR #1895, opened 2020-05-15
Fix & in patch menu categories and entries on Windows

mkruselj:


mkruselj:
And now I'm thinking, may not be a bad idea to put the find-replace function in `guihelpers`?

src/common/gui/CPatchBrowser.cpp
```diff
@@ -183,10 +207,15 @@ bool CPatchBrowser::populatePatchMenuForCategory( int c, COptionMenu *contextMen
...
             int p = ctge[i];
-            // sprintf(name,"%i. %s",p,storage->patch_list[p].name.c_str());
-            sprintf(name, "%s", storage->patch_list[p].name.c_str());
+
+            #if WINDOWS
+               string patch_name = storage->patch_list[p].name;
+               findReplaceSubstring(patch_name, string("&"), string("&&"));
+            #endif
+
+            name = patch_name;
```
baconpaul: Well so obviously this won’t work outside of windows since you only define patch name on windows. Move the string patch name to outside the windows block and Mac and Linux will be ok

mkruselj: Yes just noticed that and pushed a fix!

It was a very huge facepalm I made. Then I figured out I don't need `patch_name` at all, I can just use `name` declared beforehand!

PR #1899, opened 2020-05-16
Add scene name to automation parameters, rename controller>macro

mkruselj:


baconpaul:
So as I’m sure you saw this doesn’t compile mac and linux. Honestly I’m surprised it compiles on windows! 

Anyway printf(“%s”) takes a char* not a std::string. You convert by using .c_str(). So 

```
Std::string foo = “Hey”;
Sprintf( txt, “This is my std string %s”, foo.c_str() )
```

PR #1906, opened 2020-05-18
Zoom button added to status panel, main menu code cleanup, std tuning/kbm available when tuning loaded but disabled

### LINKED ISSUE #1905: Access certain tuning options when custom tuning is loaded but disabled
We should be able to access the "Set to standard tuning option" when custom tuning is loaded but is simply disabled. This is current behavior (Tuning file is loaded):

image

image

Same thing applies for keyboard mapping.

From Paul on Slack:
In the code, change SurgeGUIEditor line 4320 to:
`! this->synth.isStandardTuning || tuningCacheForToggle.size() > 0`

Same for kbm, about 13 lines down.

mkruselj:
* order of submenus in the main menu slightly modified
* since we have the MPE button, the main menu doesn't need to say if MPE is enabled/disabled
* moved the USE_DEV_MENU define all the way to the top of the cpp
* changed every addEntry("-") to addSeparator() instead for clarity
* each submenu in the main menu now has its own function, to clean up that part of the code a bit
* closes #1905

### DIFF
--- a/src/common/gui/CStatusPanel.cpp
+++ b/src/common/gui/CStatusPanel.cpp
@@ -20,18 +20,21 @@ void CStatusPanel::draw( VSTGUI::CDrawContext *dc )
    std::string labs[numDisplayFeatures];
    labs[mpeMode] = "MPE";
    labs[tuningMode] = "Tuning";
+   labs[zoomOptions] = "Zoom";
    int y0 = 11;
    int boxSize = 13;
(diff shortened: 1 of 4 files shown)


mkruselj:
Also, this PR will close #1905 too!

baconpaul:
Looks great; checked for leaks and all clear. Merging.

PR #1925, opened 2020-05-22
Vector LFO UI on Linux

baconpaul:
1. Make it so the Vector LFO UI on Linux can be activated
2. Add some basic timing data to the Vector LFO UI


baconpaul:
I'm not gonna merge this am just gonna leave it here for @tank-trax to do some testing

PR #1939, opened 2020-05-23
Step seq UI/popup touchups, new zoom levels, fix for typein insets on Win

mkruselj:


src/common/gui/SurgeGUIEditor.cpp
```diff
@@ -745,11 +745,6 @@ void SurgeGUIEditor::refresh_mod()
    }
    // ctnvg	frame->redraw();
    // frame->setDirty();
-#if LINUX
-   frame->invalid();
-   // Turns out linux is very bad with lots of little invalid rects in vstgui
-   // see github issue 1103
-#endif
```
baconpaul, line 752: Why did you remove this? It’s an important Linux optimization!

baconpaul: Looks like you weren’t against master. Pull master, check out this branch again, do a “git rebase master” then force push and you should be ok. But need to do that before I merge obvs

mkruselj: Yeah not sure how I ended up against master... reinstated now!

src/common/gui/SurgeGUIEditor.cpp
```diff
@@ -5053,7 +5061,7 @@ R"HTML(
        {
           int oct_offset = 1;
              oct_offset = Surge::Storage::getUserDefaultValue(&(this->synth->storage), "middleC", 1);
-          char notename[16];
+          char notename[8];
```
baconpaul, line 5064: Oh sorry this 16 -> 8 is gonna break Ubuntu 20 also and looks like another merge (it was in the same diff as the Linux opt above). As soon as Ubuntu 20 is in azure I will activate it

PR #1944, opened 2020-05-24
Note name typein, about box tweak

mkruselj:
* note names for now only work on split key and keytrack root
* both sharps and flats are supported
* about box has the Surge github link colored now (aboutbox.link skin color)

### DIFF
--- a/src/common/Parameter.cpp
+++ b/src/common/Parameter.cpp
@@ -1764,11 +1764,11 @@ bool Parameter::set_value_from_string( std::string s )
       {
          ni = std::stoi(c);
       }
-      catch (std::invalid_argument const &e)
+      catch (const std::invalid_argument &e)
       {
          ni = val_min.i - 1;   // set value of ni out of range on invalid input
       }
-      catch (std::out_of_range const &e)
+      catch (const std::out_of_range &e)
       {
          ni = val_min.i - 1;   // same for out of range input
       }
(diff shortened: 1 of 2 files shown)


src/common/Parameter.cpp
```diff
@@ -1764,9 +1764,13 @@ bool Parameter::set_value_from_string( std::string s )
       {
          ni = std::stoi(c);
       }
-      catch (std::invalid_argument const& e)
+      catch (std::invalid_argument)
       {
-         ni = val_min.i - 1;   // throw things out of range on invalid input
+         ni = val_min.i - 1;   // set value of ni out of range on invalid input
+      }
```
baconpaul, line 1770: You would want the const on the second exception also

baconpaul: And crazily you want const std::invalid_argument &e, not std::invalid_argument const& e, which means something different. Const is hard.

src/common/Parameter.cpp
```diff
@@ -1793,7 +1870,7 @@ bool Parameter::set_value_from_string( std::string s )
       return false;
    }
 
-   auto nv = std::atof(c);
+   float nv = std::atof(c);
```
baconpaul, line 1873: curious why you changed auto to float here?

mkruselj: I had the same try-catch as for ints but then reverted it and forgot to reinstate `auto`.

PR #1974, opened 2020-05-28
Increase scrolling speed to correct speed

VincyZed:
This just fixes the new scrolling feature on LFO shape. It was 2x too slow, meaning it took 2 wheel notches to increase or decrease by 1 type. It's now 2x faster, meaning it only takes 1 notch to increase or decrease, bringing it exactly like all the other button rows and columns in the GUI.

Let me know if there would be a more elegant way to fix this, btw.


baconpaul:
That’s perfect!

PR #1976, opened 2020-05-28
Further CMake Improvements

baconpaul:
1. Installed assets are both build local for out of source builds
   and in a well named directory on all platforms
2. Version generation happens at compile time reliably
3. Strip the linux vst3 dll
4. Improve some versioning schemes


tank-trax:
just curious, are you not stripping the mac osx vst3 build as well?

baconpaul:
Oh I guess I should! I’ll look when I next sit down at my laptop

PR #2000, opened 2020-06-01
Check VST3 process inputs count and nullptr.

rncbc:
[vst3-process-inputs-fix] Check VST3 process inputs count and nullptr.

Fixes Surge.vst3 to be instantiated, run and process audio correctly on qtractor >= 0.9.13  (Linux VST3).

### DIFF
--- a/src/vst3/SurgeVst3Processor.cpp
+++ b/src/vst3/SurgeVst3Processor.cpp
@@ -366,7 +366,7 @@ tresult PLUGIN_API SurgeVst3Processor::process(ProcessData& data)
 {
    CHECK_INITIALIZED
 
-   if (data.numInputs == 0 || data.numOutputs == 0)
+   if (data.numOutputs == 0)
    {
       // nothing to do
       return kResultOk;
@@ -377,7 +377,7 @@ tresult PLUGIN_API SurgeVst3Processor::process(ProcessData& data)
    int32 numChannels = 2;
    int32 numSamples = data.numSamples;
 
-   surgeInstance->process_input = data.numInputs != 0;
+   surgeInstance->process_input = data.numInputs != 0 && data.inputs != nullptr;
 
    float** in = surgeInstance->process_input ? data.inputs[0].channelBuffers32 : nullptr;
    float** out = data.outputs[0].channelBuffers32;
@@ -433,7 +433,7 @@ tresult PLUGIN_API SurgeVst3Processor::process(ProcessData& data)
          surgeInstance->process();
       }
 
-      if (surgeInstance->process_input)
+      if (surgeInstance->process_input && in)
       {
          int inp;
          for (inp = 0; inp < N_INPUTS; inp++)


baconpaul:
First: Thank you! Looks reasonable.

So basically: qtractor reports a lack of input routing by just setting the input array to null not by setting num_inputs to 0? Is num_inputs just left as uninitialized memory or something which is why the check fails?

Happy to merge just want to understand the behavior.

Thank you

rncbc:
@baconpaul 

> So basically: qtractor reports a lack of input routing by just setting the input array to null not by setting num_inputs to 0? Is num_inputs just left as uninitialized memory or something which is why the check fails?

qtractor does not support side-chaining and previous released versions just presented `data.numInputs = 1` but `data.inputs[0].channelBuffers32 = nullptr`, which was causing it to crash immediately on instantiation; this has been corrected recently on qtractor git master, by having `data.numInputs = 0` but then Surge.vst3 bails out on `SurgeVst3Processor::process()` too early, making it a pure silent instrument plugin ;)

this PR fixes both cases.

baconpaul:
Great. Merged. Thank you!!!

tank-trax:
I get these errors when loading VST3

Screenshot_20200601_123719



```
(zenity:7535): Gtk-WARNING **: 12:37:36.068: Theme parsing error: gtk.css:73:46: The style property GtkScrolledWindow:scrollbars-within-bevel is deprecated and shouldn't be used anymore. It will be removed in a future version
IPlugFrame is a runloop 0x55a8634cfda0
Constructing a SurgeBitmaps; Instances is 2
Reloading skin  name=dark-mode.surge-skin/
Surge Error
Host Error
Your host failed to zoom VST3

(zenity:7539): Gtk-WARNING **: 12:37:41.821: Theme parsing error: gtk.css:68:35: The style property GtkButton:child-displacement-x is deprecated and shouldn't be used anymore. It will be removed in a future version

(zenity:7539): Gtk-WARNING **: 12:37:41.821: Theme parsing error: gtk.css:69:35: The style property GtkButton:child-displacement-y is deprecated and shouldn't be used anymore. It will be removed in a future version

(zenity:7539): Gtk-WARNING **: 12:37:41.821: Theme parsing error: gtk.css:70:34: The style property GtkCheckButton:indicator-size is deprecated and shouldn't be used anymore. It will be removed in a future version

(zenity:7539): Gtk-WARNING **: 12:37:41.821: Theme parsing error: gtk.css:73:46: The style property GtkScrolledWindow:scrollbars-within-bevel is deprecated and shouldn't be used anymore. It will be removed in a future version
```

tank-trax:
Also upon opening Init is highly distorted. Moving back to the original Init and everything is OK.

Here are two examples. How Surge loads the Init in Qtractor and how Init should be

[Init-examples-qtractor.zip](https://github.com/surge-synthesizer/surge/files/4712413/Init-examples-qtractor.zip)

rncbc:
please get rid of all gtk2 binding ... if you're on a gtk based DE (eg. gnome, dammit) go in qtractor menu > View > Options... > Display > Custom > .Style theme: I strongly advise you to get away from "(default)" ! please chose "Fusion".

on the same vain, do make sure View > Options... > Display > Options > Use desktop environment native dialogs are OFF.

hth.
cheers

tank-trax:
I am using KDE and Dark Breeze theme. 

Use desktop environment native dialogs was already OFF.

Same VST3 Zoom error using Fusion

baconpaul:
That zoom error comes if the host hands us an IPlugFrame as a container which returns false on resizeView. If you are at default size of 100% and don't resize you will never hit the code path but if your default size is not 100% you will hit it on startup

```
        Steinberg::IPlugFrame *ipf = e->getIPlugFrame();
        if (ipf)
        {
            Steinberg::ViewRect vr( 0, 0, newW, newH );
            Steinberg::tresult res = ipf->resizeView(e, &vr);
            if (res != Steinberg::kResultTrue)
               Surge::UserInteractions::promptError("Your host failed to zoom VST3", "Host Error");
        }      
```

From this I conclude that the `IPlugFrame` instance which is being handed me by qtractor does not return kResultTrue.

Looking at the code in qtractor-git I'm not sure what condition makes it bail but that's where I'd look first.

baconpaul:
@rncbc the reason surge gives that error is you are not allowing re-entrant zooms. That is, if you are in a resize and a resize gets called you return false. the m_resizing thing

The VST3 API doesn't apply this constraint I don't think. Since resize will call onSize and onSize gets an opportunity to take the given size and make a different size (for instance to keep aspect ratio) you can get resize inside a resize

If I allow resizeView to continue with m_resizing true (so change `m_widget || m_resizing` to just `m_widget`) everything works and qtractor supports the same clip and scale that reaper, cubase, etc... all support on linux and elsewhere.

h2h.

rncbc:
in any case, the issue is not related to this PR merge whatsoever.

it's a GUI thing and rather benign at this stage if I'm allowed to say :)

tbh. this qtractor vst3 implementation has been only tested with the vst3sdk sample plugins _and_ the awesome u-he plugins (most if not all of them)

tank-trax:
I can confirm that building qtractor with

```
		if (!m_widget)
			return kResultFalse;
```
resolves the zoom issue

- the Gtk-WARNING's are gone

- although there was this `IPlugFrame is a runloop 0x55587f036bb0`

- the init error persists

Issue #1514, opened 2020-01-26
Crash: FL Studio crashes as soon as playing this patch


VincyZed:
**Describe the bug**
Surge makes FL Studio crash when playing this patch:

**To Reproduce**
Here's an example patch with **only the last step needing to be done (step 6)** to make the crash happen:

[SURGE_CRASH.zip](https://github.com/surge-synthesizer/surge/files/4113119/SURGE_CRASH.zip)

Steps made to create this patch:
1. Extend Scene Pitch parameter.
2. Set LFO1 to Envelope mode.
3. Route LFO1 with maximum extended value (168) for the Scene Pitch.
4. In LFO1. set all 6 LFO EG parameters to 0.
5. Route Keytrack to LFO1 Magnitude to the highest value possible (1.00). At this point, you should still be able to play a note without any issues.
6. While spamming a note (I spammed C5 on my MIDI keyboard), slowly increase the Decay parameter of LFO 1. I found that past 0.060 (sometimes a bit more), FL Studio will crash.

**Please let us know your surge version**
Latest nightly, VST3 x64

**Desktop (please complete the following information):**
Windows 10 x64, FL Studio 20.6.1 x64.

**Additional details**
FL Studio will freeze, and a single CPU thread will be maxed out 100%. It will then show the "Stopped Responding" window.

baconpaul:
I look forward to debugging this.

Without running or looking at the code or any other such thing, I bet that we are overrunning the pitch table because the note after modulation is longer and we aren’t checking for an upper bound properly. That is we are effectively playing note 2000 or something and running off the end of the array. But that’s just a guess I”m putting on the github here as a record keeper, in case I’m right, in which case, it will seem super smart. If I’m wrong I promise I won’t edit this comment out. Chuckle.

VincyZed:
:D Reading your guess, that would make sense!

baconpaul:
OK in Hosting AU it doesn't crash but it does go super overloaded click mad when I hit decay above that point. Probably an internal NAN crashes FL. let me find out why.

baconpaul:
Alright I was wrong but I don't quite know what it is yet

Something is making a massive DC offset and pushing the onset of the sound out further, just resulting in a click.

Here's the left and right waveform (I think - this is hacked together in the regrets) at delay 0, delay 0.049 and 0.131 or so. As you can see the saw is riding some sort of big DC offset when it starts which does decay away.

[...]

PR #2008, opened 2020-06-02
Replace a while with an fmod in sin oscillator

baconpaul:
Addresses part of the thread at #1514


baconpaul:
ahh no this is wrong. fmod does the wrong thing on negatives.

PR #2017, opened 2020-06-03
Fix LFO Modulation / Deform bug

### LINKED ISSUE #2016: LFO deform problem, cross LFOing
**Describe the bug**
It is hard to locate what the cause of this problem is but when doing cross LFOing between for example two LFOs and involving deform surge sometimes breaks/stops/clips.

**Please let us know your surge version**
all versions from old to newest.

**To Reproduce**
Just take deform on LFO1 (SEQ 1) to over 50% and surge will break/stop/clip
[surge breaks LFO1 deform 50%.zip](https://github.com/surge-synthesizer/surge/files/4720356/surge.breaks.LFO1.deform.50.zip)

Just take deform on LFO1 (SEQ 1) to over 50% and surge will give a sound that indicates the problem but not break
[surge breaks LFO1 deform 50% 2.zip](https://github.com/surge-synthesizer/surge/files/4720357/surge.breaks.LFO1.deform.50.2.zip)

This one just breaks/stops/clips, directly when played.
[surge broke 2.zip](https://github.com/surge-synthesizer/surge/files/4720359/surge.broke.2.zip)

**Expected behavior**
No break/stop/clip, when doing cross LFOing in interesting ways.

baconpaul:
When LFOs modulate other LFOs especially their rates the phase
can go not only >1 but also >2 and then the -=1 takes too
long to come back down, causing random noise and glitches
(and things like cubic interpolation to 'phase 74' which tives
you modulators like '1e6'. Fix by using fmod in extreme cases.

CLoses #2016

### DIFF
--- a/src/common/dsp/LfoModulationSource.cpp
+++ b/src/common/dsp/LfoModulationSource.cpp
@@ -1,4 +1,5 @@
 #include "LfoModulationSource.h"
+#include <cmath>
 
 using namespace std;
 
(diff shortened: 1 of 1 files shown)


VincyZed:
Fixes #1514

PR #2027, opened 2020-06-04
Implement default author setting

### LINKED ISSUE #1898: Default author
We should be able to set who the default author is when storing patches. Get this in userconfig.xml and refer to that always when showing Store Patch dialog. User can of course override in the dialog, but the field would be autopopulated every time based on what's saved as author in userconfig.

mkruselj:
* closes #1898

### DIFF
--- a/src/common/gui/SurgeGUIEditor.cpp
+++ b/src/common/gui/SurgeGUIEditor.cpp
@@ -3385,12 +3385,20 @@ void SurgeGUIEditor::valueChanged(CControl* control)
    case tag_store:
    {
       patchdata p;
+
       p.name = synth->storage.getPatch().name;
       p.category = synth->storage.getPatch().category;
       p.comments = synth->storage.getPatch().comment;
       p.author = synth->storage.getPatch().author;
+
+      string s = Surge::Storage::getUserDefaultValue(&(this->synth->storage), "defaultPatchAuthor", "User");
+
       if (p.author.empty())
          p.author = synth->storage.defaultname;
+      
+      if (p.author != s)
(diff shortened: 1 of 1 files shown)


baconpaul:
So I think we should make some changes to this feature before we merge it. Not big ones but important ones. My primary concern is you blow away the author of a patch even if no default author is set; and if a default author is set you obliterate the prior author without an option.

Also looking at the code it turns out that surge *already has* a default name and comment option which we should just re-use

So I think the right change here is

1. in SurgeStorage.cpp we load defaultname and defaultsig from the snapshot. That is at line 371 of SurgeStorage call userprefs for default name and defaultsig rather than the e->Attribute, which is left over from when config.xml was writable. You can use (this) as your SurgePointer at that point and will be fine

2. Default the default name to "" not to "User" so we can test if it is blank.

3. remove your get in SurgeGUIEditor and also remove your set of p.author

4. Add a "set default comment" also

At this point you will use the data structure in storage and have it be writable. Cool. If you save with a blank author it overrides with the default and the code is a bit more uniform.

Now the other thing is: what to do if the patch has an author and you have a default. Right now you just clobber independent of the default. That's wrong.

I think we should do the following

1. if the default is "" don't do anything. That requires no code right now
2. if the default is !"" and the patch author is "", use the default. (That will already happen in the code at line 3402
3. If the default is ! "" and the patch author is ! "" what to do. I propose the following
    a. if the author begins with defaultAuthor, do nothing
    b. if the author does not begin with default author make the author 'defaultAuthor (was priorAuthor)'

See what I mean?

Thanks!

PR #2041, opened 2020-06-06
Filters, Ranges in the Sin Oscillator

### LINKED ISSUE #1661: Negative and Extended Feedback on Sine and FM2/3 oscs
Here's a very neat trick Yamaha did with their FM implementation on Reface DX: the feedback parameter can go negative. What this does is feed back the squared output of the operator back to itself, which creates a square wave-like waveform, instead of the usual sawtooth. @baconpaul Maybe that trick could be retrofitted to FM2 and FM3 and Sine oscs in Surge?

### LINKED ISSUE #2039: Add a LP and HP biquad to the sin oscillator
We have two slots. Lets use em!

baconpaul:
1. The Sin Oscillator gets a high cut and low cut
2. The Sin Oscillator feedback range is -1,1 rather than 0,1
   This may impact saved DAW automations of this parameter.
3. Parameters can be activatable
4. High Cut and Low Cut are activatable

Closes #2039
Closes #1661

### DIFF
--- a/src/common/Parameter.cpp
+++ b/src/common/Parameter.cpp
@@ -193,6 +193,7 @@ void Parameter::clear_flags()
    temposync = false;
    extend_range = false;
    absolute = false;
+   deactivated = false;
 }
 
 bool Parameter::can_temposync()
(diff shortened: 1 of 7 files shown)


baconpaul:
oh yeah - gotta remember to stream that don't I!

PR #2066, opened 2020-06-08
Rename LF/HF Damp to LF/HF Damping

### LINKED ISSUE #2065: Reverb 2 - consistent low/high cut filters
Currently Reverb 2 is the only effect that has LF/HF Damp parameters rather than Low/High Cut that delay, chorus, sine osc have. Also the value reported is different (% vs Hz). It'd be nice to have this consistent in Surge.

mkruselj:
* closes #2065


baconpaul:
good solution.

PR #2100, opened 2020-06-11
Fixed scrolling on stepped (int and bool) sliders

### LINKED ISSUE #1931: Scrolling inconsistencies with int sliders and number fields
**Description**
Scrolling is inconsistent on integer sliders and number fields. For instance, when scrolling on Uni Voices, it skips values, and when scrolling on number fields, it repeats values quite often (every 2-3 scroll steps).

**Surge version**
Latest nightly, VST3 x64

**Expected behavior**
Scrolling on integer value sliders and number fields should act like 1 scroll step = 1 increase or decrease of the integer value.

**Screenshots**
Slider wheel bugs

On the above GIF, I scroll through the Poly field consistently, but the result is not consistent (skips very often).
Also, this is easier to see, but Uni Voices skips values.

**Config:**
 - OS: Windows 10 x64
 - Host: FL Studio 20.7 x64

Although this should be an easy fix, I tried and looked in CSurgeSlider.cpp and messing around with range calculations in it also changed the behavior for float sliders (which work correctly). 
I also tried to tweak the range calculations in CNumberField.cpp but I messed things up even more.

VincyZed:
Scrolling on int and bool sliders now works correctly.

Closes #1931.


src/common/gui/CSurgeSlider.cpp
```diff
@@ -791,7 +791,16 @@ bool CSurgeSlider::onWheel(const VSTGUI::CPoint& where, const float &distance, c
       wheelInitiatedEdit = true;
       beginEdit();
    }
-   *edit_value += rate * distance;
+   
+   if (isStepped == true && !(buttons & kControl))
+   {
+      *edit_value += distance / (intRange);
```
baconpaul, line 797: One tiny change. If intRange is 0, on some systems, this can cause an exception which would crash the DAW and the plugin. intRange shouldn't be zero but it's not guaranteed. I would make this `if( intRange && isStepped = true && !(buttons&kControl))`

baconpaul: also isStepped == true can be replaced with just isStepped

VincyZed: Didn't think about that. Good idea.

PR #2115, opened 2020-06-13
Allow OSC Sync to work past 40 semitones

VincyZed:
Fixing the Absolute Mode crash in the oscillator clamped OSC Sync at 40 semitones. This allows it to work past it.


baconpaul:
Thanks!

PR #2135, opened 2020-06-17
Allow real-time phase scrubbing in a static LFO

VincyZed:
This simple change allows real-time phase scrubbing in a LFO if its rate is deactivated (0 Hz).

What still needs to be done is to make it scrub through the Step-Sequencer steps.


baconpaul:
@VincyZed I'll pull this change into a fresh branch and add the SS fix to it but probably not tonight. Great idea!!

VincyZed:
Sounds good! Thank you :)

PR #2138, opened 2020-06-17
Add units to mod amounts, tweak order of certain FX ctrls, added ct_sendlevel

### LINKED ISSUE #1991: Modulation has Units
My old high school physics teacher, when asking a question like "you go 5m/s for 3 seconds how far did you go", if you answered "15" would say "Your mom sent you to the store to buy 3".

Screen Shot 2020-05-30 at 12 54 31 PM

Or in this case, 3.31

I plumb it out, @mkruselj puts in units. It is not obvious at all what the units are in all cases so a default of just a float isn't bad. But maybe we can do better in some cases.

mkruselj:
* ct_amplitude still creates some problems for mod amount units
* after fixing ct_amplitude, this PR closes #1991
* effects which had Width after Mix slider have the order reversed now
* Low/High Cut in ring modulator get their own section
* Width in Reverb 2 is now bipolar dB just like all other widths (but code needs tweaking so that it behaves like width in all other FX)

### DIFF
--- a/src/common/Parameter.cpp
+++ b/src/common/Parameter.cpp
@@ -728,8 +728,14 @@ void Parameter::set_type(int ctrltype)
       valtype = vt_float;
       val_default.f = 0;
       break;
-   default:
+   case ct_sendlevel:
+      val_min.f = 0;
+      val_max.f = 1.5874;
+      valtype = vt_float;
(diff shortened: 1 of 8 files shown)


baconpaul:
All looks good!

PR #2155, opened 2020-06-18
Negative feedback on FM2/FM3 oscs, cleaner code with enums for oscdata

mkruselj:


src/common/dsp/FMOscillator.cpp
```diff
@@ -78,38 +100,54 @@ void FMOscillator::process_block(float pitch, float drift, bool stereo, bool FM,
...
+   oscdata->p[fm_m2amt].val.f = 0.0f;
+   oscdata->p[fm_m2ratio].val.f = 1.0f;
+   oscdata->p[fm_m3amt].val.f = 0.0f;
+   oscdata->p[fm_m3freq].val.f = 0.0f;
+   oscdata->p[fm_fb].val.f = 0.0f;
+}
+
+void FMOscillator::handleStreamingMismatches(int streamingRevision, int currentSynthStreamingRevision)
+{
+   oscdata->p[fm_fb].set_type(ct_osc_feedback);
```
baconpaul, line 134: You need to only do this if streamingReivsion = 12 not in all cases. (New patches shouldn't do this reset).

src/common/dsp/FMOscillator.cpp
```diff
@@ -185,28 +229,36 @@ void FM2Oscillator::process_block(float pitch, float drift, bool stereo, bool FM
...
+   oscdata->p[fm2_m2amt].val.f = 0.0f;
+   oscdata->p[fm2_m2ratio].val.i = 1;
+   oscdata->p[fm2_m12ofs].val.f = 0.0f;
+   oscdata->p[fm2_m12ph].val.f = 0.0f;
+   oscdata->p[fm2_fb].val.f = 0.0f;
+}
+
+void FM2Oscillator::handleStreamingMismatches(int streamingRevision, int currentSynthStreamingRevision)
+{
+   if (streamingRevision <= 12)
```
baconpaul, line 260: Like this! Do this in FM3 also :)

PR #2163, opened 2020-06-20
Glide flex: adds constant rate and glissando modes, 3 curves

### LINKED ISSUE #472: Constant rate mode for portamento
Currently, portamento in Surge always works in constant time, which means that no matter the distance between two notes, portamento time will always be the same.

I am favoring the constant rate portamento instead, which is going to result in longer portamento times the wider the interval between two notes is. I find it more musical while performing.

Would be nice to have this down the road sometime.

mkruselj:
* closes #472
* TODO: streaming and clipboard
* ALSO: fixed audio in init patches having flipped mono/stereo


src/common/dsp/SurgeVoice.cpp
```diff
@@ -330,15 +361,40 @@ void SurgeVoice::uber_release()
...
-   // if(zone->portamento_mode) ratemult = 12/(0.00001+fabs(key-portasrc_key));
+   float const_rate_factor = 1.0;
+   int quantStep = 12;
+   
+   if (!storage->isStandardTuning && storage->currentScale.count > 1)
+      quantStep = storage->currentScale.count;
+
+   // portamento constant rate mode (multiply portamento time with every octave traversed (or scale length in case of microtuning)
+   if (scene->portamento.porta_constrate)
+      const_rate_factor = (1.f / ((1.f / quantStep) * fabs(state.getPitch() - state.portasrc_key))) + 0.00000001;
```
baconpaul, line 372: If this +0.0000001 is supposed to suppress divide by zero then it is in the wrong place in the ()s. Also it is too small for a float. You want 0.00001 more likely.

src/common/dsp/SurgeVoice.cpp
```diff
@@ -200,8 +214,25 @@ void SurgeVoice::legato(int key, int velocity, char detune)
       state.portasrc_key = state.getPitch();
    else
    {
-      state.portasrc_key =
-          ((1 - state.portaphase) * state.portasrc_key + state.portaphase * state.getPitch());
+      // exponential or linear key traversal for the portamento
+      float phase;
+      switch (scene->portamento.porta_exp)
+      {
+      case -1:
```
baconpaul, line 221: We could add an enum in SurgeStorage.h with names like porta_linear, porta_exp and port_log then use those rather than 0,1,-1 consistently

src/common/dsp/SurgeVoice.cpp
```diff
@@ -330,15 +361,40 @@ void SurgeVoice::uber_release()
...
+
    if (state.portaphase < 1)
    {
-      state.pkey = (1.f - state.portaphase) * state.portasrc_key +
-                   (float)state.portaphase * state.getPitch();
+      // exponential or linear key traversal for the portamento
+      float phase;
+      switch (scene->portamento.porta_exp)
+      {
+      case -1:
```
baconpaul, line 383: And similar use the enum here

src/common/dsp/SurgeVoice.cpp
mkruselj: I see it now too, thanks!

src/common/dsp/SurgeVoice.cpp
mkruselj: Indeed - no magic numbers!

PR #2179, opened 2020-06-21
Complete AU Preset layer exposed

### LINKED ISSUE #405: AU Preset Names
currently, for instance, Logic or Live do not show the "currently selected surge preset name" in the daw.

baconpaul:
Onlky factory presets
Better naming
Correct ordering
Closes #405

### DIFF
--- a/src/au/aulayer.cpp
+++ b/src/au/aulayer.cpp
@@ -115,6 +115,23 @@ void aulayer::InitializePlugin()
       // This allows us standalone performance mode. See issue #146 and comment below tagged with
       // issue number
       plugin_instance->time_data.ppqPos = 0;
+
+         sub3_synth* s = (sub3_synth*)plugin_instance;
+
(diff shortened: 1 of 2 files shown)


mkruselj:
Don't you think it should also contain 3rd party presets, since this is also considered a part of Factory?

baconpaul:
So I chose to do factory only because this is the Logic UI

Screen Shot 2020-06-23 at 8 47 42 AM

And this is the surge UI

Screen Shot 2020-06-23 at 8 47 49 AM

If you choose a preset outside of factory it correctly renames the logic track though, which is the desired behavior.

mkruselj:
Gotcha.

This is why I'd like to see 3rd party presets just merged into the factory. I mean they are installed in the factory data folder and everything, yet are treated specially... eh - preset browser will sort that out, someday...

baconpaul:
yeah indeed.

PR #2237, opened 2020-06-29
Add slider.negative.modulation color

### LINKED ISSUE #2147: Show the negative part of modulation range in a different color
Currently both sides of a bipolar modulation amount are colored the same.

image

Here's how ye olde massive did it:

image

It is much easier to see when you've flipped over the modulation amount this way, so it would be great to see this in Surge, too.

mkruselj:
* closes #2147

### DIFF
--- a/src/common/gui/CSurgeSlider.cpp
+++ b/src/common/gui/CSurgeSlider.cpp
@@ -309,8 +309,6 @@ void CSurgeSlider::draw(CDrawContext* dc)
                    << " md=" << moddn << std::endl;
       */
 
-      printf("Modup: %.3f | Moddn: %.3f\n", modup, moddn);
-
       std::vector<CRect> drawThese;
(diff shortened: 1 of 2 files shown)


src/common/gui/CSurgeSlider.cpp
```diff
@@ -302,12 +309,19 @@ void CSurgeSlider::draw(CDrawContext* dc)
                    << " md=" << moddn << std::endl;
       */
 
+      printf("Modup: %.3f | Moddn: %.3f\n", modup, moddn);
```
baconpaul, line 312: we should remove this printf

mkruselj: Oh snap I forgot to remove it - thanks!

baconpaul:
Cool feature!

PR #2308, opened 2020-07-06
Found and fixed more skin issues

itsmedavep:
Main GUI had a VRule that was wrong on the left side of filter configuration
Approached drawing the hover states differently. This was after Paul had reports of errors drawing on Linux.
Update for the R & L buttons buy cutoff and resonance sliders buttons
Fixed ruler tricks on LFO when tempo synced.


baconpaul:
I think your changes to the hovers are going to hurricane out again. I'll pull and look tomorrow.

baconpaul:
Yeah @itsmedavep your two hover changes will override (and be the same as roughly but not work on linux) mine so I'll pull this into a branch and then revert those. May see this PR closed unmerged but it's just because I'm shuffling.

baconpaul:
#2310 swept this in with a few tweaks to bring up to main

PR #2309, opened 2020-07-06
Assortment of minor UI tweaks in classic and dark skins

mkruselj:
* F2 Offset Mode and Resonance Link have improved icons (both skins)
* their positioning is also improved so the slider thumbs don't overlap with them
* Dark: step sequencer has a darker shade of blue outside of loop bounds rather than gray
* Dark: slight tweak on major/minor step colors for step seq
* Width of FX preset name label now related to X position of FX preset jog buttons


baconpaul:
Have a non-conflicted hand merged version. Closing this.

PR #2314, opened 2020-07-08
Fixes a rounding issue with host automation not updating the active/deactivated slider state in VST2

K0rrid0r:
Glad to do my first PR.

Fixes a rounding issue with host automation not updating the active/deactivated slider state properly that still existed in VST2 after commit b8f60a51 was merged.

Automating with for example a square waveform or S&H would not update the slider. It is now fixed and I have double checked several times.

How it was

How it is now after this change.

### DIFF
--- a/src/common/gui/SurgeGUIEditor.cpp
+++ b/src/common/gui/SurgeGUIEditor.cpp
@@ -705,7 +705,7 @@ void SurgeGUIEditor::idle()
                auto sv = synth->getParameter01(j);
                auto cv = cc->getValue();
 
-               if ((sv == cv) && ((tag == fmconfig_tag || tag == filterblock_tag)))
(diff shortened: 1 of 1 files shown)


mkruselj:
Looks fine to me but I'm not sure if changing that condition makes it update UI too often. Paul would know that better.

baconpaul:
Yeah that looks right I just want to check one thing when at my computer

baconpaul:
Oh also toy should add yourself to the authors file and push that on this branch!

baconpaul:
Yeah it used to be sv != cv and when we fixed it we accidentally changed it so the change is good.

We could I suppose check if the state has changed before we do the disablement but the draw is really cheap and only happens on an idle so merging this!

PR #2342, opened 2020-07-12
Rudimentary support for ARM cross-compilation

baconpaul:
Add the appropriate hooks for ARM cross compilation, make them work
with CMAKE, etc..

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -17,3 +17,6 @@
 [submodule "libs/tuning-library"]
 	path = libs/tuning-library
 	url = https://github.com/surge-synthesizer/tuning-library
+[submodule "libs/sse2neon"]
+	path = libs/sse2neon
+	url = https://github.com/DLTcollab/sse2neon
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -358,9 +358,15 @@ elseif( UNIX AND NOT APPLE )
                       COMMAND python ${CMAKE_SOURCE_DIR}/scripts/linux/emit-vector-piggy.py ${CMAKE_SOURCE_DIR} ${CMAKE_BINARY_DIR}/lintemp
                     )
 
+  if( LINUX_ARM_CROSS )
+    set( cxml src/linux/ConfigurationXmlARM.S )
+  else()
+    set( cxml src/linux/ConfigurationXml.S )
+  endif()
+
   set(SURGE_OS_SOURCES
     src/linux/UserInteractionsLinux.cpp
-    src/linux/ConfigurationXml.S
+    ${cxml}
     ${CMAKE_BINARY_DIR}/lintemp/ScalablePiggy.S
     )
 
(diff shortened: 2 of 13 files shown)


baconpaul:
@falkTX @hexdump0815 FYI this merges most of the @falkTX  changes along with some hexdump suggestions, and builds using a cmake toolchain cross compiling on ubuntu. All still a bit rough but when we ship 1.7.0 should be able to allow you to make a KX toolchain file which we can merge back into master (vs the premake-based patch you had before). If you get a native-compile toolchain file and want to mod the CMake accordingly just let me know. But I didn't want that PR ignored when we hit 1.7.0.

Also I activated a NEON/ARM cross compile of headless in our pipelines so ARM should build reliably moving forward.

falkTX:
Thanks for the merge, but one thing that would be nice to have is a way to specify ARM target without enforcing the compiler flags.
Right now you added `-march=armv7-a -marm -mfpu=neon` which will be there unconditionally. This is really suboptimal, as:
1. raspberry pi 3 is not `-march=armv7-a` but `-march=armv7ve`
2. aarch64 does not have FPU options, so this will fail to build anything
3. -marm is only relevant when selecting between arm or thumb modes, so it does not do much here

My personal recommendation is to simply not set any flags at all when it comes to ARM.
Wherever building for armhf (32bit with neon) or aarch64, the arguments are just not compatible.
Plus there are far too many devices out which use an always-evolving ARM architecture, setting such flags is better left to the one building the project.

For reference, the flags I use in the kxstudio repositories are `-march=armv7ve -mcpu=cortex-a7 -mfloat-abi=hard -mfpu=neon-vfpv4` for armhf and `-march=armv8-a -mcpu=cortex-a53`.
This is the minimal spec we can do reasonably, if a system cannot run binaries built with these flags, it is likely going to have completely terrible performance.

hexdump0815:
@baconpaul cool to see native arm support being on the list for surge 1.7.0

@falkTX according to the gcc doc armv7ve is simply armv7-a plus virtualization extensions which we definitely do not need for surge, so armv7-a should be more generic. did you ever actually successfully run any surge build with any kind of -mfpu=neon* enabled? for me anything with neon set crashed with unaligned access (ubuntu 18-04/gcc7) so that i had to got back to "-mfpu=vfpv3-d16 -mno-unaligned-access"

falkTX:
> @falkTX according to the gcc doc armv7ve is simply armv7-a plus virtualization extensions which we definitely do not need for surge, so armv7-a should be more generic

True, but my point on specifying armv7ve is that if you mix armv7-a with -mnative or a few others (dont recall now), you will get compiler warnings about incompatible flags.

> did you ever actually successfully run any surge build with any kind of -mfpu=neon* enabled? for me anything with neon set crashed with unaligned access (ubuntu 18-04/gcc7) so that i had to got back to "-mfpu=vfpv3-d16 -mno-unaligned-access"

I never got surge to work on armhf, only aarch64.
Let me try with `-mfpu=vfpv3-d16 -mno-unaligned-access`, thanks!

hexdump0815:
i think -mnative should never really be used in a generic build as it introduces a direct dependency to the build machine (i.e. it can generate different options for different build machines)

baconpaul:
Yeah cool! So I don't have very strong feelings or (really) any knowledge about this! I am super duper happy if you guys figure out what to pull for the right flags and arch. I just didn't want to not have a merge for 1.7.0. (Also I'm thinking of dec when I get an ARM MacBook and want to get surge running too....)

But now that I have it crudely isolated (we have toolchains, we are in the pipeline, and we are in cmake), I am super duper open to merge changes to the ARM flags. Like I said my primary goal is: when falk puts surge in KX at 170 that we don't need to apply an ARM patch and when I finally pull surge rack to the 170 surge underlier that hexdump can avoid massive patches to get arm rack building. And the mac thing...

So what's the right thing? All ears!

And of course, thanks for your help and collaboration.

baconpaul:
Oh and thinking: The "right" way to do this is to have a cmake toolchain file for each version we want. So I should have pushed the -mneon or whatever into the cross compile cmake file and factored accordingly. Then adding a new flag flavor is simply adding a new toolchain file.

Lemme do that tonight or tomorrow and refactor the cmake a bit. Then we can talk about other toolchain configurations we would want. Sound good?

falkTX:
> i think "-mnative" should never really be used in a generic build as it introduces a direct dependency to the build machine (i.e. it can generate different options for different build machines)

Of course. I know this. I mean if someone builds surge and adds these flags, there will be warnings. They can be very confusing. When you are building for yourself, it is normal to use the native option, specially useful on ARM devices as they are not that powerful usually.

I still think the best option for now is to not have any ARM related flags in the cmake project.
We can specify the flag ourselves using env vars, manually. This can be done for the CI/CD builds

baconpaul:
Right

So what I’ll do is factor this all into the toolchain file and make it so it compiles in my cross compile env. Then we can easily add and use (and commit to master) other toolchains as appropriate 

Ill grab a pi4 on Amazon one day and see how much I can get rendering properly - for now I just want the patch size you both need to build to be substantially minimized!

And thank is again for all the info. Will update here when the slight refactor is done. Maybe a day or two

PR #2364, opened 2020-07-22
Right Mouse Button hung jogfx button

baconpaul:
Make the RMB on jogfx return 1.


baconpaul:
need store also. Got it now. Thanks!

PR #2393, opened 2020-07-29
Re-introduce the LV2 idlefactor

### LINKED ISSUE #2391: LV2 uiextension
Hey @falkTX and @jpcima 

when building for clang 9 over in #2386 we noticed that

```
/usr/ports/audio/surge-synthesizer-lv2/work/surge-release_1.7.0/src/lv2/SurgeLv2Export.cpp:158:15: error: expression result unused [-Werror,-Wunused-value]
              "    lv2:extensionData ui:idleInterface .\n";
              ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
```

this constant was not in the string because of the ; above. Not wanting to change the manifest I just commented it out. But I wonder if we need this at that line; and also if that could be causing #1975 so wanted to make it a separate issue since we fixed the clang problem over in that other issue.

baconpaul:
Closes #2391


jpcima:
It's good yes, it passes the lint check.

PR #2416, opened 2020-08-02
Added 24dB/oct subtype for band reject (notch) filter

mkruselj:


src/common/dsp/QuadFilterUnit.cpp
```diff
@@ -751,7 +751,10 @@ FilterUnitQFPtr GetQFPtrFilterUnit(int type, int subtype)
       }
    }
    case fut_br12:
-      return IIR12Bquad;
+      if (subtype < st_BR24)
```
baconpaul, line 754: So this is a bit nitpick but since (for good reason) you are re-using values in the enum (that is, st_BR12Legacy is a value alias for st_ROUGH or whatever) the enums are no longer well ordered. This cast to int will work but it is idiomatically odd since usually enums don't do that so you read this as 'if subtype is any of the enums before st_BR24 in enum space' and the compiler thinks 'if subtype cast to int is any of the enums with value below...

The code isn't wrong but I would much prefer if this line said

```
if( subtype == st_BR12 || subtype == st_BR12_Legacy )
```

Also the display name is "Mild" and the enum name is "leg" and I think "leg" means "Legacy"? Can we pick a consistent name for the enum value?

Super glad this works! Cool stuff.

PR #2434, opened 2020-08-05
Cursor hiding rework

mkruselj:


mkruselj:
This is WIP because the issue is hard and there are too many weird edge cases where cursor doesn't reappear where it should.

baconpaul:
Ahh I kinda wanted this branch to stick around. Can you restore it and add it to the issue?

mkruselj:
> Ahh I kinda wanted this branch to stick around. Can you restore it and add it to the issue?

Oops, that was an unintended delete! Restored.

Issue #2435, opened 2020-08-06
Typein width when high-precision value readouts are enabled


mkruselj:
...needs tweaking:

![image](https://user-images.githubusercontent.com/2393720/89536302-faa17280-d7f7-11ea-9509-6f423a20dec0.png)

baconpaul:
As I mentioned on slack

`Parameter::get_display_of_modulation_depth` checks the precision thing but then doesn't use it in the printfs!

mkruselj:
Sure - but this is more about the _width_ of the typein dialog and those values being clipped at edges.

mkruselj:
Also from ENV1 on KvR: https://www.kvraudio.com/forum/viewtopic.php?p=7850910#p7850910

baconpaul:
Yeah - and things like units of semitones can make it overrun.

I'm not sure we want it wider.
Perhaps we want it stacked.

Lemme noodle.

baconpaul:
<img width="246" alt="Screen Shot 2020-08-06 at 10 24 29 PM" src="https://user-images.githubusercontent.com/13387561/89601960-b59b3180-d833-11ea-9876-c07bb0b9e9e9.png">

Branch `baconpaul/stack-mod-ed-2435`

play with it a bit? I think it's the right answer. If you agree lets merge it and close this.

mkruselj:
That looks great and better than it is now. Merge is in order :)


BTW is it like that for all params or just the long ones?

baconpaul:
All modulators get the extra lines; assignments do not. Modulators typeins are already better because they get the "from" line and stuff.
Will merge! Thanks!

PR #2444, opened 2020-08-07
Stack the modulation edit display to avoid overflows

baconpaul:
Addresses #2435


baconpaul:
Sitting this here but it'd be good if @mkruselj agrees with me that it's the right answer before we press merge. Just saving myself some time in case he does!

PR #2456, opened 2020-08-08
Fix bug with notes clicking, convert frequency constants in code to hz

amyfurniss:


baconpaul:
OK since I merged the reposition change this has some small conflicts. Also just in case you missed it the question i asked on slack is: Have you confirmed that the new omega calculation gives the same result as 1.7.1 does when stages = 4? That is, the base case is unchanged even though the calculation is different?

Great stuff here!

PR #2480, opened 2020-08-14
Extend tempo sync all parameters to FEG and AEG

### LINKED ISSUE #2037: Add Tempo Sync for all Filter and Amp EG controls (where logically applicable)
Filter and Amp EG | Tempo Sync

In the LFO/SEQ section, one can right-click on a control and select from either Tempo Sync, or Enable Tempo Sync For All…

How complicated would it be to add the latter option to the Filter and Amp EGs, where it could be logically applied (obviously not on Sustain stage for example)? 

Enable Tempo Sync For All Controls?

mkruselj:
Closes #2037

### DIFF
--- a/src/common/gui/SurgeGUIEditor.cpp
+++ b/src/common/gui/SurgeGUIEditor.cpp
@@ -3262,38 +3262,19 @@ int32_t SurgeGUIEditor::controlModifierClicked(CControl* control, CButtonState b
 
                   int a = p->ctrlgroup_entry;
 
-                  switch (a)
+                  if (p->ctrlgroup == cg_ENV)
                   {
-                  case 0:
-                  {
-                     sprintf(prefix, "Amp EG");
-                     break;
+                     if (a == 0)
+                        sprintf(prefix, "Amp EG");
+                     else
+                        sprintf(prefix, "Filter EG");
                   }
-                  case 1:
+                  else if (p->ctrlgroup == cg_LFO)
                   {
-                     sprintf(prefix, "Filter EG");
-                     break;
-                  }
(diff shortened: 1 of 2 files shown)


mkruselj:
This commit closes #2037

src/common/gui/SurgeGUIEditor.cpp
```diff
@@ -3254,31 +3254,68 @@ int32_t SurgeGUIEditor::controlModifierClicked(CControl* control, CButtonState b
...
-                  if (a > 6)
-                     sprintf(prefix, "SLFO %i", a - 6);
-                  else
-                     sprintf(prefix, "LFO %i", a);
+                  int a = p->ctrlgroup_entry;
+
+                  switch (a)
+                  {
+                  case 0:
+                  {
```
baconpaul, line 3268: Thisn0 and 1 should be the modulator name not the content 0:or 1, no?

baconpaul: Oh no it just needs a comment. Got it.

baconpaul: OK so stylistically doing this as a switch where we mix the different entries of different control groups is what got me.

I think this code should look like

```
if( ctrglr == LFO )
   if( entry > 6 )
      SLFO
   else
     VLFO
elseif( ctrlgrp == ENV )
   if( entry == 0 )
     AEG
   else
     FEG
```

in pseudocode. Switching over those values is dangerous (what if we re-define ms_lfo1 to 0 - this would break) since they have different meanings in different context.

PR #2486, opened 2020-08-16
Reworked About page and Menu button with new logo

mkruselj:
Finetuned the background rectangles in CAboutBox

image

image


mkruselj:
image

image

Issue #2490, opened 2020-08-17
Missing link to fontconfig / Fails to build


falkTX:
Note sure if it is surge directly, or me by using `-Wl,--no-undefined` linker flag, but surge is failing at the linking stage here with:
```
[ 23%] Linking CXX shared library libsurge-lv2-dll.so
/usr/bin/cmake -E cmake_link_script CMakeFiles/surge-lv2-dll.dir/link.txt --verbose=1
[...]

PR #2492, opened 2020-08-17
Detect missing symbols; Link against fontconfig and x11 on linux

falkTX:
A fix for #2490, while also detecting future issues with missing symbols.


baconpaul:
Thank you! Merged.

PR #2558, opened 2020-09-02
Depend on xcb-keysyms

orivej:
xcb-keysyms are used in vstgui:
https://github.com/surge-synthesizer/vstgui/blob/vstgui4_7/vstgui/lib/platform/linux/x11platform.cpp#L23


baconpaul:
Thanks!

PR #2561, opened 2020-09-02
Implement multi solo (and exclusive mute)

### LINKED ISSUE #950: Implement multi-solo
Right now solo is single-solo per scene.

It would be great to have multi solo. 

With #891 the UI matches the engine, which is good, but as @sense-amr points out, multi-solo would be way better.

Bit tricky to get it right down in the SurgeVoice land and stuff so will fix the UI and then do this later; perhaps for 1.6.2 perhaps for 1.6.n (but I will tag it 1.6.2 now).

mkruselj:
Closes #950

### DIFF
--- a/src/common/gui/SurgeGUIEditor.cpp
+++ b/src/common/gui/SurgeGUIEditor.cpp
@@ -2422,12 +2422,14 @@ int32_t SurgeGUIEditor::controlModifierClicked(CControl* control, CButtonState b
       return 0;
    if (!editor_open)
       return 0;
+
    /*if((button&kRButton)&&modsource)
      {
      modsource = 0;
      queue_refresh = true;
      return 1;
      }*/
+
    if (button & (kMButton | kButton4 | kButton5))
    {
       toggle_mod_editing();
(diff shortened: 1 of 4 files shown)


src/common/gui/SurgeGUIEditor.cpp
```diff
@@ -3701,20 +3701,22 @@ int32_t SurgeGUIEditor::controlModifierClicked(CControl* control, CButtonState b
             }
          } // end vt_float if statement
 
-#if TARGET_VST3
-         auto hostMenu = addVst3MenuForParams(contextMenu, ptag, eid );
-#endif
+         #if TARGET_VST3
```
baconpaul, line 3704: Compiler directives shouldn't be tabbed like this. While technically #if can be anywhere, idiomatically almost every code base in the world keeps them at column 0 since they are compiler directives not code flow directives and don't match the same balance as the code. Can you put it back the way it was where you changed it?

src/common/gui/SurgeGUIEditor.cpp
```diff
@@ -3751,6 +3753,47 @@ int32_t SurgeGUIEditor::controlModifierClicked(CControl* control, CButtonState b
...
+               if (curr->id == p->id)
+                  curr->val.b = true;
+               else
+                  curr->val.b = false;
+
+               curr++;
+            }
+         }
+
+         synth->refresh_editor = true;
```
baconpaul, line 3795: We only need to do this refresh if our ctrltype is mute or solo. Can you move it inside the if blocks?

PR #2569, opened 2020-09-03
Make mousewheel click threshold more correct in windows

K0rrid0r:
Instead of three clicks to change step in a menu, now only one click is required.


baconpaul:
I have a more comprehensive one with wrapping and a few other fixes so I'm going to close this. Thanks.

K0rrid0r:
ok cool!

PR #2611, opened 2020-09-11
Make WT export path properly cross-platform

mkruselj:
Slightly adjust vertical position of text in filter type menu


baconpaul:
Looks good. Will merge if ci works

PR #2622, opened 2020-09-15
SurgeSynthesizer: Fix iterator usage

mvf:
Comparing iterators from different lists caused Windows debug builds to
crash on startup with a "list iterators incompatible" assertion failure.


baconpaul:
Thank you!
I'll merge both these, but if you have another coming, can you add yourself alphabetically to the AUTHORS file in a way which you would find appropriate?

mvf:
Thanks, will do!

PR #2626, opened 2020-09-16
Fix non-Release builds

mvf:
This fixes the "other" builds, e.g. Debug, RelWithDebInfo, MinSizeRel. Please see the commit message for details.

This change affects all platforms, so I tested both Release and Debug builds (open Surge editor, pick a few patches, play a bit) in the following environments:

**Linux 64-bit:** VST3, LV2 (Qtractor)
**Windows 64-bit:** VST3 (Bitwig Studio)
**macOS Catalina:** VST3 (Bitwig Studio), AU (GarageBand)

From what I've seen, Debug cuts build times in half on Windows. In my limited testing across all platforms, Debug builds seemed to render roughly 3-5x slower, which should still be fast enough for many development tasks.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,6 +8,10 @@
 cmake_minimum_required(VERSION 3.10)
 set(CMAKE_OSX_DEPLOYMENT_TARGET 10.9 CACHE STRING "Build for 10.9")
 
+if(NOT CMAKE_BUILD_TYPE)
+  set(CMAKE_BUILD_TYPE "Release" CACHE STRING "" FORCE)
+endif()
+
 project(Surge VERSION 1.7.1.0 LANGUAGES CXX ASM)
 message( STATUS "CMake Version is ${CMAKE_VERSION}" )
 message( STATUS "Compiler Version is ${CMAKE_CXX_COMPILER_VERSION}" )
@@ -43,9 +47,6 @@ add_custom_target( git-info BYPRODUCTS ${CMAKE_BINARY_DIR}/geninclude/version.cp
                            -P ${CMAKE_SOURCE_DIR}/cmake/versiontools.cmake
                            )
 add_dependencies( surge-shared git-info )
-
-
-set(CMAKE_BUILD_TYPE Release)
 
 # Set up external packages
 list(APPEND CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/cmake/")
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -510,7 +511,6 @@ elseif( UNIX AND NOT APPLE )
   set(OS_COMPILE_DEFINITIONS
     ${ARCH_COMPILE_DEFINITIONS}
     LINUX=1
-    RELEASE=1
```
baconpaul, line 513: Curious: Doesn't this need to be replaced with something like

```
 $<IF:$<CONFIG:DEBUG>,DEVELOPMENT,RELEASE>=1
```

as below?

baconpaul:
@mvf thank you! I am rerunning the failed job. It seems to be unrelated to your change (the unit tests are occasionally tripped over in their random numbers).

I did have one comment above. 

Thank you very much for this change. It's greatly appreciated. I actually ended up learning CMake as we went here (since we needed to get off of premake) and it shows... thank you for the changes.

baconpaul:
Oooh also maybe we should turn on debut builds for ci to speed up the ci checks! If you’ve edited azure yaml it should be clear but if not I’ll do it after this merges once I hear back on above

Thanks!

CMakeLists.txt
mvf: The `$<IF>` below applies on all platforms, so it doesn't need to be repeated in the `OS_COMPILE_DEFINITIONS` for Linux.

On a related note though, on Windows I had somehow added a global compile switch. Moved into `OS_COMPILE_DEFINITIONS` now for consistency.

mvf:
Glad if this helps, I'm not a CMake pro either. The syntax is ~annoying~surprising and it's hard to find up-to-date examples, but it's still the best option for C++ projects IMHO.

> [...] maybe we should turn on debug builds for ci to speed up the ci checks

I'm not sure Debug is the best fit for CI builds. It could miss issues that only occur with optimization. Now, the reverse is also true of course, e.g. with the Windows debug runtime. But if I had to choose, I'd test what we ship, i.e. Release.

I'll look into the Azure yaml. Maybe builds can be sped up even without switching to Debug.

baconpaul:
> The syntax is annoyingsurprising and it's hard to find up-to-date examples, but it's still the best option for C++ projects IMHO.

I 100% agree with both of these statements. It's great, but it's also kinda odd.

Anyway merged. Thanks!

PR #2631, opened 2020-09-17
Change mixer default state: all mutes disabled, volumes to -inf instead

mkruselj:
Update factory Templates patches with the same change


baconpaul:
Oooh lets talk about this one. Level -inf is not the same as mute in the codepath.

baconpaul:
Critically this change means we will run all 3 oscillators plus RM and noise on each voice and 0 multiply them, as opposed to just run the unmuted oscillators. I don't think it's the right change.

mkruselj:
Ha we've talked on Discord about it. I think for template patches and init state it's fine, ultimately. But we can discuss further if you want!

baconpaul:
So I ran this in a profiler

It uses 30% more CPU and if only O2 and 3 are on at -inf and 50% more when you add the RM and Noise. Even though that total amount of CPU on my system is tiny, it's a real impact.

I would sooner not merge this. Going to close it. Happy to discuss more but we have lots of low-cpu usages out there so lets have the starting points be efficient.

mkruselj:
🙁

PR #2674, opened 2020-09-23
New modulation routing layout, new modulators

mkruselj:
Implemented necessary adjustments to classic and dark skins.

Currently, breath controller, expression and sustain pedal modulators are for sure working, I am not sure if monophonic keytracks are working, I think I missed something. Random and Alternate still need to be implemented (will likely need code hints) 🙂 

Also question for @baconpaul: line 167 in SurgeVoice.cpp, do I need to have these additional assignments from `oscene->modsources` for monophonic keytracks, too? I put these here for consistency but I'm not even sure if that is defined in oscene. Also I noticed (line 154) that `ms_keytrack` rather uses `&keytrackSource`, I assume because it's per voice. But for monophonic keytracks it's likely supposed to indeed assign from `oscene->modsources`, right? Not quite understanding this part of code...


mkruselj:
Yeah monophonic keytracks are quite broken...

PR #2685, opened 2020-09-24
Add Clear MIDI settings option to menu, slight macro context menu renames

mkruselj:


baconpaul:
OK a couple of things here

1. You should *not* be updating the dawExtraState in this loop. I guided you to that code so you could see how the data structures work, but it is updated by the DAW streamers when needed, not inline in the synth. So you should remove this references

2. The reason that scene b stuff doesn't work is because the midi ctrl spans both side. Look at this code from SurgeSynthesizer.cpp

```
 if (learn_param >= 0)
   {
      if (learn_param < n_global_params)
      {
         storage.getPatch().param_ptr[learn_param]->midictrl = cc_encoded;
      }
      else
      {
         int a = learn_param;
         if (learn_param >= (n_global_params + n_scene_params))
            a -= n_scene_params;

         storage.getPatch().param_ptr[a]->midictrl = cc_encoded;
         storage.getPatch().param_ptr[a + n_scene_params]->midictrl = cc_encoded;
      }
      storage.save_midi_controllers();
      learn_param = -1;
   }
```

so your loop in the midi callback is wrong in two ways

1. It adds n_scene_params to everything, and only needs to add it to the non-global params. 
2: it updates the daw information where it doesn't need to.

```
           int n = n_global_params + n_scene_params;

           for (int i = 0; i < n; i++)
           {
              this->synth->storage.getPatch().param_ptr[i]->midictrl = -1;
              if( i > n_global_params )
                  this->synth->storage.getPatch().param_ptr[i + n_scene_params]->midictrl = -1;
           }
```

baconpaul:
Looks great. Will merge once CI passes.

PR #2705, opened 2020-09-26
Redo miniedit in VSTGUI

mkruselj:
Add new skin colors
Reduce bmp00170 to just Surge logo


baconpaul:
While it's not required for the code to work, if you are renaming a resource you should rename the windows stub ids for it also. In src/windows that's:

```
scalableresource.h:#define IDB_MINIEDIT_BG_SCALE_SVG 80170
svgresources.rc:IDB_MINIEDIT_BG_SCALE_SVG DATA "assets/SurgeClassic/exported/bmp00170.svg"
```

Those names are actually arbitrary. The thing that matters is "80170" (which is the SVG version of 170) maps to a resource id with the file. But I've been trying to keep them consistent lately by using a script.

Happy to merge otherwise.

mkruselj:
I would have done that, but VS didn't catch those references when renaming solution-wise! Might be I had "whole word only" enabled... Anyways fixing :)

PR #2707, opened 2020-09-26
Windows/MSVC: Trivial cleanups

mvf:
- Fix unused variable, unreachable code warnings
- Remove WIN_X86 define


baconpaul:
So that WIN_X86 was there to get around the PLUGIN_API problem on 32 bit VST3; PLUGIN_API defines to __stdcall on windows. Are we sure we don't need it in 64 bit? Just confirming that you've tested and understood that change and that the vst3 works both 32 and 64 bit win before I merge (since I didn't understand it when I got 32 bit working 2 years ago and still really don't know; but it did work).

Thanks

baconpaul:
Changes all look great otherwise and am happy to merge. Thank you very much!

mvf:
Yes, the `PLUGIN_API` thing works on both 32 and 64 bits. The problem was: These virtual functions were declared `PLUGIN_API`/`__stdcall` in `SurgeVst3Processor` but not in the base class. Virtual calls via the base class's `__thiscall` calling convention would crash when hitting our `__stdcall` overrides. It worked on 64-bit because there `__stdcall` has no effect.

baconpaul:
Awesome! Thanks for the explanation. As the comment showed when I “solved” it I didn’t have that patience or insight :)

Merging now

PR #2715, opened 2020-09-28
Rename vintage ladders

mkruselj:
Make char array inits more consistent in SurgeStorage.h
Rename sub3 enums


baconpaul:
This is a bigger change than Rename Vintage Ladders. It looks OK but just want to check before I merge?

Also: I thought we decided on "Tamed" not "Mild"?

PR #2716, opened 2020-09-28
Global highpass deactivatable

mkruselj:


baconpaul:
baconpaulToday at 8:24 AM
before I merge
It should work like I said
but "save"
then new surge and "load"
the same surge and "load old"
EvilDragonToday at 8:25 AM
oh it doesn't work
so if I deactivate it and load old patch it stays deactivated
baconpaulToday at 8:25 AM
OK!
EvilDragonToday at 8:25 AM
nice catch
do I need handlestreamingmismatch
baconpaulToday at 8:25 AM
i am not sure
there's no mismatch interceptor for that control right now
i need to look at the code

PR #2724, opened 2020-09-29
Windows: Switch to Unicode

### LINKED ISSUE #2579: Crash when loading "µcomputer" patch on Japanese Windows 10
**TL;DR:** Uncaught C++ exception because Surge doesn't support Unicode paths on Windows.

Related to #1699

**Description**
Trying to load the "Lead"→"µcomputer" patch on a Japanese Windows 10 crashes Surge.

* Surge Version: 1.7.main.502ec53a
* Plugin type: VST3
* Bits: 64

**To Reproduce**
1. Make sure your ANSI code page does not have the "µ" character
2. In the patch browser menu, click "Lead"→"µcomputer"
3. Observed: Crash

**Expected behavior**
The "µcomputer" patch gets loaded.

**Desktop:**
 - OS: Windows 10 Professional 64-Bit, Japanese
 - Host: VST3PluginTestHost from the VST3 SDK
 - Version: 2.9.10 Build 39

**Additional context**
The immediate cause of the crash is an uncaught `std::system_error` from `std::filesystem::path::generic_string()` called here:

```c++
void SurgeSynthesizer::loadPatch(int id)
{
   ...
   loadPatchByPath( e.path.generic_string().c_str(), e.category, e.name.c_str() ); // generic_string() throws
}
```

The exception is thrown because code page 932 on a Japanese Windows doesn't have the "µ" character. Screenshot of a debug session at the relevant `WideCharToMultiByte()` call:

surge-cropped
`_Code_page` is `0`, aka `CP_ACP`, i.e. the system default ANSI code page. `_Used_default_char` being `1` means that not all characters were representable in this code page. ([WideCharToMultiByte Docs](https://docs.microsoft.com/en-us/windows/win32/api/stringapiset/nf-stringapiset-widechartomultibyte))

`std::filesystem::path::generic_string()` seems to be used in multiple places, so this might be a broader issue. On Western versions of Windows, this should be reproducible with paths containing CJK characters.

mvf:
This makes Windows builds Unicode-aware, most notably unlocking the full
range of NTFS file system paths.

To avoid #ifdef, paths in std::strings are now UTF-8, like on other
platforms. They are converted to and from the native UTF-16 as needed.
This is mostly transparent, except when converting between fs::path and
std::string. There, the following should be observed:

- Use path_to_string() instead of .generic_path()
- Don't pass strings into fs::path API, convert them to paths first via
  string_to_path()

Otherwise, on Windows and if the path contains chars other than 7-bit
ASCII, either a system_error will be thrown, or (less likely) the path
could be mistranscoded.

This should become less and less of an issue as fs::path <-> std::string
conversions are refactored away.

Fixes #2579

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -378,13 +378,10 @@ set(SURGE_GUI_INCLUDES
 
 ### OS SPECIFIC SECTION
 if( APPLE )
-  set(SURGE_OS_SOURCES
-    src/mac/UserInteractionsMac.mm
-    )
-
   set(SURGE_OS_GUI_SOURCES
     src/mac/DisplayInfoMac.mm
     src/mac/RuntimeFontMac.cpp
+    src/mac/UserInteractionsMac.mm
     src/mac/cocoa_utils.mm
     vstgui.surge/vstgui/vstgui_mac.mm
     vstgui.surge/vstgui/vstgui_uidescription_mac.mm
@@ -448,12 +445,6 @@ elseif( UNIX AND NOT APPLE )
     set( cxml src/linux/ConfigurationXml.S )
   endif()
 
-  set(SURGE_OS_SOURCES
-    src/linux/UserInteractionsLinux.cpp
-    ${cxml}
-    ${CMAKE_BINARY_DIR}/lintemp/ScalablePiggy.S
-    )
-
   file(GLOB LINUX_VSTGUI_GLOB vstgui.surge/vstgui/lib/platform/linux/*.cpp )
 
   # Our friends at steinberg have this maddening habig of using #warning DEPRECATED in linux code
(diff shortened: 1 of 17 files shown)


mvf:
Right, so I only just now realized that Rack builds probably need to remain ANSI. Will reopen once MinGW builds are fixed.

mvf:
Turns out Rack uses its own Makefile which won't link the Unicode parts, so I guess we're fine after all. Sorry for the noise.

baconpaul:
Just bringing my comments from discord over here in case that's more convenient:

----
so I just reviewed it all
I really have only one comment which is about the CMakeFile
You moved ${cxml} out of the OS_SOURCES into the OS_GUI_SOURCES on linux; and you did the same for the various UserInteractions
The ${cxml} is needed for lots of non-OS functions so it should be in OS_SOURCES. I want to unwind ${cxml} one day - it was a mistake at the outset of the linux port - but for now it needs to stay in LINUX_OS_SOURCES
The UserINteractions I understand why you did it. I was worried headless wouldn't build! I think some of the juce submodules which aren't on by default will break but we should use that by refactoring the targets properly at the time so I'm OK with it
But I would move the ${cxml} back.
And also, and this is minor and you don't need to fix it
fs::path foo(bar)

looks right to me as does
type name(init)

but I've idiomatically converted that to
auto name = init

even though both work. NO need to change that but as I was reading it made me hiccup.
Finally: Just want to confirm you've tested those functions on windows. I was going to merge without building and testing myself since the review looks good to me. But just wanted to be sure.
The functions especially are: open files, and show HTML.
SO, like, you've loaded an SCL file into the plug and its worked
Finally finally: Thanks so much. This is a huge amount of work and a massive improvement. Wonderful!!
---

baconpaul:
Actually I just added #2729 which will solve the CXML thing so we can leave it for now and get it with that.

Issue #1249, opened 2019-10-15
CritSection runaway with windows.wt missing on linux


baconpaul:
As reported by @codecollider in #1247 if you run surge on linux with the shared assets missing, you get a runaway in the critical section recursion. That's interesting and we should debug it one day, even though the proximate problem in #1247 was solved with better error reporting.

mkruselj:
@mvf since CriticalSection is now removed and replaced with std::mutex, is this issue still relevant?

baconpaul:
It is. In fact if this happens we will deadlock instead of run so we need to understand why it happened before 18 or change the mutex to recursive which sucks

mvf:
Yeah, would love to get to the bottom of this. Spent quite some time already trying to reproduce this with the current version by deleting `windows.wt` with 8f4e694f697871f8fb6896e6aad90d7aaea7700a reverted, then trying to load "Window" into the oscillator. Couldn't get it to deadlock/recurse yet but will poke around some more.

If we can't figure this out, this old issue alone wouldn't be reason enough for me personally to consider recursive mutex. What was once a bad code path should now be barred, and to the user both an endless loop and a deadlock manifest as a hang. So even if this strikes again it's not really a regression IMHO.

baconpaul:
I agree if we can't repro it we should go ahead
I also agree we should poke at it some more
I also agree we have time
So lets leave this open and we can poke at it some
:)

mkruselj:
Here's a thought. Why don't we actually subsume windows.wt into the binary itself? Then it's always gonna be available, no way to deadlock in that case?

baconpaul:
while that would solve the proximate problem
i want to understand why it recursed
in case it would in other failure cases

baconpaul:
So I just tried for another half hour to make this happen again
I am unable to
I conclude that fixes we put in other parts of the codes to pthis runaway which was hard to reproduce anyway
So I'm going to close it.

PR #2765, opened 2020-10-02
Remove CriticalSection, replace with std::mutex

mvf:
Didn't seem to be used recursively. If it is after all, that's a bug
that needs fixing. If it can't be fixed right away, there's always
std::recursive_mutex to restore the old behavior.


baconpaul:
Oooh would have been good to chat since there are recursive uses and I’ve never trapped them.

https://github.com/surge-synthesizer/surge/issues/1249 Gives you one example

What happens now if we do recourse? Deadlock?

src/common/gui/SurgeGUIEditor.cpp
```diff
@@ -51,6 +51,10 @@
 #include <stack>
 #include <unordered_map>
 
+#if WINDOWS
+#include <windows.h>
+#endif
```
baconpaul, line 56: Curious why?

mvf: This is for the definitions of `SYSTEMTIME`/`GetSystemTime`. Before, `windows.h` was included implicitly via `SurgeStorage.h`→`CriticalSection.h`.

baconpaul: Thanks!
We should move that to std::chrono one day

mvf:
> What happens now if we do recurse? Deadlock?

Yes, which is hopefully easy to debug. Recursive mutexes are a hack and especially bad in audio applications. It seems that the current known recursive use it not intentional though. I'll look into that.

src/common/gui/SurgeGUIEditor.cpp
baconpaul: we actually didn't need the time at all so just cleaned it up and pushed

PR #2771, opened 2020-10-03
CMake: Fix build when Git is missing or failing

### LINKED ISSUE #2681: Fails w/out git
When git returns an error code configure prints this:

```
CMake Error at cmake/versiontools.cmake:66 (if):
  if given arguments:

    "STREQUAL" "main"

  Unknown arguments specified
```

Both situations when git is missing and when git fails should be handled gracefully.

mvf:
Fixes #2681


baconpaul:
Thank you!

PR #2781, opened 2020-10-05
macOS, Linux: Move _aligned_malloc/free defines to globals.h

mvf:
OS-specific defines don't belong in compile options and CMake doesn't
support function-style defines in compile definitions.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -392,8 +392,6 @@ if( APPLE )
   set(OS_COMPILE_OPTIONS
     -fasm-blocks
     -msse2
-    "-D_aligned_malloc(x,a)=malloc(x)"
-    "-D_aligned_free(x)=free(x)"
     -Wno-deprecated-declarations
     -Werror=inconsistent-missing-override
     -Werror=logical-op-parentheses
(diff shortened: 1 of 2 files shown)


baconpaul:
Smart. Should have done this forever ago
I think only the quad filter unit uses this so one day we could even get rid of it altogether and just go with alignas directives i think.

mvf:
Thanks, not sure if I'd call this "smart", but it's a prerequisite for what I'm planning next... :suspect:

baconpaul:
ooooh ... do tell!

mvf:
Basically disentangle compiler and OS flags so the codebase can be organized better. Still fiddling with it though. Will definitely get your opinion first!

baconpaul:
Exciting

I don't know if you saw on discord, but I'm in the middle of a pretty major restructure of the UI code to finish the skin engine so we may end up conflicting if we don't stage it properly. Basically: I've asked for a moratorium on SGE, SurgePatch and Skin* for a few days. Just FYI.

mvf:
Yep, saw that. Not to worry, my changes are 100% CMake ATM, so not as exciting as "data driven skin engine". :wink:

baconpaul:
OK cool. I have a few new C++ files (of course) but I can do that merge trivially I am sure.

PR #2788, opened 2020-10-06
CMake: Abolish OS_COMPILE_OPTIONS

mvf:
(This is the sales pitch. See the commit message for some boring technical details. :wink:)

This PR tries to pave the way for addressing two issues with the Surge codebase:

#### 1: Every source file is rebuilt for every target

This is not only slow, but it also means interfaces are not enforced. Anything can include anything, see all feature macros, etc., making Surge hard to maintain.

To fix this, over time, we should start breaking the codebase down into libraries. These have to declare their dependencies and interfaces and are only built once. This change facilitates this by making compile options (not definitions though!) global so they apply to all CMake subdirectories.

#### 2: Compilers are tied to OSes

Currently compile options are defined in the OS-specific CMake parts, making it hard to support one compiler on multiple platforms. So for instance, Clang builds have stricter warnings on macOS than elsewhere.

This change moves the compile options out of the OS parts and into a hierarchy of `if`s:

- Clang or GCC
  - Clang
    - Apple Clang
  - GCC
- MSVC
  - Clang-CL

Options set at each level are inherited by more specific variants of the same compiler family. For instance, `-Werror` at the "Clang or GCC" level affects all variants of both Clang and GCC on all platforms. OS-specifics can still be applied in the branches as needed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,6 +15,74 @@ if(NOT CMAKE_BUILD_TYPE)
 endif()
 
 project(Surge VERSION 1.8.0.0 LANGUAGES CXX ASM)
+
+set(CMAKE_CXX_EXTENSIONS OFF)
+if(APPLE)
+  set(CMAKE_CXX_STANDARD 14)
+else()
+  set(CMAKE_CXX_STANDARD 17)
+endif()
+
+if(CMAKE_CXX_COMPILER_ID MATCHES "Clang|GNU")
+  # Any Clang or any GCC
+  add_compile_options(
+    -Wno-multichar
+
+    # PE/COFF doesn't support visibility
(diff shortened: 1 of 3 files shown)


baconpaul:
Cool I will review this evening
Having at least the common and dsp code compile once would be great. I’ve wanted that forever but it is a lot of prying apart.... but worth doing

mvf:
Thank you! Sorry for being a bit short on Discord earlier. Had to run and was already late due to the AU issue. Reaper just managed to load my AU, so every 64-bit plugin type is now smoke-tested on every platform. Not sure what happened to that Azure job though... 🤔

baconpaul:
You weren't short! I'm kicking azure now. Seems like one of the windows jobs self-canceled?

baconpaul:
The azure failures are due to a pipeline resource being down at the last notify step; looks great otherwise so merging. Thanks!

Issue #2795, opened 2020-10-07
Channel Aftertouch is not smoothed in MPE mode


selenologist:
**Bug Description:**
In MPE mode, Channel Aftertouch is updated immediately without any smoothing, causing audible artifacts.
This is due to the use of a ModulationSource instead of a ControlModulationSource.

**Surge Version**
Noticed in NIGHTLY-2020-10-07-c91cd9c, LV2, 64-bit

**Reproduction Steps:**
Modulate filter cutoff over a large range with channel aftertouch. It will be smooth when MPE is disabled and blocky when it is enabled.

**Additional Information:**
I'm about to submit a fix ;). It was suggested that I post an issue anyway for completeness :)

selenologist:
Just making a note here that there's now a little bit of weirdness when MPE is turned on while a note is held - aftertouch seems to act like it's stuck at maximum. Toggling MPE while no notes are held fixes the issue. Interestingly, released notes (that are still consuming a voice) do not seem to cause the issue.

baconpaul:
Good to know, but I really don't think we need to worry about 'toggling MPE mode mid-voice' as a realistic performance or recording constraint. 

Thanks for the fix!!

PR #2798, opened 2020-10-07
Add smoothing to MPE Timbre.

selenologist:
ref https://github.com/surge-synthesizer/surge/issues/2795

Add smoothing to MPE timbre as well.

Some code was reorganised here to clean up and avoid repeating an if, AFAICT I haven't broken anything but could do with a second set of eyes.


baconpaul:
Looks correct to me. Happy to merge once CI completes. thanks!

Issue #2780, opened 2020-10-05
SVG gradients exported from Figma do not display


h-v-b:
**Describe the bug**

**Please let us know your surge version**
This information is all in the menu/about screen

* Surge Version: 1.8.nightly.82289b9
* Plugin type: VST2
* Bits: 64

**To Reproduce**
In Figma, create an svg resource with a gradient.
Export resource with the correct name as SVG
Add to SVG folder of your skin, load skin
**Expected behavior**
The resource should display the shapes with the gradients.

**Screenshots**
![Screenshot 2020-10-05 at 14 13 42](https://user-images.githubusercontent.com/1453415/95078229-13a1a380-0715-11eb-8de5-555444d1c856.png)

**Possible fix**
The issue could be related to the gradientUnits used by Figma on SVG export. Converting gradientUnits from userSpaceOnUse to objectBoundingBox could to the trick, as SGVs exported from Lunacy display correctly and seem to use objectBoundingBox for its gradients.
Found this trail while looking at how to convert the from userSpaceOnUse to objectBoundingBox: https://stackoverflow.com/questions/62431621/algorithm-to-convert-userspaceonuse-coordinates-to-objectboundingbox-coordinates

baconpaul:
OK there's at least two problems

The first is: our SVG parser does one pass, so the fact that you define your gradients after your paths means that they don't scan.

This is probably 'hard to fix' but maybe not impossible.

The second is that I am ignoring the user space flag when I am doing transforms. this is in my code and is probably fixable
We need to put this in an issue.

I think the fix on the first is 'always do a pass looking for defs'

that's more surgery on the SVG parser than I am up to this morning but all fixable I think

Here's the original offending svgs from @h-v-b and a modified version where I flip the sections which shows it rendering but still without the gradient (due to second issue)

[svg.zip](https://github.com/surge-synthesizer/surge/files/5327284/svg.zip)

baconpaul:
This code in nanosvg.h

```
	} else if (attr->hasFill == 2) {
		float inv[6], localBounds[4];
		nsvg__xformInverse(inv, attr->xform);
		nsvg__getLocalBounds(localBounds, shape, inv);
		shape->fill.gradient = nsvg__createGradient(p, attr->fillGradient, localBounds, &shape->fill.type);
		if (shape->fill.gradient == NULL) {
			shape->fill.type = NSVG_PAINT_NONE;
		}
	}
```

[...]

PR #2808, opened 2020-10-08
Fix SVG Gradient Problems

baconpaul:
SVG Gradient parsing was semi-broken in some cases from
Lunacy and FIGMA. Fix in the forked nanosvg; commit my test cases


baconpaul:
Will close #2780

Issue #2778, opened 2020-10-04
Mapping a CC to macro freezes Live when MIDI message stream is too dense


MartyLake:
**Describe the bug**
Mapping a CC parameter to Surge macro freeze Live when the amount of message is too high.
I suspect that it's because Surge macros are generating some kind of Gesture/Edit.

**Please let us know your surge version**

* Surge Version: 1.7.1.5270cfb
* Plugin type (VST3)
* Bits (64)

**To Reproduce**
1. Start Ableton Live
1. Load Surge in a midi track
1. Live: Create a midi clip containing midi data on CC18
1. Surge: map CC18 to macro 1
2. Live: Create an audio track, open the EDIT menu to see that the undo text is about the created audio track
3. Live: Launch clip
4. Live: open the EDIT menu, see that the undo text has changed to "macro 1..."

**Expected behavior**
Last step:
Live: open the EDIT menu to see that the undo text is about the created audio track (aka not a macro change due to midi mapping)

**Screenshots**

**Desktop (please complete the following information):**
 - OS: windows 10
 - Host Live
 - Version 10.1.13

**Additional context**
Using Ableton Live 10.1.13 on Windows, any VST that declares a gesture/edit have its parameter changes recorded in the undo stack, thus completely flooding the Live's undo when moving a parameter from a midi mapped parameter. When using high resolution midi devices (like Touché from Expressivee), this flood of parameter changes can overwhelm Live to the point of freezing.

One main interest I have for Surge is the ability to midi map, in order to bypass Ableton Midi mapping mecanism that suffers from the same behavior. Unfortunately, this bug is no go to me =(

I can provide dev hours if it can help fixing this. I have tried grepping https://github.com/surge-synthesizer/surge/search?p=2&q=beginEdit but only found gui-related code. I also tried grepping https://github.com/surge-synthesizer/surge/search?p=9&q=midi but found nothing related to midi mapping.

baconpaul:
Does this happen if the UI is closed? I bet it does not.

If not: my guess is

1. The midi application in SurgetSynthesizer and subsequent smoothing in ControllerMOdulationSource is updating the value
2. The updated value is forcing the display of the controller to re-render raising an event
3; you hit the sendParameterAutomation in SurgeGUIEditor in the valueChanged callback for tag_mod_source block. My version of SGE is undergoing some surgery but that's around like 3896 for me.

[...]

PR #2827, opened 2020-10-10
Fix: do not produce CCs when parameter automation changes

MartyLake:
Fixes https://github.com/surge-synthesizer/surge/issues/2778

Hey baconpaul, apparently, it's this line that produces parameter change on CC mapped parameters. Is it really a feature at all ? Can we just remove it ? Or maybe add an option for the end user (like the ui size or something?)

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -34,5 +34,6 @@ Paul Walker <paul@pwjw.com>
 Keith Zantow <kzantow@gmail.com>
 Vincent Zauhar <v.zauhar@bell.net>
 簡志瑋 <dppss92132@gmail.com>
+Marty Lake <matthieu@talbot.audio>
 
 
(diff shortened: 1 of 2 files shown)


baconpaul:
With this change, if you drag the CC with the mouse, does it produce latch messages still? That's what we need to test.

Lemme pull it and build and find out

baconpaul:
I think your change is correct! But we really don't need either the LV2 or the non-LV2 block. 

Can I ask you to just remove the entire #if ! LV2 blah else blah block rather than leave it commented, add yourself to the AUTHORS file in appropriate alphabetical order, and re-push? Then I'll merge.

Thank you very much!

MartyLake:
@baconpaul sure! I think I did what you asked, though it may need a double check.

PR #2840, opened 2020-10-12
BSD support fix.

devnexen:
backtrace API is in a separated (execinfo) library in these platforms.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -35,5 +35,5 @@ Keith Zantow <kzantow@gmail.com>
 Vincent Zauhar <v.zauhar@bell.net>
 簡志瑋 <dppss92132@gmail.com>
 Marty Lake <matthieu@talbot.audio>
-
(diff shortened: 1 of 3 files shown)


baconpaul:
thanks for the diff. Couple of qs

1: Seems like there's also an update to libs/lv2 - on purpose?

2: This is for the printStackTrace implementation to link basically? At some point i should make that so its only in debug build but no problem in this case.

baconpaul:
Oh also since this is (I think) your first commit, could you add yourself alphabetically to the AUTHORS file in a manner you find appropriate?

devnexen:
1/ I did not purposely updated it and this particular change can be dismissed especially FreeBSD package lv2 exists.
2/ Exactly this :-)

baconpaul:
OK cool if you can push reverting the submodule I'll happily merge!
Thanks!

baconpaul:
Thanks - all merged!

Issue #2860, opened 2020-10-14
Why crash happens when i open this preset?


K0rrid0r:
**Bug Description:**
Ok would love to know why this specific surge/daw crash happens when i open this preset (important to note is, this crash happen randomly, sometimes it does not happen and sometimes it does and i have many of those presets that "needs" to crash one time and then on reopening whole project and surge, no crash)

It's very random. And why crash one time then next time open no crash? 

You will probably have a hard time catching this error because it comes very randomly.

info from the dedugger
![crash 1 opening up preset that normally works 3](https://user-images.githubusercontent.com/56724286/95984623-c06dd600-0e23-11eb-953a-0734d0a4f04e.gif)
![crash 1 opening up preset that normally works 4](https://user-images.githubusercontent.com/56724286/95984628-c2379980-0e23-11eb-876a-91edec2925da.gif)

[PSY 3 - trigga med midi out wholetone.zip](https://github.com/surge-synthesizer/surge/files/5377935/PSY.3.-.trigga.med.midi.out.wholetone.zip)

Version 1.8.main.f61927a7 (Intel 32-bit Windows VST2 (unsupported) in Image-Line Fruity Wrapper - DEBUG BUILD. Built 2020-10-14 at 01:34:39 on local host 'KORRIDOR')

K0rrid0r:
This still occurs and now i got another crash error instead.

![crash 1 opening up preset that normally works 5](https://user-images.githubusercontent.com/56724286/96195654-97496480-0f4d-11eb-846c-75489f2d0c92.gif)
![crash 1 opening up preset that normally works 6](https://user-images.githubusercontent.com/56724286/96195658-99132800-0f4d-11eb-84d9-fc7d381ad32a.gif)

K0rrid0r:
and another one

![crash 1 opening up preset that normally works 7](https://user-images.githubusercontent.com/56724286/96195776-fe671900-0f4d-11eb-902e-01ae0439ce61.gif)
![crash 1 opening up preset that normally works 8](https://user-images.githubusercontent.com/56724286/96195778-00c97300-0f4e-11eb-87fb-e53c05c9b8f1.gif)

K0rrid0r:
and another one

![crash 1 opening up preset that normally works 9](https://user-images.githubusercontent.com/56724286/96196021-9fee6a80-0f4e-11eb-8fc3-e4c47b31bd66.gif)
[...]

PR #2876, opened 2020-10-16
Don't update UI from Audio Thread in VST2

baconpaul:
The VST2 would call udpateDisplay but call it in a way
that the audio thread would call it. That's obviously wrong
so remove that call and just have all the hosts force the update
through the refresh flag.


baconpaul:
Closes #2860

PR #2889, opened 2020-10-17
Move the oct selectors and filter menus down 1 px (vertical align)

mkruselj:
Rename ls_constant1 to ls_envelope
Reinstate Formula as Function (FUN) but make it work as envelope for the time being, "coming in 1.9!"

### DIFF
--- a/src/common/gui/SurgeGUIEditor.cpp
+++ b/src/common/gui/SurgeGUIEditor.cpp
@@ -5661,7 +5661,7 @@ std::string SurgeGUIEditor::modulatorName( int i, bool button )
       int fnum = idx % 6;
       auto *lfodata = &( synth->storage.getPatch().scene[current_scene].lfo[ i - ms_lfo1 ] );
 
-      if( lfodata->shape.val.i == ls_constant1 )
+      if( lfodata->shape.val.i == ls_envelope )
       {
          char txt[64];
          if( button )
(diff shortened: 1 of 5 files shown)


src/common/gui/SurgeGUIEditor.cpp
```diff
@@ -5688,6 +5688,15 @@ std::string SurgeGUIEditor::modulatorName( int i, bool button )
             sprintf( txt, "%s MSEG %d", (isS ? "Scene" : "Voice" ), fnum + 1 );
          return std::string( txt );
       }
+      else if( lfodata->shape.val.i == ls_function)
+      {
+         char txt[64];
+         if( button )
+            sprintf( txt, "%sFUN %d", (isS ? "S-" : "" ), fnum + 1 );
```
baconpaul, line 5695: So I'm fine having this button be active but I think the UI should show what it's doing so my only change is that this should be ENV / ENVELOPE for now.

mkruselj: The idea was to create hype for function/formula modulator :) Maybe we get a dev or two interested enough, haha!

mkruselj: Oh also the reason I renamed it to Function is because FUN is a nicer shorthand than FORM or FML. 😁

PR #2893, opened 2020-10-17
filesystem: Rewrite

mvf:
Replaces the "make it compile on macOS"-quality fallback filesystem with
a more robust implementation. It still only provides the bare minimum we
can get away with. The tests also pass against libc++ on both macOS and
Linux.


baconpaul:
This is amazing!
thank you

PR #2900, opened 2020-10-18
filesystem: Manage filesystem selection

mvf:
Moves the selection of the filesystem implementation into the filesystem
library so it is visible to other libraries.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -95,10 +95,13 @@ endif()
 set(SURGE_PRODUCT_DIR ${CMAKE_BINARY_DIR}/surge_products)
 file(MAKE_DIRECTORY ${SURGE_PRODUCT_DIR})
 
-add_subdirectory(libs/catch2)
-
 add_library(surge-shared INTERFACE)
 add_library(surge-tests INTERFACE)
+
+add_subdirectory(libs/catch2)
+add_subdirectory(libs/filesystem)
+
+target_link_libraries(surge-shared INTERFACE surge::filesystem)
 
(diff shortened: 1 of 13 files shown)


baconpaul:
Hey just FYI this seems to break the release pipeline which is I think ubuntu 18 https://dev.azure.com/surge-synthesizer/surge/_build/results?buildId=5546&view=logs&j=2f953adc-c56d-55c4-a64a-eab7c4b02235&t=72b09e0e-2846-587e-f662-8e2f4186f217

baconpaul:
OK so here's what I did

1. I did a standalone commit to return Ubuntu 18 to the CI pipeline. That's 3c03cbe278d. Standalone that fails the pipeline
2. I did a subsequent commit to revert this PR. That's commit 2e142be

I did it this way so you can revert #2 and work, but then we keep Ubuntu 18 in the pipeline once you have

It looks to me like we are just not finding std::experimental::filesystem which is what ub18 used but I didn't debug, just reverted.

Thanks!

mvf:
Sorry for this and thanks for adding Ubuntu 18.04 to the CI. Turns out each `CHECK_CXX_SYMBOL_EXISTS` call needs a new variable name or it just doesn't work. I had tested the `experimental/filesystem` codepath by commenting the first check, so it didn't trigger the issue. Can't make this up... :hurtrealbad:

baconpaul:
No problem and I figured it was something easy! Glad it is back. It is a way better solution.

baconpaul:
Oh and if you can just make sure this makes it to a nightly that'd be great.

PR #2922, opened 2020-10-20
Add smoothing to pitchbend

selenologist:
This is a bit of a hack and kind of breaks channel 0 pitch bends.

I‘ve kept #2921 separate / still open if you'd rather keep this patch out of main.

### DIFF
--- a/src/common/SurgeStorage.h
+++ b/src/common/SurgeStorage.h
@@ -915,6 +915,7 @@ public:
    float tuningPitch = 32.0f, tuningPitchInv = 0.03125f;
 
    ControllerModulationSource::SmoothingMode smoothingMode = ControllerModulationSource::SmoothingMode::LEGACY;
+   float mpePitchBendRange = -1.0f;
 
    std::atomic<int> otherscene_clients;
 
--- a/src/common/SurgeSynthesizer.cpp
+++ b/src/common/SurgeSynthesizer.cpp
@@ -885,17 +885,16 @@ void SurgeSynthesizer::pitchBend(char channel, int value)
 {
    if (mpeEnabled)
    {
-      float bendNormalized = value / 8192.f;
       channelState[channel].pitchBend = value;
 
-      if (channel == 0)
-      {
-         channelState[channel].pitchBendInSemitones = bendNormalized * mpeGlobalPitchBendRange;
-      }
-      else
-      {
-         channelState[channel].pitchBendInSemitones = bendNormalized * mpePitchBendRange;
-      }
+      /*
+      ** todo: handling of channel 0 and mpeGlobalPitchBendRange were broken with the addition
+      ** of smoothing. we should probably add that back in if it turns out someone actually uses it
+      *:)
+      ** currently channelState[].pitchBendInSemitones is now unused, but it hasn't been removed
+      *from
+      ** the code yet for this reason.
+      */
    }
 
    /*
--- a/src/common/SurgeSynthesizer.cpp
+++ b/src/common/SurgeSynthesizer.cpp
@@ -206,7 +206,7 @@ SurgeSynthesizer::SurgeSynthesizer(PluginLayer* parent, std::string suppliedData
 
    mpeEnabled = false;
    mpeVoices = 0;
-   mpePitchBendRange = Surge::Storage::getUserDefaultValue(&storage, "mpePitchBendRange", 48);
+   storage.mpePitchBendRange = (float)Surge::Storage::getUserDefaultValue(&storage, "mpePitchBendRange", 48);
    mpeGlobalPitchBendRange = 0;
 
 #if TARGET_VST3 || TARGET_VST2 || TARGET_AUDIOUNIT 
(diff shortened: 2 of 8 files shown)


selenologist:
In order to make the tests pass, some questionable hacks were needed:

1. Make sure the `storage.getPatch().dawExtraState` copy of `mpePitchBendRange` gets updated by hacking in a call to `surge->populateDawExtraState()` in the test
2. Start sampling the frequency 2000 samples later because the smoothed bend takes a little longer to reach the correct value.
3. Don't update `channelState[channel].pitchBend` in `SurgeSynthesizer::pitchBend` for channel zero so that it behaves like the old code did when `mpeGlobalPitchBendRange` remained at its default of zero.

... but lol it just failed on Azure again anyway (hack 2). Better suggestions would be appreciated.

selenologist:
It does actually behave now with the `SmoothingMode` set to `DIRECT`, but there are still some nasties in the code that should probably be discussed.

Hack 2 has been reversed but I still did the no-no of changing a test.

Not sure why the previous build failed on MacOS, these changes shouldn't have affected the envelopes.

mkruselj:
Ah, unit tests sometimes randomly barf on certain platforms exactly on envelopes. I don't know why, but Paul might. :)

baconpaul:
Some problems here. Will add comments later today but pls don't merge this one

selenologist:
Okay, moved `mpePitchBendRange` into `SurgeStorage` avoiding the issue with `dawExtraState`.

Now there isn't anything weird in the unit tests.aside from disabling the smoothing, but there are already tests that show that the smoothing code eventually reaches its target value so that should be okay.

baconpaul:
OK I just read your change and I'm still concerned about moving the bend off off the channelState.

Specifically I'm concerned that removing the bend from the channel state changes the behavior in the mono modes. I'm not sure it i the right change.

In SurgeSynthesizer we copy the channel state between voices across channels in the various mono mpe modes to make sure the various states are copied. I worry quite a lot that your change here means the bend modes won't work with the various mono modes. (Getting the mono modes working with MPE was ... difficult).

Can we write a test or do something else to allay those fears?

Or: Can you not change the channel state object and smooth inside the voice somehow?

selenologist:
Sorry, I'm not quite sure what you mean - the raw pitch bend value is still stored in channelState, same as the old code, just we no longer keep it around in semitone form as well. Then that value is used as a target for the smoother in SurgeVoice.

As I'm not sure quite what sort of bug you're expecting to see, I'm not quite sure how to write a test. But I can say that the mono modes... seeeeem to work? I can play a note, bend it, play another that takes priority, bend that, release that, and it goes back to the pitch of the first bend. This also works with portamento.

baconpaul:
OK! I'll take a look later today then and convince myself that my worries are unfounded and press merge, or come back with something more concrete! :)

Thanks for chasing all of this down.

baconpaul:
OK great I see yeah you just moved the multiply. Sorry for my confusion this morning.

And that channel[0] pitchbend message - how sure are we nothing sends that? Like: If you had two keyboards feeding surge you could send the pitchbend from a wheel or use a wheel as channel 0 pitchbend? Do you think we should restore it? (I'm happy to merge this if we then restore it even it if is unused right now).

But my real question is this: Do we want the same smoothing on MPE pitchbend as we do on other controllers? Like: Do we ever want legacy mode? Or do we just want to force fast linear? Or do we want a separate option for pitch bend smoothing? (or the MPE smoothing modes in general?)

I can merge this and try it with my linnstrument and roll but it's not 10)% clear to me that the pitch gesture being smoothed the same as a random knob is entirely desirable as a default. (Although with large pitch bend ranges I see why you want smoothing but i'd think fast linear as a default is what you want (although again I haven't had a chance to play with your branch)).

Happy to merge and open an issue on the 18 milestone on the above two thoughts or for you to tell me i'm wrong or for you to make more changes. Lemme know!

selenologist:
Channel zero pitchbend still works mostly the same way it did in the old code i.e. the same as when not in MPE mode. This is because normally `mpeGlobalPitchBendRange` stays zero, so the code I removed was basically a NOP anyway.

Actually, if I understand correctly, had Surge ever received the RPN message that sets `mpeGlobalPitchBendRange` (here https://github.com/surge-synthesizer/surge/blob/main/src/common/SurgeSynthesizer.cpp#L1016 ) to something other than zero, the old code would have applied a pitch bend twice...

Agreed that an option for how/if pitch bend should be smoothed would be desirable, but for the moment I've forced it to fast linear. Which doesn't seem to slew too noticeably on LinnStrument but it might be more of an issue on Roli.

baconpaul:
Ahh those MPE pitch bend tests use our ad-hoc frequency detector which can be a bit approximate but the failure is "real" inasmuch as pitchbend isn't applied immediately

But the frequencyForNote does play for 500 blocks before it does a frequency measure so

1: Are you sure you are on fast linear and not some other one in the regtest engine? (You can run it yourself as the surge-headless target)
2. Are you sure the pitch bend has quiesced before we measure frequency?
3. If "yes" and "yes" you can change that margin from .3 to .6 and it will work :)
4. If "no" or "no", fix and goto 1!

baconpaul:
Great. Thanks for your patience and changes. Merging

Issue #475, opened 2019-02-02
Output Master Volume can´t be saved with the preset


cybpun:
Output Master Volume can´t be saved with the preset.
With this fixed we have the possibility to level all the presets.

baconpaul:
Thanks!

Linking to #391 since I have to do that one before this one

baconpaul:
I may do this in 16 @esaruoho seems pretty important to patch leveling

baconpaul:
So turns out this is a conscious choice in the code. I'm not quite sure why but because of that, I chose to not implement it with the 1.6.2 patch change. I'll keep this issue open and revisit with 1.6.3

mkruselj:
@baconpaul let's do this, along with saving polyphony in patch, for 1.9?

baconpaul:
Yup great

baconpaul:
@mkruselj we did this yeah? Closing but reopen if I’m wrong

mkruselj:
@baconpaul We didn't AFAIK, saved it for XT because of patch balancing considerations IIRC.

baconpaul:
Oh right the reason I was confused is because the fix is the same as poly but we only fixed poly

mkruselj:
Indeedio!

PR #2932, opened 2020-10-21
Hardclip bypass

selenologist:
This patch allows the hardclipping before the main volume slider to be disabled by right clicking this slider.

Hard clipping is enabled by default to retain Surge's current behaviour.

Hard clipping state is not saved, the intention is more like giving the user the ability to hear whether their patch was clipping so it can be adjusted.

It also renames this slider to "Main Volume" in line with the renaming of the main branch.

### DIFF
--- a/src/common/Parameter.cpp
+++ b/src/common/Parameter.cpp
@@ -459,6 +459,7 @@ void Parameter::set_type(int ctrltype)
       val_default.f = 0;
       break;
    case ct_decibel_attenuation:
+   case ct_decibel_attenuation_clipper:
       valtype = vt_float;
       val_min.f = -48;
       val_max.f = 0;
@@ -979,6 +980,7 @@ void Parameter::set_type(int ctrltype)
 
    case ct_decibel:
    case ct_decibel_attenuation:
+   case ct_decibel_attenuation_clipper:
    case ct_decibel_attenuation_large:
    case ct_decibel_fmdepth:
    case ct_decibel_narrow:
@@ -1150,6 +1152,7 @@ void Parameter::bound_value(bool force_integer)
       case ct_decibel_narrow_short_extendable:
       case ct_decibel_extra_narrow:
       case ct_decibel_attenuation:
+      case ct_decibel_attenuation_clipper:
       case ct_decibel_attenuation_large:
       case ct_decibel_fmdepth:
       case ct_decibel_extendable:
@@ -2585,6 +2588,7 @@ bool Parameter::can_setvalue_from_string()
    case ct_decibel_narrow_short_extendable:
    case ct_decibel_extra_narrow:
    case ct_decibel_attenuation:
+   case ct_decibel_attenuation_clipper:
    case ct_decibel_attenuation_large:
    case ct_decibel_fmdepth:
    case ct_decibel_extendable:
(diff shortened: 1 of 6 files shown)


mkruselj:
I'd kinda want to keep it at Master Volume. We've had this discussion before and IIRC left it as is due to familiarity in audio world. I'd rather back that change out and maybe reopen the discussion on Discord and do a separate PR afterwards.

Also I definitely do think that this switch should be stored in the patch.

baconpaul:
I'm happy to merge this. But I think we should bring #475 into the 180 milestone so I'll do that, and when we do we need to save this param. (@selenologist if you want to do that it would be welcomed!)

as to master vs main two thoughts

1. An irrelevant opinion. I think most audio software will flip this terminology in the next 12 months but hardware labels will of course not. But that doesn't matter because

2. Did you know that with the volume at 0 surge will still make noise! Just put it in dual mode! It is no more the master/main volume than a bus mute is a master/main mute 

So I think the right thing to do is to call this param what it actually is - Scene Volume.

@selenologist if you want to push that tweak I'll merge this and will also bring 475 into this milestone.

mkruselj:
Renaming it to Scene Volume makes sense to me. But then... we REALLY need a proper master volume that governs eeeeeeeverything (and can be a proper mute if need be).

mkruselj:
Oh but WAIT! It's NOT individual scene volume... It does adjust the output of both scenes A and B simultaneously...

baconpaul:
I just tried it in dual mode. And it didn't.
Are you sure that's what you see?

mkruselj:
Positive, both in Reaper, S1, Live and SAVIhost VST3. Acts as master volume for both scenes.

baconpaul:
OK maybe I am using the wrong slider.
Which slider are you using?

mkruselj:
image

baconpaul:
I am using the wrong slider.

baconpaul:
But the scene volume slider works great

mkruselj:
We could rename it to Global Volume I suppose. 

We also have two Output headers so that's confusing as well.

image

baconpaul:
I like Global Volume. Like I said I think this will be a word that the software community moves away from so I'm happy with a change to Main or Global for now.

mkruselj:
Oh, this is also a good chance to provide extend mode for master volume, so that it can go -inf to, say +12?

baconpaul:
Doesn't master volume not stream?

baconpaul:
That's kind of the point of #475

mkruselj:
It's not stored in patch but it is in DAW state.

baconpaul:
Ahh that was clever of me. Then yeah we should definitely add this to the dawExtraState and stream it. Lets do that and change to Global Volume before merge. Sound good @selenologist ?

And thanks!

selenologist:
Sounds good, I'll look at changing the range and saving hardclip state later but for now I've made the label changes from the github web UI - sorry, have to replace a failed drive before I can use my main system. That's why I've been slow to reply too.

baconpaul:
OK cool I'll merge this once CI is done then and then when you are back in action I can show you where how the daw extra streaming code works.

mkruselj:
I think renaming to Scene Volume is not necessary, we should change the title of that section's header to Scene Output instead.

Issue #2408, opened 2020-07-31
Options for cursor restore position after hiding


MD-78:
**Describe the bug**
Cursor restore change in 1.7.0 was not a nice change. Cursor should be restored at current drag position but not outside of fader width (when cursor is hidden while dragging). Say I drag Cutoff fader fully to the right (but the drag itself even goes behond), the mouse cursor should be restored at the right edge of the Cutoff fader when mouse button got release. Currently it jumps back to the left - why? Looks and feels weird.

**Please let us know your surge version**
This information is all in the menu/about screen

* Surge Version: 1.7.0
* Plugin type (VST2, 3 or AU): VST3
* Bits (32/64): 64

**To Reproduce**

1. drag Cutoff fader fully to the right (the drag itself even goes behond)

**Expected behavior**
The mouse cursor should be restored at the right edge of the Cutoff fader (when drag was behond right edge) when mouse button got released. Currently it jumps back to the left - why? Looks and feels weird.

**Screenshots**
[untitled.zip](https://github.com/surge-synthesizer/surge/files/5006751/untitled.zip)

**Desktop (please complete the following information):**
 - OS: Windows 8.1 Pro
 - Host: REAPER 6.13 x64

baconpaul:
@mkruselj did you make this an option? I forget?

mkruselj:
Yes I made this option and this is how it should work, I compared to other plugins that have this feature and it's working exactly like that (i.e. Massive X).

MD-78:
This is not nice. I don't care what Massive X does. I also don't see an option in Surge for this.

Have a look at Voxengo plugins - they are perfect in this.

mkruselj:
Option is in the Menu -> User settings -> Mouse behavior -> Show cursor while editing.

MD-78:
But I don't want the cursor to be shown all the time (because you can hit screen edges).

I'm  not a programmer, that's why I bring it up here.


Current Surge behaviour mares no sense. You release the mouse button -> cursor getting moved to drag start position -> now you have to move the cursor back to the current fader position to start a new drag (this step would be unneeded when the cursor would be shown where button released).

baconpaul:
@mkruselj Ithink @MD-78‘s point is “the 1.6.6 behavior was better”.

I’m not a windows user and this cursor hiding stuff freaks me out since mac doesn’t do it so I don’t have a strong opinion, but perhaps add an option to restore the 1.6.6 behavior is the request here?

[...]

PR #2934, opened 2020-10-21
Better Cursor API now in place

baconpaul:
- A CRTP base class for CViews makes the API super easy
- Applied to MSEG Editor and retained capability
- Applied to CSurgeSlider and the hide/show is correct, including
  on macos


mkruselj:
Closes #2408.

PR #2953, opened 2020-10-23
Add Smooth stairs curve to MSEG

mkruselj:
Control point for Spike curve is now on the spike itself
Update both skin backgrounds with title change for Scene Output section (was just Output)
Move the scene hardclip toggle to scene volume slider
Rename the hardclip menu entry, add OS letter case function, add checkmark
Rename scene volume slider back to just Volume


baconpaul:
Looks good here!

PR #2959, opened 2020-10-24
Add K35 filter

selenologist:
Some issues/uncertainties:
* I used the `clampedFrequency` stuff from `VintageLadder` to stop it from exploding when the frequency goes too high. It's possible that it can be set higher than `VintageLadder`'s limit, but be cautious for now.
* Coefficient calculation probably doesn't need to use `double` but I've kept it like the original. **FIXED**
* `applyOverdrive` is implemented differently than in Odin2 - might need more oversampling to avoid artifacts.
* ... just noticed I didn't use the `SURGE_EXTRA_FILTERS` ifdef. Oh.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -208,6 +208,7 @@ set(SURGE_COMMON_SOURCES
   src/common/dsp/effect/VocoderEffect.cpp
   src/common/dsp/filters/VintageLadders.cpp
   src/common/dsp/filters/Obxd.cpp
+  src/common/dsp/filters/K35.cpp
   src/common/dsp/AdsrEnvelope.cpp
   src/common/dsp/AudioInputOscillator.cpp
   src/common/dsp/BiquadFilter.cpp
--- a/src/common/Parameter.cpp
+++ b/src/common/Parameter.cpp
@@ -2164,6 +2164,10 @@ void Parameter::get_display(char* txt, bool external, float ef)
                   case fut_obxd_4pole:
                      sprintf( txt, "%s", fut_obxd_4p_subtypes[i]);
                      break;
+                  case fut_k35_lp:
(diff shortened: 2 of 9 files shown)


selenologist:
OK, saturation has been added as per the code Odin2 is based on (where tanh is in the feedback loop).

As Surge has no way to specify the saturation parameter, it has been added as filter subtypes for now.

Note that you should reduce your pre-filter gain when using the saturation modes if you want to hear resonance, otherwise it will be drowned out - much like in a real filter!

baconpaul:
OK this is amazing. Thank you. It's still a draft so I can't merge it. But let me add a couple of review comments

1. The fasttan - I know it is only used here but can you put it in the header with the others anyway? My rough plan is "port stuff over there as we need it" and "add SSE versions as we need them" which is why it is incomplete but having it there vs inline would be super great.

2. I've added retests for the fast stuff. If you could do same for fasttan when you move it that would be great. src/headless/UnitTestsDSP. Just a simple compare over range is good.

3. To make the coefficients smooth, you should do the dC / dT thing. That is add something like     https://github.com/surge-synthesizer/surge/blob/5c5b4e906f76a12d54886ca79d6b6b3aaad44754/src/common/dsp/filters/Obxd.cpp#L106 at the top of each process. Surge will automatically calculate the change you need.

4. Similarly in your makeCoeff rather than editing cm->C directly you want to make a tmp C, fill it up, then call cm->FromDirect like vintage and OBXD do https://github.com/surge-synthesizer/surge/blob/5c5b4e906f76a12d54886ca79d6b6b3aaad44754/src/common/dsp/filters/VintageLadders.cpp#L110

3 and 4 won't make any difference at constant value of freq and res, but when modulated will avoid all the ugly step noise.

PR #2979, opened 2020-10-27
airwindows: Lib-ify

mvf:
Linux Clean Release Build Before:
```
[843/843] Running utility command for Surge-VST2-Packaged
ninja  1893.13s user 53.87s system 1095% cpu 2:57.77 total
```
After:
```
[547/547] Running utility command for Surge-VST2-Packaged
ninja  1415.95s user 41.25s system 947% cpu 2:33.79 total
```


baconpaul:
Hey thanks!
Two comments here

1. I didn't use the name 'audioeffectx.h' to avoid (1) confusing with the VST2SDK one and; (2) appearing to include the VST2SDK in our GH repo (see the various steinberg DMCA takedown sweeps based on filenames).

2. There's a script which adds a new airwin. Does it still work with these changes?

Thanks!

mvf:
Thanks for the feedback, I'll look into these points.

Regarding 1:
I wasn't aware of the significance of `audioeffectx.h`, I just reused the name to reduce the source delta to the original sources. I'll drop it and include `AirWinBaseClass.h` instead.

Regarding 2:
I hadn't looked at the script, so sources imported with it won't compile. I'll update it. If adding new effects is common, it might even be possible to make things submodule-compatible.

baconpaul:
On 1 - thanks!

on 2 - we probably aren't adding many more. But when I did the first 50 automating it was a must! Was more just curious there. We should at least update the 'how to add' doc.

Awesome change btw. Compiling is way way faster.

PR #2984, opened 2020-10-27
Implement all MSEG actions (except Flip horizontally)

mkruselj:


baconpaul:
Please don't merge this. Review comment inbound.

mkruselj:
I would have waited for your thoughts on this anyways! :)

baconpaul:
Hi!

OK so these implementations are fine but they are in the wrong place. I have endeavored to not put any direct manipulation in the menu callbacks and instead put them in the (non-gui) MSEGModulationHelper code.

So things like

```
addCb(actionsMenu, Surge::UI::toOSCaseForMenu("Half Size"),
                             [this](){
                                        for (int i = 0; i < this->ms->n_activeSegments; i++)
                                           this->ms->segments[i].duration *= 0.5;

                                        modelChanged();
                                     });
```

I would implement as follows

1. In MSEGModulationHelper.h add a free function "scaleDuration( MSEGStorage *ms, float amount )"
2. Have it do that except the 0.5 is the amount
3. Make that implementation `Surge::MSEG::scaleDuration( this->ms, 0.5 ); modelChanged();`

This way we get two things

1. A set of functions we can test independent of the UI in headless and
2. All the manipulators in one place, not scattered in callbacks

It looks like the functions you have are 'scaleDuration', 'scaleValue' and 'setAllDurationsTo' (it is fine to do the ms->total duration / ms->n_activeSegments in the menu callback)

See what I mean?

mkruselj:
Done!

Yep, it's better this way :)

PR #2986, opened 2020-10-27
airwindows: Remove references to audioeffectx.h

mvf:
An earlier change introduced a header of this name to avoid changing
include directives in the imported sources. But the original is part of
the VST2 SDK, which might get the repo flagged by filename sweeps.


baconpaul:
Thank you very much for doing that! :)

mvf:
No worries, wouldn't wanna get us into trouble!

PR #2994, opened 2020-10-28
Scene octave/pitch, related modulations and PB works with non-keytracked oscs

### LINKED ISSUE #2556: Scene pitch doesn't change the pitch of oscillators that have keytracking disabled
The main pitch controls for the individual scenes don't change the pitch of oscillators that have keytracking disabled.

I would expect the control to still work, despite keytracking being disabled.
If this is intended behavior, please ignore the report.

I'm using Surge-NIGHTLY-2020-09-01-a4152d7 in Ableton Live 10.

mkruselj:
Closes #2556

### DIFF
--- a/src/common/dsp/SurgeVoice.cpp
+++ b/src/common/dsp/SurgeVoice.cpp
@@ -307,10 +307,13 @@ void SurgeVoice::switch_toggled()
       pb *= (float)scene->pbrange_up.val.i;
    else
       pb *= (float)scene->pbrange_dn.val.i;
-   state.pitch = state.pkey + pb;
 
-   modsources[ms_keytrack]->output =
-       (state.pitch - (float)scene->keytrack_root.val.i) * (1.0f / 12.0f);
+   // scenepbpitch is pitch state but without state.pkey, so that it can be used to add
+   // scene pitch/octave, pitch bend and associated modulations to non-keytracked oscillators
+   state.scenepbpitch = pb;
+   state.pitch = state.pkey + state.scenepbpitch;
+
+   modsources[ms_keytrack]->output = (state.pitch - (float)scene->keytrack_root.val.i) * (1.0f / 12.0f);
 
    for (int i = 0; i < n_oscs; i++)
    {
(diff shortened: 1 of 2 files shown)


src/common/dsp/SurgeVoice.cpp
```diff
@@ -728,31 +731,23 @@ bool SurgeVoice::process_block(QuadFilterChainState& Q, int Qe)
...
       }
+
       if (osc2)
       {
          if (is_wide)
-         {
-            osclevels[le_osc2].multiply_2_blocks_to(osc[1]->output, osc[1]->outputR, tblock,
-                                                    tblockR, BLOCK_SIZE_OS_QUAD);
-         }
+            osclevels[le_osc2].multiply_2_blocks_to(osc[1]->output, osc[1]->outputR, tblock, tblockR, BLOCK_SIZE_OS_QUAD);
```
baconpaul, line 748: Any reason for this diff?

src/common/dsp/SurgeVoice.cpp
```diff
@@ -766,38 +761,29 @@ bool SurgeVoice::process_block(QuadFilterChainState& Q, int Qe)
...
-         {
-            osclevels[le_osc1].multiply_2_blocks_to(osc[0]->output, osc[0]->outputR, tblock,
-                                                    tblockR, BLOCK_SIZE_OS_QUAD);
-         }
+            osclevels[le_osc1].multiply_2_blocks_to(osc[0]->output, osc[0]->outputR, tblock, tblockR, BLOCK_SIZE_OS_QUAD);
          else
-         {
             osclevels[le_osc1].multiply_block_to(osc[0]->output, tblock, BLOCK_SIZE_OS_QUAD);
-         }
+
```
baconpaul, line 786: Similar any reason for this diff?

src/common/dsp/SurgeVoice.cpp
```diff
@@ -858,13 +844,10 @@ bool SurgeVoice::process_block(QuadFilterChainState& Q, int Qe)
...
       }
       if (is_wide)
-      {
          osclevels[le_noise].multiply_2_blocks(tblock, tblockR, BLOCK_SIZE_OS_QUAD);
-      }
       else
-      {
          osclevels[le_noise].multiply_block(tblock, BLOCK_SIZE_OS_QUAD);
-      }
+
```
baconpaul, line 850: Ditto

src/common/dsp/SurgeVoice.cpp
```diff
@@ -694,29 +694,32 @@ bool SurgeVoice::process_block(QuadFilterChainState& Q, int Qe)
...
 
+   auto pb = modsources[ms_pitchbend]->output;
+   if (pb > 0)
+      pb *= (float)scene->pbrange_up.val.i;
+   else
+      pb *= (float)scene->pbrange_dn.val.i;
+
+   // recalculate pitch state but without state.pkey, so that the residue can be used to add scene pitch/octave/pitch bend
+   // and associated modulations to non-keytracked oscillators
+   auto tempPitch = pb + localcopy[pitch_id].f * (scene->pitch.extend_range ? 12.f : 1.f) + (octaveSize * localcopy[octave_id].i);
```
baconpaul, line 705: Since there's a couple of brace elisions which we should fix I think it also makes sense to choose a better name than 'tempPitch'. I was thinking "sceneAndPBPitchAdj" or something. So the code below reads as 'ktkroot + sceneAndPBPitchAdj". Or some such.

mkruselj: Makes sense, I also moved it to member variable of voice state. Code is clearer now!

Issue #2966, opened 2020-10-25
MSEG edit mode: LFO vs envelope


mkruselj:
We should add an explicit switch to change the MSEG editing mode from the current (_and default_) extensible flexible envelope mode, to a single-cycle (one phase unit) custom LFO mode.

This would bring a few changes to how the editor operates specifically in LFO mode: zooming and scrolling won't be necessary as you will always see exactly one phase unit and one phase unit only. Link start/end would always be enabled and you couldn't turn it off (probably?). More?

mkruselj:
Question for @baconpaul: do you think we really need a button for this, or if we could just tuck it inside the Actions menu?

baconpaul:
I think a button would be useful since you will want to be able to see what state you are in when editing.

mkruselj:
OK then, I will add it today probably!

mkruselj:
Implementation details and food for thought:

* In LFO mode, first and last node are fixed at 0 and 1 phase unit respectively, and cannot be moved horizontally ever, only vertically (depending on start/end node linking)
* In case of Shift movement mode, it should of course still work, but within 0...1 constraints, as soon as second to last node hits phase 1 that's all it can move things about.
* In the model, nodes that were created after phase unit 1 should still remain.
* When going from envelope to LFO mode, the node closest to phase unit 1 should snap to phase unit 1
* When going from LFO to envelope mode, nothing changes
* Any Actions should only apply to nodes between 0 and 1 phase unit, and not touch nodes outside of it (gonna need to thread an argument into those free functions in MSEGModulationHelper, I suppose)
* Loop mode is always on, loop start is always 0, loop end is always 1. Maybe we should gray out the loop mode button then, and remove the loop start/end menu options and markers in this mode?

baconpaul:
So the only thing about your spec I did differently was in flipping modes
Rather than truncating points, I rescale the envelope to [0,1] and remember the prior duration; if you flip back it restores the duration.
This is kinda more natural (you have drawn an LFO and flip it to LFO mode you keep the pattern) and also buttloads easier to do in the code :)
Did the rest of it basically exactly as you laid out over in #3206 
Will leave this open after the merge but I think it's pretty close to done.

mkruselj:
Huh, that would probably result in a pretty unusable single cycle if the prior envelope was very long and contained a lot of points... Unsure if that's the best way to go, but let's test.

baconpaul:
[...]

PR #3000, opened 2020-10-29
Add MSEG edit mode parameter and required assets

mkruselj:
Addresses #2966 

Also commented out std::cout calls in SkinSupport


mkruselj:
PR/issue number #3000! 🥂 

All that's left is implementation by @baconpaul 🙂

PR #3005, opened 2020-10-29
Add different LFO deform types

### LINKED ISSUE #2914: Different deform modes (or curves) options for modulators
After I saw that deform result in the MSEG was different than in other modulators ( see #2912 ), I was thinking that it wasn't inherently a bad curve by itself, though it would be more consistent to make it behave the same way as other modulation types.

That being said, if Deform was consistent on all modulation types, I thought it would be a good idea to being able to choose different deform curve types in the Deform parameter's context menu (a bit like the options for Portamento we have right now, for instance). This would apply for all LFO types in that modulator, including the MSEG.

I think I might be able to work on this, at least eventually...

VincyZed:
- Added different LFO deform types for Sine, Triangle, and Saw (right-click on Deform, Type 1, Type 2 or Type 3).

I made those formulas based on the feedback on the LFO Display, so it could be possible that it might go slightly (and I mean very slightly, if it's the case) out of bounds.

I tested saving and loading patches (including backwards compatibility) and streaming works.

Closes #2914


mkruselj (changes requested):

baconpaul:
I agree with @mkruselj suggestion but otherwise think the change is great and would merge once you do that small update. Thank you!

VincyZed:
Sounds good, will work on that today. Thanks guys!

baconpaul:
Merged! Thanks!

mkruselj (approved):

PR #3025, opened 2020-10-31
Fix for crash when clicking on waveshaper type

mkruselj:
Potentially fixes the crash TheNerdyMusicGuy experienced with filter config in Live


baconpaul:
Great find

PR #3040, opened 2020-11-03
Windows 32-bit, macOS: Remove __asm blocks

mvf:
Messy and universally slower than compiler-generated code.


baconpaul:
Thanks! This i great and one of those things I had been putting off *forever*. Really appreciate you cleaning it all up :)

PR #3041, opened 2020-11-03
Add diode ladder filter icon

mkruselj:
Update vintage ladder icon to make it different from legacy ladder icon (diode ladder icon follows suit)


baconpaul:
Nice!

PR #3043, opened 2020-11-03
Some mod preset tweaks

mkruselj:
Capitalization (Lfo -> LFO) and spaces (StepSeq -> Step Seq)
Subfolder name changes
Separator after "Save xyz as..." only shows if we have any mod presets saved


baconpaul:
Oh sorry macos auto-suppresses separators at end of menus. Thanks!

PR #3056, opened 2020-11-05
Clean up scene index literals used in loops etc., an attempt

mkruselj:


src/common/SurgePatch.cpp
```diff
@@ -1459,7 +1461,7 @@ void SurgePatch::load_xml(const void* data, int datasize, bool is_preset)
    // ensure that filtersubtype is a valid value
    for (auto& sc : scene)
    {
-      for (int u = 0; u < 2; u++)
+      for (int u = 0; u < n_filterunits_per_scene; u++)
       {
          sc.filterunit[0].subtype.val.i =
```
baconpaul, line 1466: I think there's an old bug here that the [0] should be the [u] don't you agree? If so we should open a separate issue to investigate this line (your change is fine here).

src/vst3/SurgeVst3Processor.cpp
```diff
@@ -507,15 +507,15 @@ tresult PLUGIN_API SurgeVst3Processor::process(ProcessData& data)
...
+            for (int c = 0; c < 2; ++c) // c = channel, we have 2 (left, right)
             {
                outA[c][i] = (float)surgeInstance->sceneout[0][c][blockpos];
                outB[c][i] = (float)surgeInstance->sceneout[1][c][blockpos];
             }
          }
          else
          {
-            for( int c=0; c<2; ++c )
+            for (int c = 0; c < n_scenes; ++c)
```
baconpaul, line 518: This one is also channel not scene. So need to fix this.

mkruselj: Oh, ok! Easy. Should've assumed based on it being `c` not `s`!

src/common/SurgePatch.cpp
mkruselj: That definitely looks like a legit bug to me, because that inner `u` loop doesn't do anything other than repeat the same thing 2x...

PR #3071, opened 2020-11-06
Make the debug console message a bit nicer

mkruselj:


src/common/DebugHelpers.cpp
```diff
@@ -29,9 +29,9 @@ bool Surge::Debug::openConsole()
         winconinitialized = true;
         AllocConsole();
         freopen_s(&confp, "CONOUT$", "w", stdout);
-        std::cout << "Surge Debugging Console\n"
-                  << "This console shows stdout from the surge plugin. If you close it with X you might crash.\n"
-                  << "Version: " << Build::FullVersionStr << " built at " << Build::BuildDate << " " << Build::BuildTime
+        std::cout << "SURGE DEBUG CONSOLE\n\n"
+                  << "Displays standard text output from Surge, for debugging purposes. If you close this window, Surge will crash!\n"
```
baconpaul, line 33: Oooh "stdout" is the C++ named file handle (as opposed to stderr) so that was there on purpose! 

Doesn't matter much but if you haven't merged we should keep it saying stdout.

PR #3073, opened 2020-11-06
Initial implementation of decimal pitch bend range adjustment

### LINKED ISSUE #2192: Add cents values to Bend Depth ranges in extended mode
Excerpted from Slack discussion:

Jacky  
Is it possible to make Bend Depth ranges support floating point values?

Negative/positive implications of doing such?

If it were actually possible, and wouldn't interfere with the tuning library implementation, then, instantly, the first part of #2125 could be checked off the list.

image

Fantasy implementation of the above, might include Shift + LMB scrolling to specify precise semicents values... 
And or typeins.

baconpaul  
Where they are applied yes; where they are stored i would have to look

And yeah shift to scroll etc....

Don’t want to do it for 170 since I’m really trying to freeze new features but would be happy to look once we are stabilized. From my understanding of the code there is nothing that makes it fundamentally an int inside the synth engine. Can you add an issue?

mkruselj:
Breaks old patches (they load with PB ranges set to 0), needs work
When implemented correctly, closes #2192

### DIFF
--- a/src/common/Parameter.cpp
+++ b/src/common/Parameter.cpp
@@ -402,10 +402,10 @@ void Parameter::set_type(int ctrltype)
       val_default.i = 1;
       break;
    case ct_pbdepth:
-      valtype = vt_int;
-      val_min.i = 0;
-      val_max.i = 24;
-      val_default.i = 2;
+      valtype = vt_float;
+      val_min.f = 0.f;
+      val_max.f = 24.f;
+      val_default.f = 2.f;
       break;
    case ct_pitch_semi7bp:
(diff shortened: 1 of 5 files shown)


baconpaul:
lets not merge this - i have some feedback but don't have time now

mkruselj:
Didn't even intend to merge it. In fact I missed the option to make it a draft PR when I opened it, so feel free to mark it as such!

mkruselj:
I'm closing this one, let's revisit this for Surge XT when we're JUCEd up.

PR #3141, opened 2020-11-13
Prompt for Install Folder in Win Installers

### LINKED ISSUE #2403: Choose Installation folder in installer
**The request itself**
This should be a no brainer; let the user pick installation folder when installing Surge, instead of having it install in a generic, undisclosed folder. Surge already has a portable mode, so that should not be a challenge.

Now, here's my experience installing 1.7.0 (on Windows):

I opened the installer, hoping that at the very least, Surge would let me install it in the folder I keep the VST2 in (which is a custom folder where I keep all of my other VST in, 64-bit, 32-bit, VST2, VST3, what gives you). But instead, it prompted me nothing about destination folder and just installed itself where it wanted.
After the initial confusion, I read on the manual that Surge has a "portable mode", and that the user had to manually do a bit of tweaking to get working. Thankfully, that is definitely not a lot of work (would be much better if it was no work at all, but i'll take it), but there remained the problem that the installer didn't actually inform me **where** the VST3 was installed in the first place. I had to guess it was installed on Program Files/Common Files/VST3 (and not, say, Program Files/VstPlugins, Program Files/VST or something similar), and that took me a little while.

All in all, happy ending, I got Surge to work, on the folder it was meant to be. But it shouldn't take me this much work, and I'm posting this in hopes that the next time I have to install it will be smoother. By the way, thank you Surge Synth Team for all the hard work, 1.7.0 is amazing.

baconpaul:
Closes #2403

### DIFF
--- a/installer_win/surge-x86.iss
+++ b/installer_win/surge-x86.iss
@@ -24,10 +24,10 @@ AppPublisher={#MyAppPublisher}
 AppPublisherURL={#MyAppURL}
 AppSupportURL={#MyAppURL}
 AppUpdatesURL={#MyAppURL}
-DefaultDirName={cf}\VST3\Surge Synth Team\
+DefaultDirName={cf}\VST3\
 DefaultGroupName=Surge
 DisableProgramGroupPage=yes
-DisableDirPage=yes
+DisableDirPage=no
 LicenseFile=..\LICENSE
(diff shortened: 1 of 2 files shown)


mkruselj:
We're rather gonna disclose the .vst3 location on the final page of the installer!

PR #3159, opened 2020-11-16
Oscillator sync state is always updated now (Classic and S&H Noise oscs)

### LINKED ISSUE #2737: Moving the OSC Sync parameter from 0% after hitting a note will often make a clicking noise
**Description:**
Moving the OSC Sync parameter from 0% after hitting a note will often make a slight clicking noise.

**Surge version:**
Latest Nightly (007c08d)
VST3 x64 in FL Studio x64

**To Reproduce:**
1. Load an init instance of Surge.
2. Press down a note
3. Move the Sync parameter in the OSC section slightly. As soon as you move it, you will often hear that slight clicking noise.

**Expected behavior**
It should increase smoothly.

Here's a video demonstrating the issue:
[Video.zip](https://github.com/surge-synthesizer/surge/files/5301957/Video.zip)

mkruselj:
Closes #2737

### DIFF
--- a/src/common/dsp/SampleAndHoldOscillator.cpp
+++ b/src/common/dsp/SampleAndHoldOscillator.cpp
@@ -368,8 +368,10 @@ void SampleAndHoldOscillator::process_block(
       {
          driftlfo[l] = drift_noise(driftlfo2[l]);
 
-         while ((syncstate[l] < a) || (oscstate[l] < a))
+         while (((l_sync.v > 0) && (syncstate[l] < a)) || (oscstate[l] < a))
+         {
             convolute(l, false, stereo);
+         }
 
          oscstate[l] -= a;
          syncstate[l] -= a;
--- a/src/common/dsp/SampleAndHoldOscillator.cpp
+++ b/src/common/dsp/SampleAndHoldOscillator.cpp
@@ -165,21 +165,31 @@ void SampleAndHoldOscillator::convolute(int voice, bool FM, bool stereo)
 
    const float p24 = (1 << 24);
    unsigned int ipos;
-   if (FM)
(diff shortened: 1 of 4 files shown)


src/common/dsp/SampleAndHoldOscillator.cpp
```diff
@@ -373,12 +368,13 @@ void SampleAndHoldOscillator::process_block(
       {
          driftlfo[l] = drift_noise(driftlfo2[l]);
 
-         while ((syncstate[l] < a) || (oscstate[l] < a))
+         while (((l_sync.v > 0) && (syncstate[l] < a)) || (oscstate[l] < a))
```
baconpaul, line 371: So why do you suppress convolution for l_sync.v == 0? This is clearly the first problem!

src/common/dsp/SurgeSuperOscillator.cpp
```diff
@@ -360,7 +355,7 @@ template <bool FM> void SurgeSuperOscillator::convolute(int voice, bool stereo)
    */
    unsigned int ipos;
 
-   if ((l_sync.v > 0) && (syncstate[voice] < oscstate[voice]))
+   if (syncstate[voice] < oscstate[voice])
```
baconpaul, line 358: I'm pretty worried about this one too. Are you sure that ss[v] < os[v] is false when l_sync.v == 0?

mkruselj: This one is actually fine, regtests don't seem to trip up over it. It's just S&H noise that is being tripped up (in absolute unison mode).

mkruselj: And I think I found a proper solution for the issue in the end!

PR #3162, opened 2020-11-16
 Fine-tuning of low bpm tempo sync overdraws

K0rrid0r:
It's perfect now. Tested in 10 bpm and in 500 bpm. Works flawlessly!


baconpaul:
Oh man -1.2 is super low. Happy to merge but we do need to remember that it means envelope displays on the SS about .3hz will be wrong. We OK with that? (I am).

K0rrid0r:
can you show me what you mean with a licecap? I cant see the issue.

baconpaul:
Turn an envelope on the step sequencer (dahdsr)
As you adjust the rate the envelope moves correctly
Below rate -1.2 it won’t 
That’s ok

baconpaul:
Anyway merging. We have to choose some threshold right?

PR #3172, opened 2020-11-18
Various code cleanups, Classic osc Width parameters renamed

mkruselj:
Add enums for all individual oscillator parameters and apply them in code
Rename WT2 to Window
Replace wt2_suboscs in Window with MAX_UNISON from globals
Rename "Airwindow" to "Airwindows" everywhere
For AW parameters that are unused, use generic "Param x" name
Make sure when loading different AW effects, if param number changes, that unused parameters are set to generic name


baconpaul:
Will review today or tomorrow am

PR #3178, opened 2020-11-19
Implement retrigger properly for all oscillator types

### LINKED ISSUE #3171: Oscillator Retrigger is not implemented for a number of oscillators
As it stands currently (before 1.8 release), oscillator retrigger parameter is only valid in Classic, Wavetable and S&H Noise oscillator types. All other oscillators are blissfully unaware of this parameter and will always retrigger from 0° on every new note played.

For 1.9, "the oscillator release", enable this parameter for all oscillator types, and handle streaming so that old patches using the affected oscillator types are loaded with Retrigger button enabled, if it wasn't.

mkruselj:
Closes #3171

### DIFF
--- a/src/common/Parameter.cpp
+++ b/src/common/Parameter.cpp
@@ -2199,9 +2199,6 @@ void Parameter::get_display(char* txt, bool external, float ef)
                   case fut_k35_hp:
                      sprintf(txt, "%s", fut_k35_subtypes[i]);
                      break;
-                  case fut_diode:
-                     sprintf(txt, "%s", fut_diode_subtypes[i]);
-                     break;
 #if SURGE_EXTRA_FILTERS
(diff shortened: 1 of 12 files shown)


mkruselj:
Questions for @baconpaul:

1. Should we add a very tiny quick fade-in for sine/FM2/FM3/window? It seems that WT osc has something that de-clicks its output even with retrigger off, so maybe others should match that?
2. Method to somehow set osc retrigger from the config.xml presets. Window, Sine, FM2/FM3 and WT should have it on by default when selecting osc type.

mkruselj:
Since clang-format did its thing to the codebase, and FM3 had some work done by @baconpaul which clashed with this PR, I am closing this one in favor of #3785 which is done from scratch.

PR #3181, opened 2020-11-19
Mousewheel on osc type doesn't share the starting point for all oscs/scenes anymore

mkruselj:


src/common/gui/CSnapshotMenu.cpp
```diff
@@ -248,6 +248,7 @@ COscMenu::COscMenu(const CRect& size,
 {
    strcpy(mtype, "osc");
    this->osc = osc;
```
baconpaul, line 250: You don’t need this storage object

src/common/gui/CSnapshotMenu.cpp
```diff
@@ -278,8 +279,12 @@ void COscMenu::loadSnapshot(int type, TiXmlElement* e, int idx)
 {
    assert(within_range(0, type, num_osctypes));
    auto sge = dynamic_cast<SurgeGUIEditor *>(listenerNotForParent);
-   if( sge )
-      sge->oscilatorMenuIndex = idx;
+   if (sge)
+   {
+      auto des = storage->getPatch().dawExtraState;
```
baconpaul, line 284: Because you don’t need this state object

Issue #2397, opened 2020-07-30
"Failed to zoom VST3" message and crash when opening with custom zoom value in Studio One Pro 5


anaveragebear:
**Describe the bug**
As stated in the title, after setting a default zoom value in Surge, my computer spits out the above message and my DAW crashes on attempted reopen.

**Please let us know your surge version**
1.7.0
VST3 // 64-bit

**To Reproduce**
1. Manually enter a default value in the Surge settings menu
2. Exit the program
3. Attempt to reopen the program
4. Bug

**Expected behavior**
An infinitely repeating text box containing the text in the title, GUI does not load, DAW freezes
![Screenshot (61)](https://user-images.githubusercontent.com/68977809/88869256-45abfc00-d1e0-11ea-8672-42d7d3e3333d.png)

**Desktop (please complete the following information):**
 - OS: Windows 10 Pro v1909
 - Host: Studio One Pro 5.0.0.59360 64-bit

**Additional context**
I am a lowly music producer without a lick of coding knowledge but I love this synth and wanted to help improve it. Cheers

baconpaul:
Thanks for the report!

@mkruselj do you have studio one? Can you see if you can reproduce this in the next few days?

@anaveragebear if you don't zoom does surge work? (I know that's not a solution but it is at least a workaround!). If not, you can always grab the previous release (1.6.6) from our website.

This error will appear once if your host doesn't respond with "True" from the vst3 "resize" message. I think it happens over and over because we've never actually seen that happen before so I bet we keep on trying. If we can reproduce it we can fix that loop case.

baconpaul:
Oh and one more thing @anaveragebear - if you go to your %DOCUMENTS%\Surge directory there will be a file called "SurgeUserDefaults.xml". If you just delete that file, it will undo the default setting and you should be good to go at 100% again. (Which will be tiny but at least will work for now).

anaveragebear:
Thanks @baconpaul I'll be sure to do that.

mkruselj:
I do have S1 but v4 not v5...

mkruselj:
In v4 I don't get the endless repeating dialog - but I do get the same message if you try to zoom outside of screen boundaries (when Surge warns you that won't work and it auto-reduces the zoom level to fit at largest possible size for the current resolution).

mkruselj:
[...]

Issue #1951, opened 2020-05-26
Investigate the VST3 Scale Factor API


baconpaul:
https://steinbergmedia.github.io/vst3_doc/base/classSteinberg_1_1IPlugViewContentScaleSupport.html

This could help us a lot with some of our windows off-by-one pixels boogers i bet if the DAWs support it.

Basically use a call to this API to modify the backing store size on windows and maybe on Linux too.

baconpaul:
So I investigated briefly and alas reaper windows doesn't seem to call me, so push this out of 1.7 beta 1 until I feel like looking in more detail

Kingston1:
Cubase 11 Pro reports this correctly but not quite as expected. The result from setContentScaleFactor on the first few calls that the host sends are 1.0. Eventually it calls with the correct scale factor.

baconpaul:
Well that is certainly consistent with the bug we are seeing in that DAW.

Looks like an “if hostname == “Cubase”” is in our future in SurgeGUIEditor I guess :(

Kingston1:
Well, scaleFactor at least is consistently broken. On opening the UI (both reopen and initial open) the scaleFactor will be 1.0 for the first few calls.

On the other hand SurgeGUIEditor::onSize(Steinberg::ViewRect* newSize) will be different on initial open vs. reopen.

Kingston1:
Due to zoom stabilize across the board in https://github.com/surge-synthesizer/surge/pull/3210 this issue can probably be closed.

The scale factor that cubase reports is not really useful for anything and is not needed for even high DPI scenarios. Windows size simply works now, despite DPI setting.

baconpaul:
Yeah we implement and understand this now. Can close. Thanks so much @Kingston1 !

PR #3201, opened 2020-11-22
Bugfix/1460

Kingston1:


src/common/gui/SurgeGUIEditor.h
```diff
@@ -126,6 +130,12 @@ class SurgeGUIEditor : public EditorType,
 
    virtual Steinberg::tresult PLUGIN_API onSize(Steinberg::ViewRect* newSize) override;
    virtual Steinberg::tresult PLUGIN_API checkSizeConstraint(Steinberg::ViewRect* newSize) override;
+   virtual Steinberg::tresult PLUGIN_API setContentScaleFactor(ScaleFactor factor) override
+   {
+      scaleFactor = factor;
```
Kingston1, line 135: It will be left for future refactoring to properly use this scaleFactor, if that's even possible. Now this is just used to resolve wrong GUI size from host on initial open vs. reopen. More info here: https://github.com/surge-synthesizer/surge/issues/1951

src/vst3/SurgeVst3Processor.cpp
```diff
@@ -1007,12 +1007,13 @@ void SurgeVst3Processor::handleZoom(SurgeGUIEditor *e)
...
             {
-               std::ostringstream oss;
+			   // Leaving this here for debug purposes. resizeView() can fail in non-fatal way and zoom reset is just too harsh.
+               /*std::ostringstream oss;
                oss << "Your host failed to zoom VST3 to " << e->getZoomFactor() << " scale. "
                    << "Surge will now attempt to reset the zoom level to 100%. You may see several "
                    << "other error messages in the course of this being resolved.";
                Surge::UserInteractions::promptError(oss.str(), "VST3 Host Zoom Error" );
-               e->setZoomFactor(100);
+               e->setZoomFactor(100);*/
```
Kingston1, line 1015: Not sure if this actually fixes a problem in some host. In it's not necessary at all and in fact breaks resize when playing with the zoom menu.

baconpaul: That code is there to fix Studio One. https://github.com/surge-synthesizer/surge/issues/2397

We will need to retest this change carefully in that DAW

src/common/gui/SurgeGUIEditor.cpp
```diff
@@ -5828,8 +5830,19 @@ VSTGUI::CCommandMenuItem* SurgeGUIEditor::addCallbackMenu(VSTGUI::COptionMenu* t
 #if TARGET_VST3
 Steinberg::tresult PLUGIN_API SurgeGUIEditor::onSize(Steinberg::ViewRect* newSize)
 {
-   float izfx = newSize->getWidth() * 1.0 / getWindowSizeX() * 100.0;
-   float izfy = newSize->getHeight() * 1.0 / getWindowSizeY() * 100.0;
+   Steinberg::ViewRect currentSize = *newSize;
```
baconpaul, line 5833: So implicit here is that currentSize.top == 0 and currentSize.left == 0. I am sure they are but do we want to check somehow below (since you move the bottom and assume that is changing the width).

src/vst3/SurgeVst3Processor.cpp
Kingston1: Maybe this is the part where daw == cubase detection is needed then. I can add it.

baconpaul:
@Kingston1 tested on cubase

Testing Checklist for @mkruselj and @baconpaul 

- [ ] macOS (bp)
      - [ ] Reaper
      - [ ] Bitwig 3.3
      - [ ] Live
      - [ ] Cubase (if I can get my NFR licenses running)
- [ ] Linux (bp)
      - [ ] reaper
      - [ ] bitwig 3.3
      - [ ] juce plugin host
      - [ ] Ardour 6.4 or thereafter
- [ ] Windows (ed)
       - [ ] Reaper
       - [ ] Live
       - [ ] Studio One

src/common/gui/SurgeGUIEditor.cpp
Kingston1: Didn't seem like those values were used at all.

src/vst3/SurgeVst3Processor.cpp
baconpaul: lets see what @mkruselj sees when he tests it first. We could also put an “if studio one” just as easily.

src/common/gui/SurgeGUIEditor.cpp
baconpaul: They are used below when you can currentSize->getWidth() which is right - left.

baconpaul:
OK mac shows the same behavior main vs your branch.
That said, there's a bug in main in bitwig which is #3202. But that doesn't stop us merging this.

baconpaul:
From my tests this seems to behave just like main on mac and linux hosts. (Although main has bugs!)

mkruselj:
@baconpaul OK, did the test and there is no zoom dance in any of those hosts (plus some others that I have, like MuLab and Reason and FL20).

However, it still happens that the attempted overly large zoom level is stored in DAW state. This needs to be updated when constraining the zoom level so that it doesn't pop an error when opening the project (which currently happens in Studio One and Live, even if plugin window is not opened).

src/vst3/SurgeVst3Processor.cpp
mkruselj: S1 seems good without it now!

PR #3203, opened 2020-11-23
Update zoom factor stored in DAW state after fixing too large zoom attempt

mkruselj:
Previously, if we tried setting zoom level to a large value which wasn't possible on our current display, Surge would fix it and find largest fitting zoom level. But DAW state wasn't informed about this, so every time after closing then reopening Surge's editor it would attempt that overly large zoom again, and pop up a warning for the user that zoom level was adjusted. It gets annoying real fast.

Fix this by just updating the DAW state `instanceZoomFactor` after recursively finding the largest fitting zoom level in `SurgeGUIEditor::setZoomFactor()`.

Also, clean up some indents 'n' stuff. Because why not.


baconpaul:
Oh is there any way you can make a one line change a one line diff? All that reformatting makes thisnappear trrrifyingly large

mkruselj:
@baconpaul Done it!

baconpaul:
Great! Happy to merge when CI completes. Thank you

PR #3210, opened 2020-11-23
Fixed window zoom/resize/scaling API usage (infinite loops etc. window reopen issues)

Kingston1:
This now works perfectly on Cubase but requires more work for AU and possibly VST2. I'll get to that next.

### DIFF
--- a/src/common/gui/SurgeGUIEditor.cpp
+++ b/src/common/gui/SurgeGUIEditor.cpp
@@ -5838,6 +5838,8 @@ bool SurgeGUIEditor::initialZoom()
         // Daw does not necessarily have any idea what size to draw on init. We need to tell it.
         setZoomFactor(initialZoomFactor);
         initialZoomFactor = 100;
+
+        // Important: this is the one and only window resize operation allowed inside onSize. Infinite recursion warning!
         zoom_callback(this, true);
         return true; 
     }
--- a/src/common/gui/SurgeGUIEditor.cpp
+++ b/src/common/gui/SurgeGUIEditor.cpp
@@ -275,21 +275,19 @@ SurgeGUIEditor::SurgeGUIEditor(PARENT_PLUGIN_TYPE* effect, SurgeSynthesizer* syn
    mod_editor = false;
 
    // init the size of the plugin
-   int userDefaultZoomFactor = Surge::Storage::getUserDefaultValue(&(synth->storage), "defaultZoom", 100);
-   float zf = userDefaultZoomFactor * 0.01;
-
-   auto izf = synth->storage.getPatch().dawExtraState.editor.instanceZoomFactor;
-   if(izf != 100)
+   initialZoomFactor = Surge::Storage::getUserDefaultValue(&(synth->storage), "defaultZoom", 100);
+   auto instanceZoomFactor = synth->storage.getPatch().dawExtraState.editor.instanceZoomFactor;
+   if(instanceZoomFactor != 100)
    {
-       // If I restore state before I am constructed I need to do this
-       zf = izf * 0.01;
+	   // dawExtraState zoomFactor wins defaultZoom
+	   initialZoomFactor = instanceZoomFactor;
    }
 
    rect.left = 0;
    rect.top = 0;
 #if TARGET_VST2 || TARGET_VST3
-   rect.right = getWindowSizeX() * zf;
-   rect.bottom = getWindowSizeY() * zf;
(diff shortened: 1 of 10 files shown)


src/common/gui/SurgeGUIEditor.cpp
```diff
@@ -5833,39 +5835,40 @@ VSTGUI::CCommandMenuItem* SurgeGUIEditor::addCallbackMenu(VSTGUI::COptionMenu* t
...
+   // If the resolved currentZoomFactor changes size by more than a pixel, store new size
+   auto zfdx = std::fabs( (currentZoomFactor - zoomFactor) * getWindowSizeX() ) / 100.0;
+   auto zfdy = std::fabs( (currentZoomFactor - zoomFactor) * getWindowSizeY() ) / 100.0;
+   bool windowDragResize = std::max( zfdx, zfdy ) > 1;
 
-   if( zfd > 1 )
+   if(windowDragResize)
    {
-      setZoomFactor( izf );
+	  printf("SurgeGUIEditor::onSize setZoomFactor %f\n", currentZoomFactor);
```
baconpaul, line 5870: Probably don't want to leave this printf in here obvs

Kingston1: whoops

Kingston1:
VST2 now done. I see from the CI the very same build problem VST2 also had. Will set up my build on macOS and fix it there instead of coasting on the CI.

baconpaul:
The LV2 will have the same problem
You used to be able to actually build the LV2 on mac if you want (if you don't have a linux vm kicking around) by just setting BUILD_LV2 to true in the cmake. but it doesn't work so i leave it off of course.

baconpaul:
Doing a few confirms here now. Have a couple of minutes.

baconpaul:
Screen Shot 2020-11-24 at 2 04 32 PM

Bummer - I pull up a saved AU session on HostingAU and it gets the zoom wrong.

So definitely: lets not merge this.

baconpaul:
(yet of course - while we find it).

baconpaul:
Right so the problem is even if we have a saved zoom state in the DAW state, the default zoom wins

baconpaul:
OK here's a way to repro

1. http://ju-x.com/hostingau.html install hosting au
2. start it and load surge and see it at default zoom
3. set the zoom to 150% or what not (basically any non-default) 
4. save that hosting au session
5. quite HAU
6. Load that hosting au session 

at this point the window will be sized to the default but the contents will be sized to saved size so the UI will be truncated

baconpaul:
Cool that push fixed the AU. Lemme do a few more tests quickly

src/common/gui/SurgeGUIEditor.cpp
```diff
@@ -5197,14 +5199,14 @@ VSTGUI::COptionMenu* SurgeGUIEditor::makeZoomMenu(VSTGUI::CRect& menuRect, bool
...
+                          snprintf(c, 256, "%f", zoomFactor);
                           promptForMiniEdit(
                               c, "Enter a default zoom level value:", "Set Default Zoom Level",
                               menuRect.getTopLeft(), [this](const std::string& s) {
                                  int newVal = ::atoi(s.c_str());
-                                 Surge::Storage::updateUserDefaultValue(&(this->synth->storage),
+                                 Surge::Storage::updateUserDefaultValue(&(synth->storage),
                                                                         "defaultZoom", newVal);
-                                 this->setZoomFactor(newVal);
+								 resizeWindow(newVal);
```
baconpaul, line 5209: Looks like there's some hard tabs here? Of course when we fix the clang format this will go away but just wanted to point out you have editor set with them. Easy to fix as well. If this is my only comment we can merge and clean up later.

src/vst2/Vst2PluginInstance.cpp
```diff
@@ -705,20 +705,23 @@ bool Vst2PluginInstance::tryInit()
...
 {
-    ERect *vr;
     float fzf = e->getZoomFactor() / 100.0;
-    int newW = e->getWindowSizeX() * fzf;
-    int newH = e->getWindowSizeY() * fzf;
-    sizeWindow( newW, newH );
+    int width = e->getWindowSizeX() * fzf;
+    int heigth = e->getWindowSizeY() * fzf;
+
+	if (resizeWindow)
```
baconpaul, line 721: another hard tab

PR #3216, opened 2020-11-23
Fine-tune Shift+drag speed in MSEG editor

mkruselj:


baconpaul:
Ha! I had .1 and thought “evil will think that’s too fast’ so took it to .05. Guess I was exactly wrong

mkruselj:
It happens. :)

It was definitely way too slow at 0.05!

PR #3220, opened 2020-11-24
Disallow zooming and panning in MSEG editor LFO mode

### LINKED ISSUE #2966: MSEG edit mode: LFO vs envelope

mkruselj:
Closes #2966 

This was the only thing left from LFO mode implementation.

### DIFF
--- a/src/common/gui/MSEGEditor.cpp
+++ b/src/common/gui/MSEGEditor.cpp
@@ -1292,25 +1292,29 @@ struct MSEGCanvas : public CControl, public Surge::UI::SkinConsumingComponent, p
          }
          else
          {
-            // This means we are a pan or zoom gesture
-            float x = where.x - mouseDownOrigin.x;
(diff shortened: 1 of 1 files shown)


baconpaul:
Why do we not want panning and zooming in LFO mode?

baconpaul:
Chatted n discord. We may add this but bp thinks it could be useful to keep once pan/zoom works so lets close this unmerged now and if pan/zoom is good in lfo mode we will not add it etc.

PR #3227, opened 2020-11-24
Tweak MSEG RMB context menus

mkruselj:
Also make "Edit Value:" use toOSCaseForMenu everywhere

### DIFF
--- a/src/common/gui/MSEGEditor.cpp
+++ b/src/common/gui/MSEGEditor.cpp
@@ -1718,54 +1718,65 @@ int32_t MSEGControlRegion::controlModifierClicked(CControl* pControl, CButtonSta
 
    bool isOnOff = false;
    std::string menuName = "";
-   switch( tag )
+   switch (tag)
    {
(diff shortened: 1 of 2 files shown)


src/common/gui/MSEGEditor.cpp
```diff
@@ -1800,9 +1811,8 @@ int32_t MSEGControlRegion::controlModifierClicked(CControl* pControl, CButtonSta
          {
             auto val = op.second;
             auto men = addcb(op.first, [val, pControl, this]() {
-               pControl->setValue(val);
-               canvas->invalid();
-               invalid();
+               pControl->setValue(Parameter::intScaledToFloat(val, 100, 1));
```
baconpaul, line 1814: Uhhh ... this is wrong. What about the controls which aren't 100->1 CNumberFields where the value is 0.5?

mkruselj: Fixing based on our Discord chat! :)

PR #3246, opened 2020-11-26
Center mouse cursor to controls only when mouse is moved

### LINKED ISSUE #3139: Jumpy mouse cursor in certain cases
jumpy_cursor

When clicking on slider tray when menu is open, cursor jumps. The other one with info window is trickier to get but also happens. Same class of issue I bet!

Kingston1:
Closes #3139

### DIFF
--- a/src/common/gui/CSurgeSlider.cpp
+++ b/src/common/gui/CSurgeSlider.cpp
@@ -683,6 +683,9 @@ bool CSurgeSlider::isInMouseInteraction()
 
 CMouseEventResult CSurgeSlider::onMouseDown(CPoint& where, const CButtonState& buttons)
 {
+   startPosition = where;
+   currentPosition = where;
+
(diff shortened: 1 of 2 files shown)


src/common/gui/CSurgeSlider.cpp
```diff
@@ -777,11 +780,16 @@ CMouseEventResult CSurgeSlider::onMouseUp(CPoint& where, const CButtonState& but
       
       edit_value = nullptr;
 
-
-      if( resetPosition )
+      if (resetPosition &&
+          startPosition != currentPosition)
```
Kingston1, line 784: This comparison could be a threshold, but after using it like this for a while, it feels most natural like this.

Kingston1:
Closes https://github.com/surge-synthesizer/surge/pull/3246

PR #3258, opened 2020-11-27
Add Nonlinear Feedback Lowpass Filter

selenologist:
Adds a lowpass filter based on https://jatinchowdhury18.medium.com/complex-nonlinearities-episode-5-nonlinear-feedback-filters-115e65fc0402

Subtypes are the number of chained biquad stages in the filter. Unfortunately it won't self-resonate even if you crank the Q factor, but with four stages it's pretty resonant.

Could maybe use some tweaking, but let's maybe get some beta testers on this and see what happens?

### DIFF
--- a/src/common/dsp/filters/NonlinearFeedback.cpp
+++ b/src/common/dsp/filters/NonlinearFeedback.cpp
@@ -41,7 +41,7 @@ static inline __m128 doLpf(
    noexcept
 {
    // out = z1 + b0 * input
-   const __m128 out = z1 + M(b0, input);
+   const __m128 out = A(z1, M(b0, input));
    // nonlinear feedback = tanh(out)
    const __m128 nf  = Surge::DSP::fasttanhSSEclamped(out);
    // z1 = z2 + b1 * input - a1 * nf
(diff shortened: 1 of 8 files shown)


src/common/dsp/filters/NonlinearFeedback.cpp
```diff
@@ -0,0 +1,155 @@
+#include "globals.h"
+#include "K35.h"
```
baconpaul, line 2: Why this include?

src/common/dsp/filters/NonlinearFeedback.cpp
```diff
@@ -0,0 +1,155 @@
...
+      
+      C[nlf_a1] = 2.0 * (1.0 - phi) / a0;
+      C[nlf_a2] = (phi - K + 1.0) / a0;
+      C[nlf_b0] = 1.0 / a0;
+      C[nlf_b1] = 2.0 / a0;
+
+      cm->FromDirect(C);
+   }
+
+#define process_coeffs() \
```
baconpaul, line 93: () on a define! Any reason to not just have an inline?

PR #3278, opened 2020-11-28
Disable overwrite prompt in Bitwig on Linux

bgins:
The overwrite patch prompt causes the Bitwig UI to hang in Bitwig 3.3. Skip the prompt for now.

### DIFF
--- a/src/common/SurgeSynthesizerIO.cpp
+++ b/src/common/SurgeSynthesizerIO.cpp
@@ -378,7 +378,12 @@ void SurgeSynthesizer::savePatch()
    fs::path filename = savepath;
    filename /= string_to_path(legalname + ".fxp");
 
-   if (fs::exists(filename))
+   bool checkExists = true;
+#if LINUX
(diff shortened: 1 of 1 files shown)


src/common/SurgeSynthesizerIO.cpp
```diff
@@ -378,7 +378,12 @@ void SurgeSynthesizer::savePatch()
    fs::path filename = savepath;
    filename /= string_to_path(legalname + ".fxp");
 
-   if (fs::exists(filename))
+   bool checkExists = true;
+#if LINUX
+    // Overwrite prompt hangs UI in Bitwig 3.3
```
baconpaul, line 383: Does “bitwig” match case on Linux? Just checking

Thanks!!

bgins: Yep! I tried both "Bitwig" and "bitwig" and all lowercase turned out to be the one.

baconpaul: Great thanks. Will merge when ci is done

Issue #2446, opened 2020-08-07
Expand FX engine with more slots


mkruselj:
Once the parameter list ordeal is hopefully resolved, we could think about adding more FX slots, like so:

![image](https://user-images.githubusercontent.com/2393720/109390734-a5ec5300-7913-11eb-8000-6e9ac2e4b3e5.png)

I doubt we need more than 2 send FX, but that can be open for discussion.

magnetophon:
Yes, this is great!
4 sends would also be very welcome!

mkruselj:
That creates some UI layout issues, I would much rather keep sends at 2. We need the UI space for other more important params - more filter and osc params, more FX params...)

baconpaul:
So @mkruselj reading this issue: Seems like we are discussing putting 4 new sliders (2 more send levels and 2 more return levels) here. I know how to code those but before I do: Considering we aren't resizing the UI in XT 1.0 where would you put them?

Note if we develop a knob we could answer this question pretty easily. Happy to do so also.

mkruselj:
Same thing as waveshapers. Keep 2 sends for now then add 2 more after XT 1.0?

baconpaul:
yeah i keep running out of sends but i suppose that's the answer.

baconpaul:
I want to take a shot at this one pretty soon. Remind me: Did we want to stay at 2 sends in XT or are we going to split/mouse the sliders or squeeze in a bit more UI space?

We could also go with 4 knobs (and then use knobs for the filters also) as an approach. The knob UI itself is pretty easy the way I have the code set up.

mkruselj:
I'm down with 4 sends, if sliders related to sends 3/4 swap places with 1/2 when you select send FX slot 3 or 4. Should be fine for the time being, let's do any medium to major skin overhauls in XT 2.0?

baconpaul:
OK so merged all the slots but the back 2 sends don't work. Here's my rough plan based on chatting on discord with @mkruselj 

1. Update SceneStorage so that sends goes from [2] to [n_sends] or some such
2. Assign them in surge patch but have 3 and 4 have a special property called 'togglableWith({other-control}).toggleState(false)' or some such and have 1 and 2 have .togglableWith({}).toggleState(true)'
3. Do the same for the return sliders (although we don't need params for them since each return is a param on the FX we may need to assign the params since I think the sends are specially assigned in surge patch)
[...]

PR #3393, opened 2020-12-10
Apply n_fx_slots instead of magic number 8 in various FX related loops

mkruselj:
In preparation for #2446

### DIFF
--- a/src/common/gui/SurgeGUIEditor.cpp
+++ b/src/common/gui/SurgeGUIEditor.cpp
@@ -317,7 +317,7 @@ SurgeGUIEditor::SurgeGUIEditor(PARENT_PLUGIN_TYPE* effect, SurgeSynthesizer* syn
    vu[15] = 0;
    lfodisplay = 0;
    fxmenu = 0;
-   for( int i=0; i<8; ++i )
+   for (int i = 0; i < n_fx_slots; ++i)
    {
       selectedFX[i] = -1;
       fxPresetName[i] = "";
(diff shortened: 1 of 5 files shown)


src/common/gui/SurgeGUIEditor.cpp
```diff
@@ -668,7 +668,7 @@ void SurgeGUIEditor::idle()
          }
       }
 
-      for (int i = 0; i < 8; i++)
+      for (int i = 0; i < n_fx_slots; i++)
```
baconpaul, line 671: This 8 is not n_fx_slots. It is the size of the ctrl_refresh_queue which doesn't have a constant.

src/common/gui/SurgeGUIEditor.cpp
```diff
@@ -1268,7 +1268,7 @@ void SurgeGUIEditor::openOrRecreateEditor()
    {
       auto fxpp = currentSkin->getOrCreateControlForConnector("fx.param.panel");
       CRect fxRect = CRect( CPoint( fxpp->x, fxpp->y ), CPoint( 123, 13 ) );
-      for (int i = 0; i < 8; i++)
+      for (int i = 0; i < n_fx_slots; i++)
```
baconpaul, line 1271: This 8 is not n_fx_slots. In fact I'm not entirely sure why it is 8 (other than the only usage of this caps at 3). The size of the array we are writing to is 16 and we write at index + 1 so this should probably be 15.

PR #3408, opened 2020-12-12
Various loop marker/region positioning and drawing tweaks

mkruselj:
Also adjusted skin colors for loop region for both skins

### DIFF
--- a/src/common/gui/MSEGEditor.cpp
+++ b/src/common/gui/MSEGEditor.cpp
@@ -323,10 +323,10 @@ struct MSEGCanvas : public CControl, public Surge::UI::SkinConsumingComponent, p
          {
             int lmSize = 10;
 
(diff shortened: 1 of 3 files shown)


src/common/gui/MSEGEditor.cpp
```diff
@@ -1118,13 +1131,12 @@ struct MSEGCanvas : public CControl, public Surge::UI::SkinConsumingComponent, p
...
+               auto cx = r.left;
 
-               /*dc->setLineWidth(1);
-               dc->setFrameColor(kRedCColor);
-               dc->drawRect(r, kDrawStroked);*/
+               if (h.zoneSubType == hotzone::LOOP_END)
+                  cx = r.right;
 
-               if (cx >= drawArea.left && cx <= drawArea.right)
+               if (cx >= drawArea.left || cx <= drawArea.right)
```
baconpaul, line 1139: This or to and swap is i think a mistake no?

mkruselj: Yep!

PR #3416, opened 2020-12-13
Add makeup gain for nonlinear feedback filters

jatinchowdhury18:
The nonlinear feedback filters had an issue where the cutoff frequency drastically affected the output volume (most noticeable when using multiple filter stages). I've added a "makeup gain" to the lowpass and highpass filters, that depends on the cutoff frequency. Off the top of my head, I can't think of a similar work-around for the bandpass or notch filters. Hoping for some second (and third) opinions on how this sounds in different use cases.

From a code perspective, this fix definitely feels a bit hack-y at the moment, but I'm more worried about getting it to sound right before trying to get the code perfect. Still, please let me know if I haven't quite nailed the Surge style, or if there are preferred functions to use over `std::pow`, `std::max`, etc.

Thanks,
Jatin

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -6,6 +6,7 @@ Authors in the repository since 1.6 alpha was open-sourced (in alphabetical orde
 David Bata <oddy.o.trax@gmail.com>
 Alexandre Bique <bique.alexandre@gmail.com>
 Hendrik van Boetzelaer <http://opuswerk.com>
+Jatin Chowdhury <jatin@ccrma.stanford.edu>
 Jean Pierre Cimalando <jp-dev@inbox.ru>
 Filipe Coelho <falktx@falktx.com>
 Alec DiAstra <alecdiastra@gmail.com>
--- a/src/common/dsp/filters/NonlinearFeedback.cpp
+++ b/src/common/dsp/filters/NonlinearFeedback.cpp
@@ -81,6 +81,7 @@ static inline __m128 doNLFilter(
       const __m128 b0,
       const __m128 b1,
       const __m128 b2,
+      const __m128 makeup,
       const int sat,
       __m128 &z1,
       __m128 &z2)
@@ -110,7 +111,7 @@ static inline __m128 doNLFilter(
    z1 = A(z2, S(M(b1, input), M(a1, nf)));
    // z2 = b2 * input - a2 * nf
    z2 = S(M(b2, input), M(a2, nf));
-   return out;
+   return M(out, makeup);
 }
 
 namespace NonlinearFeedbackFilter
@@ -121,6 +122,7 @@ namespace NonlinearFeedbackFilter
       nlf_b0,
       nlf_b1,
       nlf_b2,
+      nlf_makeup,
       n_nlf_coeff
    };
 
@@ -141,7 +143,8 @@ namespace NonlinearFeedbackFilter
 
       const float q = ((reso * reso * reso) * 18.0f + 0.1f);
 
-      const float wc = 2.0f * M_PI * clampedFrequency(freq, storage) / dsamplerate_os;
+      const float normalisedFreq = clampedFrequency(freq, storage) / dsamplerate_os;
+      const float wc = 2.0f * M_PI * normalisedFreq;
 
       const float wsin  = Surge::DSP::fastsin(wc);
       const float wcos  = Surge::DSP::fastcos(wc);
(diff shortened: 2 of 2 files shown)


baconpaul:
Hey cool thank you!

std::pow and std::max are fine there (those spots are called only every block not every sample and then interpolated). Obviously in the SSE code it would have been different. Lemme grab the branch and play it quickly! Oh also, we generally ask people to add themselves alphabetically to the AUTHORS file in the manner they would like to be added when they do a first commit, if you would like to?

baconpaul:
OK so I played with this a bit and it definitely solves the problem at hand. But I think it may be just a bit too loud overall.  This patch shows how really loud the NLF in state 12 is now, for instance (top tip: once you unzip that you can just drag and drop the fxp onto surge gui from desktop on mac and win; on linux you need to put it in ~/Documents/Surge and rescan)

[nlf.zip](https://github.com/surge-synthesizer/surge/files/5685286/nlf.zip)

I think (think) this is because of your makeup. So right makeup goes like 1/(norm freq^1/3) cool. But normalizedFrequency is frequency / samplerate (so the makeup gain os SR dependent which I think is correct) but also means that you get q lot of makeup gain at lower frequencies. In that particular patch, the makeup gain is about 7 or so in the quiesced period if you play in the 3rd octave of so.

If I change it so the makeup gain has the same functional form but basically just reduce the makeup gain by a factor of 8 or so (so change it so it is 0.125f/pow etc...) then not only is it uniform across the range but also it doesn't blow out volume compared to the other filters.

baconpaul:
oh and: man this filter sounds awesome in that patch. Totally different from all the other ones. So excited to have it in the release!

mkruselj:
I'm hoping something could be figured out for bandpass and notch, though! (Grabbing dinner then checking this branch!)

mkruselj:
OK so my opinion is that Jatin's gain normalization is great for subtype 1, but there needs to be additional normalization for all the subtypes (which would I suppose change the 1.0 in 1.0 / pow() to a lower number), and then we are golden.

Turns out bandpass, notch and allpass don't need this frequency-dependant gain normalization, but they do need a little bit of subtype normalization.

baconpaul:
OK cool it is super easy for me to write a little program over in headless which runs it and finds reasonable calibrations vs an RMS against a saw. Will do that next and report back.

jatinchowdhury18:
> OK so my opinion is that Jatin's gain normalization is great for subtype 1, but there needs to be additional normalization for all the subtypes (which would I suppose change the 1.0 in 1.0 / pow() to a lower number), and then we are golden.

Would you mind explaining what "subtypes" means in this context? Sorry if I'm not caught up with all the Surge nomenclature.

> OK cool it is super easy for me to write a little program over in headless which runs it and finds reasonable calibrations vs an RMS against a saw. Will do that next and report back.

Whoa, that sounds pretty neat! Just for my own personal curiosity, would you mind sharing how you do that?

Thanks again,
Jatin

baconpaul:
> Whoa, that sounds pretty neat! Just for my own personal curiosity, would you mind sharing how you do that?

even better - i'll grab your branch and add the normalization test.

Subtypes in surge - so each filter has a bunch of different types inside. It is the number graphic that says 1 in a box. Right mouse on it and you can see all the different types.

For instance in the vintage ladder we implemented two different non linear models and added bass compensation so there's 4 'sub'types

For these filters we have taps at each of the stages and also have 3 different wave shapers so 12 types. My patch used type 12 (which is all 4 stages with an asymmetric shaper).

baconpaul:
OK I did this.

https://github.com/surge-synthesizer/surge/compare/main...baconpaul:nl-feedback

basically what I did was create a little program in the headless area which sets up surge and then runs (for low pass) the OBXD 4 pole vs the NLF and for (high pass) the built in HP24. It then figures out the RMS ratio of playing a chord for 2 seconds between the filters and splits out a normalization table. I then coded this normalization table.

When I play the synth with these normalizations in all the modes it works great.

If you can check out `baconpaul/nl-feedback` branch I think its probably how we want to go

Oh and basically to run the calibrator

1. turn off the useNormalization in the filter (don't forget that!)
2. Build the surge-headless target
3. Edit the HeadlessNonTestFunctions.cpp to set it to LP or HP.
4. run `surge-headless --no-test --generate-nlf-norms`

That shows you a bunch of ways to directly manipulate a surge engine and play it too, which I know was your real question. It's roughly what the plugin wrappers use to talk to Surge (although there are some API shortcuts I take in headless basically manipulating the patch directly rather than calling the set value function which if you are interested best discussed on discord or some such).

Lemme know how that branch sounds. 

Oh and here's the modified version fo the patch I used to test.

[NLF2.zip](https://github.com/surge-synthesizer/surge/files/5685854/NLF2.zip)

jatinchowdhury18:
The updated version on your branch sounds good to me! That patch is pretty wicked too :).

Thanks for explaining the filter subtypes, I had played around with those in the plugin, but didn't quite make the connection with the name.

Having the "headless" target is pretty neat... thanks for walking through that process as well. I'll probably end up using a similar pattern for some future projects, especially now that I'm using CMake more regularly.

baconpaul:
Cool I will get it merged then (which will involve making a different PR so let me close this one when I do that one). 

We swapped to cmake in April of this year - and I basically learned it then. Over the intervening months I have come to like it very much especially now Juce 6 uses it so well. It also integrates fantastically with CLion. Basically after 25+ years of ‘gnu make and emacs is the best’, I found that CMake + CLion are actually an upgrade (CLion is not free tho). Anyway happy to help with any questions etc... if you want.

PR #3431, opened 2020-12-14
Tweak gain compensation for NLFB LP/HP manually

mkruselj:
Rename NLFB/NLST filters to Cutoff Warp and Resonance Warp

The goal of manually adjusted gain normalization factors was to get the filter subtypes sound reasonable at default filter settings (Cutoff 50%, Resonance 0%)
Of course, resonance on NLFB LP especially will boost the volume an awful lot, so this either needs another round of gain compensations or we leave it to the users to adjust manually...


src/common/dsp/filters/NonlinearFeedback.cpp
```diff
@@ -163,57 +163,68 @@ namespace NonlinearFeedbackFilter
...
-          0.235227
+          0.813073,
+          0.962984,
+          0.865058,
+          0.808567,
+          0.774069,
+          0.885141,
+          0.786042,
+          0.734219,
+          0.700227
```
baconpaul, line 180: wow this is super big change. Please don't merge until I can try it.

PR #3442, opened 2020-12-15
Update nonlinear feedback filter normalization

jatinchowdhury18:
A few folks had pointed out how the previous normalization for nonlinear feedback filters could create some "unstable" settings, particularly for highpass filters. I adjusted the `normalizedFreq` to be between [0, 1], and tweaked the exponential factor, which fixed that issue, and then re-ran the script to generate the makeup gains. For whatever reason, I wasn't able to generate makeup gains for the HPF that I was happy with, but I think the LPF gains work pretty well for both. Could probably use a bit more testing, and second or third opinion on how they sound.

Also, I made `useNormalization` a `constexpr` since that should set at compile-time.


baconpaul:
awesome. giving it a whirl now. thank yo so much

baconpaul:
sounds great on my test playing it now and comparing it with obxd and vintage.

PR #3449, opened 2020-12-16
add inigo kennedy patches for v1.8

inigokennedy:
Some new patches ahead of the v1.8 release. New patches are sorted into appropriate subdirectories per previous reorganisation. Updated a number of previous patches also - mostly level balancing.


baconpaul:
Thank you!

Issue #3438, opened 2020-12-15
NLF LP HP Callibration


baconpaul:
As discussed on discord
we have the tools
now we need the functional form and constants
Just a placeholder in 18 so I don't forget

PR #3451, opened 2020-12-17
NL Feedback filters: add resonance makeup term for OJD subtypes

jatinchowdhury18:
Addresses #3438.

Previously, when using the Cutoff Warp filters with the OJD waveshaper, the output volume was highly dependent on the resonance control. Here I've added another normalization factor make that less drastic.

Could probably use a bit more testing, in particular using a HPF with the cutoff and resonance controls all the way down can be a little finnicky on some patches.


baconpaul:
Great; I will look this afternoon. Thank you!

baconpaul:
This is totally great. Merging!

PR #3458, opened 2020-12-17
Loop markers are now always drawn on top of loop region area

mkruselj:
Note to @baconpaul if he wants to take a look.

lmdrag

When we get to the case that loop start=loop end, I would expect when dragging either start or end markers, that the triangle being dragged would not be drawn in the joined position /|\. This is for consistency reasons, since in normal non start=end case, you can see the triangle disappearing as you drag it around. I like this behavior, but I couldn't figure out how to achieve it with the start=end case. Thanks for taking a look!


baconpaul:
ahh i see what you are saying now
hold on i bet i can fix that and push a fix up to the branch. don't merge

PR #3483, opened 2020-12-20
GCC/Clang: Fix SSE2 compiler flags

mvf:
Ask the compiler instead of checking for "not Linux on ARM". SSE2 is
implied on x86-64, so only force-enable it on i386. Also enable SSE FP
math for better performance and in order to be closer to x86-64.


baconpaul:
Oh cool also this solves the annoying mac arm -msse2 warning. Thanks. Merging!

mkruselj:
Matt is BAKK! 😺

PR #3495, opened 2020-12-22
Move MSEG stuff into the patch

### LINKED ISSUE #3485: At user option, restore snap on/off from the patch, not the daw extra state
EDIT: Over in #3487 and #3486 we ended up saving the default levels in the patch; This issue is now about saving the snap on/off and other "UI" things; and then making those either optionally overwrite your state or not.

Very late in the 18 cycle, @K0rrid0r and I had a discussion about whether snap levels (not state of if snap is on and what not, just the level values) should be stored in the patch (and by extension on the MSEGStorage model as opposed to the MSEGEditState)

* The argument against is that snap levels are UI only and so belong on the MSEGEditState in the dawState along with things like zoomFactor and currently selected oscillator. That's what informed our choice in 1.8 / patch version 15
* The argument for is that even though snap levels aren't required to evaluate an MSEG, they are an important part of the MSEG design process, so dropping them takes information about the intent of the MSEG out of the patch.

Honestly I found that argument compelling and if we were in October i would change it, but it requires a model type, patch format, and daw Extra State change, and we are at the very last commit of 1.8, so ship 1.8 the way it is and do this in 1.9.

### LINKED ISSUE #3492: Save MSEG zoom level in patch too
Zebra: doesn't save in patch
Absynth: doesn't save in patch
Arturia DX 7 V: DOES save in patch
Dune 3: DOES save in patch
FM8: DOES save in patch
Massive X: doesn't save in patch
Omnisphere: DOES save in patch

Based on this short research, and Omnisphere being probably the most popular out of the above listed plugins, we should probably follow suit!

baconpaul:
- zoom in the patch
- snap button on/off in the patch
- option to ignore the patch for snap on/off and load from the DES
- and reorgniaze the user settings menu as a result to add workflow

Closes #3485
Closes #3492

### DIFF
--- a/src/common/ModulatorPresetManager.cpp
+++ b/src/common/ModulatorPresetManager.cpp
@@ -195,7 +195,7 @@ void loadPresetFrom( const fs::path &location, SurgeStorage *s, int scene, int l
    {
       auto msn = lfox->FirstChildElement("mseg");
       if( msn )
-         s->getPatch().msegFromXMLElement(&(s->getPatch().msegs[scene][lfoid]), msn);
+         s->getPatch().msegFromXMLElement(&(s->getPatch().msegs[scene][lfoid]), msn, true);
    }
 
    if( lfotype == lt_stepseq )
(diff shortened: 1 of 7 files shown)


baconpaul:
**PLEASE DO NOT MERGE**. I want to get breakfast and a coffee and re-review this code to be 100% sure I got it right.

PR #3530, opened 2020-12-27
Deferred Asset Loader

### LINKED ISSUE #3523: Deferred Asset Manager
NIck's skin will explode the installer if we include the high res backgrounds. So ideally the skin engine could say

```
<image resource="foo" remote_url="https://bar"/> 
```

then lazy grab it when you load the skin (with  a prompt). 

So lets make a class which allows us to resolve that cross platform. This issue is not to mod skin engine but just to have a tested way to grab https assets cross platform with appropriate threading and what not.

baconpaul:
1. Implement an API to load assets into a local cache
2. Add a regets which tests this API
3. Implement on mac (Foundation), win (WinHTTP) and lin (libcurl)

Closes #3523


baconpaul:
WIP - BP will merge thanks!

PR #3536, opened 2020-12-28
Allow opening main menu by right-clicking anywhere on the GUI

mkruselj:
Similar to middle mouse modulation arming, utilize LastChanceEventCapture. Had to change one method and a member from private to public to make it happen, so gonna wait on @baconpaul to sanction the change.


baconpaul:
No problem making it public. Looks great. Clever idea. Go ahead!

PR #3557, opened 2021-01-01
Linux: Make libcurl dependency optional, decouple from target arch

mvf:


baconpaul:
Oh cool
We should also make the regtest not run if CURL isn't present on linux. 
Hold on...

baconpaul:
Yeah see line 771 of UnitTestsIO?
We should probably add after that something glike

#if LINUX && ! GET_WITH_CURL
skiptests = true
#endif 

or what not.

mvf:
Got it, I'll add that. Mostly wanted to avoid releasing 1.8 with a mandatory dep that isn't really used yet.

baconpaul:
Yeah great. Super smart.

Hey are you online? Have one tiny thing to discuss with you which kinda guides some things I'm doing on SC3.

baconpaul:
In fact perhaps the thing to do - since we changed our mind and decided to not use this in 18 - is to just comment out the entire thing to avoid the dep on win and mac also (although it is way more harmless there). Will ponder. but this change is good with the test update too - thank you!

mvf:
Alright, I'll close this then. Not sure what's up with the ARM cross pipeline. It can't find some internal libc header...

baconpaul:
Oh no we should keep this around! I can look at that ARM thing but I bet it is just because my CURL scope isn't right in the code.

baconpaul:
(Keep it open so I don't forget to pull this into a branch and do the bigger fix at least. lets not merge it)

baconpaul:
Addressed sin #3558

PR #3558, opened 2021-01-01
Disable the Deferred Asset Loader

baconpaul:
Since we are headed into 1.8 without using the DeferredAssetLoader,
do a hard-disable of it at compile time to avoid the extra deps
on various platforms


mvf:
Removing the code doesn't get rid of the additional dependencies. Maybe it's better to just revert #3530 at this point.

baconpaul:
Ann no I just forgot to commit my cmake changes

baconpaul:
Good eye

PR #3571, opened 2021-01-02
Some tiny AirWindows fixes

baconpaul:
Some tiny AirWindows fixes for some missed cases and initializers
and a small python script to traverse all the AW and print their
state


baconpaul:
Hold on one more coming

mkruselj:
Default value of C in Focus.cpp should be 0, not 0.5, too.

PR #3588, opened 2021-01-06
Move Function from Coming in 1.9 to Coming soon

baconpaul:
So many things might change
We might call it surge 2 or surge 2021.2
We might want to do a 1.9 release before function is done
We might never do a 1.9 release (but then soon would also be a lie :) )
So hedge a bit on this UI point

### DIFF
--- a/src/common/SurgeStorage.h
+++ b/src/common/SurgeStorage.h
@@ -391,7 +391,7 @@ const char lt_names[n_lfo_types][32] =
    "Envelope",
    "Step Sequencer",
    "MSEG",
-   "Function (coming in 1.9!)",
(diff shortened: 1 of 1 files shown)


mkruselj:
"Coming in a future version" would probably be the most correct, I suppose.

baconpaul:
32 chars

mkruselj:
Coming TBD?

Half a year isn't something I'd call "soon"?

baconpaul:
How about just "coming..."

mkruselj:
How about: `Function (work in progess!)`, 27 chars.

baconpaul:
Sold! Let me push that now.

PR #3606, opened 2021-01-09
Allow custom classes for CHSwitch2

### LINKED ISSUE #3602: CHSwitch2 should support custom skin classes
...so that we can add variations to the same image used multiple times on the GUI (example: jog buttons, envelope shape controls, envelope mode, etc.).

The custom class for CHSwitch2 should support tags `image`, `hover_image` and `hoveron_image`.

baconpaul:
Turns out this just didn't work and was never going to. Now we
get 'image' 'hover_image' and 'hover_on_image'

Closes #3602.


baconpaul:
DO NOT MERGE
@mkruselj reports crash in MSEG editor with this diff

PR #3610, opened 2021-01-10
Lazy load PNGs

### LINKED ISSUE #3607: PNG Deferred Load
CScalableBitmap

1. When you addPngForZoom rather than doing a std::move of a bitmap on a PNG, make a fake class which isa bitmap and has a path with a well defined type and push that
2. when you do pngZooms[level]->draw right before that check if pngZooms[level].get dynamic casts to fake class. If it does replace it (but make sure we aren't iterating when we do that).
3. voila. PNG lazy load

baconpaul:
PNG images at a given zoom level don't load until the first
time they are called on to draw (except the 100% which has to
load to prep the data structures)

Closes #3607

### DIFF
--- a/src/common/gui/CScalableBitmap.cpp
+++ b/src/common/gui/CScalableBitmap.cpp
@@ -230,7 +230,7 @@ CScalableBitmap::CScalableBitmap(std::string ifname, VSTGUI::CFrame* f)
 
     if( _stricmp( extension.c_str(), "png" ) == 0 )
     {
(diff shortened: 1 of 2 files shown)


baconpaul:
DO NOT MERGE yet. I've only tested on macOS so far. Just starting the pipeline in case it does work win and lin which I will try in a bit.

baconpaul:
Works on linux with repo head royal skin; much faster open time too.

baconpaul:
works win too. merging.

Issue #3589, opened 2021-01-06
Surge, FileSystem, C++-17, etc


baconpaul:
Over in https://github.com/surge-synthesizer/shortcircuit3/commit/2f1a3a6ba9ac87dd0c16838f12dc929b63077eaa I moved SC3 to https://github.com/gulrak/filesystem after a couple of conversations with @mvf 

For surge 1.9 I propose the following

1. We use CMAKE_CXX_STANDARD 17 everywhere.
2. This means flooring macOS at 10.12
3. But several of our compilers (macOS < 10.15; gcc < 9) still don't provide std::filesystem with C++-17
4. So we do the same linkage as above optionally on those platforms

Basically: C++ everywhere, still support our user base. 

Doing this is a better idea than what I proposed in #3463 so I will close #3463

baconpaul:
@mvf my intended first commit of surge 1.9 is going to move us to require macOS 10.12 and set our C++ standard to 17.

This means that we will have (basically) c++17-without-filesystem on mac and c++-with-misnamed-filesystem on half our linuxes.

So the only question left on this issue is: do we dump our filesystem and use the gulrak one we are using in SC3 for those platforms? It's easy to do mechanically.

Your thoughts appreciated, especially since it would involve loads of code you have touched.

baconpaul:
Ahh darnit apple silicon C++ 17 will be harder. You need macOS14 to have aligned dealloc in the arm std lib.

Window Oscillator threw an error in the M1 build that 10.12 doesn't have the dealloc. I could fix it by adding an operator new/delete like this but we need to add it to several classes so it won't be the first install

will def do it though.

```
+void* operator new(std::size_t size, std::align_val_t align) {
+    auto ptr = aligned_alloc(static_cast<std::size_t>(align), size);
+
+    if (!ptr)
+        throw std::bad_alloc{};
+    return ptr;
+}
+
+void operator delete(void* ptr, std::size_t size, std::align_val_t align) noexcept {
+    free(ptr);
+}
+
+void operator delete(void* ptr, std::align_val_t align) noexcept {
+    free(ptr);
+}
```

mvf:
> do we dump our filesystem and use the gulrak one we are using in SC3 for those platforms?

Good point, I'll prepare a change that replaces `Surge::filesystem` with gulrak. We'll keep our CMake infrastructure for include and exports hygiene and the ability to build against gulrak anywhere via CMake cache variable.

[...]

PR #3620, opened 2021-01-13
1.8.2 FIRST MERGE - Show install location, update version

### LINKED ISSUE #3414: Display install location on about screen
on Win we know it to resolve portable
on Mac we know it since bundle bundle
on Linux I'm less sure. If we don't know it though we can just skip on Linux.

baconpaul:
Show the install location of the plugin in the about screen.

Closes #3414


baconpaul:
This for a little was going to move to C++17 but see #3589 for why it didn't

PR #3655, opened 2021-01-18
Use Python3 piggy packing the Linux binary

jarkk0sakkinen:
As Python2 has been EOL for over a year already, move on using Python3. The
use of Python2 is likely cause the following buid error in recent versions
of various popular Linux distributions:

```
[ 67%] Generating lintemp/ScalablePiggy.S
/bin/sh: 1: python: not found
gmake[3]: *** [CMakeFiles/surge-vst3-dll.dir/build.make:140: lintemp/ScalablePiggy.S] Error 127
gmake[2]: *** [CMakeFiles/Makefile2:562: CMakeFiles/surge-vst3-dll.dir/all] Error 2
gmake[1]: *** [CMakeFiles/Makefile2:624: CMakeFiles/Surge-VST3-Packaged.dir/rule] Error 2
gmake: *** [Makefile:332: Surge-VST3-Packaged] Error 2
```

Do the migration by

1. Renaming `emit-vector-piggy.py` as `emit-vector-piggy`.
2. Renaming `generate-lv2-ttl.py` as `generate-lv2-ttl`.
3. Giving +x for both.
4. Using `/usr/bin/env` to find the location of the interpreter in the
   POSIX standard manner, thus making the whole compilation process more
   distro agnostic.

[*] https://pubs.opengroup.org/onlinepubs/9699919799/utilities/env.html


jarkk0sakkinen:
I've converted from Ableton Live user into Reaper user over last year, moved most of my music production to Linux, a lot thanks to recent developments (mainly PipeWire), and prefer keeping my VST's at my home directory, so using self-compiled Surge is the best bet for my needs :-)

baconpaul:
Thanks

halpy to merge this but we are pausing merges for this week while we stabilize 18. Will merge it as soon as we know our merge strategy for either 181 or 19

jarkk0sakkinen:
NP. Merge when you see appropriate. I'm running now self-compiled main branch and obviously have this patch in my fork.

jarkk0sakkinen:
I'll probably rebase this weekly basis up until it is merged as I want to run the latest and greatest...

baconpaul:
Yeah depending on our path the latest greatest might get a little rocky but I will merge this early in cycle so you can build close to 18

enjoy!

jarkk0sakkinen:
If I find any other issues the fixes will be visible in this PR (assuming that I'm capable fixing them, otherwise I'll just report an issue). If you don't want to take all at once, `git cherry-pick` is a great asset,

mvf:
Making the python-ness transparent by adding a shebang, removing the `.py` extension and making it executable is good, but the interpreter should stay `python` (i.e. "any Python") since the code is compatible with both 2 and 3. In properly configured python-3-only systems, `python` points to `python3`.

jarkk0sakkinen:
@mvf, NP I can refine this. Thanks for the feedback.

Off-topic: how can I test run the new flashier skin seen in videos and screenshots?

jarkk0sakkinen:
BTW, In my Debian system python does not point to python3, so should I consider Debian default install is improperly configured? :-) I guess I need to install `python-is-python3` package but it was not preinstalled in the default Gnome 3 desktop install.

TheNerdyMusicGuy:
@jarkk0sakkinen The new skin is in the latest release. Just right click anywhere on the interface and go to Skins -> Royal Surge.

jarkk0sakkinen:
I use self-compiled master mostly because I prefer a local install over deb and also to keep in phase where things are going. I did a compile of the master when I submitted this patch. Was it added after that, or do I do some tricks to get it working? Does it get piggy packed to the binary?

jarkk0sakkinen:
OK, now the commit has been updated (and the commit message has been changed accordingly).

mvf:
Sorry about the "properly configured" it's a bit snarky and not very accurate. For the Debian GNOME install I reckon Python was only installed as a dependency and there was no need to install a `python` symlink. So the system was "properly configured" for running GNOME, but not for running Python scripts off the internet. :smile:

Anyhow, looks good now, merging!

baconpaul:
Uhhh.... you merged this @mvf? I was hoping to wait until after 181. Are we sure it works on all our distros including RHI?

baconpaul:
@jarkk0sakkinen if you self build you will need to stage the extra content repo. The new skin is in a separate repo which isn’t a submodule. There’s a cmake rule to do it

baconpaul:
Oh I see merged with oyhton3 renamed to python. All good

jarkk0sakkinen:
@baconpaul, Thanks, I'll look into it.

baconpaul:
Or just clone surge-extra-content into your documents surge folder more easily

jarkk0sakkinen:
Sounds like better plan thank you :-)

Issue #3584, opened 2021-01-05
Add ChowDSP CHOW


jatinchowdhury18:
Add [CHOW](https://github.com/Chowdhury-DSP/CHOW) to the Surge effects rack.

CHOW is a digital distortion/compressor/rectifier effect. Frankly, this effect can be rather harsh and even "ugly" but can sometimes be useful for sound design. It also aliases quite a bit (which I consider part of the effect).

mkruselj:
Already listed it in #1694 actually :)

Would not be a bad idea to add some oversampling to it, as a parameter?

baconpaul:
Oh if you would like to 2x oversample, the RingModulator effect shows how to use the half band filter at input and output to upsample/downsample in a way which is consistent with the rest of the surge signal path. (You may want to use a different strategy or not oversample of course.).

jatinchowdhury18:
Yeah, I suppose having oversampling as an option couldn't hurt :).

mkruselj:
@jatinchowdhury18 Is this one in PR state already? Feel free to push a PR and I can go in and tweak if you're busy otherwise. Rebase first, of course :)

jatinchowdhury18:
Yeah, it's pretty close, but I wanted to do a little more testing. Hoping I can put it up for PR later today, so you all can review/tweak as desired :).

mkruselj:
Cool!

Just FYI, I am pushing a PR that adds a few enums for FX types in advance to SurgeStorage.h, which should help with those pesky merge conflicts!

![image](https://user-images.githubusercontent.com/2393720/106333130-4fdeae00-6288-11eb-9647-26257ec7b75a.png)

PR #3656, opened 2021-01-18
1.8.2 MERGEABLE - Neuron, distortion effect based on Gated Recurrent Unit

### LINKED ISSUE #3587: Add ChowDSP Gated Recurrent Distortion
Add [Gated Recurrent Distortion](https://github.com/jatinchowdhury18/ComplexNonlinearities/tree/master/GatedRecurrentDistortion) to Surge effects rack. As discussed in #1694.

I think the main difficulty with this one will be in determining the ideal parameter names and ranges.

While this is fundamentally a distortion effect, I don't think it could also be used as a Surge waveshaper, since it is not a memoryless process.

jatinchowdhury18:
Closes #3587.

Some notes:
- I added an `AddingAnFX.md` page for documentation. Please let me know if 'm missing some steps.
- I added a few classes from JUCE (e.g. `SmoothedValue`) that I can probably remove after more JUCE integration is complete. If you'd prefer, I could wait until this happens before merging this PR.


mkruselj:
We don't know precisely how long until Surge is ported to JUCE vs VSTGUI, so for the time being it probably makes sense to use our own `lag` thingie, [see here](https://github.com/surge-synthesizer/surge/blob/2560170d5270bbd79231491c0fe98529074b1b81/src/common/dsp/effect/DualDelayEffect.h#L75) for example.

Then CI should let things pass.

baconpaul:
Will look today. Pretty sure I know the problems

baconpaul:
Alright the linux problems relate to (1) template resolution in gcc and (2) JUCE includes everything in a header. Here's the patch for that:

```
diff --git a/src/common/dsp/effect/chowdsp/shared/DelayLine.cpp b/src/common/dsp/effect/chowdsp/shared/DelayLine.cpp
index d73766a6..a7fa7e3b 100644
--- a/src/common/dsp/effect/chowdsp/shared/DelayLine.cpp
+++ b/src/common/dsp/effect/chowdsp/shared/DelayLine.cpp
@@ -24,6 +24,7 @@
 */
 
 #include <algorithm>
+#include <cmath>
 #include "DelayLine.h"
 
 namespace chowdsp
@@ -86,7 +87,7 @@ void DelayLine<SampleType, InterpolationType>::reset()
 
     std::fill (this->v.begin(), this->v.end(), static_cast<SampleType> (0));
 
-    for (size_t ch = 0; ch < bufferData.size(); ++ch)
+    for (size_t ch = 0; ch < this->bufferData.size(); ++ch)
         std::fill (this->bufferData[ch].begin(), this->bufferData[ch].end(), static_cast<SampleType> (0));
 }
```

and with that applied ubuntu 20 works (and I presume the rest will too). I am almost entire sure that will fix windows 2019 also.

baconpaul:
@mkruselj that's not quite the problem. @jatinchowdhury18 pulled the juce classes into surge cleanly just they have different compile options.

baconpaul:
The mac one is the same as above plus one other thing. We currently build mac on C++14. With 19 we will move to C++17 (and drop support for macOS older than 10.12) but `std::clamp` was in the 17 not the 14 spec. For now if you replace std::clamp with the surge builtin `limit_range` it will work; or wait for me to merge up the start of 19 / move to C++17 diff

baconpaul:
Yeah that's exactly it

https://github.com/surge-synthesizer/surge/pull/3620 that's the intended first commit of 1.9. It moves the version, adds one feature closing an issue, and moves us to C++17 consistently.

If you rebase your branch to that then apply my patch above mac builds fine.

So super easy to get this building. Just need to order the commits.

baconpaul:
Ahh that C++ 17 upgrade will take a little longer. ARM mac doesn't like one thing we do.

So std::clam -> limit_range and apply the diff above and it will work.

baconpaul:
See #3589 for discussion of C++17

jatinchowdhury18:
Ah, thanks for catching these... In the future, I'll make sure to build onmy Linux machine before making a PR, that should help me to catch these types of errors earlier on!

One other question: I added a Python notebook under `scripts/ipy/` to help me tune some of the parameter mappings... do you think this would go better under the `surge-python` repo? Especially since I'm not actually generating any hard-coded tuning values that are then used in the code.

baconpaul:
Yeah scripts/ipy was from before we had the python repo. Unless they are needed to generate constants used in code or some such I would rather have them out of the code repo. Did I give you merge rights to the surge-python repo? I'm happy to if not.

jatinchowdhury18:
@baconpaul it doesn't look like I have push access for `surge-python`.

@mkruselj My `Oversampling` class actually isn't a JUCE port, it's a custom class that uses the `HalfRateFilter` for up/downsampling. I just wrapped them in a class I think will be more convenient for me, seeing as I'm expecting to need oversampling for several of the other effects I'm planning to implement.

mkruselj (commented):
> @mkruselj My `Oversampling` class actually isn't a JUCE port, it's a custom class that uses the `HalfRateFilter` for up/downsampling. I just wrapped them in a class I think will be more convenient for me, seeing as I'm expecting to need oversampling for several of the other effects I'm planning to implement.

Ah ok!

baconpaul:
@jatinchowdhury18 just gave you perms to the synth repo

On the conflict on IDs. We can sort that out when we turn the merges back on in a few days or so.

mkruselj (dismissed):

jatinchowdhury18:
Hi all! I've been working to implement another effect, the CHOW distortion effect (#3584), on [this branch](https://github.com/Chowdhury-DSP/surge/tree/chowdsp-chow). I'd like to make a PR for that effect, but it builds off some of the changes I made in this PR. Would it make sense to add those changes to this PR, or just wait until this one is merged?

baconpaul:
I will merge this one probably tomorrow around lunchtime if the 181 release works OK. Lets keep them separate if that's OK.

mkruselj:
@jatinchowdhury18 Good stuff! If you'll take a look at our [1.8.2. milestone](https://github.com/surge-synthesizer/surge/milestone/20), which is planned for beginning of March, the Aural Exciter is also on that list, if you can manage the time to port it - let me know!

jatinchowdhury18:
@mkruselj, great, that sounds pretty do-able! Depending on how much time I have, hopefullly I can get a decent start on Chow Tape by then as well :).

mkruselj:
@jatinchowdhury18 Great!

BTW for ChowTaoe it's best to wait until we have expanded the parameter count so that FX can have more than 12 parameters.

And for aural exciter we should have separate controls for left and right channel (we'll add an option to eight click link the controls).

PR #3696, opened 2021-01-22
Add some factory MSEG presets, change name of factory mod presets folder

mkruselj:
Deal with several extraneous forward slashes showing in a few dialogs on Windows


baconpaul:
looks good to me (once CI passes)

PR #3734, opened 2021-01-27
SSE inside the Resonator effect

baconpaul:
This isn't finished and is not on in config.xml, but this is
a complete implementation of how to use an SSE filter in the Resonator
Effect which @mkruselj can use to finish this effect.

### DIFF
--- a/src/common/dsp/effect/ResonatorEffect.cpp
+++ b/src/common/dsp/effect/ResonatorEffect.cpp
@@ -14,15 +14,22 @@
 */
 
 #include "ResonatorEffect.h"
+#include "DebugHelpers.h"
 
 ResonatorEffect::ResonatorEffect(SurgeStorage *storage, FxStorage *fxdata, pdata *pd)
-    : Effect(storage, fxdata, pd)
+    : Effect(storage, fxdata, pd), halfbandIN(6, true), halfbandOUT(6, true)
 {
     gain.set_blocksize(BLOCK_SIZE);
     mix.set_blocksize(BLOCK_SIZE);
+
+    qfus = (QuadFilterUnitState *)_aligned_malloc(2 * sizeof(QuadFilterUnitState), 16);
+    for (int e = 0; e < 3; ++e)
+        for (int c = 0; c < 2; ++c)
+            for (int q = 0; q < n_filter_registers; ++q)
+                Reg[e][c][q] = 0;
 }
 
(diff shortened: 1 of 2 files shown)


src/common/dsp/effect/ResonatorEffect.cpp
```diff
@@ -39,21 +46,152 @@ void ResonatorEffect::setvars(bool init)
...
+    {
+        for (int i = 0; i < n_filter_registers; i++)
+        {
+            for (int e = 0; e < 3; ++e)
+            {
+                Reg[e][c][i] = get1f(qfus[c].R[i], e);
+                coeff[e][c].C[i] = get1f(qfus[c].C[i], e);
+            }
+        }
+    }
```
mvf, line 180: GCC doesn't seem to like this part. Just pasting this here, didn't look into it.
```
../src/common/dsp/effect/ResonatorEffect.cpp: In member function 'virtual void ResonatorEffect::process(float*, float*)':
../src/common/dsp/effect/ResonatorEffect.cpp:243:41: warning: iteration 8 invokes undefined behavior [-Waggressive-loop-optimizations]
  243 |                 coeff[e][c].C[i] = get1f(qfus[c].C[i], e);
      |                                    ~~~~~^~~~~~~~~~~~~~~~~
../src/common/dsp/effect/ResonatorEffect.cpp:238:27: note: within this loop
  238 |         for (int i = 0; i < n_filter_registers; i++)
      |                         ~~^~~~~~~~~~~~~~~~~~~~
```

baconpaul: Oh man that’s a real bug. It should be r not c or n coefficient not registers. Will clean up today. Thanks. 

Which gcc version shows that? Surpsiede pipelines didn’t find

mvf: GCC 10.2.1 20201203. The pipelines also show this, but we're not using `-Werror` on GCC.

baconpaul:
Ahh! Well two things to fix then! I thought we were werrror on linux now?

will get to it today. Basically just need two loops with different bounds there

src/common/dsp/effect/ResonatorEffect.cpp
baconpaul: Anyway #3751 fixes both. Thank you!!!

PR #3744, opened 2021-01-28
Add option to link delay right channel time to left channel time

### LINKED ISSUE #3613: Link left/right times in delay FX
Problem is: everytime when using the delay-time and change the Frequency or Temposynced Steps the Left and Right Channel needs to be moved individually. It would be a great when there would be and Link Button for the Left & Right Channel. Then it could be moved simultaneously.

The advantage: easy to use and very quick to setup or change.

mkruselj:
Right-click the right channel delay time and choose option to Link to left channel. This (ab)uses the deactivatable parameters trick!
Closes #3613

### DIFF
--- a/src/common/Parameter.cpp
+++ b/src/common/Parameter.cpp
@@ -236,6 +236,7 @@ bool Parameter::can_temposync()
     case ct_lforate:
     case ct_lforate_deactivatable:
     case ct_envtime:
+    case ct_envtime_deactivatable:
     case ct_envtime_lfodecay:
     case ct_reverbpredelaytime:
         return true;
@@ -290,6 +291,7 @@ bool Parameter::can_deactivate()
     case ct_decibel_deactivatable:
     case ct_decibel_narrow_deactivatable:
     case ct_decibel_extra_narrow_deactivatable:
+    case ct_envtime_deactivatable:
         return true;
     }
     return false;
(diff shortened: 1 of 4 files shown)


src/common/dsp/effect/DualDelayEffect.cpp
```diff
@@ -65,17 +65,18 @@ void DualDelayEffect::setvars(bool init)
...
 
+    auto isLinked = fxdata->p[dly_time_right].deactivated ? dly_time_left : dly_time_right;
+
     if (init)
     {
         timeL.newValue(
             samplerate *
-                ((fxdata->p[dly_time_left].temposync ? storage->temposyncratio_inv : 1.f) *
-                 storage->note_to_pitch_ignoring_tuning(12 * fxdata->p[dly_time_left].val.f)) +
+                ((fxdata->p[isLinked].temposync ? storage->temposyncratio_inv : 1.f) *
```
baconpaul, line 74: Shouldn’t this one always be left whether linked or not?

src/common/Parameter.cpp
```diff
@@ -236,6 +236,7 @@ bool Parameter::can_temposync()
     case ct_lforate:
     case ct_lforate_deactivatable:
     case ct_envtime:
+    case ct_envtime_deactivatable:
```
baconpaul, line 239: Since this is specific for delay linking, perhaps rename this to ‘ct_envtime_linkable_delay’ since we special case the string in SGE (and would not want to use it for general deactivatable envtime)

baconpaul:
Fantastic!

PR #3759, opened 2021-01-29
Add skeleton for Combulator effect

mkruselj:
Add enums for planned effects (CHOW, Exciter, Ensemble, Nimbus)

### DIFF
--- a/src/common/dsp/effect/CombulatorEffect.cpp
+++ b/src/common/dsp/effect/CombulatorEffect.cpp
@@ -4,7 +4,7 @@
 ** Surge is made available under the Gnu General Public License, v3.0
 ** https://www.gnu.org/licenses/gpl-3.0.en.html
 **
-** Copyright 2004-2020 by various individuals as described by the Git transaction log
+** Copyright 2004-2021 by various individuals as described by the Git transaction log
 **
 ** All source at: https://github.com/surge-synthesizer/surge.git
 **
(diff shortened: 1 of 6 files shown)


src/common/dsp/effect/CombulatorEffect.cpp
```diff
@@ -0,0 +1,320 @@
+/*
+** Surge Synthesizer is Free and Open Source Software
+**
+** Surge is made available under the Gnu General Public License, v3.0
+** https://www.gnu.org/licenses/gpl-3.0.en.html
+**
+** Copyright 2004-2020 by various individuals as described by the Git transaction log
```
baconpaul, line 7: No need to redo the PR. But when I touch a file I’ve been turning 20s into 21. Only need to do it when we touch a file of course but just FYI!

src/common/dsp/effect/CombulatorEffect.cpp
```diff
@@ -0,0 +1,320 @@
...
+            }
+
+            for (int i = 0; i < n_filter_registers; i++)
+            {
+                set1f(qfus[c].R[i], e, Reg[e][c][i]);
+            }
+
+            qfus[c].DB[e] = filterDelay[e][c];
+            qfus[c].WP[e] = WP[e][c];
+            qfus[c].WP[0] = subtype;
```
baconpaul, line 144: This will be where the bug is by the way. For sure. This needs to be conditioned on filter type (or removed here)

PR #3769, opened 2021-01-31
Add DSP for Neuron comb freq. modulation

### LINKED ISSUE #3755: LFO for Comb Frequency in Neuron
@jatinchowdhury18 a request from Jacky L and we have slots

Lets add a little LFO that lets you modulate the frequency of that comb filter with some rate and depth (so 2 slots)

I'm happy to add it or happy for you to. The flanger shows how to add an LFO easily etc... but I am sure it is not even remotely surprising. Wanted to drop it here though as an issue

Only trick is we will have a choice of 'add params in the middle' which means old streamed sessions will break (which is OK - we are still in the nightlies) or 'add it at the end' which means they won't but we have to re-order the UI using that hokey API.

jatinchowdhury18:
Closes #3755

### DIFF
--- a/src/common/dsp/effect/chowdsp/Neuron.cpp
+++ b/src/common/dsp/effect/chowdsp/Neuron.cpp
@@ -77,6 +77,7 @@ void Neuron::process(float *dataL, float *dataR)
     decodeMS(M, S, dataL, dataR, BLOCK_SIZE_QUAD);
 
     outgain.multiply_2_blocks(dataL, dataR, BLOCK_SIZE_QUAD);
+    modLFO.post_process();
 }
 
 void Neuron::process_internal(float *dataL, float *dataR, const int numSamples)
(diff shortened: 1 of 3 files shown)


mkruselj:
Maybe it would be cleaner to use the same LFO implementation from flanger effect?

baconpaul:
It would be, which is why Jatin did! Just he put it in a convenient class so it is reusable as opposed to cram it inline. If you read the code you can see it’s the same. We should merge this and then back port the flanger to also use the class introduced here

baconpaul:
So I propose merge this, then move the modclsss out of chowdsp into effect then make the flanger use it too. Merging

mkruselj:
Oh! Did not notice it was actually the same. Apologies, Jatin, and what a great useful addition!

baconpaul:
Even some of the comments! I was looking thinking "oh what clever thing did jatin do" and I was like "huh i had the same idea for flanger" and then compared :)

So added an issue to refactor this into flanger.

jatinchowdhury18:
> Even some of the comments! I was looking thinking "oh what clever thing did jatin do" and I was like "huh i had the same idea for flanger" and then compared :)

Haha! Yes, my original intention was to actually refactor the code that way, but the flanger has to deal with multiple channels lfo signal, so I figured I'd just do it for Neuron for now.

PR #3780, opened 2021-01-31
Various Nimbus parameter tweaks

mkruselj:
Increase font size in value displays by 1 point

NOTE: Some parameters in Nimbus have changed their physical order, breaking patches/FX presets. Nightlies gonna nightly!


baconpaul:
Looks great

PR #3785, opened 2021-02-01
Implement retrigger consistently for all oscillator types

### LINKED ISSUE #3171: Oscillator Retrigger is not implemented for a number of oscillators

mkruselj:
Closes #3171

### DIFF
--- a/src/common/SurgeStorage.h
+++ b/src/common/SurgeStorage.h
@@ -400,7 +400,7 @@ enum adsr_purpose
 };
 
 /*
- * How does the sustain pedal work in mono mode? Current modes for this are
+ * How does the sustain pedal work in mono mode? Current modes for this are:
  *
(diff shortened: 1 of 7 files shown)


mkruselj:
Question for @baconpaul: should we add a very tiny quick fade-in for sine/FM2/FM3/window? It seems that WT osc has something that de-clicks its output even with retrigger off, so maybe others should match that?

baconpaul:
Rerunning pipeline. The unit tes fails look real tho

PR #3794, opened 2021-02-02
Add exciter effect

jatinchowdhury18:
Closes #3794. Here's the exciter effect! I'd love some feedback on how it sounds, things that could maybe be improved. I also still need to connect the attack/release time sliders to actual parameters, but TBH I'm a little scared by `Parameter.cpp`. I remember hearing somewhere, that there's a compressor effect coming as well, so maybe they could be re-used.

### DIFF
--- a/src/common/dsp/effect/chowdsp/exciter/Exciter.cpp
+++ b/src/common/dsp/effect/chowdsp/exciter/Exciter.cpp
@@ -80,9 +80,11 @@ void Exciter::set_params()
     auto drive = 8.f * std::pow(*f[exciter_drive], 1.5f) * drive_makeup;
     drive_gain.set_target_smoothed(drive);
 
-    // attack/release @TODO: connect to params
-    levelDetector.set_attack_time(10.f);
-    levelDetector.set_release_time(100.f);
+    // attack/release params
+    auto attack_ms = std::pow(2.0f, fxdata->p[exciter_att].displayInfo.b * *f[exciter_att]);
+    auto release_ms = 10.0f * std::pow(2.0f, fxdata->p[exciter_rel].displayInfo.b * *f[exciter_rel]);
+    levelDetector.set_attack_time(attack_ms);
+    levelDetector.set_release_time(release_ms);
 
     // "Mix" param
     wet_gain.set_target_smoothed(limit_range(*f[exciter_mix], 0.f, 1.f));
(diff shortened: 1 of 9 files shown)


mkruselj:
Looks like there's an include missing (std::exp used in level detector, it seems).

Regarding `Parameter.cpp`, what is frightening you? 🙂  What's the range of attack/release parameters you have here, and I assume they are exponential scaling?

baconpaul:
yeah I think you need a `#include <cmath>` to make mac happy.

Parameter.cpp is one of those bits of code which is magically simultaneously scary long and simple. It does a lot, it looks intimidating, but it doesn't do anything hard - but because of the first two it is indeed a bit of an investment :) Happy to help if you want to add a param type or some such - just let us know.

mkruselj:
Yeah you can just follow the path of, for example, `ct_decibel` and see all that applies to it.

src/common/dsp/effect/chowdsp/exciter/Exciter.cpp
```diff
@@ -0,0 +1,173 @@
...
+    toneFilter.coeff_HP(cutoff * omega_factor, q_val);
+
+    // "Drive" param
+    auto drive_makeup = std::pow(0.2f, 1.f - *f[exciter_tone]);
+    auto drive = 8.f * std::pow(*f[exciter_drive], 1.5f) * drive_makeup;
+    drive_gain.set_target_smoothed(drive);
+
+    // attack/release params
+    auto attack_ms = std::pow(2.0f, fxdata->p[exciter_att].displayInfo.b * *f[exciter_att]);
+    auto release_ms = 10.0f * std::pow(2.0f, fxdata->p[exciter_rel].displayInfo.b * *f[exciter_rel]);
```
jatinchowdhury18, line 85: I was able to set up the parameter types, and they seem to be working as expected. I'm using the `ATwoBx` scaling type. I was wondering if there's an easier to convert to/from the normalised value to the scaled value? This is my current hack, but it should be generalisable to any exponential parameter, like:
```
denormalised_value = param->displayInfo.a * std::pow(2.0f, param->displayInfo.b * normalised_value);
```

mkruselj: I think you could reuse `ct_reverbpredelaytime`, which gives you 0 to 2 seconds of range that is exponentially scaled. Or you could just create your own params based on that one.

baconpaul: Oh ha we added the ATwoBx for a formatting only thing long after the params. I never ever even thought about using the displayInfo values in the DSP code. 

There’s nothing wrong with it of course - in fact it is rather clever and we could easily add it to param (like “getDisplayNormalizedFloat” which returns the scaled float if you are one of the display types exactly as you suggest) and that’s a very good idea. But like I said it isn’t one I ever thought about at all!

Context: All the displayInfo stuff was retrofitted onto param to remove essentially a bunch of duplicated switch statements and allow us to do the reverse string -> param for typein. I always thought of it purely in gui / daw land. This is why having people who are new to the code read it is so valuable.

I propose we merge this - your change is fine - so I’ll press the button.  But if you agree with assessment above add another issue to add that function point and then we can retrofit.

Issue #3773, opened 2021-01-31
Combulator - outstanding items


baconpaul:
- [x] Lower than 43hz etc
- [x] Noise
- [x] LP filter
- [x] Soft clipper on the output
- [x] Bi-polar resonance, and resonance labeled as feedback
- [x] Linear scale on feedback?
- [x] Width slider pans comb 2 and 3 left and right
- [ ] Sample rate vs delay buffer
- [ ] absolute vs relative F2/3
- [x] mix negative without bleed through

PR #3796, opened 2021-02-03
Add lowpass, softclip and panned combs (with width) to Combulator

mkruselj:
Addresses #3773


baconpaul:
Cool! 
This looks like a hard pan on comb 2 and 3. I thought we were going to pan them based on the width slider and remove m/s though?

baconpaul:
Anyway lets chat about the pan tomorrow before you merge. Otherwise all great! Thank you.

mkruselj:
Yep they are hard panned but then mid-side decides by how much or collapses to mono 🙂

PR #3824, opened 2021-02-05
Add skin connector for positioning the Store Patch dialog

mkruselj:
Prior to this PR, the Store Patch dialog had a fixed position and size. This PR introduces a new skin connector which will allow repositioning that dialog.
No size adjustment yet (but the code to do so is commented out and ready for introducing it later on if need be).

Also updated Royal Surge skin so that the Store Patch dialog is nicely centered, and updated the hash for staging extra content.


baconpaul:
Code diff looks correct

baconpaul:
The reason for the build fails is I don’t default in the sge loop that’s on purpose

you can just explicitly add the npc as a noop

PR #3829, opened 2021-02-06
filesystem: Replace Surge::filesystem with ghc::filesystem

mvf:
This replaces the implementation with gulrak's filesystem
(https://github.com/gulrak/filesystem) on systems without a usable
std::filesystem, or on any system (including Windows), when the
SURGE_FILESYSTEM_FORCE_GHC CMake cache variable is set.

#3589


baconpaul:
Awesome!

Issue #3204, opened 2020-11-23
Remaining Code Cleanup


baconpaul:
At the start of 1.8.2, we clang-formatted the code and made the pipeline require that format.

Some things I want to do still though:

## Remove _aligned_malloc

1. Just the Phaser and the Synth for the QFB. 

## All code has the copyright header

All .cpp and .h files have the copyright header and ideally the pipelines fail if you add one that doesn’t

baconpaul:
```
$ clang-format --dry-run --Werror foo.cpp
foo.cpp:129:23: error: code should be clang-formatted [-Wclang-format-violations]
        if (rc <= 0) {
```

for the pipeline

Script from @Kingston1 which does the find/exec and is tested

[s.zip](https://github.com/surge-synthesizer/surge/files/5583242/s.zip)

baconpaul:
Here's my proposed .clang-format file

```
---
BasedOnStyle: LLVM
IndentWidth: 4
---
Language: Cpp
BasedOnStyle: LLVM
IndentWidth: 4
AlignAfterOpenBracket: Align
BreakBeforeBraces: Allman
ColumnLimit: 100
SortIncludes: False
---
```

I could be talked into a lower column limit easily. Other tweaks welcomed.

mvf:
Mostly :+1:, huge fan of pedantic CI builds. There's nothing worse than inconsistent formatting.

`.clang-format` looks good, just need to change `SortIncludes` to `false` (`False` is a syntax error over here).

This issue feels too big though. I suggest we focus on:

- Clang format
- Copyright header
- Pipeline changes to enforce the previous two
- Deletion of whole files

The others aren't really "cleanup-ey" enough IMO and should go into separate issues and commits. Moving files around in particular is messy business, so I'd hold off on that until it's time to move a file to its final location in a proper library, i.e. not `surge-shared`.

baconpaul:
Yeah the moves are all moves inside surge-shared so I might do them (I shuffled the libraries using your pattern a bit ago so surge-shared is now actually host-type free and the actual core shared code we use everywhere) but i agree probably not one massive commit.

mvf:
Moving files somewhat obscures history. It should be done with a clear goal in mind, and the "around", "elsewhere" and "or some such" in the cover letter suggest that there isn't one. The DSP stuff in particular is a prime target for lib-ification and unlikely to stay in `surge-shared` much longer anyway.

[...]

PR #3837, opened 2021-02-06
Fix clang-format check in pipeline builds

mvf:
The clang-format check was doing nothing in clean working copies (e.g.
pipeline builds) because it was diffing against HEAD.

This change makes it always check all cached .cpp and .h files in the
relevant directories. Delta checks via "git diff --name-only" require
deep clones (>330 MB and growing) and would likely be slower on Azure.
Checking everything takes <3 seconds on my machine.

@baconpaul I won't just merge this until you've had a chance to look at it. Especially `src/common/ModulationSource.h`, where I had to manually expand comments without fully knowing what they mean. :sweat_smile:

#3204


baconpaul:
Ha well I was surprised at how well we were doing with prs but I have been running git clang format before commits 

anyway looks like it failed in the pipeline with an error - do you have access to the pipeline logs?

mvf:
Yes, `clang-format` is not installed on the macOS VM. Any particular reason to run this on macOS? Otherwise I'd move it to a Linux job.

baconpaul:
How did it work before then?

but no it could run in Linux too

baconpaul:
Oh right it ran before because it sent no args into xargs in the pipeline! Doh

thanks for fixing all of this. Appreciated!!

mvf:
OK, let's try this with Linux then. `clang-format` was never called on the macOS VM because `git diff` produces no output when there are no unstaged changes.

Issue #3835, opened 2021-02-06
Slider label hiding and skin label listener


mkruselj:
Currently, labels for CSurgeSlider are hard-coded in there and not approachable to the skin engine. So, we cannot do this thing that Dave wants to do:

![image](https://user-images.githubusercontent.com/2393720/107125270-166dfa00-68a9-11eb-82f5-9c2d5407113f.png)

For the most part, we could solve this with skin labels, but there's a problem with oscillator and FX sliders, because they change their text based on osc/FX loaded.

My idea to resolve this is twofold:

1. The easy part - add an argument in the skin engine to not draw the hard-coded label at all.

2. The harder part - add something that would enable a skin label to "listen" to `CSurgeSlider::setLabel()` and use that text as its own, based on skin connector that links to a particular parameter. Example syntax:

`<label x="10" y="300" w="100" h="20" font-type="Roboto" font-size="14" color="#FF0000" text_listener="osc.param_1" />`

This would hook itself up to `osc.param_1` and just funnel whatever text changes come its way and update itself.

The FX sliders situation is way trickier. We really need to break that crappy API and simply define static panels per each FX that we would then use instead (and those panels should also include separators and VU meters, of course).

baconpaul:
A much much easier solution is to add a slider property “labelposn” which is above below left right 

your solution is way better - but the above solution is super easy to code

mkruselj:
Except positioning things above the slider will clip the label because it would be outside of the main draw context, no? I tried this with existing labels and got this:

![image](https://user-images.githubusercontent.com/2393720/107126853-9b114600-68b2-11eb-85b9-d0ece20d9480.png)

baconpaul:
Yeah ok lemme ponder 

since we always rebuild when labels switch it should be doable with a sh tac something like what you suggest

PR #3839, opened 2021-02-06
Allow labels to have dynamic values from parameter names

baconpaul:
Addresses #3835


baconpaul:
Will combine with slider text hide

baconpaul:
So pls don’t merge yet

baconpaul:
OK force pushed. If CI runs feel free to merge (or I will if I end up here seeing it done)

PR #3841, opened 2021-02-07
Unix/Linux: Don't link configuration.xml into the binary

### LINKED ISSUE #2729: Remove the linux compiled in configuration.xml
It was a good, but wrong, idea from the start. But we didn't know it at the time.

mvf:
Linking this in is a bit messy, and since Surge relies on other external
files in the data directory anway, it's also rather pointless.

Fixes #2729


baconpaul:
Great thank you 

juce has a super elegant way to include binaries so in xt we will move all platforms for that for config and window.wt but this is absolutely the right change for the 19 cycle!

PR #3857, opened 2021-02-09
Add option to toggle prev/next patch buttons constrained to current category

### LINKED ISSUE #3749: Make patch jog buttons continue to next or previous folder
Currently, the patch prev/next buttons only go through the currently selected folder, which is ok but is not the usual workflow I notice in other plugins.

I suggest moving the current behavior to right click on the prev/next buttons, and making left click just go through the whole flat preset list from the currently selected folder.

@baconpaul Doable for 1.8.2?

mkruselj:
Closes #3749


mkruselj:
Needs more work and inspection, so best not merge yet. Just wanted to get it out there.

PR #3881, opened 2021-02-12
Fix Mingw-w64 builds

mvf:


baconpaul:
Oh cool thanks.
Can you add a blurblet to the readme on how to run these builds on windows?

mvf:
Tbh, these are not very useful ATM. The tests complete, but the resulting DLLs don't load in hosts. IIRC the build also takes *forever* on Windows (like: minutes to link). I only use them for quick Windows cross builds from Linux and macOS. On there it's just a matter of passing `-DCMAKE_TOOLCHAIN_FILE=cmake/x86_64-w64-mingw32.cmake`.

baconpaul:
Oh cool thanks!

Issue #3827, opened 2021-02-05
Tuning aware filters


baconpaul:
Right now all the filters use note_to_pitch_ignoring_tuning

Would be nice if this was an option

May be 190 doable

Super useful for the combulator also so make sure to carry it there. (It’s really only important for the combs and key tracking but you know might as well do it in the general case)

PR #3882, opened 2021-02-12
Complete the tuning modes menus and tuning filter work

baconpaul:


baconpaul:
I think this will close #3827

PR #3890, opened 2021-02-15
Don't read zoom level on patch load, add option for drag-resize in VST3

### LINKED ISSUE #3834: UI zoom level problems in REAPER
**Bug Description:**
UI zoom is not behaving properly in REAPER. My default zoom level is set to 150%, and I would like it to be fixed at that level, but the zoom jumps around when I change presets or resize the parent FX window. See the video below for an example.

**Surge Version**
Surge Synthesizer
Plugin Host:	REAPER
System:	x86 CPU, Windows VST3, 64-bit
Build Info:	Built on 2021-01-27 at 12:54:59, using pipeline host 'WIN-T8F93AN0SJV'
Version:	1.8.1.9de9336

**Reproduction Steps:**
1. Set default zoom level to 150%.
2. Change presets and/or window size.

**Expected Behavior:**
Zoom level should always stay at 150%.

**Screenshots:**
zoom problems

**Computer Information (please complete the following information):**
 - OS: Windows 10
 - Host: REAPER
 - Version: 6.21

mkruselj:
Closes #3834

### DIFF
--- a/src/common/SurgePatch.cpp
+++ b/src/common/SurgePatch.cpp
@@ -1712,11 +1712,6 @@ void SurgePatch::load_xml(const void *data, int datasize, bool is_preset)
             int ival;
             TiXmlElement *p;
 
-            // This has a lecacy name since it was from before we moved into the editor object
-            p = TINYXML_SAFE_TO_ELEMENT(de->FirstChild("instanceZoomFactor"));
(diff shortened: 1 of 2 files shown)


mkruselj:
Also for some reason, the option to drag resize doesn't ever show its checkbox, so this needs to be fixed before merging...

PR #3894, opened 2021-02-15
Skin Component Model

### LINKED ISSUE #3605: Handle the CHSwitch2 Special Case in SkinSupport::reloadSkin
I'm puttingg this in late in cycle to deal with #3602 but it's clearly not the ultimate answer. Keeps the diff small tho.

### LINKED ISSUE #3782: Filter glyph placement
This one is juce-independent, really just parameterizing the glyph placement

Dave wants to put the filter glyph below the menu and have it be bigger. 

So basically add a relative offset etc.... type thing to the filter properties

[filters-2.svg.zip](https://github.com/surge-synthesizer/surge/files/5901424/filters-2.svg.zip)

Here's an SVG to test with

### LINKED ISSUE #3813: Not all of the <controls> take img= references
There is currently no external way for a skin developer to know when to use img vs bg_resource vs bg_id. 

My mental model was such that everything in <control> could accept img.....

### LINKED ISSUE #3846: A few quality of life requests for skin devs
Sorry to crash your issue with an edit dave - it's me bp Jus sticking a checklist at the top here

- [x] droplists for filers accept image="" rather than override the bmp.
- [x] everything else accepts image="" rather than override bmp
    - [ ] list
    - [ ] them
    - [ ] tho!
- [x] filter selector positioning of glyph or suppression of glyph
- [x] slider labels
- [x] grid

-------------

First off, apologies if these are contained elsewhere. I did a cursory look and didnt see these logged as requests or Ive tried to point out where I think there is already an existing request. 

**_General:_**

I would like to be able to make the droplists (for filters as an example) accept either image="" or bg_resource="" to specify the image being used as opposed to needing to overwrite the default bmpxxxx.svg. 

In fact, Id like all controls that can have an image be used for its visual style to accept image="". Id also suggest normalizing on a single approach in the XML. AKA use either _image_ or _bg_resource_ NOT both. I had difficultly with that at first and I couldn't quite work out which to use for what. 

For:

**_Menus_**
-------------
FilterSelector | ui_identifier: filter.type_1

Id like the ability to independently control the various elements. 

Meaning ability to offset the glyph (if it has to attached to the droplist control to work), the ability to size the glyph as desired, the ability to place it above, below, left, or right (I think this is a repeat of another feature request). Additionally, the ability to set the droplists dimensions and glpyh/button as needed/desired. I would like to be able to load my own image named whatever I like for the droplist visual design. Right now I have to overwrite the default bmpxxxx.svg to use my own image.

This is an example of something I was considering. The droplist is above, the "button" below showing that a notch filter was selected.

Screen Shot 2021-02-07 at 11 19 08 AM

**_Sliders_**
-------------
I think there are some improvements that can happen here beyond what you have already addressed (TY! for those label changes).  

Configuration for label orientation and spacing from slider track would be wonderful. Essential labels could go above, below, left, or right of the slider track. Some amount of control of the spacing from slider track to label would be amazing.  Id also like to be able to put the label at the very end of the track as opposed to the end of the slider throw. 

That said I think the current slider approach kinda needs  a bit of rejiggering. As an example of what I mean, I want to change the slider design to this with utilizes the entirety of the available slider track space:

Screen Shot 2021-02-07 at 8 18 35 AM

The net result is that the labels wind up on top of the control. Also and perhaps more importantly that slider design presents problems for the slider thumb placement given the current implementation. As it stands right now I cant quite get the slider thumb exactly where I want because how all of that slider drawing is done. 

**_General Working Stuff_**

I know we haven't talked about this one yet or anything. It just dawned on me as Im moving things around in the GUI right now.....

I would be amazing to be able to toggle a grid on and off on top of the actual running plug for skin devs. 

Essentially, an overlay of a grid that they can specify. AKA Its a 16px grid with subdivisions at 8 or similar. 

Why do you need that?!?!? I mean, your design tools provide that information...... right?!?!?!  

Why yes... yes they do. I keep running into situations where there is a one or two pixel variance from my design tool. So this isn't a super important or pressing request. BUT I think it could be useful for the more adventurous folks who on one hand appreciate precision while at the same time wanting to freestyle some GUI design.  Im not sure who the audience for that is beyond me but seems a reasonable request IMO.

baconpaul:
In S1.8 skin properties were ad-hoc. Add a model so we can
represent, name, enum doc them etc...

Along the way, add some properties, add a new tutorial starting
to show them, add an expansion to the skin inspector, and make
the filter menu much more configurable

Closes #3605
Closes #3813
Closes #3782
Closes #3846


baconpaul:
Make sure to add an "Addresses #3894" when doing the squash merge.

PR #3896, opened 2021-02-15
Windows: Replace fopen() with std::filebuf, remove UTF-8 hack

mvf:
With the removal of this hack, fopen() can now only see ASCII paths on
Windows since it expects ANSI and Surge strings are UTF-8. Use C++ API
that takes an fs::path instead (std::filebuf, std::fstream).

This is the first step on the way from "Approach B" to "Approach A" suggested in [#2579](https://github.com/surge-synthesizer/surge/issues/2579#issuecomment-688421813).


baconpaul:
Neat
When this journey is over in surge, we should reflect on

1. What it means for ShortCircuit
2. How all of this interacts with the juce::File API which SurgeXT will have at some edges (although in that case I think the answer is “grab the UTF8 string and pass it to the surge API, but just want to mention it)

PR #3897, opened 2021-02-16
ChowDSP Tape Part 1

### LINKED ISSUE #3585: Add ChowDSP ChowTape
Add [ChowTape](https://github.com/jatinchowdhury18/AnalogTapeModel) to the Surge effects rack. As mentioned in #1694 

ChowTape is a an analog tape simulation, based on a corresponding [academic paper](http://dafx2019.bcu.ac.uk/papers/DAFx2019_paper_3.pdf).

The original has many parameters, so I think there's two ways to deal with this. One option is to split the effect into several mini-effects, for example Tape Hysteresis, Tape Loss, Tape Wow/Flutter. The other option is to hardcode some of the parameters, and only expose the most important/noticeable ones.

If we were to go with the second option, here's the parameters I think I should include:
Drive
Saturation
Bias
Tape Speed
Spacing
Thickness
Gap
Wow Rate
Wow Depth
Flutter Rate
Flutter Depth

jatinchowdhury18:
Hi all,

I wanted to make a PR for merging the Chow Tape effect, even though it may not actually get merged for a while.

Currently I've implemented the tape hysteresis engine as well as the loss filter section. I've tried to tune the hysteresis engine to be very stable, while also not being too CPU heavy. For the loss filters I've tried to minimise the latency as best I can. That said, I was hoping to open this branch up for testing so I could get more feedback on those things.

I also haven't yet implemented the parameter types for the loss filter parameters, so those might be a bit clunky for now.

Thanks!

fixes #3585


baconpaul:
Cool! Feel free to leave the PR hanging out here so you can use CI to test builds and stuff too. Just let us know when you think you have something ready or want a test!

mkruselj:
You might need a `git submodule update --init --recursive` then rebase?

baconpaul:
Looks more like a missing include of <Cmath>

baconpaul:
Best way to get testing feedback is to push it into the nightlies. You OK if I merge even incomplete here? We can always take it out if you want before we ship 1.9!

jatinchowdhury18:
Sure, merging at this point seems fine! That way I can keep working on it incrementally, and getting feedback from the nightly testers. And like you said, we can always remove before 1.9 if we want to.

baconpaul:
OK! Merged! Awesome!

PR #3898, opened 2021-02-16
Add hover to the Wavetable menu

### LINKED ISSUE #2330: Add hoverability for widgets that don't have them yet
- [x] Status buttons
- [x] LFO type
- [x] OSC and FX menu
- [x] Wavetable selector+arrows
- [ ] FX grid
- [x] Step sequencer prev-next button triangles

baconpaul:
Closes #2330


mkruselj:
@baconpaul I should've conveyed this better - the L/R arrows and the WT name area should have separate hovers, since they are separate hitzones with separate actions. Currently the whole strip changes color which is quite garish IMO. Still, better to have a hover than not.

PR #3913, opened 2021-02-17
Some speedups for the non-release pipeline

### LINKED ISSUE #3912: PR Pipeline Speedup
1: only the unit tests need to be a release build; the other targets can build debug (faster)
2: the pr -> master needs to run everything but the merge-onto-master doesn't. Assuming that maintainers don't push onto master without running PR we can probably drop everything except codequal and a couple of canary builds (win vst3, mac au, lin vst3) in the second stage run.
3: --parallel

baconpaul:
Closes #3912


baconpaul:
Please don't merge this. Thank you :)

PR #3920, opened 2021-02-17
Modern osc correct unison offsets, better volume comp

mkruselj:
Waveform mix double-click default value is 0% now
Add sine snapshot preset

### DIFF
--- a/src/common/dsp/DPWOscillator.cpp
+++ b/src/common/dsp/DPWOscillator.cpp
@@ -32,9 +32,9 @@ void DPWOscillator::init(float pitch, bool is_display)
         if (n_unison > 1)
         {
             float mx = 1.f * u / (n_unison - 1); // between 0 and 1
-            unisonOffsets[u] = (mx - 0.5);
-            mixL[u] = (1.0 - mx) * oosun;
-            mixR[u] = mx * oosun;
+            unisonOffsets[u] = (mx - 0.5) * 2;
+            mixL[u] = (1.0 - mx) * oosun * 2;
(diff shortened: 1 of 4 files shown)


src/common/dsp/DPWOscillator.h
```diff
@@ -66,17 +66,17 @@ class DPWOscillator : public Oscillator
 
     // character filter
     bool dofilter = true;
-    double charfiltB0, charfiltB1, charfiltA1;
-    double priorY_L, priorY_R, priorX_L, priorX_R;
+    double charfiltB0 = 0.f, charfiltB1 = 0.f, charfiltA1 = 0.f;
```
baconpaul, line 69: 0.0 not 0.f for doubles.

baconpaul: (not that it matters; don't redo the PR for just this. but i will fix it up when clion complains at me :))

src/common/dsp/DPWOscillator.cpp
```diff
@@ -115,7 +115,7 @@ void DPWOscillator::process_block(float pitch, float drift, bool stereo, bool FM
              * So rather than lagging, back compute with the current dPhase
              * which makes us way more stable under phase changes
              */
-            double sBuff[3], sOffBuff[3], triBuff[3];
+            double sBuff[3]{}, sOffBuff[3]{}, triBuff[3]{};
```
baconpaul, line 118: we are actually fine with this memory uninitialized since it is initialized below. Curious why you added initializers? They do add extra instructions in the inner loop here. Can you remove?

PR #3935, opened 2021-02-19
Clear assigned modulation when changing oscillator type

### LINKED ISSUE #2224: Modulation not cleared when changing osc type
I noticed that modulation is not cleared when you change osc type. You can end up with stuff that doesn't make sense, like slider background gets blue for integer sliders in sine osc which aren't modulatable. 

Maybe in cases when the parameter name is the same when switching from osc type A to osc type B, it should be retained (WT/window morph, skew horizontal or vertical, not sure off-hand if there are any others).

mkruselj:
Closes #2224

### DIFF
--- a/cmake/stage-extra-content.cmake
+++ b/cmake/stage-extra-content.cmake
@@ -3,7 +3,7 @@
 #
 
 set(SURGE_EXTRA_CONTENT_REPO https://github.com/surge-synthesizer/surge-extra-content.git)
-set(SURGE_EXTRA_CONTENT_HASH 43219da451762265ab1e5f1664d83f7215eb9581)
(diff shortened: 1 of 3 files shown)


baconpaul:
We have to test the load paths carefully here

1: if you load patches does mod stick
2: same for daw session load 
3: and snapshot xml
4: a push through the daw preset mechanism
5: daw automation of the type

before you merge can you make sure you check all those carefully? Thanks!

mkruselj:
1. When loading a patch that uses the same osc type but different mod assignments, they load properly and as exxpected
2. Modulation sticks!
3. Modulation resets if using the menu and picking a different osctype, otherwise it sticks!
4. Modulation sticks!
5. Modulation sticks!

PR #3936, opened 2021-02-19
Some DPW branch removal (but only some)

baconpaul:
Move the DPW inner kernel to a template so we can branch less.
Not zero, but less.


baconpaul:
DO NOT MERGE. More coming. Thanks.

PR #3952, opened 2021-02-22
Fix a Tape crash on Linux (and probably win32 too)

baconpaul:
Blocks used in SSE need to be 16 byte aligned. The GCC layout
of this class knocked fadeBuffer out causing a core dump on
linux. Add alignas() directive


jatinchowdhury18:
Oh nice! Thanks for catching this :).

PR #3973, opened 2021-02-25
Integrate libsamplerate ahead of Nimbus/Twist SR correction

baconpaul:
The eurorack modules need SR conversion to and from 48khz.
This integrates libsamplerate and writes an assertive test
which shows it is integrated properly, but does not fix the
nimbus SR issue, just adds the dep, build, and test.


baconpaul:
**DO NOT MERGE**

PR #3977, opened 2021-02-26
Replace a bunch of unsafe string operations with safer alternatives

selenologist:
This isn't yet completely finished, but as I have to go do something else now and this touches such a wide span of code, I want to get these changes in before any other code changes.

strcpy and strncpy have been replaced with a wrapper function strxcpy which uses snprintf to safely do a string copy.
Some magic constants have been replaced and in general string ops are hardened a little more.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -32,7 +32,7 @@ if(CMAKE_CXX_COMPILER_ID MATCHES "Clang|GNU")
   # Any Clang or any GCC
   add_compile_options(
     -Wno-multichar
-    -Wno-format-truncation # squelch warning about snprintf truncating strings (see PR #3977)
+    -Wformat-truncation=0 # squelch warning about snprintf truncating strings (see PR #3977)
     # Targetting Windows with GCC/Clang is experimental
     $<$<NOT:$<BOOL:${WIN32}>>:-Werror>
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -32,7 +32,6 @@ if(CMAKE_CXX_COMPILER_ID MATCHES "Clang|GNU")
   # Any Clang or any GCC
   add_compile_options(
     -Wno-multichar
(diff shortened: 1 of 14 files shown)


selenologist:
Sigh. Alright so it wasn't just that AirWindowsEffect.h that threw the error.

So looks like we need `-Wformat-truncation=0` to squelch the warning/error, but I'll wait for someone else's thoughts on that.
(then AirWindowsEffect.h needs to go back to how it was in the first commit of this PR)

baconpaul:
So that format overflow error only happens with gcc > 8 or 9 or something. It is a new check. It is also what caused things like me growing the prefix to 19 rather than 16 then truncating later. I think if we have good and strict use of snprintf it is not a check we need. My only question is if I do

```
auto a = 1234567890;
char t[16];
snprintf(t, 16 "%i%i%i", a,a,a);
```

do I get a properly null terminated string? If the answer to that is "yes" then we can modify CMakeLists.txt to turn the warning off (it should be pretty clear where to do it at the top in the gcc block) and get on with our lives.

wdyt?

selenologist:
The only case `snprintf` should fail to write a null byte, is if `size` is zero - so we're good.

I'll make the changes - good to know re prefix, I'll put that back down to 16 too.

CMakeLists.txt
```diff
@@ -32,7 +32,7 @@ if(CMAKE_CXX_COMPILER_ID MATCHES "Clang|GNU")
   # Any Clang or any GCC
   add_compile_options(
     -Wno-multichar
-
+    -Wno-format-truncation # squelch warning about snprintf truncating strings (see PR #3977)
```
baconpaul: You need this at line 73 in the gcc block; this is gcc+clang both block

selenologist: Oops, thanks for that

baconpaul:
What a huge undertaking. Thanks!

PR #3996, opened 2021-02-27
Add init() call to suspend() method of effects that didn't do it so far

### LINKED ISSUE #3950: Implement "Suspend" properly on all FX
Self-explanatory, pretty much.

mkruselj:
Closes #3950


baconpaul:
I looked at the code here. I'm not at all sure that the Tape suspend -> init call is right. More confident on Chow and Exciter.

Vocoder is fine though!

but we should knock out tape or test it pretty extensively - the init does quite a bit of allocation we may not want at a play boundary.

mkruselj:
OK, can easily remove it from Tape and merge!

PR #4006, opened 2021-03-01
Add a toggle for Keytrack/Retrigger across all oscs in the scene

### LINKED ISSUE #4005: keytracking on/off for all osc's
discussed with @baconpaul on discord 1/3/2021 to add feature to set keytracking switchable on/off for all osc's
thanks
dave

mkruselj:
Closes #4005


mkruselj:
@baconpaul The only line I'm not sure about is L3148 `repushAutomationFor(p);` Ideally that `p` would need to be the one related to a particular parameter ID, right? Currently it would just repush automation for the currently selected `p` n_osc times since it's in a loop. I don't have an idea on how to fix it so that it repushes automation for each of the keytrack/retrigger params... I won't be merging this until that particular quandary is resolved.

PR #4026, opened 2021-03-04
Dyanmic Phase/Shuffle label on LFO

baconpaul:
This was, I think, @mkruselj's biggest request and one I kept
forgetting. :)


mkruselj:
Well, maybe not the _biggest_ request I had, but definitely one of the biggest annoyances 🙂

PR #4035, opened 2021-03-05
Fix Neuron crashing with Surge XT

jatinchowdhury18:
Basically, the problem was as follows:
- `Neuron` allocates the memory for it's delay lines in the `init()` function.
- With Surge XT, Neuron's `process()` sometimes gets called before `init()`, or maybe in parallel threads.
- `process()` tries to push samples into the delay line, but the delay line doesn't have it's memory allocated yet!
- segfault

For now, my solution is to allocate the required memory in the constructor, rather than in `init()`, but if this is an issue that potentiall affects other effects as well, we may need to take a look at which threads are calling each function, and make sure we don't try to `process()` an effect before it has been `init()`ed.

### DIFF
--- a/src/common/dsp/effect/chowdsp/Neuron.cpp
+++ b/src/common/dsp/effect/chowdsp/Neuron.cpp
@@ -40,8 +40,8 @@ void Neuron::init()
 
     os.reset();
 
-    delay1.prepare(dsamplerate * os.getOSRatio(), BLOCK_SIZE, 2);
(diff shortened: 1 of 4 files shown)


baconpaul:
Yes thanks we should merge this fix (init should really just clear not allocate memory if possible) but it shouldn’t be necessary - the situation you describe is indeed a bug in surge xt which I haven’t been able to Pin down.

PR #4065, opened 2021-03-11
Sine, FM2, FM3 with Extend on feedback by default

### LINKED ISSUE #4064: FM feedback default to extended range
Please consider making the FM Feedback slider default to the extended range. 

The current default range provides very limited impact to the sound. When setting it to extended, some really great sound design capabilities open up. I have been using Surge for 13 years (!), and today I discovered this option for the first time.

At a minimum the ability to extend the range should be mentioned in the manual.

Thanks!

-Bill

mkruselj:
Multiple columns for FX preset subfolders (Windows only)
Fix typo in FX preset streaming version token name
Closes #4064

### DIFF
--- a/resources/data/configuration.xml
+++ b/resources/data/configuration.xml
@@ -13,12 +13,12 @@
     </type>
     <type i="2" name="Wavetable" retrigger="0" />
     <type i="7" name="Window" retrigger="0" />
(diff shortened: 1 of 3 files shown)


baconpaul:
Looks great!

PR #4069, opened 2021-03-12
Turn off the menu on effect settings

baconpaul:
Noone liked it and it was a bad idea


mkruselj:
I liked it and it just needed constraining to A/B boxes 🙁

baconpaul:
Yeah we can do that. As you see I left the function there just reverted the broken behavior so we could get the nightlies running again.

Issue #2089, opened 2020-06-10
Coupled slider deactivation


mkruselj:
Book of work:

- [x] Phaser: if Count parameter == 1, show Spread as semitransparent (it shouldn't have a Deactivate option, just show as semitransparent because it's non-functional in this case)
- [x] Tape: Deactivate the whole Hysteresis section if right-click on Drive deactivate option is toggled
- [x] Tape: Deactivate the whole Loss section if right-click on Speed deactivate option is toggled
- [x] Tape: Deactivate the whole Degrade section (when it's implemented) if right-click on Amount deactivate option is toggled
- [x] EQ: Deactivating one band should deactivate all 3 sliders pertaining to it

__________

Related mostly to the new parameters in the rotary speaker, but could be useful elsewhere, who knows.

It would be great to have a way to deactivate multiple parameters, while the activate/deactivate toggle is only on one slider. Basically a linked relationship.

This would mean that both Amp related sliders in rotary speaker effect would be semitransparent when drive is deactivated.

mkruselj:
Just adding a related note here based on what @baconpaul said on Discord - the mechanism that is going to be implemented to support this feature request can be reused to do other things, like renaming parameters on the fly based on certain other parameter values (i.e. LFO Phase / Shuffle becomes only Phase or only Shuffle based on LFO type being Step Seq or not).

baconpaul:
Turns out what I said on discord was wrong and we should be a bit more tactical and just do activation as one step.

baconpaul:
So the three use cases to solve here right now are

1: When drive is deactivated in the rotary, so is waveshape
2: When phaser stages != 4 then deactivate some other stuff
3. Pattern to make that at least not tedious

mkruselj:
5. Deactivate LFO rate and phase sliders when in Envelope mode.

baconpaul:
sigh this one is so tricky. I don't know a good way to structure it. Like: I just don't know where to put the logic.

here's the problem. When we deactivate something we set the deactivated to true and rebuild. Fine. 

So at that point we have a param. We need to know the other params to deactivate. How do we know that? Well clearly you would want some data structure adjacent to the parameter where you can ask for 'deactivation buddies'. 

[...]

PR #4071, opened 2021-03-12
[WIP] Integrate BBD emulation into ensemble

jatinchowdhury18:
Still a bit of a WIP, but I wanted to post this so that:
1. You all could give it a try :).
2. Hopefully get some help with performance optimisation.
3. Give myself a bit more motivation to get it finished.

The AA filter is sounding pretty good, and I added a "Nonlinearity" parameter as well, to control how much of the nonlinear characteristic from the BBD is audible.

Right now, I have 3 BBD stage settings: 512, 1024, and 2048 stages. I also left the sinc delay in there, since I think it sounds pretty good as a "cleaner" version of the effect. Right now, I have a weird hack with the "Stages" slider so that it selects the Sinc delay from 0-25%, 512-stage BBD from 25-50%, and so on. I feel like there was a conversation about this that I've forgotten about, but were we planning to use a drop-down menu, or something, to control the number of stages/delay type?


mkruselj:
We could definitely use a menu to select number of BBD stages. The parameter ctype needs to be an integer for that.

baconpaul:
The two build fails seem to be (1) forgetting an `#include <memory>` in a header and (2) if constexpr being a c++17ism which I think can be fixed fairly well with a switch.

I’ll pull today and look and perhaps also add the menu. You OK if I do a merge if I get it building?

baconpaul:
OK @jatinchowdhury18 if you check out `baconpaul/bbd-ensemble` and specifically commit 869d9ed7ab3ed667cf28bfe3bc4b2c967f8bd9d7 on top of your branch you can see that I got it fixed on mac and linux and added a dropdown.

I can push that to your repo if you want or you can just grab the commit too.

But the 512, 1024, and 2048 all crackle from CPU in debug on my MBP. So definitely some performance work to be done!

baconpaul:
Oh and annoyingly the if constexpr thing couldn't become a switch; it had to become a template specialization. but i did that!

baconpaul:
oh and finally i ran clang-format so codequal won't fail which adds some format-only diff noise. sorry bout that not being a separate commit.

jatinchowdhury18:
Oh awesome! Thanks for the assists :). I just pushed your commit onto this branch (I can squash it down later before merging the PR).

I'll try taking a deeper look at performance today... It's kind of tricky since the BBD runs at it's own sample rate, and especially when you get to a higher number of stages, that sample rate can get to be very large.

By the way, the numbers of stages that I chose are very much arbitrary. Did we have a particular set of stage choices that we were shooting for?

jpcima:
> By the way, the numbers of stages that I chose are very much arbitrary.

They're usually power-of-2 but not always. The smallest one that I am aware of is 128.
TCA350Y was used in many machines, it had 185.

mkruselj:
> Did we have a particular set of stage choices that we were shooting for?

128, 256, 512, 1024, 2048. If we're wild we can do 4096 and 8192, unless CPU-prohibitive.

jatinchowdhury18:
@mkruselj, yeah, I'll add those sizes, and I wanted to attempt some performance optimisations before merging as well.

baconpaul:
yeah i can look in a profiler too and see what i can see.

jatinchowdhury18:
Okay, so I was able to add the 128 and 256 stage delay lines, and make a handful of performance improvements. That said, it's still more expensive than I would like it to be, plus some of the code I added is pretty messy. I'd like to come back to it (probably tomorrow) and clean things up a bit, but if either of you want to take a look in the meantime, please go ahead!

mkruselj:
clang-format :)

jatinchowdhury18:
Ah crap, I keep forgetting I haven't installed clang-format yet on this machine... will rectify....

jatinchowdhury18:
Also, is it possible to disable the "AA Cutoff" and "Nonlinearity" sliders when we're in Sinc delay mode?

mkruselj:
That's a tricky subject. It's basically issue #2089 but @baconpaul says it's a gnarly one...

PR #4077, opened 2021-03-13
Readme tweak

mynameismuhl:
makes linux CMake example behave as advertised
& reappearence of some table content in browser view


baconpaul:
Thank you!

PR #4105, opened 2021-03-17
Large amount of work on BBDEnsembleEffect

jatinchowdhury18:
- Rename and reorder certain Ensemble params, add Clock Rate param
  - An SSE Complex Number class
  - BBD filters using SSE parallelisation
  - BBD nonlinearity using templates
  - Implemented BBD nonlinearity for Sinc delay
  - BBD nonlinearity using SSE parallelisation


baconpaul:
Ahh looks like an `#include <memory>` missing somewhere so you aren't finding `std::unique_ptr` on linux

PR #4112, opened 2021-03-17
Implement clock rate and feedback parameter for BBD Ensemble

jatinchowdhury18:
Clock Rate: I tried to implement this based on @mkruselj's notes in `surge-development`. Not sure I nailed exactly what he was going for, but at the very least I got the center delay time to be correctly proportional to the clock rate and number of stages. I chose 256 stages for the Sinc delay mode, but that's completely arbitrary.

Feedback: I added a feedback path around the delay lines, along with a highpass filter to prevent DC build-up. There's a fair amount of hand-tuning in this implementation, which maybe someone with better ears than mine could try to improve.

I also tried to mangle Paul's parameter deactivation code to work for the input filter slider... I'm not totally sure I implemented it correctly though?

### DIFF
--- a/src/common/dsp/BiquadFilter.cpp
+++ b/src/common/dsp/BiquadFilter.cpp
@@ -73,13 +73,9 @@ void BiquadFilter::coeff_LP2B(double omega, double Q)
             (w_sq * w_sq) + (M_PI * M_PI * M_PI * M_PI) + w_sq * (M_PI * M_PI) * (1 / Q - 2);
         double G1 = min(1.0, sqrt((w_sq * w_sq) / den) * 0.5);
 
-        double cosi = cos(omega), sinu = sin(omega),
-               // alpha = sinu*sinh((log(2.0)/2.0) * (BW) * omega / sinu),
-            alpha = sinu / (2 * Q),
-               // G1 = 0.05, //powf(2,-log(M_PI/omega)/log(2.0)),
-               // set aa to 6 db
+        double cosi = cos(omega), sinu = sin(omega), alpha = sinu / (2 * Q),
 
-            A = 2 * sqrt(G1) * sqrt(2 - G1), b0 = (1 - cosi + G1 * (1 + cosi) + A * sinu) * 0.5,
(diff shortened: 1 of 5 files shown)


baconpaul:
i generally make the throwaway object a static in case the param lives longer than the effect. also in this case you can leave the adjacent param unused. but otherwise looks spot on!

jatinchowdhury18:
> i generally make the throwaway object a static in case the param lives longer than the effect. also in this case you can leave the adjacent param unused. but otherwise looks spot on!

I was thinking about that though... what happens if there's multiple instances of the effect? If the object is static, how will it know which instance of the effect `getValue()` is being called for? Also, what is the adjacent param?

baconpaul:
It doesn't matter. The getValue has no state and is passed the parameter. You just need to look up your control group. Look at how I do it in EuroTwist.cpp or at how it works in SurgePatch with the lfoDeactivation thing

The 'adjacent parameter' is for when you have explicit vs value based deactivation. One of them is primary and the menu to deactivate shows up on all of them.

I dropped a little doc in 'doc/DynamicBlah' which kinda explains it.

mkruselj:
@jatinchowdhury18 You think the input filter should deactivate itself if in sinc mode? Dunno, might be useful to have it still...

jatinchowdhury18:
Actually, yeah I think @mkruselj is right. From the user side it makes sense that the input filter will still work in sinc mode (even though it'll have to be different under the hood). I can go ahead and do that.

mkruselj:
Yeah just checked the code, there's no need for this tbh. Let's leave the filter in, but just make it regular deactivatable (see for example Treemonster filters).

jatinchowdhury18:
Well, in that case, I feel like it probably shouldn't be deactivateable at all, since there's not really a way to avoid that processing in BBD mode. (Remember the filters in BBD mode are integrated into the BBD sample rate conversion)

mkruselj:
Right. Yep, that would be the way to go!

baconpaul:
https://github.com/surge-synthesizer/surge/pull/4114

check out that PR @jatinchowdhury18 for the pattern I'm recommending for deactivation.

PR #4143, opened 2021-03-21
Tape: implement degradation and tone control

jatinchowdhury18:
I've now implemented the tone control, and chew/degrade sections from ChowTape. @mkruselj would you mind giving this a try and see if the Chew/Degrade match up with the parameter mappings you had in mind?

A note about the tone control: in the original plugin I use opposite frequency responses on each side of the tape distortion, similar to what you would have with a pre/post-EQ in an actual tape machine. In Surge, I found this made it pretty hard to hear the effects of the tone change, so I've removed the post-tape filters for now.


mkruselj:
Perhaps the Tone should be only after the tape rather than before the tape?

Eother way, merged it for now! Any tweaks can be done down the road 🙂

PR #4152, opened 2021-03-22
A number of tweaks on Alias oscillator

mkruselj:
Bit Depth has its own ctype, range 1-8 bits, floating point
Bit Depth applied at the output of the oscillator rather than at phase accumulator
Restored `<random>` header but seeded it with `rand()` instead of using a system call
Reference to number of waves not hard-coded anymore but resolved through `extern`

Additionally, renamed waveguide... to stringosc... where applicable

### DIFF
--- a/src/common/dsp/AliasOscillator.cpp
+++ b/src/common/dsp/AliasOscillator.cpp
@@ -16,13 +16,13 @@
 // This oscillator is intentionally bad! Not recommended as an example of good DSP!
 
 #include "AliasOscillator.h"
-#include "DebugHelpers.h"
+#include <random>
 
-// says ao_n_types is undefined. makes no sense, whatever, for now I'm hardcoding it. TODO FIX.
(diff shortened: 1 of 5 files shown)


src/common/dsp/AliasOscillator.cpp
```diff
@@ -160,8 +162,14 @@ void AliasOscillator::process_block(float pitch, float drift, bool stereo, bool
 
             phase[u] += phase_increments[u] + fmPhaseShift;
         }
-        output[i] = vL;
-        outputR[i] = vR;
+
+        // bitcrush output
+        auto bits = limit_range(localcopy[oscdata->p[ao_depth].param_id_in_scene].f, 1.f, 8.f);
+        auto quant = powf(2, bits);
```
baconpaul, line 168: Oh goodness call this pow and divide outside the loop since it is constant across the loop!!

mkruselj: Done! That was stupid.

PR #4162, opened 2021-03-23
Change LFO Shuffle to Phase when in Step seq mode and Rate is deactivated

VincyZed:
When in StepSeq waveform **and the rate parameter is deactivated**, switch the label of the Phase/Shuffle parameter back to Phase. This change was made because when rate is deactivated, this parameter scrubs through the stationary waveform like it does with other LFO shapes, instead of "shuffling".


mkruselj:
You need to run clang-format. There's an add-on for Visual Studio that does it, install it then Ctrl+R, Ctrl+D on the changed file, save and push commit.

VincyZed:
Will do tonight. Thanks for pointing this out. :)

Issue #4117, opened 2021-03-17
Scene Hard Clip is off by +4dB


mynameismuhl:
Linux/64bit/Reaper
Version:	1.9.main.99908055
steps to reproduce:

- Turn off your speakers
In Surge choose :
- Global Hard Clip Disabled
- Scene Output level 0dB
- Scene A Hard Clip at 0dBFS
- choose latch to play sound
- pull gain all the way up
- meters in Reaper peak at +4dBFS

To test Hard Clip at +18dBFS
you must turn off automute in Reaper first
(Options - Preferences - Mute/Solo)
Meters will peak at +22dBFS

Global Hard Clip clips at correct values.

mkruselj:
Hmmm, can confirm this in SPAN too. Curious.

baconpaul:
Another great report thank you.

I wonder if it is something to do with clipping before we downsample from the oversample rate? Unlikely. I also notice the global amplitude comes after the scene clip. Wonder if that’s it also?

Can look before we ship 1.9 for sure though!

mynameismuhl:
So i read into /doc(s)
still i don't understand the signal flow

- HPF sits post Scene Output? If I deactivate it in above scenario 
meters now peak at 3.1dB (resonance?).

- I can send individual Scene Outputs to the daw pre masterFX,
excluding FX Return but still Scene Output is not independent
from Global Volume. So GV is not a stereo bus fader but more a VCA Fader
controlling all outputs?

OR

- insert FX are post HPF so post Scene Output 
but applied when scene is individualy routed to daw 
so SceneOutput != SceneOutput 
?

any of this true?

Tl;dr: 
confused.by --routing -h

mkruselj:
HPF is at the end of scene output before insert FX.

mynameismuhl:
Thanks,
and sorry my comment was kind of a mess.
What i meant was:
Is HPF located post Scene-Output-Attenuation/clipping?
Because if yes you found aprox.1dB.
The remaining 3dB could then be a mono-summing/pan-law issue somewhere post clipping.
But just guessing.

mkruselj:
HPF is before output attenuation/clipping (you can see it in the code before the line I posted above).

baconpaul:
So I am wondering (but am still not sure) if this is just a mislabeled menu. Let me explain.

[...]

PR #4163, opened 2021-03-23
move last sceneout-hardclip-stage post insert

mynameismuhl:
adresses remaining issues of #4117
- prevents global hardclip from having influence on individual scene outputs
- sceneout-hardclip now clips post insert


mkruselj:
Looks like it's not properly clang-formatted still...

PR #4164, opened 2021-03-23
Add memory access modes to Alias. Also, default cmake to RelWithDebInfo.

selenologist:
Insane feature: `AliasOscillator` can use areas of Surge's memory as a wavetable.

Needs better choices of memory region.

Tacked on the switch to `RelWithDebInfo` in `CMakeLists.txt` too.


baconpaul:
Hi!

Can we wrap that RelDebInfo in an

``
if( AZURE_PIPELINE )
   set( Release )
else()
   set( RelDebInfo )
endif()
```
perhaps?

Also if you want, you can add another argument to package-vst3.sh which is 1 or 0 and add a `$<IF$<CONFIG:RELEASE>,1,0>` to the script and strip but that's extra.

PR #4166, opened 2021-03-23
Linux CMake Lists

baconpaul:
RelDebInfo in user builds, Release in pipelines, don't strip
DEBUG builds


baconpaul:
I'm not going to merge this. I'm not at all convinced that RelWithDebInfo == Release performance on all our platforms.

If you want RelWithDebInfo put it on your build command line.

PR #4171, opened 2021-03-24
TapeEffect: improve output gain and chew dry/wet time calculation

jatinchowdhury18:
Fixed some issues with the "Chew" part of the tape effect (one of the random number generators used for calculating dry/wet times was not initialized right). I also adjusted the default parameters and added a makeup gain so that when you don't get a volume drop when initializing the effect.

@mkruselj, I played around with moving the tone filters to after the hysteresis. I ended up choosing not to, but I'm not married to that decision. If you think it sounds better after, I'd be happy to move it.


mkruselj:
After some testing, it's fine if tone is before hysteresis, I think. I will double check once again but I think we good.


Thanks for the fixes!

PR #4212, opened 2021-03-29
Make Alias Additive calculations fixed point and less frequent

selenologist:
No more expensive sinf()! Uses the pre-calculated sine table. Harmonic amplitudes are now also normalised.

This breaks patches. Nightlies gonna nightly!


mkruselj:
I'd appreciate @baconpaul's code review before merging. It seems fine to me otherwise, but yeah I'm not exactly a good metric for code reviews. :grin:

baconpaul:
Looks good

Issue #4176, opened 2021-03-24
Loading FX preset with deactivated param doesn't work when changing FX types


mkruselj:
After #4175 is merged:

1. Open Surge
2. Open EQ3 and deactivate one band
3. Save FX preset

Then:

1. Reopen Surge
2. From no FX loaded, load your saved EQ FX preset -> band will not be deactivated
3. Load the same FX preset again -> band will be deactivated as it should

FX lifecycle shenanigans, apparently.

mynameismuhl:
This fixed it for me but is tested on linux only

baconpaul:
Yeah that fix would break other things alas. The right fix is quite a bit trickier and invokes making sure fx presets are consistently applied after the new fix init ctrltypes are called.

baconpaul:
So ignore me completely. Your fix is right and I misread your diff :( Getting coffee now

https://github.com/surge-synthesizer/surge/blob/main/doc/FXLifecycle.md

There’s the foll documentation on the fx lifecycle. As you can see when you load from snapshot we apply values then call init_ctrltypes but not init_defaultvalues.

You are exactly correct that moving the deactivated state from init_ctryltypes into init_defaultvalues fixes EQ. And I misread your diff as only doing half of that!! (Sorry again)

So the way to close this issue is simply to go through all the REST of the FX and make sure that when we default deactivated we do it in `init_defaultvalues` and not `init_ctryltypes` like your diff does.

I’ll merge your diff when CI is done. @mkruselj if you wanna scan the rest of the FX for this that’d be awesome or I can do it when I’m back!

mynameismuhl:
Cool. Happy this worked.
I can also apply that to the rest of the fx where needed
if you are not doing that already @mkruselj ?

baconpaul:
Thanks! And again sorry for th initial misread!

mkruselj:
> 
> 
> Cool. Happy this worked.
> I can also apply that to the rest of the fx where needed
> if you are not doing that already @mkruselj ?

Feel free to do the rest of FX (and add yourself to AUTHORS file while at it, as mentioned in your PR) :slightly_smiling_face:

PR #4214, opened 2021-03-30
read deactivated param in eq preset on first load

mynameismuhl:
adresses #4176


baconpaul:
Oh unfortunately that is not the correct fix. That will leave the param in whatever state was adopted before which is also not what we want

the fix to #4176 Is to make sure deactivated state is applied after init ctrltypes is called for all fx

so unfortunately can’t merge this, I’m sorry

baconpaul:
I’m sorry ignore me
This change is totally correct.
Merging when ci passes, 
Will update 4176

mkruselj:
@mynameismuhl Please add yourself to AUTHORS file in your next commit! :)

PR #4215, opened 2021-03-30
Fix dry-wet mix in ring modulator effect

mkruselj:
Now it's executed after downsampling, rather than within oversampling branch.


baconpaul:
Looks good!

PR #4216, opened 2021-03-30
move default deactivated to init_defaultvalues

### LINKED ISSUE #4176: Loading FX preset with deactivated param doesn't work when changing FX types

ghost:
in fx with deactivatable parameters
authorised myself

closes #4176


baconpaul:
Looks good, and welcome to the authors list!

PR #4225, opened 2021-04-01
BBDEnsembleEffect: Adjust input filter freq to depend on clock rate

jatinchowdhury18:


src/common/dsp/effect/BBDEnsembleEffect.cpp
```diff
@@ -61,11 +61,13 @@ int ensemble_num_stages(int i)
...
-float calculateFilterParamFrequency(float paramVal, SurgeStorage *storage)
+float calculateFilterParamFrequency(float *f, SurgeStorage *storage)
 {
-    return std::min(2 * 3.14159265358979323846 * 440 *
-                        storage->note_to_pitch_ignoring_tuning(paramVal),
-                    25000.0);
+    auto param_val = f[BBDEnsembleEffect::ens_input_filter];
+    auto clock_rate = f[BBDEnsembleEffect::ens_delay_clockrate];
+    auto freq = 2.0f * (float)M_PI * 400.0f * storage->note_to_pitch_ignoring_tuning(param_val);
+    auto freq_adjust = freq * std::pow(clock_rate * 0.01f, 0.75f);
```
jatinchowdhury18, line 69: @mkruselj This line controls the connection between clock rate and the anti-aliasing filters, getting closer to the sound that you had shared examples of. The `0.75f` factor is something I had tuned by hand, but I don't have my headphones with me at the moment, so you may want to take another look at that factor.

mkruselj: Will check it out, thanks!

Issue #4226, opened 2021-04-01
That windows menu split thing


baconpaul:
Just remembering this for 19 @mkruselj

mkruselj:
Linking:

https://github.com/steinbergmedia/vstgui/issues/194
https://www.kvraudio.com/forum/viewtopic.php?f=33&t=563088

baconpaul:
I think those bugs arent getting fixed though right? So off by default surge option which doesn’t call the api still the plan?

mkruselj:
Or take it out completely, I guess... What a bummer. We can still do this last minute so I'd wait to see if we get some replies eventually.

baconpaul:
It is super duper useful. Why not make it an off-by-default option now and just let folks know in surge dev? Easy to flip it if we get a response.

But I'm not sure how widespread the problem is. Just worried about beta testers running into this as a first bug and not going further during our test push basically.

and yeah, what a bummer :(

mkruselj:
It only seems to happen in 20H2. And to be honest, I wonder if it also happens with the latest VSTGUI at head!

PR #4228, opened 2021-04-01
Add option (off by default) for multi-column patch browser menu (Win only)

mkruselj:
Addresses #4226


baconpaul:
Looks correct to me. Thanks and again what a bummer :(

PR #4237, opened 2021-04-02
Support single cycle wavetables

mkruselj:


baconpaul:
So before I merge this I want to fully understand why this line 

```
  float newlevel = distort_level(
        oscdata->wt.TableF32WeakPointers[mipmap[voice]][tableid][state[voice]] *
            (1.f - tblip_ipol) +
        oscdata->wt.TableF32WeakPointers[mipmap[voice]][tableid + 1][state[voice]] * tblip_ipol);
```

doesn't break. I think it is because we pad out the wavetable with copies bu twnat to confirm that in a debugger. Please don't merge until I take the time to do that.

baconpaul:
The reason this works is twofold

1: BuildWT adds padding in the pointer arrays to take you up to 3 tables no matter what
2: The oscillators basically clamp the table at ntables so the above line is always called with tblip_ipol = 0

So fine to merge.

PR #4245, opened 2021-04-03
Fix clipboard not taking ExtraConfigurationData into account

### LINKED ISSUE #4244: copying alias with additive to another osc does not remember additive settings
copying alias with additive to another osc does not remember additive settings: is that expected behaviour?

mkruselj:
Closes #4244

Also use a constant instead of a magic number for number of Alias osc additive partials.


mkruselj:
Don't wanna merge until @baconpaul takes a look!

baconpaul:
The coy and paste stuff is great thank you!
The constant is wrong. It should be a static constexpr member of aliasoscillator so it’s usage is scoped not a new global in storage

PR #4247, opened 2021-04-04
Template optimisations for Alias Oscillator.

selenologist:
Alias oscillator is now 'done' :sunglasses:


selenologist:
Oh hang on a sec, let's take that FM depth smoothing out too. One moment/

PR #4261, opened 2021-04-07
Deactivatable low/cut filters in a number of FX

mkruselj:
This PR adds an option to disable low/high cut filters in effects that previously didn't have this option. Also includes streaming version handling, of course.
Also, use μm as unit for microns in Tape.
Also, Reverb 1 filters aren't using slowlag variant anymore.


baconpaul:
All looks good here

PR #4265, opened 2021-04-07
Return of the multi-column patch browser on Windows!

mkruselj:
The reason why multi-column menu didn't show properly on some machines was because they didn't have any user presets.
So some checks I did were quite wrong and not thought through very well.
Now we have a much more bulletproof solution, plus we now also construct the menu more properly as far as Factory and 3rd Party headers, and separators, in any combination of any of them missing/being removed by the user.

Thanks to KvR user `lkjb` for helping in debugging! [Link to post here!](https://www.kvraudio.com/forum/viewtopic.php?p=8076473#p8076473)

### DIFF
--- a/src/common/gui/CPatchBrowser.cpp
+++ b/src/common/gui/CPatchBrowser.cpp
@@ -92,8 +92,6 @@ CMouseEventResult CPatchBrowser::onMouseDown(CPoint &where, const CButtonState &
     int root_count = 0, usercat_pos = 0, col_breakpoint = 0;
     auto patch_cat_size = storage->patch_category.size();
 
-    printf("3P: %d | USR: %d\n", storage->firstThirdPartyCategory, storage->firstUserCategory);
-
     if (single_category)
     {
         contextMenu->setNbItemsPerColumn(32);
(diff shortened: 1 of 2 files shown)


src/common/gui/CPatchBrowser.cpp
```diff
@@ -87,8 +87,12 @@ CMouseEventResult CPatchBrowser::onMouseDown(CPoint &where, const CButtonState &
 
     int main_e = 0;
     // if RMB is down, only show the current category
-    bool single_category = button & (kRButton | kControl);
+    bool single_category = button & (kRButton | kControl), has_3rdparty = false;
     int last_category = current_category;
+    int root_count = 0, usercat_pos = 0, col_breakpoint = 0;
+    auto patch_cat_size = storage->patch_category.size();
+
+    printf("3P: %d | USR: %d\n", storage->firstThirdPartyCategory, storage->firstUserCategory);
```
baconpaul, line 95: Ooops! Left  sprintf in!

baconpaul:
All looks good except that printf

Issue #4210, opened 2021-03-29
Modulator Envelope Presets Saving to LFO Folder


bboxdave:
When I save a preset that I created for the Envelope type modulator using the Save env as... menu item within the Surge interface, it saves to the user subdirectory Documents\Surge\Modulator Presets\LFO instead of Documents\Surge\Modulator Presets\Envelope . Presets for the other modulator types save to the appropriate subdirectories (LFO, MSEG, Step Seq).

Surge Synthesizer
Plugin Host:	Bitwig Studio
System:	Windows 64-bit VST3 on         Intel(R) Core(TM) i7-3770 CPU @ 3.40GHz
Build Info:	Built on 2021-03-28 at 02:32:17, using pipeline host 'WIN-S2NGKRSV6OL' with 'MSVC-19.16.27045.0'
Version:	1.9.nightly.081cbbf

Windows 10, Bitwig 3.3.6

mkruselj:
It's essentially a part of LFO because it's applied to the whole LFO waveform. just for envelope the LFO waveform is a constant 1.

Not exactly a bug, and considering people already might have presets done with it, I wouldn't want to change that folder nowžć.

PR #4312, opened 2021-04-16
Put LFO envelope mode presets into Envelope subfolder when saving

mkruselj:
Temporarily do the same for Function presets


mkruselj:
Closes #4210

Issue #4387, opened 2021-04-24
1.9.0: UnitTestsDSP fails in modified arch build


dvzrv:
**Bug Description:**
When running the unit tests for 1.9.0 the UnitTestsDSP fails.

**Surge Version**
This information is found on the About screen, which you get to from the bottom right menu

* Version: 1.9.0
* Plugin Type: build
* Bitness: 64-bit

**Reproduction Steps:**
1. Build tests
```
# set datapath to local dir for testing
sed -e 's|/usr/share/Surge|resources/data|' -i src/common/SurgeStorage.cpp
# build surge-headless (test-suite)
cmake -DCMAKE_INSTALL_PREFIX='/usr' \
           -DCMAKE_BUILD_TYPE='None' \
           -W no-dev \
           -B build-test \
           -S .
make VERBOSE=1 -C build-test
```
2. Run `build-test/surge-headless`:

```
# surge-headless: 1.9.0 built: 0 0

~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
surge-headless is a Catch v2.9.2 host application.
Run with -? for options

-------------------------------------------------------------------------------
Every Oscillator Plays
  Oscillator type Wavetable
-------------------------------------------------------------------------------
/build/surge/src/surge-1.9.0/src/headless/UnitTestsDSP.cpp:781
...............................................................................

/build/surge/src/surge-1.9.0/src/headless/UnitTestsDSP.cpp:806: FAILED:
  REQUIRE( sumAbsOut > 1 )
with expansion:
  0.0f > 1

===============================================================================
test cases:      59 |      58 passed | 1 failed
assertions: 4643968 | 4643967 passed | 1 failed
```

**Expected Behavior:**
All unit tests pass

**Screenshots:**
n/a

**Computer Information (please complete the following information):**
 - OS: Arch Linux
 - Host: n/a
 - Version: n/a

**Additional Information:**
- gcc 10.2.0

baconpaul:
Hmm well we run all the tests on every build so def something about your setup
That test will fail if the wavetable oscillator can’t find wavetables. You sure that see is working?
I our next release I have en environment var to override the data path which will make things much easier.

dvzrv:
Thanks for the hint!
[...]

PR #4394, opened 2021-04-25
Force a WT selection in the DSP test

baconpaul:
May address #4387. Makes the test stronger in any case.


baconpaul:
Pls don't merge without talking with me first. thx.

baconpaul:
This solves the problem and is mergable but I’m just leaving it open so I remember to merge it into main also before I click merge. Just FYI.

Issue #4335, opened 2021-04-21
Surge XT Locations


baconpaul:
Surge XT gets config.xml, paramdocumentation.xml and windows.wt from the binary. But the factory patches still need a home. So

- [x] Move from Surge to Surge XT for factory
- [x] Test windows portable mode to find SurgeDataXT
- [x] Change user settings file to SurgeXTUserSettings
- [x] Build cmake rules to do a local install

mkruselj:
Can we also do FX presets as a separate folder structure in the binary rather than crammed into config.xml? Would be much easier to expand the factory FX presets then.

baconpaul:
yeah sure. but lets open a separate issue for that.

PR #4414, opened 2021-04-25
UserDefault goes to Enum; XT gets new File

### LINKED ISSUE #4085: UserPref Keys should be variables
the user prefs have "fooBarHootie" as keys but that should really be UserPrefs::FooBarHootie consistently.

baconpaul:
1. UserDefault API is no longer an unrestrained string but is
   instead an enum. Closes #4085
2. SurgeXT puts its user prefs in a SurgeXT file. Addresses #4335


baconpaul:
@mkruselj as discussed on discord if you want to do a review of this quickly would be welcomed.

PR #4419, opened 2021-04-26
remove old dependencies from CMakeLists.txt, need no more with JUCE

KottV:
I think it's safe to delete these lines in xt-aplha branch. At least it frees from installing all Cairo and XCB stuff in pipilenes.


baconpaul:
Thanks!!

PR #4421, opened 2021-04-26
remove Cairo and other dependencies from CMakeLists.txt (pkgconfig) 

KottV:
oops, I didn't delete most (in)valuable lines, now it's really won't ask for deps.
and while I'm not sure for what that line:
```
   if( NOT LINUX_ON_ARM )
     # pkg_check_modules(CURL REQUIRED libcurl)
   endif()
```
I keep it


baconpaul:
Oh that line is for the deferred asset loader which we never turned on and now we are in juce we can do differently if we decide to do it. I will add an issue to remove the program unit. Thanks!

PR #4431, opened 2021-04-26
Remove the DeferredAssetLoader

### LINKED ISSUE #4422: Remove deferred asset loader
We would do it differently in juce if we did it anyway

baconpaul:
We never used it, and we would do it differently now we have
the Juce libraries available anyway

Closes #4422


baconpaul:
want this one to run to 100% ci before we merge

PR #4444, opened 2021-04-28
Add (off-by-default) AUv3 extensions

baconpaul:
This lets us build an AUv3 but AUv3 is wierd so leave it
off for later exploration

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,7 +5,7 @@
 cmake_minimum_required(VERSION 3.15)
 cmake_policy(SET CMP0091 NEW)
 set(CMAKE_MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>")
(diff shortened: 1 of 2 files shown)


jdiercks:
Has anyone tried to make Surge work as an AUv3? I am not a developer, so I don’t really know what obstacles might exist, but there is a pretty significant population of musicians on the iOS platform who would LOVE to see this!! :-)

baconpaul:
Unfortunately the App Store distribution isn’t compatible with gpl3 software so the only path for surge to iOS is self build and not redistribute.

jdiercks:
Ah, I see. Guess I should have seen that coming, as it's the same reason the Surge VCV modules cannot be ported to miRack. Pity.

baconpaul:
miRack is also a 5 year old version of the rack codebase, but yes Apple will reject GPL3 apps unless apparently you lie about them, which we wouldn't do of course.

jdiercks:
All makes sense, thanks for the info. Still sad, but so it goes. Lots of other good apps to choose from on iOS, and I'm still glad to know that desktop users have a high-quality open-source option like Surge available.

PR #4450, opened 2021-04-30
Folder-related tweaks and fixes

mkruselj:
Make "open folder" actions work properly
Move "Open current skin folder..." to right-click (dev mode) menu
Add "Install new skin..." option that will open %USERFOLDER%/Skins
Add .clang-tidy to .gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,7 +31,7 @@
 *.out
 *.app
 
-# Visual studio
+# Visual Studio
 obj/
 *.sln
 *.vcxproj
@@ -63,7 +63,7 @@ fxbuild/
 # IntelliJ IDEA
 .idea
 
-# linux
+# Linux
 Makefile
 surge-*.make
 premake-stamp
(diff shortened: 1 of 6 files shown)


baconpaul:
Please do not merge. File start as process is not the correct api

baconpaul:
So 

https://forum.juce.com/t/open-containing-folder-for-files-or-dirs/14243/3

I'm sure you read that as well as I did

The problem with startAsProcess is if you pass a file which is actually a process it will run it. That would be 'bad' and also allow people to, say, change surge config in a way that it launches arbitrary executables when that function is chosen.

It seems, if launchInBrowser doesn't work on windows (sigh) that

```
auto f = juce::File(path);
if( f.isDirectory() )
   f.startAsProcess();
else
   f.revealToUser();
```

is the safe alternative. We should probably add a small function to SGE then 

```
void SurgeGUIEditor::openFolder(const juce::File &f)
{
if( f.isDirectory() )
   f.startAsProcess();
else
   f.revealToUser();
}
```

then your uses become

```
this->openFolder(juce::File(this->synth.storage.userDataPath))
```

PR #4484, opened 2021-05-04
squelch sqlite3 error: function may return address of local variable

tank-trax:


baconpaul:
oh we should merge this to main not xt-alpha
i can move it for you later tonight though (or you can resubmit against main)
thanks!

Issue #4097, opened 2021-03-16
SurgeFX Init can ask for incorrect Note


mynameismuhl:
Linux/64bit/Reaper
NIGHTLY-2021-03-16-6da806f

Here is what i did:

- open Reaper in terminal
- new project
- load surge on new track
- remove surge from track
- insert SurgeEffectsBank (same track)
- crash (Segmentation fault (core dumped))

syslog says this:
[coredump.txt](https://github.com/surge-synthesizer/surge/files/6149527/coredump.txt)

mynameismuhl:
I forgot to mention:
This only happens if you load and remove Surge first.
Just inserting fx bank without that step works fine.

baconpaul:
Hmm thank you. That note tuning table is a static member of SurgeStorage so I wonder if the create / destroy / recreate i leaving it somehow out in the cold. Excellent report. Will look.

baconpaul:
Hi @mynameismuhl 

I just tried this at head 7e5975e1312 and had no problem running both plugin in either reaper or juce audio plugin host on linux.

I did fix a pretty big leak the other day on linux but that doesn't seem like the same thing.

Can you just confirm that you still see this at head and if so, I'll figure out a way to perhaps understand what you see that I don't?

baconpaul:
Oh and @tank-trax if you are online sometime and have a minute, can you see if you can run both latest surge and surge-fx in a single session without a crash too? Thanks!

mynameismuhl:
yes i still see this with nightly 0c44a80
but only following the exact steps in first post
lots of other combinations like running both together or loading fxplugin first 
don't crash here now.

baconpaul:
oh yeah. that does it here!

baconpaul:
oh and now i see it. stay tuned.

baconpaul:
cool what a truly excellent bug report. thank you.

baconpaul:
So here's what is happening

SampleRate is a global. I mean I don't have to say much more than that.
But destroying the surge leaves that global in a weird state
Reloading an SFX reads it
At play time this is fine - we do the right thing
but there's an init path which uses sample rate now
and that sees a 0 or a wierd number
and makes an array index

So the solution is to check the samplerate and if it is 0 set it to 44100 at construction time so we at least have a non-crashing value.

baconpaul:
f31146d contains a fix to this problem. Thanks again!

tank-trax:
[...]

PR #4499, opened 2021-05-06
Samplerate Startup Path adjustment

### LINKED ISSUE #4132: JUCE has late samplerate startup - address
See #4097 for the problem that caused

Lets make it a member of storage

baconpaul:
When Storage is constructed, set a samplerate. This
generally doesn't matter since prepareToPlay will set it
to the rigth thing, but this means if juce calls a function
before a process block with samplerate, we can end up with
at least a mostly valid set o finternal tables.

Closes #4132


mkruselj:
@baconpaul Wouldn't this one also close #4097?

baconpaul:
yes

PR #4510, opened 2021-05-07
Consistent naming of files in src/gui

mkruselj:
@baconpaul git log --follow does follow here, but I'm not sure why my file renames aren't recognized as moves but as new files entirely. My previous commit with filter filename changes was perfectly alright, even though some of the renamed filter files also contained line changes, but over here it's kinda all over the place? Feel free to close PR if this destroys things too much regarding git blame or something. Would like to know why git is being fussy here...


baconpaul:
Yeah I don’t want to destroy history for a case change. Don’t know why git gets it wrong but lets close this.

PR #4525, opened 2021-05-08
Creation of RPM package builder.

gnac:
Added a script (make_rpm.sh) to support the ability to build an RPM on rpmbuild enabled systems.
Added a custom build rule (surge-xt-rpm) to CMakeLists to invoke the make_rpm script.
Added a build step to make_deb.sh to create a tarball of the same folder hierarchy used to build the deb package.
Modified call to find shared libraries to use -exec instead of piping to xargs as xargs was barfing on spaces. Alternatively, we could have passed -0 to xargs.
Added some additional build instructions to Readme.

### DIFF
--- a/scripts/installer_linux/make_deb.sh
+++ b/scripts/installer_linux/make_deb.sh
@@ -73,7 +73,7 @@ gzip -9 -n ${PACKAGE_NAME}/usr/share/${SURGE_NAME}/doc/changelog.Debian
 cp ${SOURCEDIR}/LICENSE ${PACKAGE_NAME}/usr/share/${SURGE_NAME}/doc/copyright
 cp -r ${SOURCEDIR}/resources/data/* ${PACKAGE_NAME}/usr/share/${SURGE_NAME}/
 
-# Copy the VST3 bundld 
+# Copy the VST3 bundle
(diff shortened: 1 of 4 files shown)


scripts/installer_linux/make_deb.sh
```diff
@@ -83,15 +83,21 @@ cp -r "${INDIR}/Surge XT Effects" ${PACKAGE_NAME}/usr/bin/
 # cp -r ../build/surge_products/Surge.lv2 ${PACKAGE_NAME}/usr/lib/lv2/
 
 # set permissions on shared libraries
-find ${PACKAGE_NAME}/usr/lib/vst3/ -type f -iname "*.so" | xargs chmod 0644
```
gnac, line 86: This was barfing on my system due to spaces in the vst3 path, eg "Surge XT". xargs can take the -0 parameter to ignore spaces as delimeters, but I believe using -exec chmod should be more universal.

baconpaul:
All looks great.
What do I need to apt-get on ubuntu to get rpm build tools in the pipeline?
if CI all runs (which it should) I'll merge this and mod the release pipeline once it is through one pass without the build. sound good?

gnac:
> What do I need to apt-get on ubuntu to get rpm build tools in the pipeline?
"apt-get install rpm"
 This includes the necessary rpmbuild command.

PR #4553, opened 2021-05-14
RPM Build Name formatting

gnac:
rpmbuild does not allow dashed in the version number. So we replace dashes with dots. However that causes the output to be in a different format then all of the other builds.
This change preserves the original version format and uses that in the name when moving the rpmbuild output file to the dist folder.
eg instead of surge-xt-x86_64-9.NIGHTLY.2021.05.08.90e93f2.rpm, we now get surge-xt-x86_64-NIGHTLY-2021-05-08-90e93f2.rpm.

Addresses 4524


baconpaul:
Thank you!

PR #4637, opened 2021-06-03
Set default patch from main menu, faster infowindow fades

mkruselj:


src/gui/widgets/PatchSelector.cpp
```diff
@@ -423,5 +424,9 @@ void PatchSelector::loadPatch(int id)
     }
 }
 
+const int PatchSelector::getCurrentPatchId() { return current_patch; }
+
+const int PatchSelector::getCurrentCategoryId() { return current_category; }
```
baconpaul, line 429: This const is in the wrong place

It should be

```cpp
int PS::getPatchID() const { }
```

src/gui/widgets/PatchSelector.h
```diff
@@ -82,6 +82,9 @@ struct PatchSelector : public juce::Component, public WidgetBaseMixin<PatchSelec
     void loadPatch(int id);
     int sel_id = 0, enqueue_sel_id = 0;
 
+    const int getCurrentPatchId();
```
baconpaul, line 85: and here too

src/gui/widgets/ParameterInfowindow.h
```diff
@@ -65,7 +65,7 @@ struct ParameterInfowindow : public juce::Component, public Surge::GUI::SkinCons
     int getMaxLabelLen() { return 30; }
 
     int countdownHide{-1}, countdownFade{-1}, countdownFadeIn{-1};
-    static constexpr int fadeOutOver{10}, fadeInOver{4};
+    static constexpr int fadeOutOver{3}, fadeInOver{3};
```
baconpaul, line 68: if you are going to make this change, please just remove the fade mechanism instead

PR #4659, opened 2021-06-09
ModulationSourceButton to juce

baconpaul:
The modulation source button ported to juce using zero vstgui
and with better life cycle. All features work except drag and drop
which displays visually but doesn't respond.


baconpaul:
Oh shoot don’t merge. I didn’t finish up the alternates

PR #4688, opened 2021-06-27
Add implementation for the "Threeler" filter

jatinchowdhury18:
Here's the "Threeler" filter! I think there's still a couple more optimisations to be found, and some testing to be done, but I think this implementation is at the point that it's ready for PR and review :).

Also, I remember there was some talk about naming the filter something else, but I don't remember what that name was?

Thanks,
Jatin


baconpaul:
Looks great. Let me merge it now. Reviewed the apparatus carefully and the DSP passingly at best but lets get it into the code and play with it. 

Don’t remember the name. @mkruselj do you?

mkruselj:
Think it was Thriller or 3-wheeler.

PR #4690, opened 2021-06-28
More Escape Removal

baconpaul:
This commit removes the reliance on VSTGUI::IControlListener and
the like and sets up a new callback interface which is way smaller
and entirely in the surge namespace. Also clean up some thing slike
frame->juceComponent() which are no longer needed


baconpaul:
DO NOT MERGE

PR #4719, opened 2021-07-05
Can drag around LFO shapes now

mkruselj:


baconpaul:
what happens with this feature if you have lfo type param latched out of curiosity?

PR #4736, opened 2021-07-08
Make the user data path orderly and rename to Surge XT

### LINKED ISSUE #4735: User WT/Presets columns still show even if they don't exist
If we have any user content in user data folder, _except_ patches themselves, the patch browser still displays the User Patches column still shows, but is empty.

Same thing happens in wavetable list. In fact, if all I have in user content folder is a custom skin, I get this:

image

image

And this is my user content folder:

image

mkruselj:
This PR creates a clean slate for Surge XT user data path:
* new folder name for user data path - `Documents/Surge XT`
* wavetables get their own subfolder - `Documents/Surge XT/Wavetables`
* patches get their own subfolder - `Documents/Surge XT/Patches`
* MIDI settings got a whitespace now - `Documents/Surge XT/MIDI Settings`
* FX presets are not in FXSettings anymore but instead - `Documents/Surge XT/FX Presets`, keeping it consistent with how we call it in the GUI
* exported wavetables will be saved to `Documents/Surge XT/Wavetables/Exported`
* since exported WTs are now within the main user wavetables folder, removed the "Open exported wavetables folder..." option, and renamed "Save wavetable to file" to "Export wavetable to file", to drive the point home and close the circle

Closes #4735 as a consequence of adding subfolders for each type of content.

Also, removes some unnecessary conditions and local variables from PatchSelector.cpp.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -169,6 +169,7 @@ juce_add_binary_data(surge-shared-binary
         HEADER_NAME SurgeCoreBinary.h
         SOURCES
         resources/data/configuration.xml
+        resources/surge-xt/README_UserArea.txt
         resources/data/windows.wt
         resources/data/paramdocumentation.xml)
 set_target_properties(surge-shared-binary PROPERTIES
--- a/resources/surge-xt/README_UserArea.txt
+++ b/resources/surge-xt/README_UserArea.txt
@@ -1,17 +1,17 @@
-This is your Surge-XT User Area
+This is your Surge XT user area!
 
-With Surge-XT, we are using a more structured user area which
+With Surge XT, we are using a more structured user area which
 separates the storage of patches, wavetables, FX settings,
-MIDI settings, and Modulator Settings. Each of these types of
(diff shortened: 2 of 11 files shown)


mkruselj:
I will not be merging this one, definitely want a second pair of eyes on it, although I tested it locally and everything shakes pretty well.

baconpaul:
Ahh this only gets win not mac and lin. Will need a bit before I can finish and merge it.

mkruselj:
Hmmm how come it doesn't work on Mac/Lin? We don't have UserInteractions per OS here, right?

baconpaul:
Here's the things I outlined on Discord that I want to add to this before we merge (and am happy to do so)

1. Make a single function in the codebase to setup the directory if it isn't there or is incomplete
2. When you set it up, include a "README.txt" about the layout
3. Set an 'option to migrate' flag to false but at some point in the future use that to offer a one-shot on migrating the old surge directory automatically
4. Open an issue to write that migration assistant

PR #4750, opened 2021-07-14
Resolve GCC error

baconpaul:
This commit should *fail* CI since it activates the warning
in question


baconpaul:
Please make sure full ci passes before a merge here

PR #4754, opened 2021-07-15
Clean up SurgeParamConfig

### LINKED ISSUE #4615: Rebuild the SurgeParamConfig.h
This let us do param configuration with vstgui like constants without having vstgui in the way

when we move to justifying after that we can probably kill this file or consolidate it into the skin model

mostly a note-to-self for @baconpaul

baconpaul:
See the comment in the header. But make it generally more
safe and reasoabnel. But not vanished

Closes #4615


baconpaul:
Wow it makes no sense to me that this blows up unitntests. But don’t merge until I figure it out obvs

baconpaul:
Oh man it’s the accidental xml change. Will fix in am

Issue #1964, opened 2020-05-27
Waveshaper flex tasks


mkruselj:
Here are some thoughts on how we could improve the waveshaper, after 1.7 release, possibly!

Here's the work to do to close this
- [x] Which shapers do we want in Dist and Rot. @mkruselj to provide a list here based on latest merge
- [ ] UI work
    - [ ] Front Pane
        - [x] move the button over and make it a switch which launches the preview 
        - [ ] Preview to proper overlay 
        - [x] Mouse Hide when dragging
        - [x] Final Assets for classic
        - [x] Dark and Royal - will we need a hover asset or is text color enough?
    - [ ] Popover gets just cleaner and positioned better
    - [ ] Then close this issue!

K0rrid0r:
I would let the multi-switch stay and add a drop down meny after the DIGI somehow (smart design change required) because host automating between the waveshapes in the multi-switch is an amazing feature.

I'm going to do a design test right now. AN di will update this comment also. hold on!

mkruselj:
You could still host automate things I reckon - it'd just be a visual change. I am not a fan of the behavioral change that your suggestion is introducing.

K0rrid0r:
How would you go on about host automating things in a dialog list?, it is not possible any wear in surge as of now? Or I'm missing something.

baconpaul:
you can host automate the filter and that shows up in a menu (as well as a switch). if we turned off the ui element you could still host automate it.

K0rrid0r:
![use](https://user-images.githubusercontent.com/56724286/83062691-5664b800-a05f-11ea-8344-79096ccd7c81.jpg)

That waveform curve under the drive slider could maybe be turned into a inverse gain compensation button, at least be used for something for saving space.

SVG function change.

mkruselj:
That waveform curve has absolutely no visual clue that it would do gain compensation. It needs to be designed properly, possibly a more elaborate skin modification.

baconpaul:
yeah making it a menu with a clear dropdown indicator is a good idea.
it may also not be all that hard.

baconpaul:
(gain compensation is tricky as I shared with evil on slack)

mkruselj:
I mean the menu doesn't have to look like a menu, it could actually be an icon of the actual curve that is dropdownable/mousewheelable. Kinda like Osc FM/Filter config routing?

This would be far better than the multi-switch that is there now.

baconpaul:
[...]

PR #4776, opened 2021-07-27
AntiDeriv form for the Rectifiers

baconpaul:
The +ve -ve and fw rectifiers are now cast in antiderivative
form removing last nights analog diode form. More importantly
a general case AntiDeriv WS kernel is in the code base so I
can do the cheby's next.

Addresses #1964


baconpaul:
DO NOT MERGE

PR #4791, opened 2021-07-30
Add hysteresis mode options for Tape

jatinchowdhury18:


baconpaul:
This is great. I think you need to set the deform in init ctrltypes also though before you merge 

we should leave issue open since we need to stream the deform type in fx presets. I can code that up if you want after fixing above or if you want to it is src common fxpresetmanager

jatinchowdhury18:
Figured I'd take a stab at it myself :). I tried to make it so that if the xml for an effect preset does not have a deform_type field, the default value will be used. It seems to work in my testing, but a second pair of eyes would probably be nice too.

mkruselj:
Looks good to me, but let's wait for @baconpaul before merging!

baconpaul:
Ha thanks that's great and looks perfect yeah!

Merging.

Issue #4657, opened 2021-06-09
Allow Dual mode MIDI configuration as option (defaulting to 1/2/3 to not break old sessions)


khr128:
macOS Big Sur 11.4. Surge 1.9. When I choose Dual mode, I expect both scenes to make sound, only the Scene A is audible, when I set gain for scene A to 0 and gain for scene B to the max, there is no sound.

baconpaul:
Are you sending midi on channel 1?

In dual mode there is midi routing between the channels.

I just set up S1.9 in LPX 10.6 on macOS 11.4 and sent the following events to a dual mode surge. I got both / A / B over the section.

<img width="574" alt="Screen Shot 2021-06-09 at 9 47 57 AM" src="https://user-images.githubusercontent.com/13387561/121366852-ce6aec80-c907-11eb-9339-9c39df1651f5.png">

khr128:
I am sending midi on channel 2 (from Arturia BeatStep Pro). Is there a way to change midi routing in dual mode so that both scenes are audible on midi channels other than 1?

khr128:
OK, after reading the manual about midi routing in Dual mode, I understand that it is setup so that on channel 2 only Scene A is playing and on channel 3 only Scene B is playing. I did not know that and I was sending midi on ch2, got only Scene A by design (whether it is a good design is questionable IMHO).

baconpaul:
I am not sure if it is an obvious design but, it is an ancient one. we can't change it without breaking projects.

We could however make it configurable per session defaulting to 1/2/3. If you think that's of interest reopen the issue and we can turn it into a feature request although wouldn't be top of stack.

khr128:
Thank you for your assistance, Paul. It really helped me understand the issue. I do think the design is a bit quirky. It really should be a user configurable option. One can really see that it can be useful occasionally, but as a default setting it creates confusion. 

Yes, please make it a feature request, but I agree that it’s not urgent.

mkruselj:
Let me see if I can hack this one in for 1.1.n, actually.

baconpaul:
ok but be careful. that channel mask and mpe interact in uncertain ways. and it's not all in one spot. the easy fix may be incomplete so test quite a few scenarios just fyi.

baconpaul:
oh you said 1.1.n. Oh yeah plenty of time for that :)

baconpaul:
sorry i thought you said 1.1.0 :)

PR #4805, opened 2021-08-06
WIP: Basic support for octave-per-channel mode

### LINKED ISSUE #4151: Let different channels transpose octaves for microtuning support
Surge has great tuning support, but one problem is that the MIDI standard only has 128 notes per channel. This is 10 octave in 12, but as soon as you start using even a modestly large EDO like 22 or 31, it is much less (only 6 octaves of 22 and 4 of 31). As a result you quickly run out of notes.

I would prefer this solution be the same as Pianoteq's, which has a "multi-channel MIDI layout mode" that you can put on. If you do, then channel 2 transposed up an octave from channel 1, and channel 3 transposed up 2 octave, etc. Also, channel 16 is transposed an octave below. So it looks like this:

Channel 1: main tuning
Channel 2: +1 oct
Channel 3: +2 oct
Channel 4: +3 oct
Channel 5: +4 oct
Channel 6: +5 oct
Channel 7: +6 oct
Channel 8: +7 oct
Channel 9: +8 oct
Channel 10: -7 oct
Channel 11: -6 oct
Channel 12: -5 oct
Channel 13: -4 oct
Channel 14: -3 oct
Channel 15: -2 oct
Channel 16: -1 oct

I would suggest this kind of thing. That way, we just go to the next MIDI channel to get another octave of notes past the end of the 128 note range. This setup lets you hypothetically play in any equal temperament up to 128 notes per octave, which is pretty far beyond the largest equal temperaments that are frequently used (such as 72 or 96).

I don't know if there's already some way to do this in Surge (is there?), but this would be best for my setup. I have a few scripts written that already use the Pianoteq style of doing things so this would be great.

Of course Surge already has another mode that uses different MIDI channels for some kind of MIDI split mode, so this would need to be mutually exclusive from that.

cgibbard:
See "Use MIDI Channel for Octave Shift" on the Tuning menu.

Only works correctly so far when tuning is applied at MIDI input, and has no special treatment for scene masking yet.

Aims to resolve #4151

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -15,6 +15,7 @@ The Emu (J Riley Hill) <http://jrileyhill.com>
 Matthias von Faber <http://github.com/mvf>
 Jani Frilander <http://github.com/janifr>
 Amy Furniss <http://github.com/amyfurniss>
+Cale Gibbard <cgibbard@gmail.com>
 Brian Ginsburg <https://github.com/bgins>
 Nathan Kopp <nk.sg@nathankopp.com>
 Korridor <fabian.kempe@gmail.com>
--- a/src/common/SurgePatch.cpp
+++ b/src/common/SurgePatch.cpp
@@ -2095,6 +2095,12 @@ void SurgePatch::load_xml(const void *data, int datasize, bool is_preset)
                 }
             }
 
+            p = TINYXML_SAFE_TO_ELEMENT(de->FirstChild("mapChannelToOctave"));
+            if (p && p->QueryIntAttribute("v", &ival) == TIXML_SUCCESS)
+            {
+                dawExtraState.mapChannelToOctave = (ival != 0);
+            }
+
             p = TINYXML_SAFE_TO_ELEMENT(de->FirstChild("midictrl_map"));
             if (p)
             {
@@ -2559,6 +2565,11 @@ unsigned int SurgePatch::save_xml(void **data) // allocates mem, must be freed b
         TiXmlElement mpn("mappingName");
         mpn.SetAttribute("v", dawExtraState.mappingName.c_str());
         dawExtraXML.InsertEndChild(mpn);
+
+        // Revision 17 adds mapChannelToOctave
+        TiXmlElement mcto("mapChannelToOctave");
+        mcto.SetAttribute("v", (int)(storage->mapChannelToOctave));
+        dawExtraXML.InsertEndChild(mcto);
 
         /*
         ** Add the midi controls
--- a/src/common/SurgeStorage.cpp
+++ b/src/common/SurgeStorage.cpp
@@ -529,6 +529,7 @@ bailOnPortable:
     twelveToneStandardMapping =
         Tunings::Tuning(Tunings::evenTemperament12NoteScale(), Tunings::KeyboardMapping());
     isStandardTuning = true;
+    mapChannelToOctave = false;
     isToggledToCache = false;
     for (int q = 0; q < 3; ++q)
         togglePriorState[q] = false;
--- a/src/common/SurgeStorage.h
+++ b/src/common/SurgeStorage.h
@@ -739,6 +739,8 @@ struct DAWExtraStateStorage
     std::string mappingContents = "";
     std::string mappingName = "";
 
+    bool mapChannelToOctave = false;
+
     std::unordered_map<int, int> midictrl_map;      // param -> midictrl
     std::unordered_map<int, int> customcontrol_map; // custom controller number -> midicontrol
 
@@ -1066,6 +1068,8 @@ class alignas(16) SurgeStorage
     Tunings::Scale currentScale;
     Tunings::KeyboardMapping currentMapping;
     bool isStandardTuning = true, isStandardScale = true, isStandardMapping = true;
+
+    std::atomic<bool> mapChannelToOctave; // When other midi modes come along, clean this up.
 
     enum TuningApplicationMode
     {
--- a/src/common/SurgeSynthesizer.cpp
+++ b/src/common/SurgeSynthesizer.cpp
@@ -322,7 +322,8 @@ int SurgeSynthesizer::calculateChannelMask(int channel, int key)
             break;
         }
     }
-    else if (storage.getPatch().scenemode.val.i == sm_single)
+    else if (storage.getPatch().scenemode.val.i == sm_single &&
+             !storage.getPatch().dawExtraState.mapChannelToOctave)
     {
         if (storage.getPatch().scene_active.val.i == 1)
             channelmask = 2;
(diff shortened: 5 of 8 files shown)


cgibbard:
Nice, got some kind of numerical floating point issue with the tests only on Windows... I could try to restrict the range on that one a bit further I suppose, but it seems a bit suspicious.

mkruselj:
This FR also kinda requires #4657 to be implemented, otherwise MIDI channel usage gets all over the place.

baconpaul:
@cgibbard cool I will review this am
@mkruselj @cgibbard is Cale on discord snd has been chatting with me in dev and micro tuning

mkruselj:
Just noticed!

baconpaul:
First the code looks great. Clean, well written, totally mergeable if we fix up one or two things. Thank you and welcome to the team!!

Let me rerun that windows job otherwise we will have to actually use windows. I know terrible right? But I have a VM. Let me know if you do also. Although further down these comments I think I figure it out so keep reading :)

@mkruselj is correct that we also need to fix the channel masking stuff. In the original issue I made a comment about this. Basically you have to temporarily force this mode out of channel split mode and avoid the dual mode singles. That's this cryptic comment: https://github.com/surge-synthesizer/surge/issues/4151#issuecomment-804367962

Basically in 'dual' mode surge pushes channel 2 to scene A only and channel 3 to scene B only. the issue @mkruselj linked fixes this more in general but I proposed in original issue when we are in 'force octave per' mode we just don't do that. I really think it is just at https://github.com/surge-synthesizer/surge/blob/0fcdf447c2ec61cb41731a43e03a58d1ce0a2064/src/common/SurgeSynthesizer.cpp#L325 to add a `&& ! storage->yourNewVariable` but requires some testing. That is probably also why channel 2 failed in your test now I think of it - it is not playing a sine and it blew up.

Now finally, now that I look at it, I think we also need to think do we store this setting in the patch or on the DAW Extra State. In my original comments I suggested patch (which you did) but I think that's wrong. I think, like the SCL/KBM, it should be in the daw extra state. @mkruselj if you agree I'll write up some comments on how to do that.

mkruselj:
Yep, I agree it should be in DAW state, not the patch.

baconpaul:
And ok yeah that windows failure is real and I bet it is from The channel 2 going to the wrong scene handling. Interesting - I wonder what I have scene 2 set to in the sine patch - so definitely need to fix that.

commenrs on daw extra state in a minute

baconpaul:
OK @cgibbard yeah I agree with @mkruselj this is a "daw extra state" thing. The 'daw extra state' is basically the things which are streamed in your DAW session (so when you save a reaper or ardour project or whatever) which are not part of the patch. Determining which is which is a perennial problem (not just for surge - other audio devs also have a hard time with it). The reason this setting is a daw extra is that you want to separate the performance (daw) from the sound (patch) for this I think.

So here's the change you would make

1. in SurgeStorage.h there's a `DAWExtraStateStorage` - right after the tunings just add a bool
2. In SurgeSynthesizer.cpp there's a function `populateDAWExtraStateStorage`. There you want to copy the value of your state variable onto the storage object.
3. In SurgeSynthesizer.cpp there's a function `loadDAWExtraState` right below where you want to do the opposite of step 2 - load from the state onto the synth
4. Move your streaming code in SurgeStorage (which streams into the patch) into streaming into the dawExtraState region. At main that's around line 1948 but look at line 2033 to see how we handle MPE and so on - yours is the same. Similarly around line 2511 is the write side. 
5. clang-format, fix the other channel mask thing, and if you want rebase and force push, and then if CI works I'd be happy to merge!

Thanks so much

cgibbard:
Ah, that makes sense, and agrees with what would probably be convenient for me, since this sort of expresses how my keyboard/DAW is communicating with Surge more than it expresses something about a particular patch. I'd rather not have to toggle it every time I loaded a new patch anyway. I'll give it a shot at some point today.

cgibbard:
Followed along with the instructions and force pushed. At first I thought I was going to be wholesale moving the field from SurgeStorage to the DAWExtraStateStorage, but then steps 2 and 3 wouldn't have made sense, so hopefully I got the interpretation right. Feels a little strange to have the state in two places now, but seems to roughly work at least.

Also, we still have the kind of odd Windows-only numerical issue it seems. I wonder if there's something different about how the floating point code is compiled on Windows. If this has something to do with the second scene not producing audio, shouldn't it also fail the same way on Linux and Mac?

I wonder what would be involved in making this work with tuning after modulation mode. It feels like if we don't fix its behaviour, we should at least disable that mode in the GUI when this is turned on. Presently, this has some effect on tuning after modulation, but whatever it's doing doesn't seem correct.

Perhaps we ought to visually represent the Chan Split mode being disabled as well, when this is turned on, but I'm not sure if the future plans for MIDI modes will render it a bit moot.

baconpaul:
Oh I wonder if it is % operator on - number returning neg on win.

Try ( chanoff+16 ) % 16 maybe?

we also don’t want that cmake change I don’t think.

I have some time later today to peek on ein if neither work. All look great

baconpaul:
Yeah that will e it for sure. From stack overflow quoting the spec:

> The binary / operator yields the quotient, and the binary % operator yields the remainder from the division of the first expression by the second. If the second operand of / or % is zero the behavior is undefined; otherwise (a/b)*b + a%b is equal to a. If both operands are nonnegative then the remainder is nonnegative; if not, the sign of the remainder is implementation-defined.

so

```cpp
 for (int chanOff = -2; chanOff < 3; chanOff++)
         { // Only checking reasonable octaves because frequencyForNote actually examines the
           // waveform
             int useChan = (chanOff+16)%16;
             INFO("chanOff is " << chanOff << " yielding " << useChan );
             f1 = frequencyForNote(surge, 60, 2, 0, useChan % 16);
             f2 = frequencyForNote(surge, 60, 2, 0, (useChan + 1) % 16);
```

and it explains the differences since MSVC will give you a channel of -2 for -2 value as opposed to 14. (Or what not) but gcc and clang won't.

battaglia01:
Hi everyone - this looks good, but noting that modulo 16 above, just to check, what is the behavior for channel 16? It's supposed to be that ch. 2 is an octave above ch. 1, and ch. 3 is 2 octaves, but also that ch. 16 is an octave below, ch. 15 is two octaves below, ch. 14 is three octaves below, etc, with it wrapping around at channel 8.

baconpaul:
That modulo is just in a test so shouldn’t impact usage. Do you see otherwise?

cgibbard:
Also, it does work as you describe. I have a Lumatone and have layouts set up for PianoTeq as well ;)

battaglia01:
@cgibbard great stuff! What's the best way to play around with this right now? Just compile the main branch?

mkruselj:
Yes!

cgibbard:
Yeah, and then under the Tuning menu, make sure it's set to "Apply Tuning at MIDI Input" and "Use MIDI Channel for Octave Shift"

ztnl:
Have any Lumatone owners been able to confirm if the portamento is working continuously across all channels as originally requested in #4151 ?

If so, an audio demo would be greatly appreciated.

PR #4806, opened 2021-08-06
Windows/MSVC: Specify UTF-8 encoding (fixes build on CJK systems)

mvf:
Surge and 3rd-party sources are de facto UTF-8. However, MSVC defaults to the build machine's codepage, which is never UTF-8, unless the user opted into the Beta. Nobody does, because it breaks some apps.

On CJK systems, the encoding mismatch caused [warning C4819](https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-1-c4819) in various places, breaking the build:

:cn: JUCE, Surge
:jp: JUCE
:kr: JUCE, MTS-ESP, Surge, stmlib

This change sets source and execution charset to UTF-8, fixing the build and avoiding possible runtime surprises due to encoding of literals. It also enables charset validation, i.e. MSVC will now warn on invalid UTF-8. This happens with VST2, which has ISO-8859-1 copyright © symbols in its headers. So we also silence C4828, but only when VST2 is present.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -119,6 +119,8 @@ if(MSVC)
 
     # Build with Multiple Processes (Clang-cl builds use Ninja instead)
     $<$<CXX_COMPILER_ID:MSVC>:/MP>
+    # Set source and executable charsets to UTF-8
+    $<$<CXX_COMPILER_ID:MSVC>:/utf-8>
   )
 endif()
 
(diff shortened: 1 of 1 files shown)


mkruselj:
The man is alive! 🙂

baconpaul:
yay welcome back and thanks!

and yes utf8everywhere :)

mvf:
Thanks guys, hope there's some stuff I can help with again, now that the JUCE dust has started to settle a bit. 😀

baconpaul:
Definitely! Plenty of work to be done in several spots - let us know what’s interesting in or out of the milestone!

Right now I’m finishing off accessibility stuff and will probably spend time on patch browser too. Formula modulator, startup assets in dll, smaller issues all out there and welcome any and all contribution of course :)

Issue #4385, opened 2021-04-24
JUCE menus and labels deal incorrectly with Emoji


baconpaul:
I have a "The magic of 🥓" patch category which doesn't render properly
@selenologist has a version of config.xml with japanese characters that also doesn't
If you load ucomputer 

I've confirmed we get proper UTF-8 all the way down to PopupMenu.addEntry

It's somewhere in juce labels. If I load " Sine 🌊.fxp" from "The magic of 🥓" the UI I get is this:

<img width="591" alt="Screen Shot 2021-04-23 at 9 55 23 PM" src="https://user-images.githubusercontent.com/13387561/115943560-9f470b80-a47e-11eb-9f50-60b29676e41e.png">

and in 1.9 I get this

<img width="516" alt="Screen Shot 2021-04-23 at 9 55 56 PM" src="https://user-images.githubusercontent.com/13387561/115943575-b1c14500-a47e-11eb-9ead-90431d406c89.png">

baconpaul:
A better fix is to use TextLayout in the LnF of Combo and PopupMenu which will find fallback fonts when glyphes are not available in the current one.

from https://forum.juce.com/t/embedding-unicode-string-literals-in-your-cpp-files/12600/2

baconpaul:
https://forum.juce.com/t/how-to-render-an-emoji-with-juce-graphics/45699

mkruselj:
@baconpaul Are we closing this one as JUCE has limits here?

baconpaul:
not yet

baconpaul:
Really need to wait for JUCE on this one, alas. Bouncing out of XT1

baconpaul:
This is in the JUCE 7 feature list. So let me close it here since there's nothing to be done other than upgrade juce in the future.

PR #4809, opened 2021-08-06
PatchSelector: Draw patch name as UTF-8

mvf:
`juce::Graphics::drawText()` takes a `juce::String`. Implicit conversion from `char*` assumes the data is ASCII, while the one from `std::string` assumes it's UTF-8. So use the latter to preserve the encoding.

Before:
Screen Shot 2021-08-06 at 22 32 53
After:
Screen Shot 2021-08-06 at 22 34 16

Related to #4385

### DIFF
--- a/src/gui/widgets/PatchSelector.cpp
+++ b/src/gui/widgets/PatchSelector.cpp
@@ -65,7 +65,7 @@ void PatchSelector::paint(juce::Graphics &g)
 
     // patch name
     g.setFont(Surge::GUI::getFontManager()->patchNameFont);
-    g.drawText(pname.c_str(), pbrowser, juce::Justification::centred);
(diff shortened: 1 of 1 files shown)


baconpaul:
Cool
There’s a million more drawtexts I’m sure we need to address too now I think of it!

mvf:
It's a bit of a hack though. Ideally we should keep our text in `juce::String`s in the JUCE-facing code to avoid copying. They also have some nice features such as `equalsIgnoreCase()`.

baconpaul:
Yeah I’ve avoided Having the juce data types that early in the code so far since I didn’t want to link juce to the Python api initially and stuff but I might have broken that accidentally when I made config.xml use the juce binary load mechanism 

but somehow changing storage to go from std::string to juce::string seems a bit off in a way I can’t quite explain yet

PR #4816, opened 2021-08-09
Drop JuceHeader.h, include JUCE modules directly

mvf:
JuceHeader.h is a generated header that includes all enabled JUCE
modules. This was bloating the preprocessed sources of many Surge TUs by
hundreds of KBytes each (on Linux), slowing down the build somewhat.

Also, JuceHeader.h is generated at build time, impairing IDE navigation
and completion until halfway through an initial build.

This change includes only the required JUCE modules directly, which
is supported in CMake projects as per the documentation.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -557,7 +557,7 @@ target_include_directories(surge-shared
         ${SURGE_COMMON_INCLUDES}
         ${OS_INCLUDE_DIRECTORIES}
         )
-target_compile_definitions(surge-shared PRIVATE ${OS_COMPILE_DEFINITIONS} )
+target_compile_definitions(surge-shared PUBLIC ${OS_COMPILE_DEFINITIONS} )
 
 if( BUILD_HEADLESS )
   add_executable(surge-headless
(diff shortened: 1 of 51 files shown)


mkruselj:
Lovely!

mvf:
Thanks, the build time savings are fairly tiny though. My Linux debug build went from ~2m45s to ~2m37s. Of course it's all in Surge code, which gets built incrementally all the time, and it might be more noticeable on other platforms. For me the main benefit is CLion not being dazed and confused in fresh projects because of the missing header. `JuceHeader.h` also masked the fact that we were not applying our `OS_COMPILE_DEFINITIONS` outside of `surge-shared`. 

Hm, and `clang-format` 12 wants to make curly brace related changes all over the place. Seems some default has changed since version 10 that's running in the pipelines. Gotta look into that...

baconpaul:
Cool I will try and look tonight
I almost never use go to header in clion and instead use go to symbol which is why I wasn’t bothered much but a good change indeed

mvf:
> I almost never use go to header in clion and instead use go to symbol

Same here, and that "go to" function − <kbd>Ctrl+Click</kbd> or <kbd>gd</kbd> − wasn't working for JUCE types until that header was built, e.g. things would look like this after loading a fresh project:

clion
It's not a huge deal, just a bit annoying for CMake plumbers who build clean all the time. :sweat_smile:

PR #4831, opened 2021-08-11
Fix cross and MinGW builds, LuaJIT everywhere

mvf:
Makes LuaJIT cross-capable and enables it unconditionally. Restores at
least these build targets (build-tested x86_64 hosts in parentheses):

- Linux/armv7l (Linux)
- MinGW-w64/x86_64 (Windows native, Linux, macOS)
- MinGW-w64/i686 (Windows native, Linux) - N/A on macOS, needs 32-bit

Windows Surge XT Standalone cross-compiled from macOS starts up and seems to be functional:

image

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -90,7 +90,6 @@ if(CMAKE_CXX_COMPILER_ID MATCHES "Clang|GNU")
         -Wno-free-nonheap-object # https://github.com/surge-synthesizer/surge/issues/4251
         -Wno-return-local-addr # squelch sqlite3 error: function may return address of local variable
     )
-    add_link_options("-Wl,-z,noexecstack" )
   endif()
 endif()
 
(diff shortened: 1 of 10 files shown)


baconpaul:
sure you see the windows fails. just fyi - I have a regtest in the unit test framework which makes sure lua loads and can do 1+1 and more, so if you get a clean CI you have a high degree of confidence you have it right I think. Especially if you leave mac more or less alone (mac is super hairy because of the cross compile onto arm).

mvf:
That's great to know. Think I know why the Windows builds fail. Didn't test multi-config generators, and they use different output paths. :hurtrealbad:

baconpaul:
Aww man I just noticed this diff

```
- message(STATUS "Escape *was* possible!" )
```

which is fine but i was wondering how long i would leave that there. Ha

mvf:
TBT, I actually pondered a bit if I should remove this memento of the battle against the VST SDK. :sweat_smile: But then decided to yank it since it could be a tad confusing for folks who aren't aware of the history.

PR #4834, opened 2021-08-12
SurgeSynthesizer: Remove unused wide getters, enable C4477 on MSVC

mvf:


baconpaul:
Oh thanks!
I wonder if there are other vst3 only points we can remove

PR #4876, opened 2021-08-20
Fix alignment of some arrays that are loaded by SSE

selenologist:
Some arrays that were passed to `_mm_load_ps` could potentially be unaligned.

I found four - but because spotting these is somewhat difficult (as the definition of each array passed to a potentially-unaligned SSE function must be looked up) there may be more in there. I also only checked `_mm_load_ps` calls.


baconpaul:
Once it is aligned you can change the subsequent loadu to a load and save a cycle or two

selenologist:
Good point - missed those now-unnecessary `storeu`. Fixed.

Were there some unneccessary `loadu` you spotted?

baconpaul:
I always confuse store and load and your change is what I meant :)

PR #4893, opened 2021-08-23
Fix incorrect painting of global FX bypasses

### LINKED ISSUE #4892: Incorrect FX bypasses after slot expansion
Pics will be self-explanatory:

image

image

Seems to be UI only, DSP-wise it's fine.

mkruselj:
Closes #4892

Also fixes color of send 3/4 sliders in Dark skin.

### DIFF
--- a/src/gui/widgets/EffectChooser.h
+++ b/src/gui/widgets/EffectChooser.h
@@ -107,8 +107,8 @@ struct EffectChooser : public juce::Component, public WidgetBaseMixin<EffectChoo
                 return true;
             break;
         case fxb_scene_fx_only:
-            if ((fxslot >= fxslot_send1 && fxslot == fxslot_global2) ||
-                (fxslot >= fxslot_send3 && fxslot == fxslot_global4))
+            if ((fxslot >= fxslot_send1 && fxslot <= fxslot_global2) ||
+                (fxslot >= fxslot_send3 && fxslot <= fxslot_global4))
                 return true;
             break;
         }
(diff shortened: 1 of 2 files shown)


src/gui/widgets/EffectChooser.h
```diff
@@ -102,11 +102,13 @@ struct EffectChooser : public juce::Component, public WidgetBaseMixin<EffectChoo
...
             return true;
         case fxb_no_sends:
-            if ((fxslot == 4) || (fxslot == 5))
+            if ((fxslot == fxslot_send1) || (fxslot == fxslot_send2) || (fxslot == fxslot_send3) ||
+                (fxslot == fxslot_send4))
                 return true;
             break;
         case fxb_scene_fx_only:
-            if (fxslot > 3)
+            if ((fxslot >= fxslot_send1 && fxslot == fxslot_global2) ||
```
baconpaul, line 110: Did you really mean al == in the second?
This will only be true for fxslot == global2 or global 4
I think you mean >= fxslot_send1 && <= fxslot_global2

baconpaul:
think it has a tiny bug?

src/gui/widgets/EffectChooser.h
baconpaul: yeah you definitely meant <=. Pushed a fix to your branch and will merge. Good catch!

mkruselj: Yes that was an unfortunate typo, I did mean to cover the range so <= was definitely needed there 🙂

PR #4905, opened 2021-08-24
Build LuaJit if build host is ARM

baconpaul:
The build host assumed x86_64 but not assumes `uname -m`


baconpaul:
Please let ALL of CI run before merging

PR #4935, opened 2021-08-27
PatchDB: Fix corrupt paths in patch database

mvf:


baconpaul:
Looks great. Can I use u8 string rather that path to everywhere now? If so I will change.

and thanks!,

PR #4951, opened 2021-08-29
Restore Graphical MSEG State

### LINKED ISSUE #4835: MSEG State restored properly not displayed properly
Open MSEG Editor
change movement to shift
you get shift
close and open
you get shift
but you see single
because the button doesn't set to the state
i bet

cannero:
First the canvas and the controls are created, the reference to each other is set later.
canvas = make; controls = make; canvas->controls = controls; controls->canvas = canvas
So in the first rebuild in controls the canvas member is a nullptr and the time edit mode value is taken from the state.

The loop mode 'On' was not set as integer division was used.

Closes #4835

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -31,6 +31,7 @@ Esa Juhani Ruoho <esaruoho@gmail.com>
 Mikko Saarinki <https://github.com/kingston1>
 Jarkko Sakkinen <jarkko.sakkinen@iki.fi>
 Jeremy Sagan <info@sagantech.biz>
+Sebastian Schwarzhuber <https://github.com/cannero>
 Jeff Smith <whydoubt@gmail.com>
 Jon Spruyt <http://jonspru.yt>
 Markku Tavasti <tavasti@iki.fi>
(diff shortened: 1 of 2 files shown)


mkruselj:
@cannero Awesome! Please add yourself to AUTHORS file, as well! 👍

cannero:
One thing I haven't asked is why the canvas is passed as nullptr in the MSEGControlRegion constructor, this change was done in #4671.

baconpaul:
Thank you!
The reason 4671 broke it is because I ported to juce incorrectly :) (and the vstgui and juce semantics are a bit different)

cannero:
Thank you @mkruselj for the hint and thank you @baconpaul for clarification!

During my tests I have found that the right click menu on the buttons in the MSEG control section has no effect on the state. Is this important or can we first leave it like this?

mkruselj:
Oh, that's a separate bug - should probably be taken care of in a separate PR.

Your PR failed our code quality test - please make sure to run clang-format before pushing! 🙂

baconpaul:
@mkruselj the clang format failure here is in your prior push not tbis one.  You merged a badly formatted diff (and ci failed on it too!)

I will merge this then perhaps you can clean that up?

Issue #4390, opened 2021-04-24
The TWS Tuning UI is all in JUCE


baconpaul:
And is all based on the tuning library

https://youtu.be/DJLDF6qZUX0

baconpaul:
So the reason not to bring the TWS visualizer into surge, of course, is because if people migrate towards MTS-ESP it won’t work; and the current html visualizer we have is quite good. This *may* be not worth it but I’ll keep it open for a bit anyway.

mkruselj:
It shouldn't be a problem to gray out the tuning visualizer overlay if ODDSound is enabled, though?

baconpaul:
Oh no that would be easy of course.
My point is: if most of our micro tuning users are on ODDSound by the time we are there it may not be a feature that is that useful. “Adding a menu which is 14% better for 20% of our users” may just not make the cut.

mkruselj:
Oh right, that is a very valid point.

baconpaul:
Changed my mind. Doing this.

mkruselj:
We can close this since it's integrated now. We know there's design work left to be done!

baconpaul:
Oh I was going to hold it open to hang commits as I finish that work if that's OK?

PR #4981, opened 2021-09-03
Integrate the TWS tunign UI

baconpaul:
Integrate the TWS tuning UI and hook it up
Lots ot fix here still but lets commit ot this
since it woudl be awesome
and not much work from where we are


baconpaul:
Addresses #4390

PR #5024, opened 2021-09-08
Add skin colors for virtual keyboard and tempo panel

mkruselj:
Also add skin color for text editor caret across the board.
Make highlighted text match normal text across the board.
Allow only up to 3 characters and only numbers in tempo typein.
Reduce width of Tempo panel.
Select all text upon clicking Tempo typein and release focus on pressing Enter.


baconpaul:
giveAwayKbdFocus is a 6.1 api and LV2 is still at 6.0.8
protect it in an #if SURGE_JUCE_ACCESSIBLE and it will be OK

PR #5037, opened 2021-09-09
Properly install the LV2 targets

### LINKED ISSUE #5014: cmake --install of LV2 uses the wrong file-name
As discussed on the chat:

The LV2 have their spaces removed from the .so files, but the installer expects the default filenames and thus fails to find them.

Please adjust cmake so it knows what filename to expect when installing.

baconpaul:
Closes #5014


baconpaul:
Merging since CI doesn't exercise the install path anyway

PR #5051, opened 2021-09-11
Copy instead of move borrowed path (fixes double-free SIGSEGV)

selenologist:
Fixes an invalid borrowing of memory beyond the lifetime of `fs::directory_entry f`. Attached is a Valgrind log of Surge being loaded in Ardour where the double-free was observed that led to this fix.

Though I'd appreciate another set of eyeballs on this fix, the placement-new is a little ugly so there might be a better way to do this.

[valgrind.txt](https://github.com/surge-synthesizer/surge/files/7147565/valgrind.txt)


selenologist:
Making a note here that this was the wrong approach, a better fix will follow soon (likely from @baconpaul)

PR #5053, opened 2021-09-11
don't merge but leave open to remind us to fix Force use of GHC filesystem on Linux 

selenologist:
Resolves path lifetime issue like #5051, but without the ugly code change.

Does raise some interesting questions about the behaviour of GNU/Linux libc++ vs others though.


baconpaul:
Replaced with #5055

PR #5096, opened 2021-09-15
CMake: Refactor into subdirectories, projects

mvf:
Opening as a draft for now, comments welcome. Bit of a whopper, but doing this incrementally is not really feasible.

---

Our `CMakeLists.txt` is pushing 1.5k lines of spaghetti and has become unmaintainable. This change splits Surge into subprojects to try and:

- Increase readability (targets & properties instead of variables)
- Reduce and gain insight into dependencies
- Reduce redundancy, potential for regressions and merge conflicts
- Make building things optional via `SURGE_BUILD_*` cache variables

Behavior shouldn't have changed too noticeably. Some targets were merged or renamed to declutter the IDE a bit. The pluginval stuff still needs cleaning up, but I didn't want to make this even bigger.


baconpaul:
Alright! Let me look tonight or tomorrow and offer comments. Thank you for doing this. It was inevitable that we would. But few things I want to peruse before we merge

baconpaul (commented):
We actually don't use FindLibSndFile any more

src/cmake/lib.cmake
```diff
@@ -0,0 +1,89 @@
+# vi:set sw=2 et:
```
baconpaul, line 1: I think the plugin validation could move to here right?

src/common/CMakeLists.txt
```diff
@@ -0,0 +1,324 @@
...
+  version.cpp.in
+  version.h
+)
+
+if(APPLE)
+  target_compile_definitions(${PROJECT_NAME} PUBLIC
+    MAC=1
+    MAC_COCOA=1
+    COCOA=1
+    OBJC_OLD_DISPATCH_PROTOTYPES=1
```
baconpaul, line 253: I'm 99% sure we only need MAC=1 now and the rest are from vstgui days

src/headless/CMakeLists.txt
```diff
@@ -0,0 +1,47 @@
...
+  main.cpp
+)
+
+target_link_libraries(${PROJECT_NAME} PRIVATE
+  samplerate
+  surge::catch2
+  surge::surge-common
+)
+
+find_package(LibSndFile ${PACKAGE_OPTION})
```
baconpaul, line 37: We only use libsndfile in one place so we can remove this and the cmake and when I need it again i can replace it with the juce wav support no problem. Lets clean this dependency up while we are doing this (the one place is in src/headless/ and its a test routine I don't use any more really)

src/cmake/lib.cmake
mvf: Yap, that's the plan. Just takes testing on all three platforms so I'll do this in another change.

src/common/CMakeLists.txt
mvf: Makes sense. Always glad to get rid of cruft!

src/headless/CMakeLists.txt
mvf: Great, this should also get rid of the scary "this libsndfile is linked against 10.15" warning on macOS.

PR #5101, opened 2021-09-16
CMake and File Moves, continued

baconpaul:
1. Remove some cycles discovered when CMake Shuffle happened
2. Use consistend product directory names in the source tree
3. Get it building again with those changes in place


mvf:
Would be great if we could keep the formatting in the CMakeLists. Or if we decide to change it, put the whitespace changes into a separate commit.

mvf (changes requested):
As mentioned, unaffected CMakeLists should not be changed by this.

baconpaul:
Oh darnit that’s clion auto formatting these on save for me

baconpaul:
So short version is “grrrr no that’s not easy.” I have to revert them all and re apply diffs with formatting off - let me see if I have time doe that 

if not I’ll close this and do it some other day

PR #5115, opened 2021-09-18
Draw more text as UTF-8

mvf:
Fixes mojibake in PatchDB dialog.

Basically eb477226d264784eaf45ac304247a2b80cc99603, only this time I bothered grepping... :sweat_smile:


baconpaul:
Thanks!
I merged this but I'm also in the middle of completely rewriting this, so the code you fixed is getting removed this weekend. May be better to stay out of the patched viewer for a few days while i get it into non-garbage shape (my goal for this week)

baconpaul:
(I will make sure to use text.c_str or the juce UTF8 api for all strings though :) )

PR #5124, opened 2021-09-20
macOS: Switch to C++17 (like the other platforms)

mvf:
Library support is spotty on our current deployment target 10.11 "El
Capitan", but a consistent language level across platforms does make
life easier for devs, especially those who aren't on Mac.


baconpaul:
Last time I tried this the resulting synth didn’t run on 10.11 - so you have any way to test that?

also there was the arm destructor thing with 17 and target 11 - did a universal build work?

But if ci finishes we can merge and see what happens in beta period

baconpaul:
Oh ci andwwred my question yeah

so we really can’t merge tbis. I remember something like You either have to hand code custom operator news or target 10.14 which is too high

baconpaul:
https://github.com/surge-synthesizer/surge/issues/3589 Yeah here was the convo we had a year or so ago

mvf:
Ah fiddlesticks, I remember now :sweat_smile:, thanks for digging up that thread. Would still be great if we could go C++17 on macOS for XT.

If you want to give this a try, I see two ways to fix the build:

- Set deployment target to 11.0 for Apple Silicon only
- Implement the aligned allocators

I'm going with the latter. Per-architecture deployment targets work, but they're only available with the otherwise lackluster Xcode generator.

> Last time I tried this the resulting synth didn’t run on 10.11 - so you have any way to test that?

Good point, unfortunately not. Maybe there's someone on Discord who can.

baconpaul:
If we can set per-arch yeah that's great
I think I use the Xcode generator in the release pipeline also
let me review before we merge. thank you.

mvf:
Of course, I'm not just merging this.

Regarding testing: On Intel, the linker drops our aligned allocators because the conventional `operator new` already satisfies our maximum alignment of 16. So I increased the alignment on one of the classes to 32 and stepped through them to verify that they work.

baconpaul:
OK so a couple of comments

1. Clever solution
2. perhaps a comment in the macOS aligned file though as to why it exists though linking to that original issue?
3. Also that filename doesn't seem to match our other filename conventions and is sort of odd hanging out in src/common. Perhaps 'src/common/os-specific' and then sme file name in there or some such
4. In theory I think the -faligned-malloc could be on only for arm. Is there a generator expression we could use on `OSX_ARCHITECTURES` to have this on and compiled on arm and not x86? or does it not matter at all?

baconpaul:
Your comment means we don't need my fourth point. Just put it in the comment in the second point :)

mvf:
Good points, thank you! I'll look into them when I come back later tonight.

mvf:
The aligned `new`/`delete` doesn't belong into `surge-common` at all, so I put them a new `surge-platform` library. I was considering something like this anyway for the pending `fs::path` changes. It will help consolidate platform-specific code (such as `getDLLPath`) behind a clean interface and get rid of some `#if PLATFORM`.

Turns out C++17 could potentially improve performance on ARM Mac. Since `new` alignment on there is smaller than our `alignas(16)`, allocations were likely under-aligned sometimes, requiring additional memory accesses for SIMD operations.

baconpaul:
Oh yes an excellent solution. I really like the surge-platform design. I'll merge this now.

PR #5148, opened 2021-09-23
Respect the Show cursor while editing option for MultiSwitch

mkruselj:


src/surge-xt/gui/widgets/MultiSwitch.cpp
```diff
@@ -30,16 +30,17 @@ MultiSwitch::~MultiSwitch() = default;
 
 void MultiSwitch::paint(juce::Graphics &g)
 {
-    juce::Graphics::ScopedSaveState gs(g);
     auto y = -valueToOff(value) * heightOfOneImage;
     auto t = juce::AffineTransform().translated(0, y);
+    juce::Graphics::ScopedSaveState gs(g);
```
baconpaul, line 35: while i think it is OK here, moving state pushes out of order with transforms is not a safe operation in general. I would prefer the scope save is at the top of the function

PR #5162, opened 2021-09-26
Disable dragging with middle mouse button

### LINKED ISSUE #5108: Middle-click also includes left-click gestures in some situations
- If you double-middle click on a slider when mod routing is disabled for instance, it will have mod routing mode still armed instead of arming then de-arming it.
- Middle-clicking then holding and dragging on a slider all in one motion allows to start dragging the modulation slider. This isn't bad actually, I quite like that, but unfortunately it leads to unwanted behavior when doing it again from the modulated state (the slider will jump and stuff, you can try it out).
- Middle click+drag on button rows also works when it probably shouldn't ?

This has lead to some strange and unintended gestures and actions from testing on my side. Is this something we can fix?

cannero:
Middle mouse button click in widgets should only enable mod editing.
Disabled all other actions where dragging or double click was possible.
Closes #5108

### DIFF
--- a/src/surge-xt/gui/widgets/EffectChooser.cpp
+++ b/src/surge-xt/gui/widgets/EffectChooser.cpp
@@ -273,6 +273,11 @@ void EffectChooser::mouseUp(const juce::MouseEvent &event)
 
 void EffectChooser::mouseDrag(const juce::MouseEvent &event)
 {
+    if (supressMainFrameMouseEvent(event))
+    {
+        return;
+    }
+
     if (event.getDistanceFromDragStart() > 2)
     {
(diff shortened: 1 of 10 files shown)


cannero:
There are also middle clicks possible for example in the `TypeinParamEditor` or on the slider in the waveshaper analysis overlay, but there the modulation mode is not enabled. I didn't change there anything, if that's okay?

baconpaul:
This looks great! And yes fine to skip those other spots. I’ll merge this once ci is done. Thank you!!

baconpaul:
Thank you again! All merged

PR #5167, opened 2021-09-27
platform: Implement Paths::homePath()

mvf:


src/platform/unix/Paths.cpp
```diff
@@ -31,4 +32,12 @@ fs::path binaryPath()
     return fs::path{info.dli_fname};
 }
 
+fs::path homePath()
+{
+    const char *const path = std::getenv("HOME");
+    if (!path || !path[0])
+        throw std::runtime_error{"The environment variable HOME is unset or empty"};
```
baconpaul, line 39: Ooooh that's an awful big throw to not be caught isn't it?

baconpaul: Oh ignore this.
We were doing this before.
Wow what an awful big throw to be uncaught!

mvf: Trudat. But then again, `HOME` being off is truly `exception`al I'd say. Some DAWs might not even start in such an environment.

Issue #3596, opened 2021-01-08
M/S transform effect in FX or scene output section


Oortseam:
Hi!

I'm speaking of a channel transformer performing the next algorithm:  Lnew=M=0.5*(L+R);  Rnew=S=0.5*(L-R)

It could be very useful in many situations. For example, it allows to build complicated unison-based pads with clear and straight control over mono and stereo part (to prevent undesired phase effects at low end and to have nice mono compatibility).

Now I'm using external plugin for this and it does the work nice. But it would also be useful if I had a possibility to store this as a part of preset (as I have many ones that I could share with others but now they need to be additionally processed with external plugin to sound right)

As I think, it wouldn't be a hard task. Just one more effect in FX group or RMB-menu controlled option for Pan or Width slider in Scene output section.

mkruselj:
So, basicall< mid-side encoder. But then you also need a decoder to get the proper stereo signal back. So this definitely requires more FX slots as a prerequisite (#2446).

baconpaul:
We have a m/s encoder decoder which are SSE coded already in the code base, so while I agree with @mkruselj that this would require slot expansion to be practical, it is a trivial matter to code.

The M/S encode is what is used in the 'width' parameter in many of the effects.

From the flanger code for instance:

```
  float M alignas(16)[BLOCK_SIZE],
         S alignas(16)[BLOCK_SIZE];
   encodeMS(dataL, dataR, M, S, BLOCK_SIZE_QUAD);
   width.multiply_block(S, BLOCK_SIZE_QUAD);
   decodeMS(M, S, dataL, dataR, BLOCK_SIZE_QUAD);
```

So really the thing you are after is an FX with two params

1: 'encode/decode' toggle
2: 'mix' (which is output = origin + encoded) default to 100% range -100 to 100.

Seems to me.

mkruselj:
Or, maybe a better way would be to add M/S processing option to each effect, when right-clicking the Output/Mix slider (depending on the effect), where we would select standard processing, process sides only, process mid only?

Oortseam:
mkruselj   Yes! To have hands completely free you should have M/S encoder also with a possibility to place it where you like at the circuit. 
BUT even the ecoder only that can be placed at the existing FX slot may be a great help.  

As an example, here is a setup for nice FM stereo bass with no phase artifacts I use now with SURGE. It looks like this:

[OSC2 (FM2 mode)]->(fm modulation)->[OSC1 (SIN mode)]
[...]

PR #5173, opened 2021-09-30
Add MS Tool Effect

mynameismuhl:
This attempts to solve issue #3596 and expands the idea a little.
- It has switchable I/O (LR->MS->LR, LR->MS, MS->LR)
- HP, LP and a semi parametric band per M and S channel
  (each activatable)
- separate gain for M/S
- L/R balance

possible use cases:
- sandwich fx between en- and decoder
- decode sources defined as m/s (2 oscillators or scenes)
  (incl. the option to mono low frequencies that was requested in the issue)
- eq/filter mid and side channels of l/r-stereo content
- etc.

It is currently in  a state where I'd like to hear opinions
and ask for a review

Adresses #3596

### DIFF
--- a/doc/Adding An FX.md
+++ b/doc/Adding An FX.md
@@ -1,4 +1,4 @@
-# How to add a effect
+# How to add an effect
 
 This tutorial will demonstrate how to add an example effect called `MyEffect`.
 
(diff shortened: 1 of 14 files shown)


baconpaul:
Wow thank you! I have a bit of time in the morning us time so let me pull this down and do a review then.

mynameismuhl:
just added a "last-minute-change" to the peak-eq bandwidth
(i had it at 0.5 which was not very useful)

baconpaul:
Wow what an immaculate PR @mynameismuhl - you even updated the documentation. Thank you for clean and careful work.

I think folks may have some feedback on names and stuff like that but from a DSP and code structure perspective this is basically spot on. So let me merge this and get it into the nightlies and then we can see if we have a tweak or two on the presentation before we finish.

Also

Saw unison 2 extended range to 1200cents
MS Encode -> Delay set channel R -> MS decode

gives you a delay on the high voice and not the low voice

Which is really fun

PR #5182, opened 2021-09-30
A bunch of UI/UX fixes

mkruselj:
* Adjusted margins and size of Patch Store dialog
* Param typein now uses correct skin color for its text labels
* Miniedit now has same font as param typein, is wider, has better
  spacing and positioning of OK/Cancel buttons, and has
  focused outline color assigned to match the look of param typein
* Swapped all uses of isCtrlDown() with isCommandDown()
* Switch and MultiSwitch didn't pass key modifiers to controlModifierClicked()

### DIFF
--- a/src/surge-xt/gui/widgets/Switch.cpp
+++ b/src/surge-xt/gui/widgets/Switch.cpp
@@ -59,9 +59,10 @@ void Switch::mouseDown(const juce::MouseEvent &event)
         return;
     }
 
+    notifyControlModifierClicked(event.mods);
+
     if (event.mods.isPopupMenu())
     {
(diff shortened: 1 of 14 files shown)


src/surge-xt/gui/widgets/Switch.cpp
```diff
@@ -59,9 +59,10 @@ void Switch::mouseDown(const juce::MouseEvent &event)
         return;
     }
 
+    notifyControlModifierClicked(event.mods);
+
```
baconpaul, line 63: We need to guard this with the event type we want to send. We can't send it unprotected for all mouse downs. Same in MultiSwitch

PR #5187, opened 2021-10-01
All Notes Off support

baconpaul:
AllNotes/AllSounds off message was ignored in the JUCE port. Ooops!
Clearly surge should separate these messages, but Surge 1.9 didn't and
Surge XT1 won't either. But this commit removes the regression that
the message was ignored.


mkruselj:
@baconpaul I did make Surge distinguish between All Notes Off and All Sound Off! That's PR #4731

baconpaul:
how does that ever get called? Did you confirm that it is called by the surge path?
but great!
but also we should make those externally callable members on SurgeSynth :)
but let me close this

PR #5192, opened 2021-10-02
Extended PB range

cannero:
- changed pitch bend parameter formatting for input field
- scale by 100 for pb when extended range is activated in `NumberField` and shift is not pressed
- use setter function for extended property in fx processor

### DIFF
--- a/src/common/Parameter.cpp
+++ b/src/common/Parameter.cpp
@@ -3179,7 +3179,7 @@ void Parameter::get_display(char *txt, bool external, float ef) const
         case ct_pbdepth:
         {
             if (extend_range)
-                snprintf(txt, TXT_SIZE, "%i Cents", i);
+                snprintf(txt, TXT_SIZE, "%0.2f Keys", i * 1.f / 100);
             else
                 snprintf(txt, TXT_SIZE, "%i Keys", i);
         }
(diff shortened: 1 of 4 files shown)


baconpaul:
Thanks! I have a bit of a busy day today but will try and peek - if not very first thing tomorrow

cannero:
No problem, like I said in discord, this is more like a hack, maybe you see a better solution.

baconpaul:
So this looks super close to what i thought would work. - let me pull it down tomorrow and see if i can figure out what's going wrong.

baconpaul:
Ahh I see what you mean and there was one API which could get you there. Your change actually works and is correct but there's a cleaner way to deal with that weird float conversion which Surge uses.

I pushed a commit up to your brand (f66d737) which does the change. Happy to merge this but if you want to look this morning also happy to wait to see if you have questions or comments.

And thanks for diving in here!

cannero:
Nice, that's better 👍. I have seen `intScaledToFloat` when I searched why the float is handled in this way, but never made the leap that the function could be used there.
Just one question, do we need the first `bounceToInt` call when the parameter is changed, can the value be changed in the mean time?

baconpaul:
I think we probably don't need it, but wasn't 100% sure so put it in to make sure the iValue matched the value before we applied yeah. So let me go ahead and merge this. Thank you!

PR #5195, opened 2021-10-02
Add HP for S channel to Conditioner

mynameismuhl:
- A deactivatable Highpass Filter for the Side Channel in Conditioner
  (named "Side Low Cut" for consistency)
  to roll off low frequencies in the stereo width.
- set threshold to -6dB in init preset to avoid level jump

This was a test in preparation for the ms tool.
I thought I'd ask before I throw it away...
It is patch compatible with 1.9
(but not compatible with old user-fx-presets)

Adresses #3596


baconpaul:
This is cool and we should merge it
but also i should make it so the streaming version thing happens with user presets.
let me open an issue for that.

PR #5249, opened 2021-10-13
Some UI tweaks

mkruselj:
Move JUCE color assign from SGE to SJLAF
Add a missing repaint on mouse move in MSEG Draw mode
The outer border of standalone window now finally follows titlebar color
(but doesn't update on skin change sadly)

### DIFF
--- a/src/surge-xt/SurgeSynthEditor.cpp
+++ b/src/surge-xt/SurgeSynthEditor.cpp
@@ -18,14 +18,13 @@
 #include "RuntimeFont.h"
 #include <version.h>
 
-using namespace juce;
-
 //==============================================================================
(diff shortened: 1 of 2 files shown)


baconpaul:
Added a comment. The I've avoided `using namespace juce` in the code almost everywhere (and would like to have it everywhere except the one spot with a comment). Can you remove that and also shrink the diff to what actually changed as a result? Thanks!

baconpaul:
(In fact I'm going to PR to remove it from TuningOverlays, which I had ported from TWS, now!)

baconpaul:
OK #5251 removes the 2 using namespace juce in the code which were there for no real reason. (The one in the look and feel has a reason which is documented there).

PR #5270, opened 2021-10-18
Add Spring Reverb effect

jatinchowdhury18:
Adding a Spring Reverb effect! Still TODO:

- performance improvements
- parameter tuning
- make sure my changes to `chowdsp::DelayLine` didn't break the other effects that use it (Neuron, Tape)

I should be able to finish this up over the next few days, but any help/testing is also much appreciated :)

### DIFF
--- a/src/common/CMakeLists.txt
+++ b/src/common/CMakeLists.txt
@@ -335,6 +335,7 @@ PUBLIC
   PRIVATE
   surge-lua-src
   surge-platform
+  surge-juce
   taocpp::pegtl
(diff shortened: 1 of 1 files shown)


jatinchowdhury18:
Hmmm, on GCC, it looks like the CI running into an error in UnitConversons.h, that I don't fully understand...

baconpaul:
OK I think that makes sense yeah!
Would you like me to review this now or is it still something you want to tweak a bit?

PR #5361, opened 2021-11-10
Ask Apple permission to use their microphones for Standalone plugins

jatinchowdhury18:
So I was trying to experiment with using Surge XT Effects as a Standalone effect with my guitar as input. It turns out that on Mac there's some extra permissions you need to ask for so that the plugin can access the input signal from your microphone or sound card.


baconpaul:
Oh thanks!
Funnily I turned this on for awwbees in bespoke but didn't think to turn it on in surge. funny that :)

PR #5385, opened 2021-11-13
If MTS is used, rename "Apply tuning to filter cutoff" menu entry

mkruselj:


baconpaul:
So the line immediately below your change guarantees that this will never be used since visible will always be false since mts can’t do retune all just retune st midi

Issue #5408, opened 2021-11-15
Modal miniedits are below overlays in Z-order


mkruselj:
1. open MSEG or formula overlay
2. rename a modulator
Result: you go modal but the overlay is on top of miniedit

Modal dialogs should always be forced on top of everything else since they steal context.

baconpaul:
hmm I don't see this but I will add a toFront anyway!

PR #5411, opened 2021-11-15
Force MiniEdit to front always

baconpaul:
This is the fix to #5408, but I couldn't reproduce
for a retest with this in place.


mkruselj:
@baconpaul All good and fixed now!

baconpaul:
Awesome!

PR #5486, opened 2021-11-25
Various minor CMake tweaks

mkruselj:
ASIOSDK_DIR environment variable now works on its own, without requiring BUILD_USING_MY_ASIO_LICENSE to be set in cmake command.
A number of wording and formatting fixes in messages.
VST2 plugin will now also be moved to artifacts folder, like all others.


src/cmake/lib.cmake
```diff
@@ -16,15 +16,15 @@ function(surge_juce_package target product_name)
   foreach(format ${SURGE_JUCE_FORMATS})
     add_dependencies(${pkg_target} ${target}_${format})
   endforeach()
-  foreach(format AU Standalone VST3)
+  foreach(format AU Standalone VST VST3)
     if(NOT SURGE_COPY_TO_PRODUCTS)
```
baconpaul, line 20: Should we add lv2 here also?

mkruselj: Sure, why not!

PR #5490, opened 2021-11-25
Fix for building with clang-cl on Windows

rovingeye:
This fixes clang-cl errors about NOMINMAX and WIN32_WINNT already being defined by juce_BasicNativeHeaders.h when building.

```
D:\GitHub\build\surge\libs\JUCE\modules\juce_core/native/juce_BasicNativeHeaders.h(129,10): error: 'NOMINMAX' macro redefined [-Werror,-Wmacro-redefined]
 #define NOMINMAX
         ^
<command line>(26,9): note: previous definition is here
#define NOMINMAX 1
        ^
```

```
D:\GitHub\build\surge\libs\JUCE\modules\juce_core/native/juce_BasicNativeHeaders.h(137,11): error: '_WIN32_WINNT' macro redefined [-Werror,-Wmacro-redefined]
  #define _WIN32_WINNT 0x0602
          ^
<command line>(38,9): note: previous definition is here
#define _WIN32_WINNT 0x0601
        ^
```

### DIFF
--- a/src/common/CMakeLists.txt
+++ b/src/common/CMakeLists.txt
@@ -280,10 +280,8 @@ elseif(UNIX)
 elseif(WIN32)
   target_compile_definitions(${PROJECT_NAME} PUBLIC
     WINDOWS=1
-    NOMINMAX=1
     WIN32
     _USE_MATH_DEFINES
-    _WIN32_WINNT=0x0601
     _USRDLL
     VA_SUBTRACTIVE_EXPORTS
     USE_LIBPNG
--- a/src/common/SurgeSynthesizer.cpp
+++ b/src/common/SurgeSynthesizer.cpp
@@ -13,6 +13,8 @@
 ** open source in September 2018.
 */
 
+#define NOMINMAX
+
 #include "SurgeSynthesizer.h"
 #include "DSPUtils.h"
 #include <ctime>


baconpaul:
Thanks!

I just want to look at one tbing before we merge this if that’s ok!

rovingeye:
I am wondering if I should have defined _WIN32_WINNT=0x0601 like I did NOMINMAX

baconpaul:
> I am wondering if I should have defined _WIN32_WINNT=0x0601 like I did NOMINMAX

Yeah
I think we want to conditionally define these in cmake if we are msvc and not if we are clang

but what I really want to check is if I can remove that old windows cruft in surge synth that has us using windows.h there at all then not worry about this. I was just looking at that yesterday

baconpaul:
Oh how did you do a clang build on windows btw?

baconpaul:
Like if I just use srs::thread here https://github.com/surge-synthesizer/surge/blob/534f423932def024e6e7dce502758f8810921be7/src/common/SurgeSynthesizer.cpp#L3661 I don’t need to include windows.h at all

rovingeye:
> Oh how did you do a clang build on windows btw?

I have LLVM installed, but the included tools for visual studio seem to work too. I just set the cmake_c_compiler and cmake_cxx_compiler to clang or clang-cl now and run in a developer prompt and it links everything properly

rovingeye:
> > I am wondering if I should have defined _WIN32_WINNT=0x0601 like I did NOMINMAX
> 
> Yeah
> I think we want to conditionally define these in cmake if we are msvc and not if we are clang
> 
> but what I really want to check is if I can remove that old windows cruft in surge synth that has us using windows.h there at all then not worry about this. I was just looking at that yesterday 

That seems like a better fix for nominmax, didn't realize it was possible to remove that stuff.

Interested to see if setting conditionally stops clang-cl from error on that step (I don't understand why it errors when clang and msvc are both ok with it)

baconpaul:
OK so it's too big a change to replace the few uses of windows.h in the codebase consistently this late.
Let me try a slightly different approach here

baconpaul:
the command line is

```
cmake -Bignore\cl -GNinja -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_BUILD_TYPE=Debug
```

just for my future ref

baconpaul:
OK so the answer here is not to remove the defiens but make em consistent. I'm going to close this and open a new one in a few minutes. Thanks for pushing this!

PR #5493, opened 2021-11-25
Add a couple more tests of porta/tuning/channel-per-octave

cgibbard:


baconpaul:
Siince 6these fail for now you want to set the 'true' argument after the mode to 'false' and it will skip the assertion they work; then CI can work and we can merge and then we can fix them!

PR #5498, opened 2021-11-25
Porta-Gliss in MIDI and MTS mode hits scale notes

### LINKED ISSUE #5483: Glissando Retrigger Tuning Behavior
Related finding for both microtuning modes: With Glissando and Retrigger At Scale Degrees checked, the degrees appear to be related to 12-equal. Tested with ED2-31 and a glide across 31 MIDI Notes, plays 12 retriggered notes.

 In RETUNE_AT_MIDI/MTS mode portamento is in key space so retrievers when you run over a key. If you work in RETUNE_AFTER_MODULATION it should work. But we should open a separate issue, since I have this particular one closed. I'll do that now.

This one has been there since 1.9; it may be tricky to fix. The code in question is this

```cpp
 if (scene->portamento.porta_retrigger)
            if (floor(state.pkey + 0.5) != state.priorpkey)
            {
                state.priorpkey = floor(state.pkey + 0.5);
                state.porta_doretrigger = true;
            }
```

which in retune at midi and its mode has to be retuned-key-aware not key-aware

@ztnl

baconpaul:
Previously it hit the internal 12TET notes

Closes #5483

### DIFF
--- a/src/common/dsp/SurgeVoice.cpp
+++ b/src/common/dsp/SurgeVoice.cpp
@@ -344,11 +344,7 @@ void SurgeVoice::legato(int key, int velocity, char detune)
 
         state.porta_doretrigger = false;
         if (scene->portamento.porta_retrigger)
-            if (floor(state.pkey + 0.5) != state.priorpkey)
-            {
-                state.priorpkey = floor(state.pkey + 0.5);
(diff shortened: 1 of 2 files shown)


ztnl:
Thanks for this. Love it!

PR #5522, opened 2021-11-28
Add "Knock" parameter to spring reverb

jatinchowdhury18:
Added this parameter so that you can "shake" the spring reverb. Not sure that "shake" is the right name, since it's more like if you were to knock against a spring reverb unit, rather than actually pick it up and shake it.

i made the new parameter a boolean, since that seemed to make sense internally, but then I found that I wasn't able to modulate the parameter. Are boolean parameters not modulate-able,  or am I just not doing it right?

### DIFF
--- a/src/common/dsp/effects/chowdsp/SpringReverbEffect.cpp
+++ b/src/common/dsp/effects/chowdsp/SpringReverbEffect.cpp
@@ -40,7 +40,7 @@ void SpringReverbEffect::process(float *dataL, float *dataR)
             clamp01(*f[spring_reverb_spin]),
             clamp01(*f[spring_reverb_damping]),
             clamp01(*f[spring_reverb_chaos]),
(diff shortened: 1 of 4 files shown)


mkruselj:
7ndewd boolean params are not modulatable in Surge. We have an issue about it, it's a tricky one from multiple angles.

mkruselj:
Oh and if it's not behaving like shake, then of course it makes sense to rename the parameter to Knock, or whatever 🙂

baconpaul:
yeah this is also how we dealt with 'freeze' in nimbus. make a float and either look for value or look for transition across 0.5 type thing.

I'll wait to merge this one then yeah?

jatinchowdhury18:
Using `ct_float_toggle` like in Nimbus seems to work well! I think this one should be ready to merge, unless there's some issue I might have missed.

baconpaul:
Looks great to me!

PR #5523, opened 2021-11-28
Set up pluginval targets that work on Apple and Windows

jatinchowdhury18:
Adjusted the pluginval targets so that they work on my Mac and Windows machines. Haven't been able to test on Linux yet.

Any suggestions to clean up my CMake hackery are greatly appreciated!


baconpaul:
Looks good to me! I'll merge it with a set of merges later this morning. Thanks!

PR #5540, opened 2021-12-02
CMake Installs resources; Group name tweak

### LINKED ISSUE #5537: Restore install rules on unix for assets as well
I think this is just lib.cmake gets this line

```
    install(DIRECTORY "${CMAKE_SOURCE_DIR}/resources/data/" DESTINATION share/surge-xt)
```

but I don't have a linux bootable with me right now to test

baconpaul:
1. CMake installs resources. Closes #5537
2. CHange the gropu name "Global / FX" to "Global And FX" to not
   confuse "/" with a group delimeter.


mkruselj:
I would have used & instead 🙂

PR #5587, opened 2021-12-06
Fix background for keytrack SVG in Dark skin

rovingeye:
Fixed BG & resized to match retrigger SVG after goobits fix.


mkruselj:
Fixed this in my local Figma template, proper version of SVG coming soon. Thanks for the heads up, though!

PR #5594, opened 2021-12-07
Correct API for Overlays on Scene Change

### LINKED ISSUE #5590: Tear-Outs/Ins When Switching Scenes In Dual Mode
Version: Surge XT 0.99.nightly.f017900
Build: 2021-12-06 @ 20:51:57 on 'fv-az223-812/pipeline' with 'MSVC-19.29.30137.0' using JUCE 6.1.2
System: Windows 64-bit VST3 on        Intel(R) Core(TM) i7
Host: Reaper @ 48.0 kHz

Realizing there is some ongoing work and discussion about the behavior of tear-outs/ins in general, thought I'd ask a related question before creating an issue about an observed behavior...

Q: Is it normal, or otherwise known behavior, for MSEG tear-out states to revert to tear-in upon switching Scenes?

Notice in the attached GIF, how the MSEG tear-out reverts to a tear-in state when switching between Scenes in Dual mode, with a patch that uses MSEGs in both Scenes.

Tear-Out state when switching between Scenes

baconpaul:
We were using 'show' as opposed to 'refresh' in one spot
still, so the overlay would pop in (mostly).

Closes #5590

### DIFF
--- a/src/surge-xt/gui/SurgeGUIEditorValueCallbacks.cpp
+++ b/src/surge-xt/gui/SurgeGUIEditorValueCallbacks.cpp
@@ -286,7 +286,7 @@ void SurgeGUIEditor::changeSelectedScene(int value)
 
         if (ld->shape.val.i == lt_mseg)
         {
-            showOverlay(SurgeGUIEditor::MSEG_EDITOR);
+            refreshOverlayWithOpenClose(SurgeGUIEditor::MSEG_EDITOR);
         }
         else
         {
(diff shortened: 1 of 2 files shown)


mkruselj:
Hmmm if we go from scene A MSEG torn out to scene B Formula torn out, it doesn't work (on scene B it goes back in). Should it not work like that since MSEG and Formula basically share the same overlay?

baconpaul:
Yeah will reopen and fix

PR #5600, opened 2021-12-08
Avoid a startup transient due to denorm noise

### LINKED ISSUE #4900: Denormal Noise is visible out the back of Surge
This was a DeRez problem with loads of investigation

I was able to fix DeRez in 04e38b3 but the small value 1e-15 noise which is there to fight denodm problems still causes a problem

I think the @mynameismuhl fix along with our FTZ flags mean we are fine to set it to 1e-25 and I will do that once we ship XT1.0 but too late in cycle for that now.

-----

1. Fresh session
2. Load up "TNMG/Sequences/One Node IDM"  <- do you hear a glitch at 'load'
3. Toggle th enext patch button forward and back (so unload and reload One Note IDM). <- Do you hear a glitch at 'load'
4. Bypass all FX (so put FX bylass all the way to 'all') and then repeat step 3. Does the sound go away?

it seems to be the airwindows derez

baconpaul:
Slight denormalization noise in the warmup phase of some
aurwindows (esp DeRez) would cause a pop. Avoid that by having
either the first 500 blocks (~1/3 second) or time until first note
whichever is smaller clear the output with zeros, then add the
noise after

Closes #4900

### DIFF
--- a/src/common/SurgeSynthesizer.cpp
+++ b/src/common/SurgeSynthesizer.cpp
@@ -350,6 +350,11 @@ void SurgeSynthesizer::playNote(char channel, char key, char velocity, char detu
         }
     }
 
+    /*
+     * Force the denormalizer to clear with noise even if we are still waiting, since we have a
+     * voice
+     */
+    blocks_since_unhalt = blocks_until_first_denorm_noise + 1;
     midiNoteEvents++;
 
     if (!storage.isStandardTuning)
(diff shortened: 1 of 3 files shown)


baconpaul:
It would be wonderful if someone with an x86 could test this before i merge
basically: does one note idm pop on load any more?
that's the test

PR #5651, opened 2021-12-19
Fix a Pedal Polymode problem

### LINKED ISSUE #5649: Pedal and voice count
Placeholder so I remember 

pedal down
Play 60
Release 60 
Play 60 -> now have two voices
Release pedal
-> should have one voice

apparently we have 2

maybe fix for 1.0. Let’s see how gnarly it is

baconpaul:
In poly mode

pedal down
note 60 on
note 60 off
note 60 on
pedal release

would result in 2 voices playing not one. While this is not
wrong per-se, its not how other instruments (like pianos)
woudl work and it's not how most other synths work, so choose
the convention that we switch to 1 note playing for XT

Closes #5649

### DIFF
--- a/src/common/SurgeSynthesizer.cpp
+++ b/src/common/SurgeSynthesizer.cpp
@@ -402,10 +402,12 @@ void SurgeSynthesizer::playNote(char channel, char key, char velocity, char detu
         {
             for (auto &h : holdbuffer[sc])
             {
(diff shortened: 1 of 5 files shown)


baconpaul:
Please don't early merge; want to make 100% sure all the regtests pass

PR #5681, opened 2021-12-29
Resolve some LFO related magic numbers

mkruselj:
Up for @baconpaul's review. We can hold this off until XT 1.1, but I think I got these right.

### DIFF
--- a/src/common/dsp/modulators/MSEGModulationHelper.h
+++ b/src/common/dsp/modulators/MSEGModulationHelper.h
@@ -19,7 +19,9 @@ struct EvaluatorState
     }
     int lastEval = -1;
     float lastOutput = 0;
-    float msegState[n_lfos_voice] = {0};
+    // 6 is NOT the number of LFOs, but number of MSEG state elements!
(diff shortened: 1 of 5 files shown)


src/common/dsp/modulators/MSEGModulationHelper.h
```diff
@@ -19,7 +19,7 @@ struct EvaluatorState
     }
     int lastEval = -1;
     float lastOutput = 0;
-    float msegState[6] = {0};
+    float msegState[n_lfos_voice] = {0};
```
baconpaul, line 22: This one is wrong. That 6 just happens to be the total number of state elements i need to evaluate the mseg forward for a particular voice. If you want you can make it

```
static constexpr int n_msegStateElements = 6;
msegState[n_msegStateElements];
```
but don't merge this change since it is a bit deceiving.

baconpaul: (you could also imagine calling his data structure 'msegStateRegisters')

PR #5766, opened 2022-01-15
(XT 1.1) Make certain osc controls semitransparent for Audio In

### LINKED ISSUE #5638: Make Pitch/Osc Octave/Keytrack/Retrigger semitransparent for Audio In oscillator
Since they are inconsequential and do nothing for Audio In, they should be considered disabled. And we already show disabled things as 50% opaque 🙂

mkruselj:
Closes #5638

### DIFF
--- a/src/common/SurgePatch.cpp
+++ b/src/common/SurgePatch.cpp
@@ -534,18 +534,12 @@ SurgePatch::SurgePatch(SurgeStorage *storage)
     // Assign the dynamic deactivation handlers
     static struct OscAudioInDeact : public ParameterDynamicDeactivationFunction
     {
-        // TODO: Don't forget to add osc phase here once we add it in XT 2.0!
         const bool getValue(const Parameter *p) const override
         {
             auto cge = p->ctrlgroup_entry;
             auto osc = &(p->storage->getPatch().scene[p->scene - 1].osc[cge]);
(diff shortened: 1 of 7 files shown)


src/common/SurgePatch.cpp
```diff
@@ -531,15 +531,37 @@ SurgePatch::SurgePatch(SurgeStorage *storage)
...
+    static struct OscAudioInDeact : public ParameterDynamicDeactivationFunction
+    {
+        // TODO: Don't forget to add osc phase here once we add it in XT 2.0!
+        const bool getValue(const Parameter *p) const override
+        {
+            auto cge = p->ctrlgroup_entry;
+            auto osc = &(p->storage->getPatch().scene[p->scene - 1].osc[cge]);
+            auto res = osc->type.val.i == ot_audioinput;
+
+            osc[cge].pitch.deactivated = res;
```
baconpaul, line 544: Do you need to do these modifications here? I would have thought the return value was enough if you consistently use 'appears_deactivated' vs '.deactivated'?

mkruselj: Good point. Will try!

mkruselj: Indeed that does work. Thanks!

PR #5767, opened 2022-01-16
Move hovers for the default skin into the binary

### LINKED ISSUE #5747: Default skin fully in memory
Currently we have all hovers and tempo sync overlays for the default skin installed to the file system, and the rest is being read from the binary. It would be nicer to just have a completely in memory default skin. Steps to do that:

1. Move out all SVGs from resources/data/skins/default.surge-skin to resources/classic-skin-svgs
2. Update resources/surge-xt/memory-skin.xml with tempo sync overlay setup
3. Add a flag to the skin to use hover assets from memory, which is false by default and only true for the in memory skin
4. Modify the hover resolution code to read that flag and use assets from the binary
5. Test all the skins and merge

mkruselj:
Also execute shenanigans required in SurgeImage to actually make it happen. Thanks to @baconpaul's hints!

Closes #5747


src/surge-xt/gui/SurgeImageStore.cpp
```diff
@@ -123,6 +125,12 @@ void SurgeImageStore::addEntry(int id)
     SurgeImage *bitmap = new SurgeImage(id);
 
     bitmap_registry[id] = bitmap;
+
+    if (auto *h = SurgeImage::createFromPrefix("hover", id))
+    {
+        std::string name = fmt::format("hover{0}", id);
```
baconpaul, line 131: Shouldn’t this format be set up to do a 00032

mkruselj: Yeah, you're right!

mkruselj:
Things not working:

- hovers actually don't load at all (even though they are embedded properly and `createFromPrefix` enumerates them correctly)
- tempo-sync overlays also don't work because `resource` argument in in-memory skin XML is a filename, not an internal asset name

baconpaul:
lemme see if i can finish this off this afternoon to remove this long lingering issue.

baconpaul:
OK if CI passes this is mergable and closes the issue.

PR #5796, opened 2022-01-19
Remove manually coded denormal handling

### LINKED ISSUE #5722: Revise and remove manual denormal handling
Since we have FTZ enabled and all that, there is no need for Surge to force out denormals using `clear_block_antidenormalnoise`, `flush_denormal` and `flush_sample_denormal`, it seems.

In my initial tests I didn't notice any significant difference in CPU usage between current build of Surge and the one with all anti-denormals measures removed. But let's test this more and decide which of those methods should we axe specifically.

mkruselj:
Minor other cleanups (removing unused comments in a few places)

Closes #5722 and #4900


baconpaul:
OK I spent some time with this tonight doing the tests I wanted to do on ARM and I think it is fine to merge. Let me resolve that trivial conflict and then get it into the 1.1 pipeline.

PR #5799, opened 2022-01-19
(XT 1.1) Move Surge XT to sst-plugininfra for TinyXML

baconpaul:
Making it parallel with SC


baconpaul:
Ahh have to fix the fx too

Issue #5802, opened 2022-01-19
Touch-based input issues


mkruselj:
So this one is gonna be a bit difficult to explain, but using my Surface Pro 4, I can notice these weirdities:

- [ ] Context menus (i.e. patch browser, filter type, osc type, FX type, main menu) don't always open under the location where I pressed
- [x] Long press to right click doesn't work
- [x] Touchscreen mode doesn't set the mouse sensitivity to Exact in the back end, so if we're at default Legacy setting, moving sliders doesn't follow finger movement either

baconpaul:
So the long-press-to-right-click after seeing jatin's solution is really just a timer.
maybe i could do that so it printed in the slider on a long hold and you could go from there?
sound good?

baconpaul:
OK so for the main menu position try this

replace this code

```
if (tag == tag_settingsmenu)
        {
            useDevMenu = true;
            showSettingsMenu(juce::Point<int>{}, control);
            return 1;
        }
```

with 

```
if (tag == tag_settingsmenu)
        {
            useDevMenu = true;
            showSettingsMenu(control->asJuceComponent()->getTopLeft(), control);
            return 1;
        }
```

PR #5806, opened 2022-01-20
Implement touchscreen long press on sliders

baconpaul:
1. Add a LongHoldMixin
2. Implement that on slider
3. In touch mode, as a result, you end up with a long hold
   opening the RMB menu

Addresses #5802

### DIFF
--- a/src/surge-xt/gui/SurgeGUIUtils.cpp
+++ b/src/surge-xt/gui/SurgeGUIUtils.cpp
@@ -38,6 +38,12 @@ bool showCursor(SurgeStorage *storage)
     return sc || tm;
 };
 
+bool isTouchMode(SurgeStorage *storage)
+{
+    bool tm = Surge::Storage::getUserDefaultValue(storage, Surge::Storage::TouchMouseMode, false);
+    return tm;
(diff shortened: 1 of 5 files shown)


mkruselj:
Can be merged, but 2500 ms is too long for a right click. This is actually an adjustable setting on Windows but to get it we'd need to deal with registry, so meh. But reducing it to 750 ms would be nice, before merging.

At any rate will need more work down the road (context menus pop up in the wrong place).

baconpaul:
Surge::Storage::FxUserPreset::getPresetsForSingleType(typeid) gets you al the presets by type.

SurgeStorage has a std::unique_ptr<FxUserPreset> as a member

so call getPresetsForSIngleType and iterate over it.

baconpaul:
But what you actually want to do is call

`FxMenu->menu.showMenuAsync`

So how do you get the FxMenu? Well `SurgeGUIEditor::fxMenu` is a unique_ptr to it. If it is not null you can use the code

```
if (sge && sge->fxMenu)
   sge->fxMenu.menu.showMenuAsync()
```

Issue #5818, opened 2022-01-23
Modulating a modulator workflow improvement


RaymondBaker:
**Is your feature request related to a problem? Please describe.**
It'd be nice if the workflow to modulate the lfo 1 rate/deform/etc with lfo 2 was more intuitive. The arrow buttons are pretty confusing and the drag modulator feature doesn't bring up the modulation amt slider

**Describe the solution you'd like:**
If ctrl + click on a modulator could put it into an active state and no matter where you click on the mod section the original one is still active. (until ctrl clicked again) you could use that to modulate other modulators with ease

example workflow: ctrl click lfo 2 (maybe repurpose the arrow to do the same thing) then click lfo1 and modulate the rate with lfo 2 then click lfo3 and modulate some of its params with lfo 2. Ctrl click lfo 2 to get out of this mode 

**Describe alternatives you've considered:**
I thought about having the drag and drop menu having a button to bring up the sliders but I think that's less intuitive than above. 

P.S loving XT so far. If you want this feature I may be able to implement it if you can point me in the right direction :)

**Additional context:**
![image](https://user-images.githubusercontent.com/40651354/150702661-e8e60358-ac6d-494f-8460-b7b0de19cdcd.png)
![image](https://user-images.githubusercontent.com/40651354/150703025-069b8e21-f75f-4114-a623-1b1d12008759.png)

baconpaul:
so basically 'ctrl click on the 'away' modulator is the opposite of the orange arrow'?

Today: click LFO 2, arm it, orange arrow LFO 1 and you can drag LFO 1 rate to modulate from LFO 2 fine.
Today: Clcik LFO1, drag LFO2 onto rate and you get the typein. fine but you aren't armed.
(There's also the add moduation from option and the mod list approaches)

Suggested Addition: click LFO1, control-click LFO2, LFO2 arms and LFO1 stays active for target?

Or: Control-click on the non-selected modulator arms it without selecting it basically

Just making sure I get it.

I  like that idea, but I know we have gestures already used so if that's right lets let @mkruselj weigh in also.

Your offer to implement it is kind. If I understood it and we decide to do it (with control or some other modifier) I will put some code hints here. Not super hard but not super easy alas. That code is kinda messy. But its testable.

mkruselj:
My thinking is that #5062 is a better solution for this problem.

RaymondBaker:
Yep that's what I was suggesting

[...]

PR #5824, opened 2022-01-25
Ctrl+click on modbutton arms the unfocused LFO

RaymondBaker:
Implementing the workflow talked about here
https://github.com/surge-synthesizer/surge/issues/5818

It'd be cool if the currently selected lfo had a different colour on the matrix when in this mode but I'm not sure how that would be done.

image
image


mkruselj:
Please clang-format the modified files, otherwise the PR cannot be merged.

Also, please add yourself to AUTHORS file along with this PR. Thank you for contributing!

baconpaul:
Oh that currently selected skin color is a good idea. If you open a new issue for that I can add a quick hint or two

baconpaul:
And thanks!

mkruselj:
TBH I think we already have way too many colors for modbuttons, so I would be cautious about slapping more.

baconpaul:
And other than the clang format the change looks spot on! Happy to merge once authors and format changes are in. Thank you!

PR #5886, opened 2022-02-13
Menu Colors in Skins with a default Dark/Light also

### LINKED ISSUE #5861: Menu tweaks
1. Menus have a light mode (skin based? User pref?  Oth?)
2. The font doesn’t seem to be lato

baconpaul:
New feature in skins menu - menus always dark/light/follow skin
Colors in skin which map to current dark menus by default
Light mode works and sticks

Closes #5861

### DIFF
--- a/src/common/SkinColors.cpp
+++ b/src/common/SkinColors.cpp
@@ -64,6 +64,7 @@ namespace Menu
 const Surge::Skin::Color Text("effect.menu.text", 0x000000, 0xFF),
     TextHover("effect.menu.text.hover", 0x000000, 0xFF);
 }
+
 namespace Grid
 {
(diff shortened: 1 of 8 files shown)


mkruselj:
I think I would've preferred per skin colors for the menus rather than a global option, but easy to modify down the road.

baconpaul:
If we had user skin color overrides we could of course get rid of the sticky option

mkruselj:
Is there any way we can get OS light/dark setting then follow it, then just have a single menu option to toggle using menu skin colors from the skin?

baconpaul:
Ha it's funny. I just committed to sstplugin infra saying 'it will all work we just gotta read the registry'

the registry is the api. that's one of the worst things about windows, but it's also true!

PR #5900, opened 2022-02-22
Fix modbutton DnD atop active overlays popping up typein

### LINKED ISSUE #5797: Drag and Drop Modulator with MSEG Open
Drop a modulator onto the MSEG window and it pops for the coordinate at that location

unknown

mkruselj:
Closes #5797


mkruselj:
@baconpaul Will wait for your code review before merge.

baconpaul:
My one comment is I would make the loop over overlays a standalone bool member fn of sge called isPointOverlay to avoid the copy and paste. Decl in the sge header and impl in sgeoverlays then change the loop to if is in && modblah

mkruselj:
I think I can do that 🙂

baconpaul:
Both bits of code look like

```
for each thing
    if point contains thing
         do x
```

basically right? So I suggest 

1. make a new function which looks like

```
bool isOverOverlay
   for each thing
        if point contains thing
             return true;
  return false
```

then modify your code to be

```
if (isOverOverlay(point))
   do x
```

PR #5917, opened 2022-02-27
Migrate filters code to sst-filters

jatinchowdhury18:
Here it is! A first pass at migrating the Surge filter code to the new sst-filters submodule.

From the limited testing I've done so far, everything seems to be working okay, but I'm planning to do a more comprehensive side-by-side test a little later on. The Combulator and Resonator effects use the filters also, so I'll need to make sure to test those as well.

A couple notes:
- To get the `TuningProvider` system to work, I had to make a couple modifications to SurgeStorage.
  - `MIDI_0_FREQ` is passed along from the `Tunings` library.
  - `note_to_omega_ignoring_tuning()` now accepts an extra argument for the current sample rate.
- The original filters were accessing the global sample rate variables, which the new submodule filters cannot access.
  - There's now `FilterCoefficientMaker::setSampleRateAndBlockSize()`, which is called when the sample rate changes.
  - `FilterCoefficientMaker::updateState()` passes the sample rate info along to the filter state.
- `FilterCoefficientMaker::updateState()` also passes the coefficients from the coefficient maker to the filter state.


src/common/SurgeSynthesizer.cpp
```diff
@@ -2086,6 +2086,17 @@ void SurgeSynthesizer::setSamplerate(float sr)
...
+    for (int s = 0; s < n_scenes; s++)
+    {
+        auto iter = voices[s].begin();
+        while (iter != voices[s].end())
+        {
+            SurgeVoice *v = *iter;
+            assert(v);
+            v->sampleRateReset();
+        }
+    }
```
baconpaul, line 2099: This is fine of course but does surge work if we change sr with playing voices? I guess now it won’t break here :)

src/common/dsp/QuadFilterWaveshaper.h
```diff
@@ -0,0 +1,20 @@
...
+    __m128 R[n_waveshaper_registers];
+    __m128 init;
+};
+typedef __m128 (*WaveshaperQFPtr)(QuadFilterWaveshaperState *__restrict, __m128 in, __m128 drive);
+WaveshaperQFPtr GetQFPtrWaveshaper(int type);
+/*
+ * Given the very first sample inbound to a new voice session, return the
+ * first set of registers for that voice.
+ */
+void initializeWaveshaperRegister(int type, float R[n_waveshaper_registers]);
```
baconpaul, line 20: Good change! But maybe we should do another step of putting the waveshapers in the filters module now I think of it? But we can do thst separately

baconpaul (commented):
Did a review. Zero changes I see and happy to merge if the tests pass. A few little comments inline thinking about next steps
Awesome!

baconpaul:
So I eyeballed the changes and they look like correct translations. Didn’t look into why ci failed. Exciting stuff!

jatinchowdhury18:
Thanks! It looks like the failures are just because I didn't update the unit tests and Python bindings, plus I missed a couple of clang-format files. I'll clean that up tonight or tomorrow, and then go through the manual testing process.

src/common/dsp/QuadFilterWaveshaper.h
jatinchowdhury18: I wasn't sure if the waveshapers should go into the filters submodule, or maybe they belong in their own submodule? From what I could tell in extracting the filters code, they don't seem to be particularly connected.

baconpaul: Yeah we could make them separate true! Submodules are easy enough

mkruselj: +1 on having waveshapers in a separate submodule!

baconpaul: Ok!

baconpaul:
So @jatinchowdhury18 I'm happy to merge this if you are. What do you think? We have to do it one day!

jatinchowdhury18:
Went through my testing plan tonight and fixed a couple bugs. Should be good to go now!

baconpaul:
OK merging then! Wahey! 

I have a plan for oscillators I'll share on discord. :)

PR #5924, opened 2022-03-02
"Activate individual scene outputs" overhaul

### LINKED ISSUE #5657: "Activate individual scene outputs" option stored per instance
Actually I wonder if this option should maybe only be displayed for FL Studio, and just have it enabled always in other DAWs, since they don't have problems with scene outputs being active all the time and they only enable them when you route those outputs to some tracks...

mkruselj:
Remove activateExtraOutputs user option
Add it to DAWExtraState instead
Show the menu option to enable scene outputs only for FL Studio

Closes #5657

### DIFF
--- a/src/common/UserDefaults.cpp
+++ b/src/common/UserDefaults.cpp
@@ -71,9 +71,6 @@ void initMaps()
             case UseODDMTS:
                 r = "useODDMTS";
                 break;
-            case ActivateExtraOutputs:
-                r = "activateExtraOutputs";
-                break;
             case MonoPedalMode:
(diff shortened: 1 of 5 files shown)


src/common/UserDefaults.cpp
```diff
@@ -71,9 +71,6 @@ void initMaps()
             case UseODDMTS:
                 r = "useODDMTS";
                 break;
-            case ActivateExtraOutputs:
```
baconpaul, line 74: Removing this means people who have it set in their XML will get a popup; we should leave it in and ignore it with a comment.

PR #5929, opened 2022-03-04
Replace RMB bypass in FX grid with FX presets menu

### LINKED ISSUE #5613: Replace RMB on FX grid slot with context menu
Currently we use right-click to toggle bypass of an effect slot in the FX grid. And we use double-click on waveshaper curve display and filter type icons. Inconsistent!

What's worse, in order to load an effect, you need to first select the slot you want, then move mouse down and pick an effect. Inefficient!

So for 1.1, replace RMB bypass on FX grid slot with a context menu (same as the FX preset menu, for simplicity), ~~and add double-click to bypass~~. Easy!

mkruselj:
Massive QOL improvement, this. Closes #5613.


mkruselj:
Oh also pinging @VincyZed for manual update!

Issue #3617, opened 2021-01-11
3D Waveform for OSC (when morphing waveforms)


artyfunk:
Like discussed in DIscord. I write it down here. We wish to have a Optical visualisation of the Waveform complete lengt and have the ability to morph between them.

Like Serum:
![Image 3763](https://user-images.githubusercontent.com/76977285/104245859-4f12e300-5465-11eb-8678-69f9acf31e8b.jpg)

or Pigments:
![Image 3762](https://user-images.githubusercontent.com/76977285/104245889-5d60ff00-5465-11eb-93c2-7ba993564943.jpg)

or Vital:
![Image 3764](https://user-images.githubusercontent.com/76977285/104245915-6782fd80-5465-11eb-8ca9-20ae4461b27a.jpg)

This would show what is coming next in the Waveform and would make the life little bit easier.
I understand its not THAT important but good to have.

So if you like maybe you could do it. Would be great to see this feature.

baconpaul:
`baconpaul/19-3dwavetable` branch does this

![shaded](https://user-images.githubusercontent.com/13387561/104342122-7ea70700-54c8-11eb-885a-8cdbda8b106e.gif)

but i think it is kinda bad tbh for most of our wavetables. (I'm not yet drawing the oscillator output line there). Not convinced this is something we really want.

baconpaul:
Oh that's at 250% on a mac retina display, so it is 5x denser than a windows 100% user would get.

mkruselj:
@baconpaul This might be a nice one to maybe cram into 1.1 as a sort of a surprise? It seems like it's pretty much there already, it just (sorry for using that word!) needs the WT frame drawing occurence tweak as suggested above. Maybe? 🙂 

(I'm trying to relieve the XT2 milestone a little bit.)

baconpaul:
Sure

artyfunk:
it needs a actual display highlight. i mean the actual waveform (part of the complete waveform) should be thicker and other color.

mkruselj:
That's a given.

baconpaul:
Well not the actual waveform. The selected frame of the wavetable

baconpaul:
Is there anything left to do here?
The 'actual waveform' is a bit tricky - we don't know exactly how many samples constitute one cycle easily but we can do it. 
But is there anything else we want?

I'm sort of temped to close this as done then let people argue about the look and feel.

wdyt?

mkruselj:
[...]

PR #5953, opened 2022-03-07
Add context menu for WT display

mkruselj:
Addresses #3617

Also fixes help link not working for Alias Additive Options context menu.
Also RMB on wavetable jog now doesn't jog but opens WT context menu instead.


mkruselj:
Oh and of course, a ping to @VincyZed for manual update (right-click the wavetable display!)

PR #5968, opened 2022-03-12
Windows installer improvements

rovingeye:
- Create start menu shortcut icon for Surge XT Effects standalone
- Only create shortcuts if standalone version is installed
- Fix 32 bit file path for shortcuts
- Update destination from {fonts} to {autofonts} as per Inno Setup recommendation


rovingeye:
I would prefer to install to userfonts to avoid duplicates (I have them installed locally already) but Inno Setup recommends against this with administrative installation and is only supported for Windows 10 and up. Unsure how to this without separate installers

baconpaul:
Awesome thanks!

Issue #5811, opened 2022-01-22
Drop-Down Menus are small on a 4k screen


tsigoulas:
**Bug Description:**
All the Drop-Down Menus are small on a 4k screen (4096x2160)

**Surge Version**

* Version: XT 1.1 nightly.8d294da
* Plugin Type: [VST3]
* Bitness: [64-bit]

**Reproduction Steps:**
click to open any drop-down menu
**Expected Behavior:**
Small drop-down Menus

**Screenshots:**
https://i.imgur.com/758xm0T.jpg

**Computer Information (please complete the following information):**
 - OS: Windows 10 64bit
 - Host: Reaper
 - Version: v.6.46
![Surge XT small size drop-down menus](https://user-images.githubusercontent.com/93413808/150655607-01119a2c-b1cd-4d39-be3d-769a91bdf3ed.jpg)
 
 - 
**Additional Information:**

baconpaul:
So the menus follow the os scaling not the plugin scaling. When I tested windows 4k at 100% the menus in surge were same Size as in reaper

I’m assuming you are after a mode where the menus are a scaled up version consistent with surge 100%? That’s a bit painful but can I first confirm your scaling and hdpi settings? 

Thanks

tsigoulas:
My OS scaling is at 200%, Surge XT scaling is at 159%, Reaper shortcut properties setting "high DPi Scaling override" is set to application.  Also, I would like to mention that  the Original Surge v1.9 is scaling the drop-down menus just fine, so it seems that something broke with the XT code. Hope that helps, thanks again for your amazing work!

mkruselj:
XT UI code is totally different from 1.9 because we have changed frameworks. 1.9 used OS based menus, whereas XT uses JUCE managed menus. These are very different, but I would not say "broken".

tsigoulas:
I see... So, is this something that can be improved in future updates? Unfortunately, to be able to use the plugin on a high resolution screen is only possible with WIndows Magnifier, at least for my old eyes :). Thanks!

baconpaul:
Maybe! 

It is trivial to fix if I want everyone else’s menu to be way too large - but obviously we don’t! So gotta figure out the windows hdpi shenanigans 

if you could record every scale setting you have (os and reaper) in this issue I can look

there may be a way to make them “optionally way too big” which you would turn on and others would turn off. (I say “way too big” because really they would be).

tsigoulas:
[...]

PR #5977, opened 2022-03-18
fix for popup menu scaling on hidpi display

sadguitarius:
Hi all, I've cobbled together a fix for #5811 that hopefully works for other platforms than mine (Windows 10), and also hopefully isn't a trash pile in terms of code quality. This is my first attempt at contributing to the Surge repo, and I'm only somewhat familiar with Juce and its mysteries. I tried to keep variable names and code style more or less in line with what I could find in the Surge code. Please let me know if there's anything that should be done to further clean this up. 

The one issue I'm still running into is that there doesn't seem to be an obvious way to control the scaling of the popup menus within the Lua editor. Does anyone know how to go about this?

Thanks!

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -18,6 +18,7 @@ Jani Frilander <http://github.com/janifr>
 Amy Furniss <http://github.com/amyfurniss>
 Cale Gibbard <cgibbard@gmail.com>
 Brian Ginsburg <https://github.com/bgins>
+Ben Goldwasser <https://github.com/sadguitarius>
 Nathan Kopp <nk.sg@nathankopp.com>
 Korridor <fabian.kempe@gmail.com>
 Mario Krušelj a.k.a. EvilDragon <http://github.com/mkruselj>
--- a/src/surge-xt/gui/SurgeGUIEditor.cpp
+++ b/src/surge-xt/gui/SurgeGUIEditor.cpp
@@ -2218,7 +2218,6 @@ juce::PopupMenu::Options SurgeGUIEditor::popupMenuOptions(const juce::Component 
     }
 }
 
-
 void SurgeGUIEditor::showZoomMenu(const juce::Point<int> &where,
                                   Surge::GUI::IComponentTagValue *launchFrom)
 {
--- a/src/surge-xt/gui/SurgeGUIEditor.cpp
+++ b/src/surge-xt/gui/SurgeGUIEditor.cpp
@@ -2130,7 +2130,8 @@ void SurgeGUIEditor::effectSettingsBackgroundClick(int whichScene, Surge::Widget
                                synth->storage.getPatch().isDirty = true;
                        });
 
-    fxGridMenu.showMenuAsync(juce::PopupMenu::Options(), Surge::GUI::makeEndHoverCallback(c));
+    juce::Point<int> cwhere = c->asJuceComponent()->getBounds().getBottomLeft();
+    fxGridMenu.showMenuAsync(optionsForPosition(cwhere), Surge::GUI::makeEndHoverCallback(c));
 }
 
 void SurgeGUIEditor::controlBeginEdit(Surge::GUI::IComponentTagValue *control)
@@ -2193,6 +2194,7 @@ void SurgeGUIEditor::toggleMPE()
 juce::PopupMenu::Options SurgeGUIEditor::optionsForPosition(const juce::Point<int> &where)
 {
     auto o = juce::PopupMenu::Options();
+    o = o.withTargetComponent(juceEditor);
     if (where.x > 0 && where.y > 0)
     {
         auto r = juce::Rectangle<int>().withWidth(1).withHeight(1).withPosition(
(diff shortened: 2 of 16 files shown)


baconpaul:
First, thank you very much for contributing!

I want to give this a whirl on my mac and also review it, but am traveling today. I'll try and make time though among the various activities - perhaps this evening US East Coast time.

I did take a quick skim though. Let me run it on mac but here's a couple of things.

I see you had to thread the editor everywhere. Is that just for the optionsForComponents method? I wonder if the change could be smaller if that was a static in GUIHelpers and also if it automatically generated the point if not included.

That is rather than

```
auto p = cwhere->getLowerLeft()
sge->opstionsForBlah
```

instead do

```
Surge::GUI::MenuOptionsForBlack(chwhere)
```

which internally does

```
auto p = cwhere->getLowerLeft()
// contents of that method
```

also we've generally been using `auto` so `juce::Point<int> x = c->getLower()` is increasingly idiomatic in our code as `auto x = c->getLower()`

baconpaul:
Oh also you should add yourself to the AUTHORS file :)

mkruselj:
Thanks for your contribution, and you're also welcome on our Discord server! 🙂

sadguitarius:
Thanks @baconpaul and @mkruselj! 

I absolutely agree that it's not ideal to have all those pointers to the editor. The reason I did it that way was for the `juce::PopupMenu::Options::withTargetComponent` method, which according to the documentation seems to be the canonical way to get popup menu scaling to behave itself. Since this method needs a reference to an actual instance of a component, I'm guessing it can't be made static as is, but I don't doubt there a better way to get the same thing done. Any ideas?

I do get the point about bringing some of that repeated code into the optionsForPosition methos

sadguitarius:
Thanks @baconpaul and @mkruselj! 

I absolutely agree that it's not ideal to have all those pointers to the editor. The reason I did it that way was for the `juce::PopupMenu::Options::withTargetComponent` method, which according to the documentation seems to be the canonical way to get popup menu scaling to behave itself. Since this method needs a reference to an actual instance of a component, I'm guessing it can't be made static as is, but I don't doubt there a better way to get the same thing done. Any ideas?

I do get the point about bringing some of that repeated code into the `optionsForPosition` method. I think it's pretty much either attaching to the bottom left of another component or using mouse position, so maybe it could be done with just a couple of different method signatures. I'll have to look again at the code, since some of that stuff is a little weird with all the layers of component translations and might need some edge cases. I guess I was trying not to mess with existing methods too much, but if this checks out on Mac as it is now, I'll go in and clean things up a bit more.

Do you think it would be better to pull some of this stuff into a dedicated class?

sadguitarius:
I went ahead and did a little cleanup and it reads much better now. I also found a few menus I'd missed before. One thing-- I can't for the life of me figure out why the oscillator octave popup is showing up either at the top or bottom of the entire window. The same control for the scene octave doesn't do this. What am I missing?

baconpaul:
OK just gave this a whirl

The thing is: this makes the menu appear directly under ever control instead of at the mouse click. I actually think that's an improvement but I'm not sure.

The thing I was worried about (that mac retina zoom would be off) didn't happen.

I can confirm your oscillator switch issue. Let me see if I can debug that.

baconpaul:
Screen Shot 2022-03-18 at 5 53 46 PM

That is, no matter where you click on this selector the menu always positions here.

baconpaul:
@sadguitarius do you mind if I push a diff back to your branch for a potential change to the above location shift?

baconpaul:
[position.patch.txt](https://github.com/surge-synthesizer/surge/files/8307895/position.patch.txt)

Ahh your stuff is on a main ... my git got confused and I can't push to your branch.

Anyway see the above patch file. If you apply this then (1) all your changes are in effect namely everyone gets a component but

(2) the menu positioning is unchanged and the oscillator menu is fixed

If that patch works for you please push again and I'll happily merge and close the issue!

Thanks

baconpaul:
Oh and you are correct this can't be static - needs a ref to the juceEditor indeed!

sadguitarius:
Awesome, just added your diff and all's well. The menu placement thing came about kind of accidentally and I did like it a little, but it was inconsistent when the popup came close to the edge of the screen and Juce moved it around. Any idea about the popup in the Lua editor? I can't see a way to fix its scaling without creating a TextEditor subclass, which seems a bit unnecessary.

baconpaul:
Alas that menu inside the CodeComponent is created by juce with a default popup menu. Short of patching JUCE (or asking juce team for a patch in a future version which I can do once we get this merged) there's no real way to fix it.

I see clang-format is tripping up again; if you can reformat and push when CI finishes I'll merge this and can open a forum issue with juce folks.

Thanks for diving in and fixing this. Anything else in the codebase interest you?

sadguitarius:
Ah, I figured that might be the case. Anyhow, I fixed the formatting so it should be good to go.

I'll take a look at the other issues and see what looks doable. Windows hidpi scaling has been the bane of my existence, so I wanted to tackle that so I can actually use Surge comfortably. I'm very happy to continue contributing the project. Thanks for your help!

baconpaul:
Cool. We also have work to do on shortcircuit, monique, and b-step if any of those are of interest! (And stocks too if you want but that is a bit more stable than SC Mon and Step right now)

mkruselj:
I'm really curious if this menu positioning thing is also going to apply to touch input. I will test it, but it will be really cool if this PR closes #5802 too. 🙂

mkruselj:
Nope, it's actually worse now...

baconpaul:
Yes but the fix is in place
If you look in SGE::popupOptions (or whatever the function is) that takes an argument Component*, bool.
Right now it looks like

```
if !component || !bool
   where = getMouseLocation
else
   where = component->getLowerLeft
```

So modify that to be

```
if !component || !bool
   where = getMouseLocation
else
   where = component->getLowerLeft

if touch
   where = compoent->getLowerLeft

```

and see if that helps

baconpaul:
basically this change is the api fix we need just not the implementation fix

mkruselj:
Ah gotcha! Will give it a shot.

Issue #694, opened 2019-03-02
Editor Undo/Redo


baconpaul:
OK tackle this so here's the checklist. As always items on the list without a check are known and not done; items with a check are in a branch which isn't merged or pr yet

- [x] Undo accessibility and announcements
- [x] Undoable action clears redo stack

-------

Would be great if the synth had undo/redo for edits. from 
https://www.kvraudio.com/forum/viewtopic.php?p=7330177#p7330177

j5v:
Came to post, but found this issue was already here.

- **Undo** and **Redo** could go next to **Store** in the UI.
- Storage: happy with session persistence only.
- When to store:
    - on data entry,
    - on mouse up on slider, or
    - 2s delay after scroll action, or overwrite the last history state if changing the same parameter again.
- Keyboard shortcuts: use Ctrl/Cmd+Z and Ctrl/Cmd+Y (or Shift+Ctrl/Cmd+Z), preventing pass-thru to host.

mkruselj:
> * Keyboard shortcuts: use Ctrl/Cmd+Z and Ctrl/Cmd+Y (or Shift+Ctrl/Cmd+Z), preventing pass-thru to host.

Unfortunately this one would not be happening, as dealing with keyboard events in all the various DAWs is just a humongous pain in the ass (and sometimes just doesn't work no matter what anyways)

mkruselj:
Apparently, moving to XT 1.1 milestone. 🥳

mkruselj:
- [ ] "Undo changes" action shows both on top and on bottom of Edit Keybindings overlay.
- [ ] Redo action?

baconpaul:
Ahh I forgot about redo. Will add this week before I add more undoables

PR #6004, opened 2022-04-01
Move towards editor undo

baconpaul:
This is a basic infrastucture which moves us towards undo
in an editor. Lots missing here all detailed in a checklist
in #694 but with this commit parameter changes, modulation changes,
osc and fx changes all push into the undo stack


baconpaul:
Don’t merge

baconpaul:
OK that bug is fixed. When CI completes on the new branch this can be merged. I can probably get to it in 30 minutes.

PR #6007, opened 2022-04-02
Various tweaks for keyboard shortcut manager

mkruselj:
Add Alt+B shortcut for opening the keyboard shortcut editor.
Add scaffolding for redo action.
Debug console action shows only on Windows.
Revert to JUCE default font for keyboard shortcut description


baconpaul:
This doesn’t contain your submodule change fyi

to do that you need to move the submodule pointer

cd libs sst plugin
Git checkout main
Git pull origin main
Cd ../../..
Git status

should show the lib as modified. So git add libs commit and push

baconpaul:
I actually did that for you just now and pushed it so I can take a look at row painting in a minute

PR #6013, opened 2022-04-03
Fix checkmark not showing when selecting Surge Classic skin

mkruselj:


baconpaul:
As mentioned on discord I would sooner the be done inside matches not outside so other uses are also correct

PR #6024, opened 2022-04-06
Doubleclick Renamed Modulators without annoying bug

### LINKED ISSUE #5774: Double click to rename macro/modulators
Very late in the 1.0 cycle we removed double click to rename an lfo

We did it because this gesture

LFO 6 selected
LFO 3 double click

in 1.9 does an implicit select and arm

and in XT launched the rename

So we need to figure out the state to rename only when 'intended' (namely not when also selecting) but that requires state from SGE so we pulled the feature last minute.

baconpaul:
Doubleclick renames a modulator. Cool

But this kinda sucks if you click into a modulator and then click to arm
quickly and misfire a double click

So if you have *just* selected a modulator the double click doesn't work.

Only after that

Closes #5774

### DIFF
--- a/src/surge-xt/gui/SurgeGUIEditor.h
+++ b/src/surge-xt/gui/SurgeGUIEditor.h
@@ -224,6 +224,8 @@ class SurgeGUIEditor : public Surge::GUI::IComponentTagValue::Listener,
 
     uint64_t lastObservedMidiNoteEventCount{0};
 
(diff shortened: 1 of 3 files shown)


baconpaul:
i'm going to bed and this failed coequal so i pushed it. if it is green in the am feel free to merge or I can do so when i wake up!

Issue #6137, opened 2022-05-07
Support CLAP monophonic, polyphonic modulators and note expressions


baconpaul:
Post recent merges a lot is in place. Here's what's left

- [x] Test Note Expressions in BWS 43b1 with an MPE controll
- [x] that fixme around mono mod source smoothing
- [ ] change 'mpe timbre' and 'channel at' labels depending on mpe mode or not
- [x] that '4' from BWS in Volume as tracked here https://github.com/free-audio/interop-tracker/issues/7
- [x] bugs people find of course

baconpaul:
@TjShredder reports over in clap repo and on kvj several times the clap doesn't work in bitwigh with mpe

i think this is because the clap says it supports mpe and btiwig converts this to note expressions but investigate along with this issue.

PR #6149, opened 2022-05-10
WIP: Support for Note ID, Poly and Mono modulation and more

baconpaul:
This commit allows a host of surge, if the host provides it,
to thread a note id through the voice management so you can
handle start/end note with id messages (end coming from surge
back to the daw). It also allows all the parameters and macros
to separate modulation from value optionaly.

Given that, and given the update to clap-juce-extensions,
it re-implements the ::process loop inside SurgeSynthProcessor
to have an optional clap process loop which provides polyphonic
modulation.

Addresses #6137

### DIFF
--- a/src/common/ModulationSource.h
+++ b/src/common/ModulationSource.h
@@ -307,7 +307,7 @@ class ModulationSource
     virtual bool is_bipolar() { return false; }
     virtual void set_bipolar(bool b) {}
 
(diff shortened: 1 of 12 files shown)


baconpaul:
Please *DO NOT MERGE THIS*

I want to do some more testing but also want to run builds and tests all platforms.

PR #6207, opened 2022-05-24
Always dismiss patch search results on double click of list entry

mkruselj:


mkruselj:
Don't merge, I think we don't need to do that returnKeyPressed, it would just load the patch twice. All we needed is to escape on dblclick... Am outside for a bit but will fix in an hour.

PR #6213, opened 2022-05-29
Rename default macro names for host automation

mkruselj:
From "c0:" to "M1" etc.


baconpaul:
Oh this will conflict with the accesibility change i just made that did this. I went with "Macro 1" since we have room. I'm gonna close this so I can sweep rest.

mkruselj:
Some hosts will clamp at 16 chars so with some of actually longish macro names it might not fit, but alright!

baconpaul:
Really? I thought VST3 was 64

baconpaul:
But I'll change to M and push mine.

mkruselj:
Ah right yeah VST3 is 64, but VST2 is 16. Heheh. 😅

PR #6257, opened 2022-06-07
Remove WT name nulling before extra oscdata load

### LINKED ISSUE #6248: Wavetable name not remembered after just switching to WT/Window osc type
1. Open standalone
2. Set Osc 1 to Wavetable or Window
3. Close standalone
4. Reopen standalone

Instead of Sine, wavetable name will say (Patch Wavetable). It seems the name sticks only if you literally choose it through the menu or via arrows.

mkruselj:
Closes #6248


baconpaul:
I think this is a mistake to merge. Can you hold on until I poke at it some?

I tried this last night and in some cases it fixes things, but in others i could get uninitialized memory reads

mkruselj:
Sure we can hold out on it.

PR #6305, opened 2022-06-22
Change default MIDI/MPE bend smoothing to fast linear

### LINKED ISSUE #6301: Change the default controller smoothing to something other than Legacy
Legacy controller smoothing seems to be the source of a couple of issues, which have been encountered by me and others. @mkruselj proposed that we should remove that smoothing mode in XT 2, but in the meantime for 1.1, we should probably change the default setting to some other smoothing option, like Fast Linear for instance.

mkruselj:
Closes #6301

### DIFF
--- a/src/common/ModulationSource.h
+++ b/src/common/ModulationSource.h
@@ -317,7 +317,7 @@ template <int NDX = 1> class ControllerModulationSourceVector : public Modulatio
 {
   public:
     // Smoothing and Shaping Behaviors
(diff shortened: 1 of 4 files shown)


baconpaul:
Nope keeps crashing even if i fix the tests.

I'll do a separate push to fix the test anyway but pushing this out of 1.1 milestone. A bit scary honestly but something in the tests is assuming a default somewhere. Will be tricky to debug.

PR #6323, opened 2022-06-29
Adding clap files to rpm build.

gnac:
Modified the scripts/installer_linux/make_rpm.sh file to include the clap plugin files in the rpm.
Also forced the RPM config date format. Local testing indicated that the default no longer met the config standard.


baconpaul:
Thank you!

PR #6335, opened 2022-07-10
Handle special characters (', %) in search strings

### LINKED ISSUE #6326: SQL database bug
From j5v:

I have a single quote in the patch name. It saves but then if I hit the search, it immediately shows the attached error. I guess the query isn't parameterized, and is instead string-built without escaping SQL delimiters. The workaround is to accept the error and then save with a name that doesn't break the SQL syntax.

image

baconpaul:
Closes #6326

### DIFF
--- a/src/common/PatchDB.cpp
+++ b/src/common/PatchDB.cpp
@@ -1351,6 +1351,23 @@ int PatchDB::numberOfJobsOutstanding()
 
 std::string PatchDB::sqlWhereClauseFor(const std::unique_ptr<PatchDBQueryParser::Token> &t)
 {
+    auto protect = [](const std::string &s) -> std::string {
+        std::vector<std::pair<std::string, std::string>> replacements{{"'", "''"}, {"%", "%%"}};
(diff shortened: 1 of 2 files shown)


src/common/PatchDB.cpp
```diff
@@ -1351,6 +1351,23 @@ int PatchDB::numberOfJobsOutstanding()
 
 std::string PatchDB::sqlWhereClauseFor(const std::unique_ptr<PatchDBQueryParser::Token> &t)
 {
+    auto protect = [](const std::string &s) -> std::string {
```
aaylward, line 1354: I'm new to sqlite -- is this similar to [quote](https://www.sqlite.org/lang_corefunc.html#quote)?

baconpaul: Yeah except I’m inbound from non sql so have to proctrft my c++ string

aaylward: cool :) thanks for explaining!

PR #6352, opened 2022-07-12
Fix typos

luzpaz:
Found via `codespell -q 3 -S ./libs -L adres,applys,doesnt,fo,nd,olt,ontext,ot,pard,parm,ro,ser,tbe,te,whichs,wth`


baconpaul:
Thank you.

PR #6389, opened 2022-07-20
Final changes to polyvoice modes

baconpaul:


baconpaul:
I plan to merge this, since it finishes poly mode, but #6390 shows problems in mono mode.

PR #6471, opened 2022-08-04
Install CLAP targets on Linux

### LINKED ISSUE #6467: Limux make install skips clap
Update those cmake rules….

cbix:
Closes #6467

Still building here but this should do the trick ;)


baconpaul:
Oh thank you!!

PR #6579, opened 2022-08-29
Modify CMake so Surge can be a proper sub-library

baconpaul:
We used ${CMAKE_SORUCE_DIRECTORY} too freely; so set a
SURGE_SOURCE_DIRECTORY at the top cmake and use that for
some of our paths and stuff.


baconpaul:
Please let CI fully complete before merging this. I think it's OK but want to make sure we get proper result on all platforms.

Issue #6576, opened 2022-08-27
A Basic CPU Load Implementation


danielhoey:
**Bug Description:**
Playing very many notes with multiple sine OSC with unison voices will produce a lot of clicking/fuzz/digitial distortion.

**Surge XT Version**
Surge XT 1.0.1.60228e8
64 bit
Stand alone executable

**Reproduction Steps:**
1. Initialize patch
2. Set OSC 1 to Modern/Sine, Unison Voices -> 16
3. Set OSC 2 to Modern/Sine, Unison Voices -> 16
4. Set Amp EG release to a high value (0.3s or higher)
5. Play as many note as possible for a 5+seconds.

**Expected Behavior:**
The sound produced by pressing the first key should not be significantly different then the sound produced when pressing later keys.

**Computer Information (please complete the following!):**
Windows 10 - 64bit
Stand alone executable ("Surge XT.exe")

**Additional Information**
Surge is utilizing about 25% of CPU during this process

baconpaul:
It sounds like you are running into the clipper. If you turn the amplitude down for each osc in the mixer does the problem resolfe? 

Also 25% is quite a lot of cpu. May want to take that unison count to 7.

danielhoey:
Thanks for the prompt reply. Changing the osc amplitude makes no difference
(except that the artefacts are quieter).
The magnitude of the problem reduces as the the unison count is decreased
and at 10 it disappears completely and the the cpu usages is limit to 21%
Apologies if this is expected behaviour, I didn't realise that I was
pushing the synth to a limit by selecting max unison voices.

Andreya-Autumn:
Yeah the 16 voice unison is something of a legacy option at this point. There's severely diminishing returns after around 7 or so anyway, and you're not the first to run into CPU and/or level issues with it. 

Seems like the consensus is that the next major update (not strictly scheduled but planned sometime next year maybe) will have a more sensible max to the unison parameters. :)

mkruselj:
Yeah those digital artifacts are your CPU being overloaded, not an error in Surge. Since this happens at 25% CPU,  this tells me one core (assuming you have a quad core CPU) simply got oversaturated and since Surge doesn't do multicore processing, everything falls over.

At any rate,  not a Surge bug.

baconpaul:
We really will have to do multi core when we do xt2.

danielhoey:
[...]

PR #6583, opened 2022-08-31
Display CPU load on VU meter

danielhoey:
An approximation to CPU load is displayed on the VU meter. On my dev machine the value shown accurately reflects the ability of the synth to accurately render/process the specified sound. The value is smoothed out with a running average and has falloff but it still changes very quickly. 

Related issue #6576

### DIFF
--- a/src/common/SurgeStorage.cpp
+++ b/src/common/SurgeStorage.cpp
@@ -1944,6 +1944,8 @@ void SurgeStorage::init_tables()
     // ballistics at any sample rate!
     // We have also decided to make the meters less sluggish, so we increased the cutoff to 60 Hz
     vu_falloff = exp(-2 * M_PI * (60.f * samplerate_inv));
+
+    cpu_falloff = exp(-2 * M_PI * (60.f * samplerate_inv));
 }
 
 float SurgeStorage::note_to_pitch(float x)
--- a/src/common/SurgeStorage.h
+++ b/src/common/SurgeStorage.h
@@ -1129,6 +1129,7 @@ class alignas(16) SurgeStorage
     float poly_aftertouch[2][16][128]; // TODO: FIX SCENE ASSUMPTION
     float modsource_vu[n_modsources];
     void setSamplerate(float sr);
+    float cpu_falloff;
 
     bool getOverrideDataHome(std::string &value);
     void createUserDirectory();
--- a/src/common/SurgeSynthesizer.cpp
+++ b/src/common/SurgeSynthesizer.cpp
@@ -4101,6 +4101,8 @@ void SurgeSynthesizer::process()
     memset(endedHostNoteIds, 0, 512 * sizeof(int32_t));
 #endif
 
+    auto process_start = std::chrono::high_resolution_clock::now();
+
     if (hostNoteEndedToPushToNextBlock)
     {
         for (int i = 0; i < hostNoteEndedToPushToNextBlock; ++i)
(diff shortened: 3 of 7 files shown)


baconpaul:
Hey this is fantastic thank you!

that failed ci run is because you need to run clang format on your code before merge. Since you will need to push that anyway can you add yourself to AUTHORS also? And if you show up on our discord please let us know and we will give you the team role.

I’m also sure @mkruselj will want to add skin colors but we can do that in a separate commit. Happy to merge this once the format and authors chance is added and ci passes!

mkruselj:
Yes and I suspect for the time being we might also want to toggle this via an option. In XT2 the GUI will have actual provisions to show CPU usage at all times.

baconpaul:
> Yes and I suspect for the time being we might also want to toggle this via an option. In XT2 the GUI will have actual provisions to show CPU usage at all times.

Yeah or at least only show it if it is above 50% or some such. We can tweak all that for 1.1.2 once we get the base function merged here. Super useful!

danielhoey:
I don't really think that the actual number is all that useful, it changes too quickly and there is no indication as to what it is measuring. I had an implementation that I preferred that just printed "CPU" in yellow when >75% and in red when >95% which at least indicates what is being measured and doesn't clutter the UI with unimportant info (i.e. low CPU)

baconpaul:
> I don't really think that the actual number is all that useful, it changes too quickly and there is no indication as to what it is measuring. I had an implementation that I preferred that just printed "CPU" in yellow when >75% and in red when >95% which at least indicates what is being measured and doesn't clutter the UI with unimportant info (i.e. low CPU)

I like this idea yes

mkruselj:
Can we not smooth the display rate out some so that it is not as jittery? It'd be a matter of changing the falloff constant. I'd definitely want to see the number.

baconpaul:
We have a while until 112 so let’s merge it as is and try some stuff once the clang format is fixed. The hardest part is calculating and communicating it and that’s now done.

danielhoey:
I played a bit with the falloff rate but the problem is that it is still jittery on the way up (it only affects the display when the value falls) and if you increase it too much it the old peak hangs around for a long time and it not longer accurately reflects reality. Might be worth another look at, but I think it might be better to repaint it less frequently

baconpaul:
We can also keep a buffer of `32 * 64 / BLOCK_SIZE` last elements and average instead (basically keep a ring buffer and the running total and subtract n-1 and add n then update the atomic). That would be smoother and not have the same falloff characteristics.

I still think the best thing is: merge it and all play with it a bit though :) People are used to features coming and going oddly in the nightlies and we are just past a release so its a pretty relaxed time at head right now.

mkruselj:
Right I guess the mistake here is that we use the same ballistics that a VU meter does, which is fine for audio, but not for CPU usage... Averaging a ring buffer sounds like a much better approach, yeah.

danielhoey:
I think the ballistics is probably fine. If we were displaying the value as a coloured line in the same way as the VU meter then there wouldn't be any problem. I also don't think that the ring buffer will make any difference to the stability of the number, I'm already smoothing it out in the same way (unless I'm missing something), a ring buffer would just slightly increase the accuracy of the value. 

There is an error with the smoothing though, instead of 
` int window = max_duration_usec;`
it should be something like:
`int` window = storage.samplerate / BLOCK_SIZE;`
It probably doesn't make a big difference though, so we can probably just fix it when we figure out how we want the final feature to work

mkruselj:
You're still getting asymmetric ramp up/ramp down because of the lowpass filter used which gives it thst characteristic. A linearly smoothed ring buffer would have symmetric ramp up/ramp down behavior I think, so would definitely be more appropriate for this task.

PR #6589, opened 2022-09-01
Add an option to toggle CPU meter

mkruselj:
Adjust the colors a bit (still fixed not skinnable) - 4 levels instead of 3 now
Clamped the CPU usage display to 100%.
Ordered the userdefaults enums to match the menu structure closer.

### DIFF
--- a/src/common/UserDefaults.cpp
+++ b/src/common/UserDefaults.cpp
@@ -61,6 +61,9 @@ std::string defaultKeyToString(DefaultKey k)
     case ShowGhostedLFOWaveReference:
         r = "showGhostedLFOWaveReference";
         break;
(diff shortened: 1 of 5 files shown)


baconpaul:
With this calculation method values > 100% are valid. What we are actually measuring is how much wall clock you take vs the maximum amount available for those samples at that sample rate. Just FYI. So > 100 we may just want to add a overflow meter.

mkruselj:
At least over here I get massive dropouts starting at 95% so I'm not really sure if any number above 100 really matters, though...?

baconpaul:
Ha yeah that's true. Above 100 means you are beyond-dropouts. True.

PR #6597, opened 2022-09-04
Mark scene LFO, integer and bool params as non-CLAP-polymodulatable

### LINKED ISSUE #6594: Clap plugin publishes wrong parameters as per note/id modulatable
**Bug Description:**
Parameters that should not be/won't work as per note/id modulatable, will show up as such.

**Surge XT Version**
1.1.1 nightly, Clap

**Reproduction Steps:**
1. In a host that supports per note modulation of parameters, you will see parameters like "A Scene LFO 1 Rate" or "Oscillator N octave" as such, but those either don't make sense (because they target scene level parameters) or don't work (because they are integer parameters which can't be modulated in Surge as of now).

**Expected Behavior:**
The Clap parameter flags should be set according to if the parameter actually supports per note/id modulation.

**Computer Information
 - OS: win7/mac
 - Host: Custom host
 - Version: 0.0.1

mkruselj:
Also remove two ctypes - one was completely unused (ct_bool_fm), the other is completely superfluous (ct_osccountWT, identical in usage to ct_osccount).

Closes #6594


baconpaul:
👍

PR #6608, opened 2022-09-11
Fix crash/assert with sge->effectChooser->currentClicked being -1

hfiguiere:
On Linux.

`sge->effectChooser->currentClicked` cause an assert on `cfxtype = sge->effectChooser->fxTypes[cfxid];` is it is `-1` (build with assertion on like it is the default for the flatpak). This on opening the standalone version or opening the UI on the LV2. (not sure about the VST3).


hfiguiere:
It didn't happen in 1.1.0

baconpaul:
Thanks!

I don't see why this doesn't happen with 1.1.0 but it is definitely wrong and a good guard. Appreciated.

PR #6616, opened 2022-09-19
docs: update references to Arch Linux port

grawlinson:
Hello! :wave:

Package maintainer for Arch Linux here, I've just added `surge-xt` to our repositories.

Hopefully this change to documentation is acceptable, let me know if it requires any further changes.

### DIFF
--- a/doc/Linux-and-other-Unix-like-distributions.md
+++ b/doc/Linux-and-other-Unix-like-distributions.md
@@ -66,13 +66,13 @@ Users can install it with the command: <br/>
 
 ## Surge on Arch Linux
 
-The Arch Linux port is available and maintained in the official repositories.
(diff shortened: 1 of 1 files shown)


mkruselj:
@grawlinson Thanks for the PR!

One thing I would perhaps add is link to the list of maintainers (if there is one), and definitely a direct link for the bug report issue template on your github?

haenkel:
Adding this link would be sufficient:
https://archlinux.org/packages/community/x86_64/surge-xt/

baconpaul:
Oh surge builds fine on arm64 also - just noticed the build instruction is x86 64. But I don’t understand arch so this comment may be useless

baconpaul:
Thanks so much!

PR #6626, opened 2022-09-24
Add an oscilloscope (spectrum analyzer).

mx:
Works as a tearout overlay. Accessed through the menu. Can display the spectrum of the left channel, right channel, or stereo (averaged).

Currently no waveform display.

Does require the most recent PR from sst-cpputils, not sure if that has to be merged before this PR can go in or not (or even be successfully built -- not sure how submodules work in a case like this).

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -20,6 +20,7 @@ Amy Furniss <http://github.com/amyfurniss>
 Cale Gibbard <cgibbard@gmail.com>
 Brian Ginsburg <https://github.com/bgins>
 Ben Goldwasser <https://github.com/sadguitarius>
+Morgan Kanter <https://github.com/mx>
 Nathan Kopp <nk.sg@nathankopp.com>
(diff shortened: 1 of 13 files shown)


baconpaul:
I merged SST cpputils so I think you need to repaint the submodule to main since it points at a transient commit

I don’t have time to review this am but would happpily merge and review later more extensively once the module is correct

baconpaul:
And looking at all the code except oscilloscope.cpp this is great. 
I skimmed oscilloscope.cpp briefly also so happy to merge but I do think the submodule needs a re push

mx:
Okay, just updated it to the latest HEAD in sst-cpputils.

baconpaul:
All the stuff outside the renderer looks great and the renderer on a quick skim looks probably correct but at least harmless :) 

so I’ll merge so you can start getting feedback from nightly users. We have time before a release anyway so easy to merge and change later 

thank you for joining in the fun on surge!

PR #6629, opened 2022-09-25
Additional minor Oscilloscope enhancements

mx:
This fixes the floor drawing issue (so it will always be anchored correctly) which was reported by several users. Also addresses other bits requested by users: keyboard shortcut to toggle, sizing. Also *might* help with the reported freezing issue on Mac, but I haven't been able to reproduce that so this is a guess.


baconpaul:
Look good! Will squash merge when ci is done

PR #6630, opened 2022-09-25
Tell cmake to export a compile_commands.json.

mx:
This lets people whose language server depends on its presence use it.

Useful for Emacs users like me.


baconpaul:
> This lets people whose language server depends on its presence use it.
> 
> Useful for Emacs users like me.

Do we need to add that file to the gitignore also?

mx:
> Do we need to add that file to the gitignore also?

It generates the compile_commands.json file in the build directory, so it should already be ignored.

baconpaul:
Oh duh Of course sorry

Fine to merge once ci finishes!

PR #6632, opened 2022-09-28
CPU efficiency improvement for Oscilloscope

mx:
This cuts down the CPU-hungry rendering of the Oscilloscope to be only slightly more than idle.


baconpaul:
Looks good to me!

PR #6651, opened 2022-10-14
Add a waveform display to the oscilloscope

mx:
Some limitations:

* It doesn't let you change parameters yet.
* Doesn't normalize the input data in case the gain is way up.
* Honestly, it just scrolls super fast since you view 100 ms at a time. I'm not sure how useful that is, really. I don't use waveform scope displays.

### DIFF
--- a/src/common/SurgeSynthesizer.cpp
+++ b/src/common/SurgeSynthesizer.cpp
@@ -4499,12 +4499,6 @@ void SurgeSynthesizer::process()
     accumulate_block(sceneout[1][0], output[0], BLOCK_SIZE_QUAD);
     accumulate_block(sceneout[1][1], output[1], BLOCK_SIZE_QUAD);
 
-    // Send output to the oscilloscope, if anyone is listening.
-    if (storage.audioOut.subscribed())
-    {
-        storage.audioOut.push(output[0], output[1], BLOCK_SIZE);
-    }
-
     bool sendused[4] = {false, false, false, false};
     // add send effects
     // TODO: FIX SCENE ASSUMPTION
@@ -4568,6 +4562,12 @@ void SurgeSynthesizer::process()
         break;
     case SurgeStorage::BYPASS_HARDCLIP:
         break;
+    }
+
+    // Send output to the oscilloscope, if anyone is listening.
+    if (storage.audioOut.subscribed())
+    {
+        storage.audioOut.push(output[0], output[1], BLOCK_SIZE);
     }
 
     // since the sceneout is now routable we also need to mute it
--- a/src/surge-xt/gui/overlays/Oscilloscope.cpp
+++ b/src/surge-xt/gui/overlays/Oscilloscope.cpp
@@ -36,6 +36,13 @@ float Oscilloscope::freqToX(float freq, int width)
     return xNorm * (float)width;
 }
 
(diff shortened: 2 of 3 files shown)


mx:
Also I haven't profiled this yet. At this point I just want to get the checkpoint into nightlies so people can try it and suggest any changes.

planist79:
Hi, i asked for an oscilloscope (waveform view) a few days ago on Discord, and Paul told me there was something in the making. When i read on Discord it is now implemented in the nightly i had to install it and i found the new oscillator in the options menu. 

Great! Thank you for this.
Regarding the waveform view. 
1. It reacts to sound quite slow, there is a latency of ~1 second between hearing the sound and its display in the waveform view window.
2. Also, it would be great if one could choose how many zero-crossings of the waveform is displayed. (so to see a simple saw wave i would choose 2 zero crossings. Maybe you can have a look at "Melda Oscilloscope" or "s(m)exoscope".. 
 
Thank you again!

mx:
@planist79 
Yeah, the latency of 1 second is unfortunately there to prevent missing any information. I can probably lower it to like...half a second. Making it instantaneous could be challenging though to ensure that the update rate stays exactly even.

Of course, if you just want to use a zero-crossing mode like you suggested, then none of that matters.

Is the idea with the zero-crossing thing that the scope stops once you get two, until you manually start it again? I don't actually use a waveform oscilloscope, so I don't even know what features people usually want with them. I only wanted the spectrum analyzer, haha.

mx:
Other people's PRs for functionality here are welcome btw. Like I said, I've never used a waveform scope before.

mx:
Also looks like S(m)exoscope is GPL, perhaps we're better off just importing and adapting it for the waveform display.

planist79:
You are right, waveform view is not necessarily helpful for the technical analysis of the sound you are creating - but as a visual feedback it is helpful... 

A soft synth is able to give much more visual feedback on the parameters and parameter changes (modulation, etc) than hardware synths can, just because of the big display. and this is one major advantage. 

Ideas:
1. implement the oscilloscope as "Effect" which can be placed somewhere in the fx signal chain 
2. an option to place the waveform/frequency spectrum window  permanently in the bottom right corner of the GUI (below the fx section).. IMO it would make the main GUI much more dynamic and visually pleasing.

mx:
@planist79 take a look at the latest nightly, I merged the s(m)exoscope code into the waveform view. It should behave identically.

planist79:
> @planist79 take a look at the latest nightly, I merged the s(m)exoscope code into the waveform view. It should behave identically.

Great, thank you. i get the nightly and get back to you.

planist79:
thank you. It works well! 
Some side notes
- there's still a noticeable latency, could it be lowered?00 
- could the window be resizeable?
- zero crossing option works quite well, even when changing the waveform the visualization is quite stable.

when I play around a bit more I come back to you...Thank you!!

mx:
> there's still a noticeable latency, could it be lowered?

There shouldn't be any latency at all. Do you have "sync" turned on? If so, turn it off.

> could the window be resizeable?

It's a tear-out window, so tear it out and resize.

planist79:
1. ok,
2. ok, thank you.

PR #6704, opened 2022-11-22
Merge in S(m)exoscope

mx:
This imports the s(m)exoscope code and modifies it to work with our own system and JUCE.

Only thing missing is the readout on click (you can see the old code for it #if'd out). Could use some UI love as well, as it looks pretty ugly right now. But as far as I can tell it's functional.

Parameters for the scope currently do not save. It would be kind of nice if they did.


mkruselj:
Requies clang-format to pass codequal CI build. Apart from that an another comment I left, it's looking good!

baconpaul:
The linux and macOS errors are because we have warnings set so a switch over enums requires either all cases or a default and Oscilloscope has missed one. That plus clang format and I'm more than happy to merge this and get to work on the UI and release requirements!

Issue #6343, opened 2022-07-11
FEG/AEG Analog mode attach is 2x faster than Digital mode


mkruselj:
This only happens for the attack stage - decay and release stages seem to have pretty much the same timings between the two modes.

Fix in XT2 since we're breaking compatibility - but also take care of imported patches.

mkruselj:
Closing this one since we decided XT 2.0 won't have analog mode envelope, it will have the better envelope from Shortcircuit, and MSEG.

PR #6706, opened 2022-11-23
Correct the Analog ADSR

baconpaul:
1: Add a correct (matches digital) implementaiton of ADSR
2: Make it available in a switch, off by default in surge proper


mkruselj:
Linked to #6343

PR #6717, opened 2022-11-30
Add option to focus modulator editor on Add Modulation From actions

### LINKED ISSUE #6701: LFO Selection when Adding via Menu
Chessel — Today at 5:10 PM

Hi all.  I have a suggestion to aid usability for keyboard users - the typing keyboard that is.  If I set up a parameter with modulation, could the LFO controls region switch to showing that modulator?  So if I add a modulator to my osc 1 pitch using voice 1, if I then alt full stop to the LFO controls region, it should be on the voice 1 controls.  If I then added a filter, and modulation of the cut off frequency with voice 2, and I go to the LFO controls, it would have shifted to the voice 2 controls.  So whenever I set up a new modulation the focus of the LFO controls region shifts to show that modulator.  This would save having to go to the modulator region to select the LFO.

baconpaul — Today at 5:15 PM

oh so like "add modulation from lfo1" in the menu selects lfo1?
that's a bit of a pain for the signted ui since that menu is meant to allow you to add modulators from "away" sources while looking at "current" sources. but i see why you want it.

Chessel — Today at 5:16 PM

Yes. Exactly.

baconpaul — Today at 5:17 PM

OK let me drop it in a github issue

mkruselj:
Closes #6701

### DIFF
--- a/src/common/UserDefaults.cpp
+++ b/src/common/UserDefaults.cpp
@@ -278,6 +278,10 @@ std::string defaultKeyToString(DefaultKey k)
         r = "expandModMenusWithSubmenus";
         break;
 
+    case FocusModEditorAfterAddModulationFrom:
+        r = "focusModEditorAfterAddModulationFrom";
+        break;
+
     case nKeys:
         break;
     }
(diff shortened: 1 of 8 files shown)


baconpaul:
looks good to me on a quick scan! Awesome

PR #6721, opened 2022-12-03
Fix build with mingw

falkTX:
The function `_BitScanReverse` also exists in mingw, and was causing a build failure for surge-xt in Cardinal.

### DIFF
--- a/src/common/dsp/oscillators/WindowOscillator.cpp
+++ b/src/common/dsp/oscillators/WindowOscillator.cpp
@@ -19,7 +19,7 @@
 #include <cstdint>
 #include "DebugHelpers.h"
 
-#ifdef _MSC_VER
+#ifdef _WIN32
 #include <intrin.h>
 #else
 namespace


baconpaul:
Hmm the rack toolchain builds with gcc12 in a cross compile but I think it defines win32. Does this diff work there? I think you want mscver || mingw for this condition but I don’t know what the mingw condition is

falkTX:
no need for mscv || mingw, the `_WIN32` is **always** defined on Windows.

for mingw headers we have `__MINGW32__` (always on for i686 and x64) plus `__MINGW64__` if 64bit.
but again we shouldnt need those.

baconpaul:
> no need for mscv || mingw, the `_WIN32` is **always** defined on Windows.
> 
> for mingw headers we have `__MINGW32__` (always on for i686 and x64) plus `__MINGW64__` if 64bit. but again we shouldnt need those.

Right but that’s my point. If we are win32 and not msvc then we want the alternate branch. Won’t hurt diff break that? 

Did you test this in the rack toolchain to see if it still works?

baconpaul:
So you submit a change I’m not sure if it’s right, I push back in review, and you refuse to test it because that’s good open source behavior? Come on man don’t be like that. 

Yes we have ci on surge rack but for thst we need to merge this, the. Merge the submodule up there and test if it breaks. If it does we need to undo all those merges 

or we can let this sit until I boot up a Linux box and test it in the rack env which will be post weekend

but I’m not sure the change is right (although I don’t know why this builds on rack toolchain either) so someone needs to test it before I merge it right?

falkTX:
> So you submit a change I’m not sure if it’s right, I push back in review, and you refuse to test it because that’s good open source behavior? Come on man don’t be like that.

sorry I meant I have not used the rack toolchain stuff because of those reasons. not that I refuse to test my own changes here.



> Yes we have ci on surge rack but for thst we need to merge this, the. Merge the submodule up there and test if it breaks. If it does we need to undo all those merges

understood. I thought the CI on this repo would already detect such changes.
I will see about the rack toolchain build, thought may take a bit as I have no experience with it.

baconpaul:
I can also look just not right now. I only have mac hw with me 

The ci for surge tests surge vst but it doesn’t rekick surge rack. 

But luckily surge rack is easy to test if you have a Linux box with docker. There’s a script in scripts called “drun.sh” which mounts your surge instance and starts a toolchain up. So it really is just 

1 one of us get a running Linux box with docker
2. Checkout surge rack
3. Apply this diff to the submodule 
4. Run thst script and build plugin Linux build or whatever the target is
5. If it works merge and if it doesn’t change it to be MSVC or mingw until it does 

not hard! Just step 1 isn’t physically possible for me

baconpaul:
An alternate approach would be to add a ci phase here which does a docker build. Certainly not doing that today but it would also do the necessary check. Basically check out surge rack at head repoint the submodule to the pr branch etc …. Some ugly scripting and probably not worth it since this is really only a problem for next 4 days as you get stuff running and we figure out if it is right

falkTX:
For reference, this is the build issue I have locally:

```
/home/runner/work/Cardinal/Cardinal/plugins/surgext/surge/src/common/dsp/oscillators/WindowOscillator.cpp:254:44: error: call of overloaded ‘_BitScanReverse(long unsigned int*, unsigned int&)’ is ambiguous
  254 |             if (_BitScanReverse(&MSBpos, bs))
      |                                            ^
In file included from /usr/share/mingw-w64/include/winnt.h:27,
                 from /usr/share/mingw-w64/include/minwindef.h:163,
                 from /usr/share/mingw-w64/include/synchapi.h:10,
                 from /home/runner/work/Cardinal/Cardinal/include/mingw-std-threads/mingw.mutex.h:57,
                 from /home/runner/work/Cardinal/Cardinal/include/mingw-compat/mutex:20,
                 from /home/runner/work/Cardinal/Cardinal/plugins/surgext/surge/src/common/SurgeStorage.h:28,
                 from /home/runner/work/Cardinal/Cardinal/plugins/surgext/surge/src/common/dsp/oscillators/OscillatorBase.h:18,
                 from /home/runner/work/Cardinal/Cardinal/plugins/surgext/surge/src/common/dsp/oscillators/WindowOscillator.h:18,
                 from /home/runner/work/Cardinal/Cardinal/plugins/surgext/surge/src/common/dsp/oscillators/WindowOscillator.cpp:16:
/usr/share/mingw-w64/include/psdk_inc/intrin-impl.h:1698:1: note: candidate: ‘unsigned char _BitScanReverse(long unsigned int*, long unsigned int)’
 1698 | __buildbitscan(_BitScanReverse, unsigned __LONG32, "bsr{l %[Mask],%[Index] | %[Index],%[Mask]}")
      | ^~~~~~~~~~~~~~
/home/runner/work/Cardinal/Cardinal/plugins/surgext/surge/src/common/dsp/oscillators/WindowOscillator.cpp:27:13: note: candidate: ‘bool {anonymous}::_BitScanReverse(long unsigned int*, long unsigned int)’
   27 | inline bool _BitScanReverse(unsigned long *result, unsigned long bits)
      |             ^~~~~~~~~~~~~~~
```

baconpaul:
yeah so the rack toolchain doesn't have that problem but it also doesn't uses a compile environment which it hand stacks in a docker image. 

You can proceed with your port with this patched without me merging it right? I just can't get to a toolchain to test it today. I'm sure you will have other changes (like the globals.h one for riscv I assume is coming) anyway.

falkTX:
yeah, I have quite a few changes needed, a bit all over the place.
the reason for the PRs is so any patches go to upstream, reducing maintenance burden.

one blocker issue is the sine function defined in https://github.com/surge-synthesizer/surge/blob/main/src/common/dsp/vembertech/basic_dsp.cpp#L265
from the macros around it, it is not supposed to be used if running x64 or arm, which were the only 2 supported archs anyway.
can you remove that function completely? the intrinsics contained in it causes build issues and seems to me it is not needed at all (I did an #if 0 and noticed no side effects)

baconpaul:
> yeah, I have quite a few changes needed, a bit all over the place. the reason for the PRs is so any patches go to upstream, reducing maintenance burden.
> 
> one blocker issue is the sine function defined in https://github.com/surge-synthesizer/surge/blob/main/src/common/dsp/vembertech/basic_dsp.cpp#L265 from the macros around it, it is not supposed to be used if running x64 or arm, which were the only 2 supported archs anyway. can you remove that function completely? the intrinsics contained in it causes build issues and seems to me it is not needed at all (I did an #if 0 and noticed no side effects)

Oh yeah please upstream everything. It makes life easier for all of us - but I'm also happy with a smaller number of PRs rather than single line ones if you prefer to work that way. 

ha yeah that sine implementation is from the old pre-SSE2 code which used to be in the code base and I just never removed it. That can definitely go. I thought I got all of that 3 years ago!

baconpaul:
(one item on our todo list is to just clean up all that ancient dsp code and move it a to a single submodule shared between shortcircuit and surge but that's a ways off. I did some when I did variable block size but not all of it).

baconpaul:
Anyway I have to drop off for chunk of the day now. Why don't you just use this PR as the one for "all the changes you need" and keep it going and then we can at least check the surge CI. I can find time to check the rack toolchain with the diffs - but probably not until monday -  if you don't get that setup but then we at least know the list of diffs we are considering.

falkTX:
It is tricky to push everything because I need changes in JUCE too, should I send those to your fork as well?

preferably we would have surge in a way that allows to build against a custom version of JUCE, that way I can reuse the same stuff for internal use and the surge version, with a matching config.

similarly to libsamplerate, there is a conflict because Rack uses it and surge wants that too. so that too would preferably/possibly use "system libs" instead of a vendored copy.

anyhow, I prefer to do things in small PRs, so it is also easier to review.
will send one for your JUCE copy first

baconpaul:
Surge can build against a different version of juce today so you can just do that with the same flags we need for the lv2 in the surge rack makefile 

you can do the same with libsamplerate if you want sure just add a redirect cmake option and set it in the rack dep phase

falkTX:
> Surge can build against a different version of juce today so you can just do that with the same flags we need for the lv2 in the surge rack makefile
> 
> you can do the same with libsamplerate if you want sure just add a redirect cmake option and set it in the rack dep phase

can you explain in more detail? I didnt see any flags for this.
the common cmake file has this directly:

```
surge_add_lib_subdirectory(libsamplerate)
```
with no option around it to skip it.

baconpaul:
I meant there is an option for juce so there’s no need to modify our juce submodule 

You could add an identical option for samplerate by modifying the cmake list if you want

baconpaul:
And I would merge that option off by default 

basically the “devendor” strategy is to add cmake flags to change those includes. We’ve done it for juce and would merge it for other things too

baconpaul:
hmm so the rack toolchain defined _WIN32 but not _MSC_VER so this is a change on the toolchain

I was able to do a compile with it changed to _WIN32 but I need to boot into windows to see if it loads.

Maybe I'll just look at this code and see if there's a way to rewrite the function in WindowOscillator. Honestly I'm not sure why we are using a non-portable scan at this point anyway.

baconpaul:
Anyway confirmed it works when i do a toolchain build with this diff and can load on windows

I think (think) that this is a mingw version thing since you are showing signature differences not name clashes. But it doesn't matter since the _WIN32 works in both ends.

So merging this. I'll update surge-rack to have latest submodule now and turn off MTS also.

falkTX:
> hmm so the rack toolchain defined _WIN32 but not _MSC_VER so this is a change on the toolchain

this is expected if Rack toolchain uses mingw instead of MSVC.
juce_core has bit related functions, is this one not on there too?

anyhow glad to see things merged.

baconpaul:
Oh maybe. I don't want to depend on jute-core for the oscillators but I'll open a separate surge issue for that.

baconpaul:
but anyway: rack toolchain worked before and works after. Surge builds worked before and works after. Your builds didn't work before and do work after. So happy to merge now I've tested all those facts :)

PR #6722, opened 2022-12-03
Allow building for RISC-V and Web-Assembly

falkTX:
Using simde we can enable RISC-V builds.
Depends on https://github.com/surge-synthesizer/sst-plugininfra/pull/37

### DIFF
--- a/src/common/dsp/vembertech/basic_dsp.cpp
+++ b/src/common/dsp/vembertech/basic_dsp.cpp
@@ -262,24 +262,6 @@ float sine_ss(unsigned int x) // using 24-bit range as [0..2PI] input
 
     return P * (y * abs(y) - y) + y;   // Q * y + P * y * abs(y)   */
 }
-#if !_M_X64 && !defined(ARM_NEON)
-__m64 sine(__m64 x)
-{
-    __m64 xabs = _mm_xor_si64(x, _mm_srai_pi16(x, 15));
-    __m64 y = _mm_subs_pi16(_mm_srai_pi16(x, 1), _mm_mulhi_pi16(x, xabs));
-    y = _mm_slli_pi16(y, 2);
-    y = _mm_adds_pi16(y, y);
-    const __m64 Q = _mm_set1_pi16(0x6333);
-    const __m64 P = _mm_set1_pi16(0x1CCD);
-
-    __m64 yabs = _mm_xor_si64(y, _mm_srai_pi16(y, 15));
-    __m64 y1 = _mm_mulhi_pi16(Q, y);
-    __m64 y2 = _mm_mulhi_pi16(P, _mm_slli_pi16(_mm_mulhi_pi16(y, yabs), 1));
-
-    y = _mm_add_pi16(y1, y2);
-    return _mm_adds_pi16(y, y);
-}
-#endif
 
 int sine(int x) // 16-bit sine
 {
--- a/src/common/CMakeLists.txt
+++ b/src/common/CMakeLists.txt
@@ -56,12 +56,12 @@ else()
           ../../libs/JUCE/modules/juce_dsp/juce_dsp.cpp
           )
   if (APPLE)
-    target_sources(juce_dsp_rack_sub PUBLIC
+    target_sources(juce_dsp_rack_sub PRIVATE
             ../../libs/JUCE/modules/juce_core/juce_core.mm
             ../../libs/JUCE/modules/juce_audio_formats/juce_audio_formats.mm
             )
   else()
-    target_sources(juce_dsp_rack_sub PUBLIC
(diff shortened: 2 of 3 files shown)


baconpaul:
Since we squash merge that plugininfra hash will change right.? So this will have a bad submodule.

the change is fine once we understand if the plugininfra change is fine but will probably have to change the pr once the submodule is updated

falkTX:
I can push without the submodule update if that helps

baconpaul:
Nah the yull just need to push again, leave it here and when we sort out the submodule just relish this with the right hash. Was just letting you know the change is fine

baconpaul:
Anyway if you want to update this with the correct head has for sst-plugin-infra i'm happy to merge it.

falkTX:
This should be it, I have surge running on the browser now :)

JUCE stuff and mingw fix is not pushed here, as those are a bit separate things

baconpaul:
Interesting! Does it work ok?

falkTX:
> Interesting! Does it work ok?

Let me figure out the resources first, and I will let you experiment..

src/common/dsp/vembertech/basic_dsp.cpp
```diff
@@ -262,7 +262,8 @@ float sine_ss(unsigned int x) // using 24-bit range as [0..2PI] input
 
     return P * (y * abs(y) - y) + y;   // Q * y + P * y * abs(y)   */
 }
-#if !_M_X64 && !defined(ARM_NEON)
+
+#if 0
```
baconpaul, line 266: Do you mind just deleting this entire block rather than if 0ing it?

baconpaul (commented):
All good other than the if 0 vs remove. Change that and I’m happy to merge!

src/common/dsp/vembertech/basic_dsp.cpp
falkTX: done

baconpaul: Great
When ci finishes I will merge
Thanks for upstreaming

falkTX:
Just leaving a tease...

image

will continue in the evening, need to handle a few other things for now.
still dont have resources in place properly, or rather the path to them.

baconpaul:
The resources only matter for fx presets and the wavetable and window oscillator 

wavetable and window might crash without them but classic and the filters should work

falkTX:
I figured it out, after lots of prints and trying to figure out wtf was going on: emscripten doesnt follow directory symlinks when packing resources, only file symlinks.

With that working I enabled the full packing, but ended up with 320Mb of resources! way too much..
Looking around it seems most of it comes from the 3rd party wavetables, which take 141Mb.
Can you compress those files with flac? Wavetables are short, so shouldnt take long to decode, and would save near 50% of the space.

Sadly even without the 3rd party wavetables the total resource size still ends up being large, thus browsers wont cache it.
With brotli compression and Firefox, we are just on the limit (50.7Mb > Firefox default 50Mb limit).

Anyhow, SurgeXT on Cardinal is now Live! Feel free to try it at https://cardinal.kx.studio/

baconpaul:
You can just skip the wavetables if you want - writing code to compress them is pretty painful since they are a custom format 

I will probably move to a separate download model in 2.1 since both you and rack seem to flinch at the size 

bur for now just package like basic/generated or something and add instructions to your build notes if the size is a problem.

mkruselj:
I would suggest packing the factory wavetables folder, and I don't know from 3rd party wavetables folder maybe include Venus Theory and TNMG ones. Amounts to 17 MB or so - not too terrible?

falkTX:
It is already above the maximum browser cache size, so I went with "factory" wavetables only.
For reducing the size I have been thinking of creating a font pool where modules retrieve fonts from, instead of the current way where each module has their own copy of fonts they use.

baconpaul:
Yeah I think I’m going to add a 2.1 fewture which can download the extra tables from our GitHub and place them in the user working dir we use for other settings - that should strip our assets down to a couple of wavetables we need to get start led and some svg

falkTX:
Sounds fine for general use.
Do note I have a strict rule for Cardinal where internet access is forbidden, really dont want to deal with libcurl, openssl, certificates and all that stuff... plus phone home, analytics yada yada. The lack of online connection blocks that from the start.

I think the current base wavetables are fine, the 3rd party ones I agree would be best stored elsewhere.

baconpaul:
Yup we’ll presumably whatever I write to download the wavetable programmatically will have to be done manually instead by your users but since I haven’t written it yet it’s not a problem today

falkTX:
Any blocker for merging this? I have a few other changes pending, but those will be on different PRs.

baconpaul:
Merged

PR #6723, opened 2022-12-04
Disable unwanted/not-needed sqlite features

falkTX:
The built-in sqlite has all default features enabled, we can remove some that are unused.
Specially important is the "load extension" which uses dlopen/dlsym/dlclose/etc, these are superfulous and introduce extra library links (libdl.so on Linux) without being needed.

The other ones I picked based on a quick check through the code, these seem to me the more obvious ones to take out.


baconpaul:
Thanks

the correct test for this is

build surge standalone 
Remove your patch db file in your surge directory
Start surge standalone on a system with factory patches installed 
Rescan by pressing the magnifying glass or ctrl f
Make sure typeahead search still works 

can you confirm that passes in addition to the ci tests (which also exercise SQLite)

thanks

baconpaul:
OK did a built on macOS and ran above test and it worked fine. All our cross-platform tests also passed which means we can do basic SQL on linux win and mac. Thanks for the change; will merge now

PR #6724, opened 2022-12-04
Allow to skip oddsound-mts usage

falkTX:
From what I could understand, the oddsound-mts library is only partly open-source.
All the code from https://github.com/ODDSound/MTS-ESP references a system library that must be loaded at runtime.
There is no public code for libMTS.so

I even suspect this is a GPL violation, the MTS client code is directly calling a proprietary library and Surge is using MTS client code. Would need a deeper investigation..

For now I would appreciate a way to disable this feature, I was surprised to find there is a reliance on an external library from a plugin (even if optional)
There is likely more stuff needed, but my knowledge of cmake is very lacking so I just did the code.

### DIFF
--- a/src/common/SurgeStorage.cpp
+++ b/src/common/SurgeStorage.cpp
@@ -504,6 +504,7 @@ SurgeStorage::SurgeStorage(std::string suppliedDataPath) : otherscene_clients(0)
         getPatch().scene[s].drift.set_extend_range(true);
     }
 
+#ifndef SURGE_SKIP_ODDSOUND_MTS
     bool mtsMode = Surge::Storage::getUserDefaultValue(this, Surge::Storage::UseODDMTS, false);
     if (mtsMode)
     {
@@ -514,6 +515,7 @@ SurgeStorage::SurgeStorage(std::string suppliedDataPath) : otherscene_clients(0)
         oddsound_mts_client = nullptr;
         oddsound_mts_active = false;
     }
+#endif
 
     initPatchName =
         Surge::Storage::getUserDefaultValue(this, Surge::Storage::InitialPatchName, "Init Saw");
@@ -1952,7 +1954,12 @@ void SurgeStorage::load_midi_controllers()
     }
 }
 
-SurgeStorage::~SurgeStorage() { deinitialize_oddsound(); }
+SurgeStorage::~SurgeStorage()
+{
+#ifndef SURGE_SKIP_ODDSOUND_MTS
+    deinitialize_oddsound();
+#endif
+}
 
 double shafted_tanh(double x) { return (exp(x) - exp(-x * 1.2)) / (exp(x) + exp(-x)); }
 
@@ -2378,6 +2385,7 @@ void SurgeStorage::storeMidiMappingToName(std::string name)
     }
 }
 
+#ifndef SURGE_SKIP_ODDSOUND_MTS
 void SurgeStorage::initialize_oddsound()
 {
     if (oddsound_mts_client)
@@ -2415,6 +2423,7 @@ void SurgeStorage::setOddsoundMTSActiveTo(bool b)
         tuningApplicationMode = patchStoredTuningApplicationMode;
     }
 }
+#endif
 
 void SurgeStorage::toggleTuningToCache()
 {
--- a/src/common/SurgeStorage.h
+++ b/src/common/SurgeStorage.h
@@ -1368,11 +1368,13 @@ class alignas(16) SurgeStorage
 
     void setTuningApplicationMode(const TuningApplicationMode m);
 
+#ifndef SURGE_SKIP_ODDSOUND_MTS
     void initialize_oddsound();
     void deinitialize_oddsound();
+    void setOddsoundMTSActiveTo(bool b);
+#endif
     MTSClient *oddsound_mts_client = nullptr;
     std::atomic<bool> oddsound_mts_active{false};
-    void setOddsoundMTSActiveTo(bool b);
     uint32_t oddsound_mts_on_check = 0;
     enum OddsoundRetuneMode
     {
--- a/src/common/SurgeSynthesizer.cpp
+++ b/src/common/SurgeSynthesizer.cpp
@@ -372,6 +372,7 @@ void SurgeSynthesizer::playNote(char channel, char key, char velocity, char detu
         return;
     }
 
+#ifndef SURGE_SKIP_ODDSOUND_MTS
     if (storage.oddsound_mts_client && storage.oddsound_mts_active)
     {
         if (MTS_ShouldFilterNote(storage.oddsound_mts_client, key, channel))
@@ -379,6 +380,7 @@ void SurgeSynthesizer::playNote(char channel, char key, char velocity, char detu
             return;
         }
     }
+#endif
 
     if (learn_param_from_note >= 0 &&
         storage.getPatch().param_ptr[learn_param_from_note]->ctrltype == ct_midikey_or_channel)
@@ -4090,6 +4092,7 @@ void SurgeSynthesizer::processControl()
         if (fx[i])
             refresh_editor |= fx[i]->checkHasInvalidatedUI();
 
+#ifndef SURGE_SKIP_ODDSOUND_MTS
     if (storage.oddsound_mts_client)
     {
         storage.oddsound_mts_on_check = (storage.oddsound_mts_on_check + 1) & (1024 - 1);
@@ -4104,6 +4107,7 @@ void SurgeSynthesizer::processControl()
             }
         }
     }
+#endif
 }
 
 void SurgeSynthesizer::process()
--- a/src/common/dsp/SurgeVoice.cpp
+++ b/src/common/dsp/SurgeVoice.cpp
@@ -48,6 +48,7 @@ float SurgeVoiceState::getPitch(SurgeStorage *storage)
     */
     auto res = key + /* mainChannelState->pitchBendInSemitones + */ mpeBend + detune;
 
+#ifndef SURGE_SKIP_ODDSOUND_MTS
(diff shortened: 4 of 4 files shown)


falkTX:
The comments on https://github.com/ODDSound/MTS-ESP/issues/6 confirm my suspicion that libMTS is **NOT** opensource.
So yeah, integration with it is unwelcome, at least for projects I maintain and package.

Not to mention they only publish the very basic of builds anyway (Linux/mac/win x64, no ARM, no RISCV, no BSD, no WASM...)

mkruselj:
Please run clang-format on modified files. Thanks!

baconpaul:
Also mts is open source code that works either if a dll is present or not. I have no concern with it being a gpl violation but if you do I cannot merge your changes. It is a critically important feature for us and our users.

so can you either confirm that you are happy to contribute diffs to surge knowing it uses mts and will be in the future, or if you aren’t, I can back out all your diffs 

thanks

baconpaul:
I’m happy to merge these ifdefs though if that’s amenable to you of course once you have the pr passing ci

falkTX:
> Also mts is open source code that works either if a dll is present or not

from the surge side, yes. but MTS library code is not open, that is the big important distinction.
it is akin to elastique, just with the public API being in a more permissive license.



> I have no concern with it being a gpl violation but if you do I cannot merge your changes.

Whether it is a GPL violation or not is a serious concern, not to me but to everyone as a whole.
And there is nothing to be diplomatic here, GPL purposefully and intentionally does not allow linkage against proprietary libraries.

I will fix up the code stuff, seems clang-format wants the added macros to be placed in a very specific way.

baconpaul:
I can’t risk your view leading to you making an assertion that we can’t distribute your code with mts and if you have diffs in the code base then you could make such a claim. So I can just back out your diffs 

Is that what you want?

falkTX:
> I can’t risk your view leading to you making an assertion that we can’t distribute your code with mts and if you have diffs in the code base then you could make such a claim. So I can just back out your diffs
> 
> Is that what you want?

my code or anyone else's code is the same.. it is all in a GPL project, so I dont quite get your distinction here.

anyone can make a claim that using libmts within a GPL project is not allowed by the terms of the GPL license.

do you distribute surge binaries in any license other than GPL?

falkTX:
> You know what I’m not interested in arguing about this. The libmts client works perfectly well without the associated dynamic library and doesn’t require linkage of the library in any way to function.
> 
> I’ll back out all your diffs and you can just maintain a fork.

I know this is boring and annoying, but please do take this more seriously.
We can't have the benefits of GPL only when it is convenient to us, and ignore it when it is not.
For packaging and distribution, any possible GPL violation is taken very seriously. It is the reason Debian and Fedora are refusing to package VST3 SDK, due to their ambiguous license requirements. And many other similar cases.

You can say the user is the one that loses out in the end, which is true, but this happens all over the software industry.
It is on libMTS authors to license things properly, and have an opensource version so that opensource projects can properly make use of it.

For my own and Cardinal case, I am happy with simply disabling the feature.

But it is quite sad to see this approach on a GPL licensed project to be quite honest.

baconpaul:
I take it very seriously 

I just think you are wrong about this being a gpl violation 

And if you have diffs in the code you have standing to make me have to argue that with you 

so either 

1: assert that we disagree on this and you won’t bring it up again or 
2: I’ll back out your diffs so our disagreement here doesnt require action

mkruselj:
If the DLL doesnt HAVE to be linked then there is no problem, is what Paul is saying.

baconpaul:
Well I hate appeal to authority 

my point is more that the entire client library is open and functions properly whether the library is there or not. If you have the dll you get the mts connectivity but if not you get 12tet. Moreover surge can implement the tuning without mts at all 

I just don’t see how this is a gpl violation in any reading and understand gpl well 

but I don’t want to spend time arguing about it either since folks may disagree. Xkcd 386 etc 

but I don’t want contributors to surge who disagree to feel badly about their contributions and subsequent distribution 

so that’s kinda where I am

falkTX:
I have seen this come up a few times, where an opensource GPL application was making use of proprietary bits.
When I was involved in Blender things, there was quite a lot of effort to try making proprietary rendering plugins with all sorts of craziness in an effort to bypass GPL.

And a quick search online also seems to indicate it is problematic.
https://opensource.stackexchange.com/questions/9142/gpl-application-linked-to-a-lgpl-library-loading-proprietary-plugins
https://opensource.stackexchange.com/questions/7882/can-i-use-proprietary-library-in-a-gpl-compatible-library-loaded-on-a-gpl-progra

To be clear I also dont want to argue, Surge is certainly an edge case where there is some functionality integrated with an external lib but it is all optional.
On a personal (moral, ethical) view I hate the idea of loading a proprietary library for functionality within an opensource project. But you obviously think different.

As I said, for my own and Cardinal case, I am happy with simply disabling the feature.
If/when I package surge-xt on the KXStudio repos, I will disable the feature there as well.

mkruselj:
And you would be cutting off one of the best Surge features. Heh.

falkTX:
> And you would be cutting off one of the best Surge features. Heh.

Only because libMTS is proprietary. This wouldnt be a problem if the library was properly licensed.

mkruselj:
But there is no need to distribute that lib with Surge itself! This is the other side, that Surge is listening to. It's what you get when you purchase MTS-ESP client. It is not a requirement.

falkTX:
I dont want proprietary things on my system or the ones I contribute if I can avoid it, thanks very much.

mkruselj:
So you don't need to get them then. Some KX Studio users might think differently and you'd be cutting them off just to spite your face. OK.

baconpaul:
Right so I understand both arguments @falkTX inreally do. And I’ve talked with the oddsound team about this.  I’ve concluded that this isn’t a GPL violation but I agree it is a bummer. 

Loook what I want to avoid is: I merge your change, I distribute a binary, then you show up and say “don’t distribute my diff because it has my diffs and has the MTS client and so I think it is a GPL violation”.  If you are going to do that I should just back out your stuff now. If you aren’t, knowing that we will distribute binaries with your diffs and with the MTS client code (but not closed source library), then I’m happy to merge

falkTX:
We cant predict the future, maybe libMTS expands and gets more functionality later and Surge integrates that too.
The situation is already uncomfortable as it is.

I dont have the intention of pursuing possible GPL violations when it is unclear if that is really the case or not.
But I also do not want to waive and say I wont ever talk or mention this again.

So if you feel more at ease taking off my contributions, please do so.

baconpaul:
So for total clarity

@falkTX if I merge this or any of your other changes I will use them to make a binary using only the source here. At runtime in some configurations some parts of the source here dynamically load a library which is not included here and that library is not open source. If you are ok with that I’m happy to keep merging your stuff (and would prefer to, since they are useful changes) but if that’s not I also understand but then I shouldn’t merge your changes

mkruselj:
I would also be really happy with these changes integrated, to be honest.

baconpaul:
Yeah me too

baconpaul:
Sigh but seeing falks response I don’t think I can merge his changes.

falkTX:
Same. but I do not want to close the conversation completely.
It is weekend, and we are spending our free time here, it is a waste to cause headaches now.

I truly wish the libMTS stuff would be open, really. Right now there is only a few build targets it can ever work with, so that functionality is limited by what they deem important. Always the same deal with proprietary things...
And it is not like that lib is doing magical things, so it is difficult to understand why it would remain proprietary.
Have they shown any interest on opensource the libMTS stuff?

Anyway, sorry I dont have a proper answer here.
Yes I dont like this libtMTS approach, and yes I go out of my way to ensure it is not used (because it is proprietary).
Is it a problem? depends on the POV.. a GPL violation? more for lawyers to say.

I mostly want to keep conversation about the topic of using proprietary bits open, and not outright dismiss it as "it is okay, just a small thing, ignore it" kinda deal.

baconpaul:
Ok so here’s a solution then @falkTX 

1. we agree here thst we disagree but we might both be wrong and lawyers haven’t sorted it out and we don’t want surge to be the test case 
2. I close your contribution here but then functionally re implement it including the correct cmake hooks and merge thst 
3. I continue to merge your changes elsewhere since they are good
4. You can make binaries you beleive are GPL compliant as can I but mine will have MTS client code and yours won’t 
5. We can bring this issue up on the oddsound issue tracker talking about a collection of tools not just surge (since other open source projects also include MTS)
6. We talk about it but no one hires lawyers 

thst good with you?

mkruselj:
By all means do raise your concerns with us as they come. We don't mind conversations, we encourage them.

I hope we can reach agreement and mutual middle ground to prevent forking on your side. Your changes are beneficial overall and would also help other situations that might arise too, it seems to me.

baconpaul:
Right let me take a lead on thst. I have good relations with the devs there and can try it privately first

baconpaul:
Oh and while i'm at it i can test your other change too with the sqlite so you don't have to do that rebuild.

falkTX:
> Oh and while i'm at it i can test your other change too with the sqlite so you don't have to do that rebuild.

That would be very welcome! I am yet to build surge-xt outside of rack/cardinal

falkTX:
slightly offtopic, finally cardinal+surgext builds are all starting to pass https://github.com/DISTRHO/Cardinal/actions/runs/3613397831

baconpaul:
and i actually can get to my linux box later this morning to check the open MINGW/MSVC/WIN32 one with BitReverse

baconpaul:
Oh ha, @falkTX  your diff actually still included all the MTS code also :)

Once i made the cmake changes so it wasn't even in the include path a bunch of other stuff broke. Chuckle.

Will have a diff together shortly.

PR #6725, opened 2022-12-04
Allow a SURGE build without any libMTS

baconpaul:
Adding the cmake option -DSURGE_SKIP_ODDSOUND_MTS=TRUE means that surge will not consume an libMTS code and will not present users the MTS options. The default build settings include MTS so if you want an MTS free surge, you need to actively set the option.

See the discussion in #6724

### DIFF
--- a/src/common/CMakeLists.txt
+++ b/src/common/CMakeLists.txt
@@ -4,7 +4,16 @@ project(surge-common)
 surge_add_lib_subdirectory(airwindows)
 surge_add_lib_subdirectory(eurorack)
 surge_add_lib_subdirectory(fmt)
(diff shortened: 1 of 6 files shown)


baconpaul:
@falkTX FYI

falkTX:
That is very appreciated.
Doing a test build at https://github.com/DISTRHO/Cardinal/actions/runs/3613879063

baconpaul:
I've also reached out to my collaborators at odd sound to talk about the underlying issue.

baconpaul:
Oh and I'm also going to turn this option on for surge-rack; surge-rack doesn't need the internal MTS at all.

PR #6728, opened 2022-12-04
Fix C string usage leading to buffer overflow in LTO builds

falkTX:
See https://github.com/surge-synthesizer/surge-rack/issues/690 for the discussion around it.
Only saw crashes happening with Cardinal using surge-xt modules built with LTO enabled, might be LTO uses a custom `snprintf` ?
Not exactly sure, but updated code looks good (and better/proper) to me.

### DIFF
--- a/src/common/Parameter.cpp
+++ b/src/common/Parameter.cpp
@@ -45,7 +45,7 @@ void get_prefix(char *txt, ControlGroup ctrlgroup, int ctrlgroup_entry, int scen
 {
 #define PREFIX_SIZE 16
 
-    char prefix[PREFIX_SIZE];
+    char prefix[PREFIX_SIZE + 1] = {};
 
     switch (ctrlgroup)
     {
@@ -82,7 +82,7 @@ void get_prefix(char *txt, ControlGroup ctrlgroup, int ctrlgroup_entry, int scen
 void Parameter::create_fullname(const char *dn, char *fn, ControlGroup ctrlgroup,
                                 int ctrlgroup_entry, const char *lfoPrefixOverride) const
 {
-    char prefix[PREFIX_SIZE];
+    char prefix[PREFIX_SIZE + 1] = {};
     bool useprefix = true;
     switch (ctrlgroup)
     {
@@ -186,7 +186,7 @@ Parameter *Parameter::assign(ParameterIDCounter::promise_t idp, int pid, const c
 
     strxcpy(this->name, name, NAMECHARS);
     set_name(dispname);
-    char prefix[PREFIX_SIZE];
+    char prefix[TXT_SIZE + 1] = {};
     get_prefix(prefix, ctrlgroup, ctrlgroup_entry, scene);
     snprintf(name_storage, NAMECHARS, "%s%s", prefix, name);
     posy_offset = 0;


falkTX:
> Ahh failed the clang format check again. Probably need. Space but just run clf and it should work
> 
> I would let the ci run to completion elsewhere then push the format change and I can merge it without a second run

too late! fix already pushed

baconpaul:
Ok! It’s just an extra 10 minutes

thanks for the find

baconpaul:
Oh I’m on my phone also but will merge this when ci passes.  Mind doing the submodule update to parent also so we can get it closed after that? If not I’ll be able to later tonight Ny time

falkTX:
> Mind doing the submodule update to parent also so we can get it closed after that?

what exactly do you mean? the surge-xt needs an update to get the fix, I can update the v2.0.3-cardinal branch

baconpaul:
when you update surge-rack send a pr in with the new submodule and i can merge it here is all i meant.

falkTX:
> when you update surge-rack send a pr in with the new submodule and i can merge it here is all i meant.

err I can sure, but dont see the need.
LTO is only used for release Cardinal builds, the fix this PR goes for does not affect Rack running under normal conditions so there is no rush to update it.
It is fine to update surge-xt once there is more stuff on the surge side, in my opinion.

baconpaul:
yeah i understand that but i also like the code which cardinal users report bugs on to be the code i am looking at too!

I can do it tonight though. no big deal.

falkTX:
> yeah i understand that but i also like the code which cardinal users report bugs on to be the code i am looking at too!

That will be https://github.com/CardinalModules/surge-rack/commits/v2.0.3.0-cardinal then, which is not the official repo just yet, until a bugfix branch is made.

On the good news, seems all tests and builds pass now https://github.com/DISTRHO/Cardinal/actions/runs/3614723132
I have merged Cardinal's surgext branch to main now too

baconpaul:
Cool. I'm just cleaning up the uninit warnings you sent over. i'll merge that up date surge-rack and make a branch with an appropriate name over here.

and that's it right? talk to you when 2.1 hits the library? Do you monitor packages for updates yourself?

falkTX:
> and that's it right?

mostly yes.
I see a crash when loading surge filter on the web version, need to investigate that but those crashes are difficult to debug. I thought it would be related to x86/32bit usage, but linux i686 works fine so must be something else.
Still need to try on windows 32bit (there are users on these yes! some teachers use old machines in schools), that tends to be the one where crashes happen more often. Curiously if a crash happens on win 32bit, it tends to happen on web version too.

> talk to you when 2.1 hits the library? Do you monitor packages for updates yourself?

I dont tend to monitor these too closely, so a ping when a release is near would be appreciated, so we can catch any last minute bugs.

mkruselj:
Oh yes we know 32-bit Win is still being used... But it's dwindling.

https://somsubhra.github.io/github-release-stats/?username=surge-synthesizer&repository=releases-xt

baconpaul:
i don't think i made any 64 bit only assumptions up in the rack code either so since surge proper works 32 bit win it should be OK.

Issue #6730, opened 2022-12-04
Review and ideally remediate old style C string handling


baconpaul:
We have `std::string` and we have `fmt::format`

There's a few places where we *have* to use `char *` but nowhere near as many as we do. 

# snprintf

- [ ] src/common/Parameter.cpp: 246
- [ ] src/common/SurgeSynthesizer.cpp: 9
- [ ] src/common/UnitConversions.h: 1
- [ ] src/common/PatchDB.cpp: 2
- [ ] src/common/SurgePatch.cpp: 24
- [ ] src/common/StringOps.h: 2
- [ ] src/common/dsp/modulators/FormulaModulationHelper.cpp: 6
- [ ] src/common/dsp/oscillators/TwistOscillator.cpp: 2
- [ ] src/common/dsp/utilities/DSPUtils.h: 1
- [ ] src/common/dsp/effects/NimbusEffect.cpp: 9
- [ ] src/common/dsp/effects/airwindows/AirWindowsEffect.h: 2
- [ ] src/common/dsp/effects/CombulatorEffect.cpp: 6
- [ ] src/surge-fx/SurgeFXProcessor.cpp: 14
- [ ] src/surge-xt/gui/overlays/LuaEditors.cpp: 1
- [ ] src/surge-xt/gui/overlays/MSEGEditor.cpp: 5
- [ ] src/surge-xt/gui/SurgeGUIEditorValueCallbacks.cpp: 5
- [ ] src/surge-xt/gui/SurgeGUIEditor.cpp: 4
- [ ] src/surge-xt/gui/widgets/LFOAndStepDisplay.cpp: 6
- [ ] src/surge-xt/gui/widgets/XMLConfiguredMenus.cpp: 5

# sprintf

- [ ] src/common/dsp/filters/BiquadFilter.cpp: 1
- [ ] src/surge-xt/gui/overlays/Oscilloscope.cpp: 7
- [ ] src/surge-xt/gui/UndoManager.cpp: 1
- [ ] src/surge-xt/gui/SurgeGUIEditorInfowindow.cpp: 1
- [ ] src/surge-xt/gui/SurgeGUIEditorValueCallbacks.cpp: 2
- [ ] src/surge-xt/gui/SurgeGUIEditor.cpp: 15
- [ ] src/surge-xt/gui/widgets/OscillatorWaveformDisplay.cpp: 1

# strncpy

- [ ] src/surge-testrunner/UnitTestsIO.cpp: 1
[...]

PR #6731, opened 2022-12-05
Commence updating C-style string usage to std::string

mkruselj:
Addresses #6730

### DIFF
--- a/src/common/Parameter.cpp
+++ b/src/common/Parameter.cpp
@@ -133,12 +133,18 @@ void Parameter::create_fullname(const char *dn, char *fn, ControlGroup ctrlgroup
     // note for future devs: code here used to make a buffer of 256 bytes, write into it, then
     // incorrectly truncate to NAMECHARS (which was 64 bytes) leading to unterminated strings.
     // now tfn is also NAMECHARS and unterminated strings aren't possible. But if some behaviour
-    // changes -- this comment exists.
+    // changes, this comment exists.
     char tfn[NAMECHARS];
+
     if (useprefix)
+    {
         snprintf(tfn, NAMECHARS, "%s %s", prefix, dn);
+    }
     else
+    {
         snprintf(tfn, NAMECHARS, "%s", dn);
+    }
+
     strxcpy(fn, tfn, NAMECHARS);
 }
 
@@ -1896,6 +1902,7 @@ bool Parameter::supportsDynamicName() const
     }
     return false;
 }
+
 const char *Parameter::get_name() const
 {
     // We only even want to try this for specific types we know support it
@@ -2977,14 +2984,13 @@ void Parameter::get_display_alt(char *txt, bool external, float ef) const
 
         int i_value = round(f) + ((ctrltype != ct_freq_ringmod) ? 69 : 0);
         int oct_offset = 1;
-        char notename[16];
 
         if (storage)
         {
             oct_offset = Surge::Storage::getUserDefaultValue(storage, Surge::Storage::MiddleC, 1);
         }
 
-        snprintf(txt, TXT_SIZE, "~%s", get_notename(notename, i_value, oct_offset));
+        snprintf(txt, TXT_SIZE, "~%s", get_notename(i_value, oct_offset).c_str());
 
         if (ctrltype == ct_fmratio && !absolute)
         {
@@ -2998,14 +3004,13 @@ void Parameter::get_display_alt(char *txt, bool external, float ef) const
         float f = val.f;
         int i_value = (int)(f);
         int oct_offset = 1;
-        char notename[16];
 
         if (storage)
         {
             oct_offset = Surge::Storage::getUserDefaultValue(storage, Surge::Storage::MiddleC, 1);
         }
 
-        snprintf(txt, TXT_SIZE, "~%s", get_notename(notename, i_value, oct_offset));
+        snprintf(txt, TXT_SIZE, "~%s", get_notename(i_value, oct_offset).c_str());
 
         break;
     }
@@ -3066,10 +3071,19 @@ void Parameter::get_display_alt(char *txt, bool external, float ef) const
 
 void Parameter::get_display(char *txt, bool external, float ef) const
 {
+    auto str = get_display(external, ef);
+
+    strncpy(txt, str.c_str(), TXT_SIZE - 1);
+}
+
+std::string Parameter::get_display(bool external, float ef) const
+{
+    std::string txt = "";
+
     if (ctrltype == ct_none)
     {
-        snprintf(txt, TXT_SIZE, "-");
-        return;
+        txt = "-";
+        return txt;
     }
 
     int i;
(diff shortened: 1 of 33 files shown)


src/common/SurgePatch.cpp
```diff
@@ -2923,7 +2930,7 @@ unsigned int SurgePatch::save_xml(void **data) // allocates mem, must be freed b
                             // below
                             TiXmlElement mr("modrouting");
                             mr.SetAttribute("source", r->at(b).source_id);
-                            mr.SetAttribute("depth", float_to_str(r->at(b).depth, tempstr));
```
baconpaul, line 2926: So these are not equivalent. float_to_str sets the locale to C to make sure this streams properly if you are in france or whatever with a , as a decimal separator, and std::to_string does not. If we merge this we will break streaming!

There is an STL equivalent but I'll need to figure it out.

src/common/SurgeSynthesizerIO.cpp
```diff
@@ -609,7 +609,7 @@ void SurgeSynthesizer::savePatchToPath(fs::path filename, bool refreshPatchList)
     fxp.numPrograms = vt_write_int32BE(1);
     fxp.version = vt_write_int32BE(1);
     fxp.fxVersion = vt_write_int32BE(1);
-    strncpy(fxp.prgName, storage.getPatch().name.c_str(), 28);
```
baconpaul, line 612: didn't you reverse the order here? strncpy has arguments "destination source length" so this means "fxp.prgName = storage.getPatch().name"

mkruselj: Whoops, reverting.

src/common/SurgePatch.cpp
mkruselj: Modified with updated method float_to_clocalestr() that returns std::string.

baconpaul:
<sstream> not <sstream.h>

src/common/dsp/utilities/DSPUtils.h
```diff
@@ -365,12 +367,15 @@ inline double hamming(int i, int n)
     return 0.54 - 0.46 * cos(2 * M_PI * i / (n - 1));
 }
 
-inline char *float_to_str(float value, char *str)
+inline std::string float_to_clocalestr(float value)
```
mx, line 370: I think changing this name is definitely a good idea, but I'd add a comment too saying why we've got this function. It already tripped someone up (you!) when it came to updating it to use std::string!

mkruselj: Yes, valid suggestion, makes total sense. Will add comment 🙂

src/common/dsp/effects/NimbusEffect.cpp
```diff
@@ -269,39 +270,39 @@ void NimbusEffect::init_ctrltypes()
...
-                    snprintf(res, TXT_SIZE, "%s", "Filter");
+                    res = fmt::format("{:s}", "Filter");
                 if (idx == nmb_size)
-                    snprintf(res, TXT_SIZE, "%s", "Size");
+                    res = fmt::format("{:s}", "Size");
                 break;
             case 3:
                 if (idx == nmb_density)
-                    snprintf(res, TXT_SIZE, "%s", "Smear");
+                    res = fmt::format("{:s}", "Smear");
```
baconpaul, line 297: This is just res=“smear” since strings copy properly

baconpaul (changes requested):

src/common/dsp/effects/CombulatorEffect.cpp
```diff
@@ -426,46 +427,45 @@ void CombulatorEffect::init_ctrltypes()
...
-                    snprintf(res, TXT_SIZE, "%s", "Offset 3");
+                    res = fmt::format("{:s}", "Offset 3");
                 }
                 break;
             default:
                 break;
             }
 
-            return res;
+            return res.c_str();
```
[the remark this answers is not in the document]
baconpaul: It will work because of ststic string but we need to fix the overarching api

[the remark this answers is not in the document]
baconpaul: So lemme think about this some

src/common/Parameter.cpp
```diff
@@ -3138,38 +3156,38 @@ void Parameter::get_display(char *txt, bool external, float ef) const
...
+            return txt;
         }
         case ATwoToTheBx:
         {
             if (can_temposync() && temposync)
             {
                 std::string res =
                     tempoSyncNotationValue(displayInfo.tempoSyncNotationMultiplier * f);
-                snprintf(txt, TXT_SIZE, "%s", res.c_str());
+                txt = fmt::format("{:s}", res);
```
baconpaul, line 3188: This can just be txt = res or better just return res

src/common/Parameter.cpp
```diff
@@ -3187,8 +3205,8 @@ void Parameter::get_display(char *txt, bool external, float ef) const
             {
                 if (displayInfo.customFeatures & ParamDisplayFeatures::kHasCustomMaxString)
                 {
-                    snprintf(txt, TXT_SIZE, "%s", displayInfo.maxLabel);
-                    return;
+                    txt = fmt::format("{:s}", displayInfo.maxLabel);
```
baconpaul, line 3208: No need to format for a copy

src/common/dsp/effects/NimbusEffect.cpp
mkruselj: Haha yeah, totally!

src/common/Parameter.cpp
mkruselj: Actually it doesn't need that local str::string at all, can just assign tempoSyncNotationValue to txt then return txt. Seems cleaner!

src/common/Parameter.cpp
```diff
@@ -3104,12 +3118,16 @@ void Parameter::get_display(char *txt, bool external, float ef) const
...
+
             if (ef)
             {
                 // parameter called 'len' not 'size', be on the safe side here, do - 1.
                 // It used to say just '64' anyway.
-                if (ef->formatValue(this, f, txt, TXT_SIZE - 1))
-                    return;
+                if (ef->formatValue(this, f, str, TXT_SIZE - 1))
+                {
+                    return txt;
```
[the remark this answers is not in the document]
baconpaul: it did a void return but it had mutated txt.

so the way this works is formatValue sets onto text. text is passed in. so text is modified at the return

the way your change makes it work os formatValue modifies str but then ignores that modification and returns txt 

so the formatting is ignored

I think the equivalent change is `return str` rather than `return txt`

src/common/Parameter.cpp
```diff
@@ -3433,9 +3443,9 @@ void Parameter::get_display(char *txt, bool external, float ef) const
                                 // "i & 3" selects the lower two bits that represent the stage count
                                 // "(i >> 2) & 3" selects the next two bits that represent the
                                 // saturator
-                                snprintf(txt, TXT_SIZE, "%s %s",
-                                         sst::filters::fut_nlf_subtypes[i & 3],
-                                         sst::filters::fut_nlf_saturators[(i >> 2) & 3]);
+                                txt =
+                                    fmt::format("{:s} {:s}", sst::filters::fut_nlf_subtypes[i & 3],
```
baconpaul, line 3444: Still some of these no-op formats kicking around

baconpaul:
OK I don't see any other changes and I don't think reading it again will show me another one.

Nice work @mkruselj 

when CI passes feel free to press merge :)

baconpaul (approved):

PR #6732, opened 2022-12-05
Use the defaulted copy and assignment constructors, instead of memcpy…

mx:
…, for parameters in SurgeSynthesizer.

Starting with just this file, rather than a full replacement, to ensure that everything is good with the outcome here before I change everything else.

The generated assembly code before called memcpy. Afterwards it seems to inline the copy constructor and assignment operator, using registers to move the data. I know gcc can fall back to memcpy for trivial classes, so I believe it's choosing not to do that here.


baconpaul:
Ok! This looks correct to me and the tests pass. Let’s merge it and see!

PR #6737, opened 2022-12-08
Use clang-format-12 for code-quality-pipeline-checks

haenkel:
- like we thought it would
- message what we use

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -193,7 +193,7 @@ endforeach()
 add_custom_command(TARGET code-quality-pipeline-checks
   POST_BUILD
   WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}
-  COMMAND ${CMAKE_COMMAND} -E echo About to check clang-format
-  COMMAND git ls-files -- ${CLANG_FORMAT_GLOBS} | xargs clang-format --dry-run --Werror
+  COMMAND ${CMAKE_COMMAND} -E echo About to check clang-format using clang-format-12
+  COMMAND git ls-files -- ${CLANG_FORMAT_GLOBS} | xargs clang-format-12 --dry-run --Werror
   )
 # }}}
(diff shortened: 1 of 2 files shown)


mkruselj:
Thank you!

baconpaul:
This is super cool! I just added clang-format checks to surge-rack also and to make it so I could test them locally through cmake, I did the slightly modified version as follows

```
# Clang Format checks
find_program(CLANG_FORMAT_EXE NAMES clang-format-12 clang-format)
set(CLANG_FORMAT_DIRS src)
set(CLANG_FORMAT_EXTS cpp h)
foreach(dir ${CLANG_FORMAT_DIRS})
  foreach(ext ${CLANG_FORMAT_EXTS})
    list(APPEND CLANG_FORMAT_GLOBS "':(glob)${dir}/**/*.${ext}'")
  endforeach()
endforeach()
add_custom_command(TARGET xt-rack-code-checks
        POST_BUILD
        WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}
        COMMAND ${CMAKE_COMMAND} -E echo About to check clang-format using ${CLANG_FORMAT_EXE}
        COMMAND git ls-files -- ${CLANG_FORMAT_GLOBS} | xargs ${CLANG_FORMAT_EXE} --dry-run --Werror
        )
# }}}
```

where I look for either clang-format-12 or clang-format

i didn't change other things and in ubuntu actions it found 12, but just an FYI that unless you object i may make that tweak in other spots when i try to run the codequal checks locally on my mac :)

haenkel:
Yes 12 comes with the image it just defaults to 11 (and I think 13 with 22.04).
 
No objection really, just my thoughts why I did it like this:

clang-format versions are also not really forward compatible,
so if you format and test with anything higher than 12 local you may still error in pipeline,
but with local test you don't get a message that the required version wasn't found,
 it would just choose the default which I haven't found to be 12 anywhere.
(not a problem right now, I checked the current status of the repos to also pass with 13 and 14, 
and turned cf off where it didn't, but this will change of course)

Maybe add an `if not 12 message "ALARM"` at the bottom?

PR #6738, opened 2022-12-10
Replace memcpy with std::copy and assignment operators where safe

mx:
This is all the memcpy that can reasonably be replaced without going through and refactoring uses of C-style arrays and news into C++11-and-up constructs. That seems like a reasonable exercise, but one for the future.

### DIFF
--- a/src/common/dsp/Wavetable.h
+++ b/src/common/dsp/Wavetable.h
@@ -13,6 +13,7 @@ const int max_wtable_samples = 2097152;
 #pragma pack(push, 1)
 struct wt_header
 {
+    // This struct can only contain scalar data that can be memcpy'd.
     char tag[4];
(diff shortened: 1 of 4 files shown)


src/common/dsp/Wavetable.h
```diff
@@ -13,6 +13,7 @@ const int max_wtable_samples = 2097152;
 #pragma pack(push, 1)
 struct wt_header
 {
+    // This struct can only contain scalar data that can be memcpy'd.
```
baconpaul, line 16: (since it is read directly from memory on disk just FYI as I'm sure you know)

baconpaul:
Looks good to me but I wonder why CI didn't start? Do you mind doing a quick push to see if it goes?

PR #6743, opened 2022-12-12
Replace the original set of Templates with some that are CC0-licensed.

selenologist:
Init Saw and Init Sine were created by @baconpaul and posted on Discord.

Also add CC0 `Luna/NES Triangle` wavetable that was used in `Init NES Triangle WT`.

edit: I also want to state for the record that I do not recommend the actual use of `Init Alias NES Triangle` (or indeed Alias at all) to create chiptune as it has a DC offset and produces inharmonic aliasing which is not accurate. `Init NES Triangle WT` should be more accurate for this purpose.


mkruselj:
I think plastering the whole license in the comments definitely going overboard...

mkruselj:
I've reworked this PR by recreating all factory templates (including audio in ones) from Paul's CC0 Init Saw, removed Emu and Argitoth templates wholesale, and added a License field to Save Patch dialog (#6769, #6770).

PR #6767, opened 2023-01-07
Some more work on remediating C-style strings

mkruselj:
Addresses #6730

### DIFF
--- a/src/common/SurgeStorage.cpp
+++ b/src/common/SurgeStorage.cpp
@@ -1110,6 +1110,7 @@ void SurgeStorage::load_wt(int id, Wavetable *wt, OscillatorStorage *osc)
 {
     wt->current_id = id;
     wt->queue_id = -1;
+
     if (wt_list.empty() && id == 0)
     {
 #if HAS_JUCE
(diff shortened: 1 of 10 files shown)


src/common/SurgeStorage.cpp
```diff
@@ -1110,45 +1110,66 @@ void SurgeStorage::load_wt(int id, Wavetable *wt, OscillatorStorage *osc)
...
     if (osc)
     {
-        auto n = wt_list.at(id).name;
-        strncpy(osc->wavetable_display_name, n.c_str(), 256);
+        osc->wavetable_display_name = wt_list.at(id).name;
     }
 }
 
 void SurgeStorage::load_wt(string filename, Wavetable *wt, OscillatorStorage *osc)
 {
```
baconpaul, line 1152: this is a fine change but we should open an issue to move this and a few other apis to fs::path

src/surge-xt/gui/widgets/OscillatorWaveformDisplay.cpp
```diff
@@ -573,15 +573,15 @@ void OscillatorWaveformDisplay::populateMenu(juce::PopupMenu &contextMenu, int s
     }
 
     auto rnaction = [this]() {
-        char c[TXT_SIZE];
-        strncpy(c, this->oscdata->wavetable_display_name, TXT_SIZE);
+        std::string c;
+        c = this->oscdata->wavetable_display_name;
```
baconpaul, line 577: this could be `auto c = this->oscdata` but only if we find things that require changes otherwise its ok

PR #6776, opened 2023-01-08
Add a 'Announce State' Accessible option

### LINKED ISSUE #6591: Scene and oscillator keybinds should be announced

I am frustrated when I am not sure which scene I am in or which oscillator I have selected if I am in another section of the synth.
While using Surge I would very much like the alt + s shortcuts for scenes. Alt + 1/2/3 for oscillators. used speech synthesis to tell what the status of the selection is.
Alt + s scene 1 / scene2. Alt + 1/2/3 changes the oscillator. If you are in another section like LFO I don't know after minutes if I am in one scene or another. Also for the oscillator.

It would be very useful while working if the screen reader would announce the selected scene or oscillator via the shortcuts.

If it is easier to develop. For shortcuts, the Windows sapi 5 system speech is also good. More useful if the screen reader announced it.
Thank you!

baconpaul:
If you annoucne state (default alt-0) then you get a readout of the state of your UI (scene, selected osc and modulator, selected fx, patch name).

Also MTS Main -> MTS Source in the menu

Closes #6591

### DIFF
--- a/src/surge-xt/gui/SurgeGUIEditor.cpp
+++ b/src/surge-xt/gui/SurgeGUIEditor.cpp
@@ -3547,7 +3547,7 @@ juce::PopupMenu SurgeGUIEditor::makeTuningMenu(const juce::Point<int> &where, bo
 
     auto canMaster = MTS_CanRegisterMaster();
 
-    std::string mtxt = "Act as ODDSound" + Surge::GUI::toOSCase(" MTS-ESP Main");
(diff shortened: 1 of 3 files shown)


Lo-lo78:
@baconpaul 
Works well... Maybe I don't need to announce the plugin name.
We hear what others think.
👍

baconpaul:
Oh yeah that's smart. I'll get that in my next push (which is implementing Mono noise which you also asked about in the past)

PR #6790, opened 2023-01-09
Make surgepy an installable Python package

narenratan:
Hi Surge Synth Team,

Thank you for Surge! I've been using Surge in a command line just intonation sequencer I'm writing in Python, so Surge's tuning support and Python API have been super useful :)

It would be handy for me to be able to include surgepy as a dependency for my project, so I had a look at packaging it as a Python package. It turns out there is a tool [scikit-build](https://scikit-build.readthedocs.io/en/latest/) specifically for packaging Python extensions built with CMake. In case it would be of interest, this PR is to make surgepy installable as a Python package using scikit-build. It's based on the scikit-build pybind example project [here](https://github.com/scikit-build/scikit-build-sample-projects/tree/master/projects/hello-pybind11).

With this branch you can run
```
python3 -m pip install <REPO_DIR>
```
on the surge repo to build and install surgepy as a Python package. I've used this to install surgepy in a Python virtual environment to use for development.

Also I've built a Python binary wheel for surgepy so I can install it without building it every time. I tested installing the wheel in a Ubuntu VM and it works. Using wheels it would be possible to create a surgepy package on PyPI so people could `pip install surgepy` and it would just work.

Thank you again for Surge!


baconpaul:
Oh this is super cool and I'm happy to merge it thank you. 

Only one question. You've added two new root-level files. 

1. Is it possible to only have one of those be root level and the other inside surgepy somewhere and still work with this build system?
2. Even if not, could you add some explanatory comments to both. The average person looking at this repo will be a C++ user wanting to build a VST/CLAP etc... so a comment like 'this is for the python build; see here; do this' at least would make the 'root directory' prominence a little less prominent.

Thanks!

baconpaul:
Oh also, please add yourself to the AUTHORs file along with that! 

And thanks again, This is super cool.

narenratan:
Thanks very much for looking at this! Sure I've moved setup.py and and pyproject.toml down to src/surge-python and set scikit-build's `cmake_source_dir` to point two levels up (and checked the python package still builds and the tests pass). I added some comments to them too. Also I moved test_surgepy.py down there into tests too, so src/surge-python is now a somewhat self-contained Python project.

The command to install the Python package is now
```
python3 -m pip install <REPO_DIR>/src/surge-python
```
I've updated the commit message and the notebook with the new command.

baconpaul:
Wonderful thank you. It looks great! I'll merge it once CI is complete.

Oh and if you hop into our discord (or if you are already there with a different ID - your project sounds a lot like one we've been discussing in our tuning channel) let us know and we can give you the surge synth team role as a contributor. 

Best

PR #6807, opened 2023-01-15
Expose tuningApplicationMode in surgepy

narenratan:
Add `SurgeStorage::TuningApplicationMode` enum to surgepy using pybind11's enum_. Expose `tuningApplicationMode` in surgepy with pybind11's def_property.

In Python `tuningApplicationMode` can now be accessed as
```
    >>> import surgepy
    >>> s = surgepy.createSurge(44100)
    >>> s.tuningApplicationMode
    <TuningApplicationMode.RETUNE_MIDI_ONLY: 1>
```
and set as
```
    >>> s.tuningApplicationMode = surgepy.TuningApplicationMode.RETUNE_ALL
```

### DIFF
--- a/src/surge-python/surgepy.cpp
+++ b/src/surge-python/surgepy.cpp
@@ -6,6 +6,7 @@
 #include <utility>
 
 #include "SurgeSynthesizer.h"
+#include "SurgeStorage.h"
 #include "version.h"
 #include "filesystem/import.h"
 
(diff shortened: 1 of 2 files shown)


baconpaul:
This is great

One day I should make this test run in the ci pipeline. Right now we build the python on Linux but don’t run it.

narenratan:
Thanks very much!

Running the test_surgepy.py tests needs pytest too, so a full set of commands to build the surgepy package and run tests would be
```
$ python3 -m venv surge_env
$ source ./surge_env/bin/activate
$ python3 -m pip install <REPO_DIR>/src/surge-python
$ python3 -m pip install pytest
$ pytest -v <REPO_DIR>/src/surge-python
```
and that last command exits 0 if the tests passed, 1 if they failed.

Installing the Python package like that does a full build with `-DSURGE_BUILD_PYTHON_BINDINGS=TRUE` though, so just using the shared object from the Linux build would save some time. If shared object is produced in say <REPO_DIR>/build/src/surge-python then I think
```
$ python3 -m venv surge_env
$ source ./surge_env/bin/activate
$ python3 -m pip install pytest numpy
$ export PYTHONPATH=<REPO_DIR>/build/src/surge-python
$ pytest -v <REPO_DIR>/src/surge-python
```
should work (needs numpy installing separately if we don't pip install the surgepy package).

PR #6815, opened 2023-01-22
Make typein for ct_float_toggle work

### LINKED ISSUE #6814: Nimbus "Freeze" parameter can't be modulated
**Bug Description:**
The "Freeze" parameter in the Nimbus effect cannot be modulated by a scene envelope. Confusingly, ED was able to actually make it work and even send me a patch where it is modulated, which I could load! However, myself and at least one other person confirmed that we weren't able to do it on our systems!

**Surge XT Version**
Latest nightly

* Version: Latest nightly
* Plugin Type: standalone
* Bitness: 64-bit

**Reproduction Steps:**
1. Load template "Init Saw"
2. Add Nimbus effect to first FX on chain A.
3. Click on "S-LFO 1". Change it to the template "Basic ADSR".
4. Set D to 1 second, S to 100%, every other slider to 0.
5. Try to drag and drop it onto the "Freeze" slider. Does not work. (It can successfully be dropped on any other parameter.)

**Expected Behavior:**
Freeze parameter is modulatable.

**Screenshots:**
N/A

 - OS: Windows 10
 - Host: Standalone

**Additional Information:**
N/A

mkruselj:
Closes #6814

Also rename calculate_modulation_value_from_string() to set_modulation_value_from_string() to match set_value_from_string() API.


src/common/Parameter.cpp
```diff
@@ -4870,6 +4943,30 @@ float Parameter::calculate_modulation_value_from_string(const std::string &s, st
...
+            ErrorMessageMode isLarger =
+                (rmv < -1) ? ErrorMessageMode::IsSmaller : ErrorMessageMode::IsLarger;
+
+            set_error_message(errMsg, fmt::format("{:g}", (rmv < -1) ? -bound : bound),
+                              displayInfo.unit, isLarger);
+
+            valid = false;
+        }
+
+        return mv;
```
mx, line 4968: Did you mean to return rmv here?

mkruselj: Absolutely. Thanks for catching it!

baconpaul:
Oooh but it doesn’t set the modulation value it calculates it and an external client sets it so can we undo that name change?

baconpaul:
(It doesn’t set it because the parameter doesn’t know which modulator is querying the parameter)

mkruselj:
Sure I'll revert that, maybe shorten to calc instead of calculate.

baconpaul:
that will just break rack when i pull next. if we can leave the name the same that'd be super!

PR #6825, opened 2023-01-26
Update LFO display when toggling LFO amplitude scaling option

mkruselj:


src/surge-xt/gui/SurgeGUIEditor.cpp
```diff
@@ -1167,6 +1167,11 @@ void SurgeGUIEditor::idle()
         oscWaveform->repaintBasedOnOscMuteState();
     }
 
+    if (lfoDisplay)
+    {
+        lfoDisplay->repaint();
+    }
+
```
baconpaul, line 1174: Please don’t merge tbis - it will force a full repaint on every idle 

I will look later today but why not just add a method to call repaint that you call when you need it rather than doing it in idle?

mkruselj: Updating this with a better impl!

PR #6827, opened 2023-01-27
Alt+click on an FX slot clears it

mkruselj:


baconpaul:
oooh that test failure is legit. you lose modulation when doing swaps and moves with it somehow.

can't look this am but obvs lets hold of on merging.

mkruselj:
Darn it! I have no crashes in release build... But sure, let's look into it. May just be defensive around anything done with `target` in case `source == target && FXReorderMode::NONE`.

baconpaul:
oh its not a crash. it is the test saying 'when i do the drag action the wrong thing happens'

baconpaul:
```
-------------------------------------------------------------------------------
FX Move with Modulation
  Setup a Modulated Combulator and Move It
-------------------------------------------------------------------------------
/Users/runner/work/1/s/src/surge-testrunner/UnitTestsFX.cpp:111
...............................................................................

/Users/runner/work/1/s/src/surge-testrunner/UnitTestsFX.cpp:125: FAILED:
  REQUIRE( surge->storage.getPatch().modulation_global.size() == 1 )
with expansion:
  0 == 1
```

mkruselj:
I see it, I put a good line in the wrong place. Fixing...

PR #6828, opened 2023-01-27
Build wheels for surgepy with cibuildwheel

narenratan:
Use cibuildwheel to build wheels (compiled binary Python packages) for surgepy. Wheels can be built locally by running
```
    $ pipx run cibuildwheel --platform linux src/surge-python/
```
in the surge repo (if pipx and podman are installed). More info on cibuildwheel [here](https://cibuildwheel.readthedocs.io/en/stable/).

Also do not link surgepy with the system libpython when building the surgepy Python package. This allows using the built wheels with multiple Python interpreters. The pybind11 docs recommend not linking against libpython for this reason; more info [here](https://pybind11.readthedocs.io/en/stable/compiling.html#building-manually).

No CI changes here, just changes to make cibuildwheel runnable locally.

### DIFF
--- a/src/surge-python/setup.py
+++ b/src/surge-python/setup.py
@@ -26,6 +26,13 @@ setup(
     install_requires=["numpy"],
     packages=["surgepy"],
     cmake_source_dir="../..",
-    cmake_args=["-DSURGE_BUILD_PYTHON_BINDINGS=TRUE"],
+    cmake_args=[
+        "-DSURGE_BUILD_PYTHON_BINDINGS=TRUE",
+        "-DSURGE_SKIP_JUCE_FOR_RACK=TRUE",
+        "-DSURGE_SKIP_ODDSOUND_MTS=TRUE",
+        "-DSURGE_SKIP_VST3=TRUE",
+        "-DSURGE_SKIP_ALSA=TRUE",
+        "-DSURGE_SKIP_STANDALONE=TRUE",
+    ],
     cmake_process_manifest_hook=just_surgepy,
 )
--- a/src/common/CMakeLists.txt
+++ b/src/common/CMakeLists.txt
@@ -310,10 +310,16 @@ if(APPLE)
     )
 elseif(UNIX)
   target_compile_definitions(${PROJECT_NAME} PUBLIC LINUX=1)
-  target_link_libraries(${PROJECT_NAME} PRIVATE
-    pthread
-    -Wl,--no-undefined
-    )
+  if(SKBUILD)
+    target_link_libraries(${PROJECT_NAME} PRIVATE
+      pthread
+      )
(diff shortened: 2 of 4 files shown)


src/surge-python/setup.py
```diff
@@ -26,6 +26,6 @@ def just_surgepy(cmake_manifest):
     install_requires=["numpy"],
     packages=["surgepy"],
     cmake_source_dir="../..",
-    cmake_args=["-DSURGE_BUILD_PYTHON_BINDINGS=TRUE"],
+    cmake_args=["-DSURGE_BUILD_PYTHON_BINDINGS=TRUE", "-DSURGE_SKIP_STANDALONE=TRUE"],
```
baconpaul, line 29: I'm happy to merge this but I bet we want to include some of the other 'skip' options we put in place for rack and others, like SKIP_VST3 

https://github.com/surge-synthesizer/surge-rack/blob/9f9922ddbc234fc89d4b6c1b8be4364dec72af40/CMakeLists.txt#L24

here's the list we use in rack. For python I bet we can do

```
set(SURGE_SKIP_JUCE_FOR_RACK TRUE CACHE BOOL "skip surge Juce build")
set(SURGE_SKIP_ODDSOUND_MTS TRUE CACHE BOOL "skip surge oddsound-mts build")
set(SURGE_SKIP_VST3 TRUE CACHE BOOL "skip surge VST3 build")
set(SURGE_SKIP_ALSA TRUE CACHE BOOL "skip surge Alsa build")
set(SURGE_SKIP_STANDALONE TRUE CACHE BOOL "skip surge standalone build")
```

and still hav python work just fine

baconpaul: oh let me know if you want me to merge and then test this later or if you prefer to change and test and then have it merged

narenratan: Thanks very much! Skipping more would be great. I suspect it would speed up building the surgepy Python package. It might also mean we can drop some of the 'yum install' packages for building the wheels; they were just for juce stuff if I remember.

I'll try adding those skip flags and retest - can do it as part of this PR.

baconpaul: great thanks!

the skip juce one makes a new library target you need to include called l`juce_dsp_rack_sub` which depending on the linkage pattern of surge python you may need to add explicitly to the link phase when juce is off.

narenratan: No problem at all. I pushed a version with those skip flags added and all the yum install packages removed. The surgepy Python package still builds and passes the surgepy tests, and the cibuildwheel wheel building still works. It's also easily twice as fast to build now!

baconpaul: Awesome. Will merge now

PR #6843, opened 2023-02-10
Allow custom juce path for surge-rack builds

falkTX:
This is needed in order to use a custom JUCE for surge-rack, as the JUCE dirs are hardcoded to the local files.

### DIFF
--- a/src/common/CMakeLists.txt
+++ b/src/common/CMakeLists.txt
@@ -62,18 +62,18 @@ else()
   target_compile_definitions(surge-common-binary INTERFACE HAS_JUCE=0 SKIP_JUCE=1)
 
   add_library(juce_dsp_rack_sub STATIC
-          ../../libs/JUCE/modules/juce_audio_basics/juce_audio_basics.cpp
-          ../../libs/JUCE/modules/juce_dsp/juce_dsp.cpp
+          ${SURGE_JUCE_PATH}/modules/juce_audio_basics/juce_audio_basics.cpp
+          ${SURGE_JUCE_PATH}/modules/juce_dsp/juce_dsp.cpp
           )
   if (APPLE)
     target_sources(juce_dsp_rack_sub PRIVATE
-            ../../libs/JUCE/modules/juce_core/juce_core.mm
-            ../../libs/JUCE/modules/juce_audio_formats/juce_audio_formats.mm
+            ${SURGE_JUCE_PATH}/modules/juce_core/juce_core.mm
+            ${SURGE_JUCE_PATH}/modules/juce_audio_formats/juce_audio_formats.mm
             )
   else()
     target_sources(juce_dsp_rack_sub PRIVATE
-            ../../libs/JUCE/modules/juce_core/juce_core.cpp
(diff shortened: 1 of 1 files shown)


baconpaul:
Thanks!

baconpaul:
Oh I presume another PR is inbound which sets this variable in the rack build right?

baconpaul:
(the reason this is here is because SURGE_JUCE_PATH is set as CURRENT_SOURCE_DIR which is wrong in rack; if you don't have a fix to that also let me know and I'll correct this diff)

baconpaul:
oh actually we don't need to change anything in rack if we fix one more thing in surge cmake. 

```
diff --git a/src/CMakeLists.txt b/src/CMakeLists.txt
index 7740ffe9..173ca278 100644
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -9,7 +9,7 @@ if (NOT SURGE_COMPILE_BLOCK_SIZE)
   set(SURGE_COMPILE_BLOCK_SIZE 32)
 endif()
 
-set(SURGE_JUCE_PATH "${CMAKE_SOURCE_DIR}/libs/JUCE" CACHE STRING "Path to JUCE library source tree")
+set(SURGE_JUCE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/../libs/JUCE" CACHE STRING "Path to JUCE library source tree")
 
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
 include(cmake/lib.cmake)
```

that's the extra diff we need then rack builds with this change too.

baconpaul:
i'll commit that now

falkTX:
I dont have the fix for the surge-rack side at this point, so your initiative is quite welcome (ensuring no breakage)
Thanks for the quick merge!

baconpaul:
no problem. with my change in #6844 surge rack builds fine. (And that's actually how I should have done it in the first place!)

baconpaul:
https://github.com/surge-synthesizer/surge-rack/pull/835 pulls it into surge-rack also

baconpaul:
oh if you are doing a cardinal surge upgrade also I'll probably do a 2.1.2 next week with extended envelope times in one module, just FYI. Tuesday maybe or weds it will be ready? Won't be any substantial code change from head though.

falkTX:
I hope to have a 23.02 cardinal release, but no defined date for that yet besides "sometime later this month".
I already updated to latest suge 2.1.1 in preparation for the new plugins, builds fine now mostly just dealing with this juce related things

baconpaul:
Great well I merged the change with the adjustable juce path to head so should be good 

did you get cmake build workinf or did you write an alternative makefile out of curiosity? If the later I would be happy to check in a makefile.cardinal if that’s easier for you l

falkTX:
there is a big makefile to deal with building all plugin files for all the modules. I do not use the rack-style of makefile import for building plugins.
surge stuff defined in https://github.com/DISTRHO/Cardinal/blob/main/plugins/Makefile#L1032

baconpaul:
Oh gotcha. The. Our move to cmake didn’t impact you. That’s good

PR #6852, opened 2023-02-18
Prevent modulation smoothers from running away to infinity

### LINKED ISSUE #6835: Reliable Crash, 1.1.2, Windows, Reaper latest
From William Fields

[crash-example.zip](https://github.com/surge-synthesizer/surge/files/10609352/crash-example.zip)

this minimal reaper project crashes at measure 4 when played with XT 1.1.2

baconpaul confirmed on his win 11 box.

mkruselj:
Closes #6835

### DIFF
--- a/src/common/ModulationSource.h
+++ b/src/common/ModulationSource.h
@@ -246,8 +246,11 @@ class ModulationSource
     ModulationSource()
     {
         for (int i = 0; i < vecsize; ++i)
+        {
             voutput[i] = 0.f;
+        }
     }
+
     virtual ~ModulationSource() {}
     virtual const char *get_title() { return 0; }
     virtual int get_type() { return mst_undefined; }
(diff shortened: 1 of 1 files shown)


baconpaul:
Oh isinf can be kinda expensive I think I was thinking we would fix this by clamping chanfe not isinf outout

mkruselj:
Oh we could clamp for sure what should be the clamping range?

baconpaul:
The thing to clamp is b. My fix was something like if b blows out jump to target iirc

baconpaul:
But I didn’t out that in the issue sigh sorry

mkruselj:
Right but what would be the range to test b against? What is considered as "b blew up"?

baconpaul:
20? That’s why the issue is still open o want to run it in a few cases and see what’s reasonable and what’s crazy

PR #6863, opened 2023-03-03
Stop runaway modulators on extreme changes

### LINKED ISSUE #6835: Reliable Crash, 1.1.2, Windows, Reaper latest

baconpaul:
In extreme (really computer generated) midi sequences you could force the modulator smoother to runaway since in one case it would push you outside of [target,value] when interpolating. Fix that with a clamp; and add an assert in case theres other future problems.

Closes #6835

### DIFF
--- a/src/common/ModulationSource.h
+++ b/src/common/ModulationSource.h
@@ -16,6 +16,8 @@
 #pragma once
 
 #include <random>
+#include <cassert>
+
 #include "basic_dsp.h"
 
 enum modsrctype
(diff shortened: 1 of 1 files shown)


mkruselj:
Nice, this is way better than what I hacked in! 🙂

Issue #6638, opened 2022-10-02
FX Unit shows more clearly if you violate latency implicit guarantee


baconpaul:
FX Unit by default runs 0 latency and assumes a multiple of 32 block size
If you are in 0 latency mode and get a non-multiple-of-32 block you should

1: not segfault (duh) and
2: put up a message on the UI pointing people to the option for that host

mkruselj:
Linking this one to #6524, these two issues should have the same resolution.

baconpaul:
the resolutions are different since the fx already has a latent option. This i just detection and handling.

baconpaul:
Fixed in #6866

PR #6866, opened 2023-03-03
FX Mode handles block mod buffers better

baconpaul:
In the event we are in non-latent mode and our buffer is not a multiple of block size, then

1: reset the inputs and outputs
2: Enter latent mode
3: set latency and
4: inform the host of a latency change

Addresses #6638 but some testing before we close it would be super.


baconpaul:
@mkruselj I'll need a bit of help testing this

basically do you have an FL environment with non-uniform blocks? Can you then make sure the non-latent mode is on and run it and see if it resets properly? The code looks fine but all my hosts are block-size 32 uniform.

mkruselj:
Yup sure I can try. Also clang format. 🙂

PR #6877, opened 2023-03-06
Initial oscilloscope UI tweaks

mkruselj:
Biggified the whole thing
Various text label tweaks, some parameter range and value display tweaks
Retrigger Threshold didn't gray out when Free/Internal modes were selected
Use MSEG Panel skin color for parameter area background
Larger font used for sliders, like in the rest of the GUI

### DIFF
--- a/src/surge-xt/gui/overlays/Oscilloscope.cpp
+++ b/src/surge-xt/gui/overlays/Oscilloscope.cpp
@@ -43,7 +43,7 @@ static float freqToX(float freq, int width)
 }
 
 // This could, in theory, divide by zero, so be careful with how you call it.
-static float dbToY(float db, int height, float dbMin = -100.f, float dbMax = 0.f)
+static float dbToY(float db, int height, float dbMin = -96.f, float dbMax = 0.f)
 {
     float range = dbMax - dbMin;
     // If dbMax/dbMin change and the data hasn't been updated yet due to locking, can get weirdness.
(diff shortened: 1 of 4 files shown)


src/surge-xt/gui/overlays/Oscilloscope.cpp
```diff
@@ -193,6 +206,7 @@ void WaveformDisplay::process(std::vector<float> data)
     int triggerLimit =
         static_cast<int>(std::pow(10.f, params_.trigger_limit * 4.f)); // 0=>1, 1=>10000
     float triggerSpeed = std::pow(10.f, 2.5f * params_.trigger_speed - 5.f);
+    printf("%.5f\n", triggerSpeed);
```
mx, line 209: Looks like a debug printf, should remove this.

src/surge-xt/gui/overlays/Oscilloscope.cpp
```diff
@@ -467,21 +490,27 @@ void SpectrumDisplay::resized()
...
 
 void SpectrumDisplay::updateScopeData(internal::FftScopeType::iterator begin,
                                       internal::FftScopeType::iterator end)
 {
     // Data comes in as gain.
     std::lock_guard l(data_lock_);
+
     // Decay existing data, and move new data in if it's larger.
-    const float decay = (1.f - params_.decay_rate);
+    const float decay = (1.f - (params_.decay_rate * 0.9f));
```
mx, line 503: With this change we no longer can get the instant-decay behavior when decay_rate = 1 or maxes behavior when decay_rate = 0.

src/surge-xt/gui/overlays/Oscilloscope.cpp
```diff
@@ -1260,19 +1359,23 @@ void Oscilloscope::Background::paintWaveformBackground(juce::Graphics &g)
         std::string minus = "-";
         std::string plus = "+";
         std::stringstream gain;
-        gain << std::fixed << std::setprecision(2) << 1.f / waveform_params_.gain();
+        const unsigned prec = waveform_params_.gain() > 1.f ? 2 : 0;
```
mx, line 1362: I'd explicitly label this as an unsigned int. I think raw unsigned is a deprecated C-ism (here and below).

src/surge-xt/gui/overlays/Oscilloscope.cpp
mkruselj: OK I will then just adjust the scaling to be sqrt of params_.decay_rate.

PR #6923, opened 2023-04-09
Add a few more Wavetables

Andreya-Autumn:
Taken from Crust


Andreya-Autumn:
This is Crust, to be clear: https://www.kvraudio.com/forum/viewtopic.php?t=502672

And when I say "taken from", I mean I have recorded its output and edited those recordings in Reaper to make these.

baconpaul:
Ahh right so "you recorded them using that free synth" - that's fine. let me merge it.

PR #6932, opened 2023-04-12
make ringmod have a deform type for other 2 input logic functions

morganholly:
also adds __restrict to other functions in basic_dsp.h, should somewhat optimize them

### DIFF
--- a/src/common/SurgeStorage.h
+++ b/src/common/SurgeStorage.h
@@ -178,6 +178,16 @@ enum NoiseColorChannels
     MONO = 1
 };
 
+enum CxorMode
+{
+    RINGMOD         = 0,
+    CXOR            = 1,
+    CXOR_F1         = 2,
+    CXOR_F2         = 3,
+    CXOR_F3         = 4,
+    CXOR_F4         = 5
+};
+
 enum lfo_trigger_mode
 {
     lm_freerun = 0,
--- a/src/common/SurgeStorage.h
+++ b/src/common/SurgeStorage.h
@@ -108,7 +108,15 @@ const int FIRoffsetI16 = FIRipolI16_N >> 1;
 //                                   oddsound_as_mts_main
 // clang-format on
 
-const int ff_revision = 21;
+const int ff_revision = 22;
+
+if (revision <= 21)
+    for (auto sc = 0; sc < n_scenes; ++sc)
+    {
+        scene[sc].mixer.ringmod_12.deform_type = 0;
+
+        scene[sc].mixer.ringmod_23.deform_type = 0;
+    }
 
 const int n_scene_params = 273;
 const int n_global_params = 11 + n_fx_slots * (n_fx_params + 1); // each param plus a type
--- a/src/common/SurgeStorage.h
+++ b/src/common/SurgeStorage.h
@@ -178,14 +178,14 @@ enum NoiseColorChannels
     MONO = 1
 };
 
-enum CxorMode
+enum RingModMode
 {
-    RINGMOD = 0,
-    CXOR = 1,
-    CXOR_F1 = 2,
-    CXOR_F2 = 3,
-    CXOR_F3 = 4,
-    CXOR_F4 = 5
+    rmm_ring = 0,
+    rmm_cxor = 1,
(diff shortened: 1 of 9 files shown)


src/common/dsp/SurgeVoice.cpp
```diff
@@ -992,13 +992,63 @@ bool SurgeVoice::process_block(QuadFilterChainState &Q, int Qe)
     {
         if (is_wide)
         {
-            mul_block(osc[0]->output, osc[1]->output, tblock, BLOCK_SIZE_OS_QUAD);
-            mul_block(osc[0]->outputR, osc[1]->outputR, tblockR, BLOCK_SIZE_OS_QUAD);
+            switch (scene->level_ring_12.deform_type) {
```
baconpaul: So my primary thought here is: Do we want to add a function like `SurgeVoice::ringModSources(o1L, o1R, o2L, o2R, rR, rL, is_wide, type)` so we don't have to copy this block twice?

morganholly: yeah probably worth doing

morganholly: should that function be in basic_dsp.h along with what it calls?

baconpaul: no i would make it a new member on SurgeVoice which calls the basic_dsp methods myself.

baconpaul: basic_dsp is really block operations. The switch and is wide check belongs in the voice is my thinking.

morganholly: working on this now

src/common/SurgeStorage.h
```diff
@@ -178,6 +178,16 @@ enum NoiseColorChannels
     MONO = 1
 };
 
+enum CxorMode
```
mkruselj, line 181: I would name this RingModMode instead.

src/common/SurgeStorage.h
```diff
@@ -178,6 +178,16 @@ enum NoiseColorChannels
...
+{
+    RINGMOD = 0,
```
mkruselj, line 183: Also we generally use all lowercase for enums. You'll want these something like:

rmm_default,
rmm_cxor1,
rmm_cxor2,
etc.

src/surge-xt/gui/SurgeGUIEditorValueCallbacks.cpp
```diff
@@ -2350,6 +2350,63 @@ int32_t SurgeGUIEditor::controlModifierClicked(Surge::GUI::IComponentTagValue *c
...
                         }
                     }
+                    break;
+                    case ct_amplitude_ringmod:
+                    {
+                        contextMenu.addSeparator();
+                        auto dt = p->deform_type;
+
+                        Surge::Widgets::MenuCenteredBoldLabel::addToMenuAsSectionHeader(
+                            contextMenu, "COMBINATOR MODE");
```
mkruselj, line 2360: "RING MODDULATOR MODE"

Let's name it like this for now since the GUI refers to those sliders as RM and all the documentation etc.

src/surge-xt/gui/SurgeGUIEditorValueCallbacks.cpp
```diff
@@ -2350,6 +2350,63 @@ int32_t SurgeGUIEditor::controlModifierClicked(Surge::GUI::IComponentTagValue *c
...
+
+                        contextMenu.addItem(
+                            "Ringmod (classic)", true, dt == CxorMode::RINGMOD, [this, p]() {
```
mkruselj, line 2363: So my suggestion for namese would simply be:

Default (multiply)
Continuous XOR 1...5

That's all!

src/common/SurgeStorage.h
morganholly: yea i thought of that a min ago

src/common/SurgeStorage.h
morganholly: ah i wasn't sure, tried to make it fit with the others

PR #6938, opened 2023-04-16
Open sound control 2355

pkstone:


src/common/Parameter.h
```diff
@@ -466,6 +466,12 @@ class Parameter
     void create_fullname(const char *dn, char *fn, ControlGroup ctrlgroup, int ctrlgroup_entry,
                          const char *lfoPrefixOverride = nullptr) const;
 
+    // Optional OSC name
+    std::string get_OSC_name();
+    void setOSCName(const std::string &s);
+    bool hasOSCName{false};
+    std::string oscName{};
```
mx, line 473: No need for default construction, std::string defaults to empty. So std::string oscName is fine.

src/common/SurgePatch.cpp
```diff
@@ -632,6 +632,24 @@ SurgePatch::SurgePatch(SurgeStorage *storage)
             Surge::Formula::createInitFormula(fs);
         }
 
+    // Build table of param_ptr -- storage name
+    for (auto *p : param_ptr)
+        param_ptr_by_storagename[p->get_storage_name()] = p;
+
+    /* ---- Set any OSC names here  ---- */
```
mx, line 639: Are these comments supposed to be here? There's an example function call given but I'm not sure when we'd use this.

src/common/SurgePatch.cpp
```diff
@@ -1221,6 +1240,26 @@ unsigned int SurgePatch::save_patch(void **data)
     return psize;
 }
 
+Parameter *SurgePatch::parameterFromStorageName(std::string stName)
```
mx, line 1243: This should take a std::string_view instead of a std::string (the map can still be std::string, but the method parameter should be std::string_view).

src/common/SurgePatch.cpp
```diff
@@ -1221,6 +1240,26 @@ unsigned int SurgePatch::save_patch(void **data)
...
+    if (it != std::end(param_ptr_by_storagename))
+    {
+        return it->second;
+    }
+    else
+    {
+        return nullptr;
+    }
+}
+Parameter *SurgePatch::parameterFromOSCName(std::string stName)
```
mx, line 1255: Likewise, this should take a std::string_view.

src/surge-xt/SurgeSynthProcessor.cpp
```diff
@@ -182,6 +196,19 @@ const juce::String SurgeSynthProcessor::getProgramName(int index)
 
 void SurgeSynthProcessor::changeProgramName(int index, const juce::String &newName) {}
 
+/* OSC (Open Sound Control) */
+bool SurgeSynthProcessor::initOSC(int port) {
```
mx, line 200: The caller isn't doing anything with the return value here; should it be?

src/surge-xt/SurgeSynthProcessor.cpp
```diff
@@ -452,6 +479,25 @@ void SurgeSynthProcessor::processBlockMidiFromGUI()
     }
 }
 
+void SurgeSynthProcessor::processBlockOSC()
+{
+    while (true)
```
mx, line 484: So, this is why the RingBuffer class has a popall. Suggest a slight logic change.

```
auto messages = oscRingBuf.popall();
for (const auto& m : messages) {
  float pval = m.val;
  // etc...
}
```

src/surge-xt/SurgeSynthProcessor.h
```diff
@@ -14,6 +14,9 @@
 #include "SurgeStorage.h"
 #include "util/LockFreeStack.h"
 
+#include "stmlib/utils/ring_buffer.h"
```
mx, line 17: This is a superfluous include; you're using the SST ringbuffer not this.

src/surge-xt/SurgeSynthProcessor.h
```diff
@@ -328,6 +332,28 @@ class SurgeSynthProcessor : public juce::AudioProcessor,
...
+        } type{PARAMETER};
+        Parameter *param;
+        float val{0.0};
+        std::string str;
+
+        oscMsg() {}
+        oscMsg(Parameter *p, float v) : type(PARAMETER), param(p), val(v) {}
+    };
+
+    // LockFreeStack<oscMsg, 4096> oscRingBuf;
```
mx, line 352: Remove this comment.

src/surge-xt/gui/SurgeGUIEditor.cpp
```diff
@@ -4733,6 +4737,73 @@ juce::PopupMenu SurgeGUIEditor::makeMidiMenu(const juce::Point<int> &where)
...
+    oscSubMenu.addSeparator();
+    oscSubMenu
+        .addItem(
+            Surge::GUI::toOSCase("Change OSC port (" + std::to_string(portnum) + ")"),
+            [this, portnum]() {
+                const auto c{std::to_string(portnum)};
+                promptForMiniEdit(
+                    c, "Enter a new value:", "OSC Port #", juce::Point<int>(10, 10),
+                    [this](const std::string &c) {
+                        int newPort = ::atoi(c.c_str());
```
mx, line 4760: This is a little dangerous. They could enter any value and atoi can fail.

You can use std::stoi directly (no need for atoi(c.c_str()). But you'll want to catch any exception that is thrown, and either message the user that they need to provide a valid port number or leave a default value in newPort.

Also, port numbers max out at 65535 and min out at 0. So make sure you're checking for that max value. As it is right now, a user can enter any valid integer value, even if it's not a valid port number.

src/surge-xt/gui/SurgeGUIEditor.cpp
```diff
@@ -4733,6 +4737,73 @@ juce::PopupMenu SurgeGUIEditor::makeMidiMenu(const juce::Point<int> &where)
...
+    if (alreadyListening) {
+         oscSubMenu.addItem(Surge::GUI::toOSCase("Stop OSC listener"),
+            [this]() {
+                juceEditor->processor.oscListener.stopListening();
+            });
+   } else {
+         oscSubMenu.addItem(Surge::GUI::toOSCase("Start OSC listener"), [this]() {
+             int defaultOSCPort = Surge::Storage::getUserDefaultValue(
+                 &(this->synth->storage), Surge::Storage::OSCPort, DEFAULT_OSC_PORT);
+             std::cout << "default osc port: " << defaultOSCPort;
```
mx, line 4789: This looks like a diagnostic message and should be removed.

src/surge-xt/gui/SurgeGUIEditorHtmlGenerators.cpp
```diff
@@ -411,6 +412,223 @@ th {
...
+        if (st_str[1] == '_')
+        {
+            if (st_str[0] == 'b')
+                continue; // 'b_...' entries not added to vector
+            else if (st_str[0] == 'a')
+            {
+                st_str[0] = '*';
+            }
+        }
+        sortvector.push_back(new oscParamInfo(p, st_str, p->get_full_name(), p->ctrlgroup));
```
mx, line 551: This is a memory leak. You're instantiating new oscParamInfos, but you're never deleting them.

Suggest you do not use "new" at all. Instead, sortVector should simply be: std::vector<oscParamInfo> sortVector.

Then do not create "new oscParamInfo(...)", but instead just "oscParamInfo(...)".

src/surge-xt/osc/OSCListener.cpp
```diff
@@ -0,0 +1,122 @@
...
+
+#include "OSCListener.h"
+#include "Parameter.h"
+#include "SurgeSynthProcessor.h"
+#include <sstream>
+#include <vector>
+#include <string>
+
+
+using namespace Surge::OSC;
```
mx, line 24: All this stuff is supposed to be inside of Surge::OSC. So instead, what you need to do is this:

namespace Surge {
namespace OSC {

// ... rest of the file

}  // namespace OSC
}  // namespace Surge

mx (changes requested):
Mostly minor stuff.

src/common/SurgePatch.cpp
pkstone: I think that's the mechanism that will be used to splice in OSC names on an as-needed basis. I could be wrong about that, though.

src/common/SurgePatch.cpp
pkstone: See comment on parameterFromStorageName.

src/surge-xt/gui/SurgeGUIEditor.cpp
pkstone: Now wrapped with #ifdef DEBUG

src/surge-xt/gui/SurgeGUIEditorHtmlGenerators.cpp
pkstone: 'std::vector sortvector;' gives 'no instance of constructor "std::__1::vector" matches the argument list'

src/common/SurgePatch.cpp
pkstone: Keeping as 'string' for now.

src/surge-xt/gui/SurgeGUIEditorHtmlGenerators.cpp
pkstone: I've changed the comparator, but still get the same error on the 'std::vector sortvector;' declaration. I don't get how it can be called without the template arg.

baconpaul: I think github mangled the comment and he meant std vector open type close

I can’t do backticks on my phone. But change the vector from having type be a foo* to a foo and drop the new

pkstone:
Heh. I just figure that out, thank you!

From: Paul ***@***.***>
Date: Sunday, April 16, 2023 at 11:44 AM
To: surge-synthesizer/surge ***@***.***>
Cc: Philip Stone ***@***.***>, Author ***@***.***>
Subject: Re: [surge-synthesizer/surge] Open sound control 2355 (PR #6938)

@baconpaul commented on this pull request.

________________________________

In src/surge-xt/gui/SurgeGUIEditorHtmlGenerators.cpp<https://github.com/surge-synthesizer/surge/pull/6938#discussion_r1167998067>:

> +    std::string st_str;

+    int currentCtrlGrp = endCG;

+

+    for (auto *p : synth->storage.getPatch().param_ptr)

+    {

+        st_str = p->get_OSC_name();

+        if (st_str[1] == '_')

+        {

+            if (st_str[0] == 'b')

+                continue; // 'b_...' entries not added to vector

+            else if (st_str[0] == 'a')

+            {

+                st_str[0] = '*';

+            }

+        }

+        sortvector.push_back(new oscParamInfo(p, st_str, p->get_full_name(), p->ctrlgroup));

I think github mangled the comment and he meant std vector open type close

I can’t do backticks on my phone. But change the vector from having type be a foo* to a foo and drop the new

src/common/SurgePatch.cpp
mkruselj: You're not wrong about that 🙂

src/surge-xt/gui/SurgeGUIEditorHtmlGenerators.cpp
pkstone: Fixed.

src/surge-xt/SurgeSynthProcessor.cpp
pkstone: Added proper error handling/notification.

src/surge-xt/gui/SurgeGUIEditor.cpp
pkstone: Did sanity check on user input, with error reporting.

mx (changes requested):

src/surge-xt/SurgeSynthProcessor.cpp
pkstone: Oops, you're right. Good catch.

src/common/SurgePatch.cpp
```diff
@@ -675,6 +693,7 @@ SurgePatch::SurgePatch(SurgeStorage *storage)
 #endif
 }
 
+
```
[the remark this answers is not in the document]
pkstone: I don't get this one; I only see one blank line between the functions

src/surge-xt/SurgeSynthProcessor.cpp
```diff
@@ -478,6 +505,22 @@ void SurgeSynthProcessor::processBlockMidiFromGUI()
     }
 }
 
+void SurgeSynthProcessor::processBlockOSC()
+{
+    auto messages = oscRingBuf.popall();
+    for (const auto &om : messages)
+    {
+        float pval = om.val;
+        if (om.type == oscMsg::PARAMETER)
```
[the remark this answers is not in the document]
pkstone: Yes. More to come, but only /param/.. for now.

baconpaul:
Happy to merge this and clean this up later, but if you *are* going to make another change can we change 

```
#ifdef SURGE_HAS_OSC
```

to

```
#if SURGE_HAS_OSC
```

which means we can make this 0/1. Then change `globals.h` to say

```
#ifndef SURGE_HAS_OSC
#define SURGE_HAS_OSC 1
#endif
```

that way I can compile clients with `-DSURGE_HAS_OSC=0` to turn of OSC support, which I might need for the VCV Rack implementation or others perhaps.

If this is the only change just leave it for now and we can do it when we face that submodule update for rack in the future. But if you have other diffs getting this one too would be lovely.
Thanks

src/common/globals.h
```diff
@@ -60,7 +60,7 @@ static inline int _stricmp(const char *s1, const char *s2) { return strcasecmp(s
 #error You must compile with -DSURGE_COMPILE_BLOCK_SIZE=32 (or whatnot)
 #endif
 
-#define SURGE_HAS_OSC
+#define SURGE_HAS_OSC 1
```
mx, line 63:
```suggestion
#ifndef SURGE_HAS_OSC
#define SURGE_HAS_OSC 1
#endif
```

mx (changes requested):

mx (approved):

PR #6940, opened 2023-04-16
Add a dependency on sst-basic-blocks; move Lanczos Resapmler

baconpaul:
The Lanczos Resapmler is a useful small kernel resampler which is well tested and owrking. Move it to sst-basic-blocks and make surge proper depend on this shared module.


baconpaul:
lets merge this after the OSC stuff gets all merged to avoid conflicts for phil and surgo

PR #6948, opened 2023-04-17
Fix pitch bend becoming stuck when toggling MPE mode during held bend note

### LINKED ISSUE #6490: Enabling MPE mode while a pitch bend is active causes the pitch bend to stick forever
**Bug Description:**
Enabling MPE mode while a pitch bend is active causes Surge to remain detuned, because in MPE mode pitch bend is handled differently. Surge will remain detuned.

**Surge XT Version**
This information is found on the About screen, which you get to from the bottom right menu.

Version: Surge XT 1.1.HEAD.beab6b17 (filter lib wasn't clean - rest of build is)
Build: 2022-07-26 @ 15:56:33 on 'rainbow/local' with 'Clang-14.0.6' using JUCE 6.1.6
System: Linux 64-bit VST3 on  AMD Ryzen 7 2700X Eight-Core Processor
Host: Ardour @ 48.0 kHz

**Reproduction Steps:**
1. While MPE mode is disabled, hold a note and apply pitch bend.
2. While the note is held and pitch bend is applied, enable MPE mode.
3. You can now stop doing pitch bend/
4. Surge is still detuned?

**Expected Behavior:**
Maybe when we toggle MPE mode we should reset pitch bend.

mkruselj:
Closes #6490

### DIFF
--- a/src/surge-xt/SurgeSynthEditor.cpp
+++ b/src/surge-xt/SurgeSynthEditor.cpp
@@ -585,6 +585,7 @@ void SurgeSynthEditor::setPitchModSustainGUI(int pitch, int mod, int sus)
     auto pw = dynamic_cast<VKeyboardWheel *>(pitchwheel.get());
     if (pw)
     {
+        printf("%d\n", pitch);
         pw->value = pitch;
         pw->repaint();
     }
--- a/src/surge-xt/SurgeSynthEditor.cpp
+++ b/src/surge-xt/SurgeSynthEditor.cpp
@@ -585,7 +585,6 @@ void SurgeSynthEditor::setPitchModSustainGUI(int pitch, int mod, int sus)
     auto pw = dynamic_cast<VKeyboardWheel *>(pitchwheel.get());
     if (pw)
     {
-        printf("%d\n", pitch);
         pw->value = pitch;
         pw->repaint();
     }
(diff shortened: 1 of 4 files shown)


src/surge-xt/SurgeSynthEditor.cpp
```diff
@@ -585,6 +585,7 @@ void SurgeSynthEditor::setPitchModSustainGUI(int pitch, int mod, int sus)
     auto pw = dynamic_cast<VKeyboardWheel *>(pitchwheel.get());
     if (pw)
     {
+        printf("%d\n", pitch);
```
baconpaul, line 588: You left a printf here

src/surge-xt/gui/SurgeGUIEditor.cpp
```diff
@@ -2697,11 +2697,15 @@ long SurgeGUIEditor::unapplyParameterOffset(long id) { return id + start_paramta
 void SurgeGUIEditor::toggleMPE()
 {
     this->synth->mpeEnabled = !this->synth->mpeEnabled;
+
     if (statusMPE)
     {
+        this->synth->resetPitchBend(-1);
+
```
baconpaul, line 2704: Do we want to reset if we toggle on also? So always do a pitch bend reset when changing between modes?

src/surge-xt/SurgeSynthEditor.cpp
mkruselj: Doh!

src/surge-xt/gui/SurgeGUIEditor.cpp
mkruselj: Yeah I think it's fine in both cases tbh. When toggling on is what tripped Luna up, but it's fine to have it when toggling off too.

Issue #6946, opened 2023-04-17
Consolidate basic_dsp


baconpaul:
We have basic-dsp in

surge/src/dsp
sst-filters filters/utilties
shortcircuit src/vembertech

and who even knows which functions are called

SC has started moving them to basic-blocks. Do the rest.

PR #6953, opened 2023-04-18
Move the lions share of basic_dsp to basic-blocks

baconpaul:
In prepartion for moving out the FX and more sharing with SC, move a big chunk of basic_dsp to sst-basic-blocks submodule, add tests there, port, etc...

Addresses #6946


baconpaul:
**DO NOT MERGE**

I want to do a close review after stepping away from it a bit.

PR #6971, opened 2023-04-21
Uniform code comment header and include guards

baconpaul:
1. Add a uniform code header in all cpp and h files under src/
2. Use a uniform header guard (no more pragma once; well defined unique names)
3. Add scripts to make sure we can do this again any time we want


baconpaul:
please let me merge this - i want to rewrite the comment tx.

PR #6980, opened 2023-04-22
Renamed pretty much all OSC param names

mkruselj:
Sideeffect: Renamed Level parameter in mixer to Volume to be consistent with scene output and global volume params.

Also removed setOSCName() method and hasOSCName member, as a consequence. The way things are implemented here completely obsoleted this method.

### DIFF
--- a/src/common/Parameter.cpp
+++ b/src/common/Parameter.cpp
@@ -164,7 +164,7 @@ void Parameter::set_name(const char *n)
 }
 
 Parameter *Parameter::assign(ParameterIDCounter::promise_t idp, int pid, const char *name,
-                             const char *dispname, int ctrltype,
+                             const char *dispname, const std::string &altOSCname, int ctrltype,
 
                              const Surge::Skin::Connector &c,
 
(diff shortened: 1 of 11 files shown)


pkstone (commented):
In my work this morning, I made the tokens array (in OSCListener::oscMessageReceived()) zero based like this:

   std::getline(split, each, '/');
    for (std::string each; std::getline(split, each, '/'); tokens.push_back(each)) ;

then all the tokens[] references start from zero (i.e. all have to be decremented by one in the code).

This is purely fiddly, but thought I'd mention it. I never liked that 'first token will be blank' comment. :-)

mkruselj:
> In my work this morning, I made the tokens array (in OSCListener::oscMessageReceived()) zero based like this:
> 
> std::getline(split, each, '/'); for (std::string each; std::getline(split, each, '/'); tokens.push_back(each)) ;
> 
> then all the tokens[] references start from zero (i.e. all have to be decremented by one in the code).
> 
> This is purely fiddly, but thought I'd mention it. I never liked that 'first token will be blank' comment. :-)

That sounds great! Could merge that work after this PR, then?

pkstone:
Yes, except I'm midstream on other stuff so I'll have to be careful about what I push.

PR #6981, opened 2023-04-22
Pitch slider set to 0 in Absolute mode now tunes correctly with Scala loaded

### LINKED ISSUE #6978: Absolute mode on pitch is also a tiny bit borked with scl/kbm.
Bring up two modern saw oscs, and load any unequal tuning with scl (kbm not necessary).

Make one of them absolute pitch, but keep its pitch at 0. 

Play some different notes. You'll hear the root of the scale is perfectly in tune, but there's a small amount of beating on others, indicating that absolute mode makes the osc interpret tuning very slightly wrong. Note that this happens even in "apply tuning at MIDI input" mode.

Andreya-Autumn:
Paul's fix worked. :)

Fixes #6978

### DIFF
--- a/src/common/dsp/SurgeVoice.h
+++ b/src/common/dsp/SurgeVoice.h
@@ -108,9 +108,14 @@ class alignas(16) SurgeVoice
     ** But in absolute mode you need to find the virtual note which would
     ** map to that frequency shift.
     */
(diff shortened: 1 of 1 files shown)


Andreya-Autumn:
Adresses #6978

PR #6988, opened 2023-04-23
rename cxor and other functions

morganholly:
SILM is scale invariant linear modulator


morganholly:
additionally the descriptors for each function have been updated

PR #6996, opened 2023-04-26
Ignore playhead if it exists in standalone

mkruselj:
This makes any tempo syncable stuff work in standalone, and makes the Tempo field in VKB also work as expected.


baconpaul:
LGTM

PR #7002, opened 2023-04-29
Open sound control 2355

pkstone:
OSC enhancements: /tuning/scl, /tuning/kbm, /settings/path/scl, and /settings/path/kbm added. OSC html page updated.


src/common/SurgeStorage.cpp
```diff
@@ -2269,6 +2269,36 @@ bool SurgeStorage::resetToCurrentScaleAndMapping()
     return true;
 }
 
+void SurgeStorage::loadTuningFromSCL(const fs::path &p)
+{
+    try
+    {
+        retuneToScale(Tunings::readSCLFile({p.u8string()}));
```
mx, line 2276: I don't think you need the braced initialization here (and below).

src/common/SurgeStorage.cpp
```diff
@@ -2269,6 +2269,36 @@ bool SurgeStorage::resetToCurrentScaleAndMapping()
...
+
+void SurgeStorage::loadMappingFromKBM(const fs::path &p)
+{
+    try
+    {
+        remapToKeyboard(Tunings::readKBMFile({p.u8string()}));
+    }
+    catch (Tunings::TuningError &e)
+    {
+        remapToConcertCKeyboard();
```
mx, line 2295: Why does the above remap to 12TET on error, and this one to ConcertC? Should probably be consistent.

src/surge-xt/gui/SurgeGUIEditorHtmlGenerators.cpp
```diff
@@ -446,8 +446,10 @@ struct oscParamInfo
 // Sort function for displaying parameters (below)
 bool compareParams(const oscParamInfo &opl, const oscParamInfo &opr)
 {
-    int lcg = opl.p->ctrlgroup;
-    int rcg = opr.p->ctrlgroup;
+    int cgroup_order[10] = {0, 1, 2, 4, 3, 5, 6, 7, 8, 9};
+
+    int lcg = cgroup_order[opl.p->ctrlgroup];
+    int rcg = cgroup_order[opr.p->ctrlgroup];
```
mx, line 452: I don't understand the necessity of cgroup_order here. It looks like it's mapping 0-9 to 0-9.

src/surge-xt/gui/SurgeGUIEditorHtmlGenerators.cpp
```diff
@@ -652,6 +754,7 @@ code {
                         </table>
                     </div>
                 </div>
+                </div>
```
mx, line 757: Is this right? It's the same indent level as the one above it.

mx (changes requested):

src/common/SurgeStorage.cpp
pkstone: This one is a keyboard mapping (.kbm) which is distinct from a plain tuning (.scl). Its default is 'ConcertCKeyboard()'.

src/surge-xt/gui/SurgeGUIEditorHtmlGenerators.cpp
pkstone: Four and three are transposed. It was just a way to arrange the control groups on the HTML with a better flow.

src/surge-xt/gui/SurgeGUIEditorHtmlGenerators.cpp
pkstone: Looks like it's an extra. I removed it.

src/common/SurgeStorage.cpp
pkstone: Will fix.

mx (approved):

PR #7003, opened 2023-04-29
Create a new option for overlays to not initial-grab keyboard focus.

### LINKED ISSUE #6927: the Time Scaling slider on the Oscilloscope is active when opened
happens for all skins

image
image

it dismisses only after the slider has been budged or focus changed (for example if any item is pressed/moused down), then it paints mouse overs and gestures as it should

mx:
Exempt the oscilloscope from initial-grab keyboard focus.

Fixes #6927


baconpaul:
Looks good to me! If it’s ready merge away and I can sweep into 121 once we have a hash

PR #7025, opened 2023-05-16
fix: include correct headers for gcc 13

grawlinson:
Due to changes in upstream [gcc][0], surge no longer builds from source (a.k.a. FTBFS).

This patch adds the required headers throughout the project, thankfully it is only one location!

[0]: https://gcc.gnu.org/gcc-13/porting_to.html#header-dep-changes


baconpaul:
Thanks!

PR #7027, opened 2023-05-21
Polymodfix

Xenakios:
This hopefully fixes the somewhat esoteric use case where polyphonic Clap parameter modulation is to be applied per channel or per key, instead of per note id.

Please disregard the additional old commits, only the latest one applies. My fork repo has some kind of problem that adds additional commits.


baconpaul:
Thanks
I’ll cherry pick this onto a clean branch run format and merge it later tonight or tomorrow unless someone beats me to it,

baconpaul:
In #7028

Issue #5663, opened 2021-12-23
Make juce::AlertWindow look like our miniedits


mkruselj:
So that styling and `dialog` skin colors are all picked up the same way.

This means they should have a titlebar just like miniedits and overlays do, the whole nine yards with Surge icon in top left etc. What is currently the headline of AlertWindow should be printed centered in the titlebar.

baconpaul:
When we do this, give them a component argument too so they center over the screen. This will fix Jacky's problem that a popup opened on primary when he was working on secondary.

baconpaul:
So the way to do this is with an AlertOverlay basically and dump the juce alert windows.

baconpaul:
So here’s My thought 

1. copy the typein overlay we use for prompt for mini edit 
2. Replace all the calls to juce alert box with a call to sge-> show alert 
3. Make that call show overlay
4. Style accordingly 

along the way you may want a developer menu option to call the function in the various modes so you can test without causing an actual error 

so Basically dup the about or prompt modal overlays for this purpose

baconpaul:
So this is actually quite a bit of work if we are going to do all the accesibility work also. Shall we move this out of the milestone?

mkruselj:
I still haven't given up on doing this, just was busy on NI side of things...

PR #7037, opened 2023-05-29
Use custom alert window for deleting user patches (#5663)

tejas1993:
Following up from discord, as a first step for https://github.com/surge-synthesizer/surge/issues/5663 . I figured we could start with a single use case and migrate the rest after.  I used MiniEdit as a guide for Alert. 

Since it's my first pull request for surge, let me know if there's any process I haven't followed. Attached screenshot of the alert box.

Testing: 
User testing by creating/deleting patches multiple times. Checked functionality for both Okay and Cancel. 

Surge Delete Patch Alert

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -39,6 +39,7 @@ Jarkko Sakkinen <jarkko.sakkinen@iki.fi>
 Jeremy Sagan <info@sagantech.biz>
 Chad Scarborough <github:ChadScarborough>
 Sebastian Schwarzhuber <https://github.com/cannero>
+Tejas Singh <https://github.com/tejas1993>
 Jeff Smith <whydoubt@gmail.com>
 Jon Spruyt <http://jonspru.yt>
 Phil Stone <https://github.com/pkstone/>
--- a/src/surge-xt/CMakeLists.txt
+++ b/src/surge-xt/CMakeLists.txt
@@ -111,6 +111,8 @@ target_sources(${PROJECT_NAME} PRIVATE
   gui/overlays/MSEGEditor.h
   gui/overlays/MiniEdit.cpp
   gui/overlays/MiniEdit.h
+  gui/overlays/Alert.cpp
+  gui/overlays/Alert.h
   gui/overlays/ModulationEditor.cpp
   gui/overlays/ModulationEditor.h
   gui/overlays/Oscilloscope.cpp
--- a/src/surge-xt/gui/SurgeGUIEditor.cpp
+++ b/src/surge-xt/gui/SurgeGUIEditor.cpp
@@ -43,6 +43,7 @@
 
 #include "overlays/AboutScreen.h"
 #include "overlays/MiniEdit.h"
+#include "overlays/Alert.h"
 #include "overlays/MSEGEditor.h"
 #include "overlays/LuaEditors.h"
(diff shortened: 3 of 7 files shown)


mkruselj:
Several comments from my end:

1. Somehow your IDE is not picking up our clangformat file, or is not set up to apply formatting on file save. See if you can set that up and push a commit with formatting updated.
2. We will probably want to add some margin around the text so that it doesn't almost reach right up to the borders. Same for OK/Cancel buttons. See the MPE pitch bend range miniedit as an example:

image


As for the procedure: please add yourself to AUTHORS file along with this PR. Thank you for the contribution and see you on Discord! 🙂

baconpaul:
OK this is excellent. Thank you for contributing. There's a few things

1. The code looks fantastic. There's one change I would make though. You replaced `juce::AlertWindow::showOkCancelBox(` with `sge->alertWindow`. SGE is exactly the right place to put it but if in the future we add other alert boxes (OK only, informational, etc...) we will want some way to communicate a flavor. So I would suggest either (1) have `sge->alertOKCancel` or (2) have `AlertOverlay` have an enum for window styles. Usually I would do the second but that will make the Alert header included very broadly so I think I would do the first; or perhaps have SGE own the enum. Something along those lines
3. UI: I thought the standard OK / Cancel box on most UIs had the buttons right aligned and close to the bottom not centered. But I could be wrong. Happy to merge without changing that and have us iterate though.
4. Looking at your diff there were a couple of bug-like things in the old code I think we may want to fix. `fs::remove_file` can throw an `fs::exception` I think so we should catch that and ignore it silently. And the lambda that binds `this` can get upside down if the ui element gets deleted. I think that's fine here but I've been migrating our JUCE lambdas slowly from `[this]() { this->foo; }` to `[w = juce::Component::SafePointer(this)]() { if (!w) return; w->foo; }` since that interacts with the UI lifetime in the event a window closes etc...
5. Any consideration of an `onCancel` function which defaults to nullptr? We obviously don't need it here though and can add it later if we do.

Tiny / administrative things

1. Our formatting check failed. If you can just run clang-format over all the files that would fix that immediately. (If you need help setting up clang format, just let us know your OS. On linux you can apt get it, on macOS you can brew install it, and I think on MSVC it installs if you install clang).
6. Can you add yourself to the AUTHORS file along with the PR?

Really the only ones absolutely key to merge are the clang-format check, the AUTHORS file, and the name of the SGE method at this point. The others are for consideration.

nd thank you again! Super appreciated.

tejas1993:
Thanks for the review!

Changes: 

1. Layout - added margins as suggested Surge Delete Patch Alert - with margins
2. Method name - changed as suggested to alertOkCancel
3. Exception handling for fs::filesystem_error - consistent with other parts of the code.
4. Formatting (though somehow running the formatter introduced a change that failed the format check again - I may have this incorrectly configured, will fix). 
5. Authors

For the other suggestions - I agree with all of them. I'll continue working on migrating other alerts to this overlay, add the additional functionality when needed and explore the options for standardizing paint for similar components. Could these be incremental changes in subsequent PRs?

baconpaul:
Yay! Thanks @tejas1993 

if you are on our discord (or you join) please let us know so we can give you the team role.

PR #7041, opened 2023-05-31
Migrate no icon alerts in SurgeGUIEditor and add OverlayUtils for drawing windows

tejas1993:
Following up on https://github.com/surge-synthesizer/surge/pull/7037 
Moving a couple more alertOkCancels to the custom overlay and adding utility for the code to draw the window as discussed in the above pull request. Having the Alert and MiniEdit overlays use this - though subsequently other overlays may use this as well. My preference for the implementation of this is as a utility method, but I'm happy to add an abstraction layer instead if you prefer. We could also add more common code for buttons etc to this in the next PRs.

Questions:

1. Thoughts on standardizing between "Ok"/"Cancel" and "Yes"/"No" for these? This PR uses only the former - if we want to definitely allow both, I'll update this PR. 
2. Could you point me to how to test all the changes for this? I love this synth but I am new to it so I've never used the functionality this PR edits. 

Additionally - I'll be travelling starting tomorrow mid-day Seattle/US time and will be back Monday. So it will be difficult for me to contribute during this time. Next week however, I should be able to continue work on this and look at some of the short circuit work pointed to in discord.

### DIFF
--- a/src/surge-xt/CMakeLists.txt
+++ b/src/surge-xt/CMakeLists.txt
@@ -113,6 +113,8 @@ target_sources(${PROJECT_NAME} PRIVATE
   gui/overlays/MiniEdit.h
   gui/overlays/Alert.cpp
   gui/overlays/Alert.h
+  gui/overlays/OverlayUtils.h
+  gui/overlays/OverlayUtils.cpp
   gui/overlays/ModulationEditor.cpp
   gui/overlays/ModulationEditor.h
   gui/overlays/Oscilloscope.cpp
--- a/src/surge-xt/gui/SurgeGUIEditor.cpp
+++ b/src/surge-xt/gui/SurgeGUIEditor.cpp
@@ -5520,7 +5520,7 @@ void SurgeGUIEditor::promptForMiniEdit(const std::string &value, const std::stri
 }
 
 void SurgeGUIEditor::alertOKCancel(const std::string &title, const std::string &prompt,
-                                   std::function<void()> onOk)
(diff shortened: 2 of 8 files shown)


baconpaul:
Thanks for the update on your travels. One of the joys of this project is we live a no deadlines no timelines lifestyle here but good to know! Safe travels also.

Toi test the first one, drop a skin directory (named foo.skin) onto surge. You can get that from the library and unzip it.

To test the second one, grab a zipped skin from the skin library https://surge-synthesizer.github.io/skin-library and drop the zip onto surge.


On the code: OverlayUtils.h needs a #ifdef guard around the code.

Thanks for doing this!

mkruselj:
Regarding OK/Cancel vs Yes/No, I feel like OK/Cancel should be used for loading operations and errors (well, errors should basically just have the OK button only usually), and Yes/No when the dialog is outright asking the user a question "do you want to xyz".

tejas1993:
Thanks for the testing instructions! 
MPE Pitch Bend
Save Skin Zip
Install Skin

baconpaul:
Looks good to me. You ready for us to merge this?

baconpaul:
oh at some point we may just want to add an `alertYesNo(string, string) { alertOKCancel(string, string, YESNO); }` one line alias in SGE.

tejas1993:
Yes good to merge! I'll add the alias in the next PR if it's ok.

baconpaul:
Perfect

PR #7047, opened 2023-06-14
Migrate additional AlertWindows to new UI

tejas1993:
Also added onCancel function for Alerts and alertYesNo alias as was discussed in previous pull requests.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,55 +1,55 @@
-[submodule "libs/tuning-library"]
-	path = libs/tuning-library
-	url = https://github.com/surge-synthesizer/tuning-library
-[submodule "libs/simde"]
-	path = libs/simde
-	url = https://github.com/simd-everywhere/simde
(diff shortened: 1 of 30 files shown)


mkruselj:
Lovely!!

tejas1993:
I think depends what the completion criteria is! This should be all the ok/cancel and yes/no alerts. 

There is https://github.com/surge-synthesizer/surge/blob/605b717d054c59049e045a7541aaa1449724b5c0/src/surge-xt/gui/SurgeGUIEditor.h#L842 which if I'm correct has a never-ask-again toggle 

and some message boxes eg.https://github.com/surge-synthesizer/surge/blob/605b717d054c59049e045a7541aaa1449724b5c0/src/surge-xt/gui/SurgeGUIEditor.cpp#L578

mkruselj:
Right ok so those can be sorted out in another PR, so let me merge this one. Thank you!

PR #7049, opened 2023-06-17
Implement multiple virtual keyboard layouts

### LINKED ISSUE #6688: Virtual keyboard shortcuts configuration
In Surge XT, the shortcuts configuration could be enhanced.

Example 1: The Surge XT manual precises: "Note: Some actions in Surge XT are done by doing Alt + Drag or scroll wheel. On certain Linux distributions, those shortcuts and gestures may cause conflicts between Surge and the desktop environment. We decided we couldn't give up alt-drag for all platforms because one window manager in one distribution used it by default. Therefore, it's often possible to disable that global gesture in desktop environments, and would be the easiest way to solve that possible issue."

In fact, when we use a desktop environment and other applications with their own requirements, wo don't always want to change our shortcuts for just one application, even if it is Surge XT that we love a lot!

On Linux, the key "Alt" in Surge could be changed by the "Windows" key for example but it should always be a user choice.

Example 2: On my desktop environment, the "Shift + F10" does not work and this combination can't be changed in the shortcuts configuration panel while it is an important shortcut.

Example 3: How can we configure the keys of the virtual keyboard? By trials and errors, I discovered that "}" and "c" are my keys to change the octave, that "a" give a C note, and so on. Perhaps i could be configured somewhere but I didn't found where.

Example 4: Some people use DVORAK or other keymap that fits their needs. All the shortcuts often needs to be configured to have a pleasant experience.

Long live to Surge XT! 

Configuration: Debian Linux, XFCE desktop environment, BÉPO keymap.

mkruselj:
Closes #6688

### DIFF
--- a/.github/ISSUE_TEMPLATE/release-checklist.md
+++ b/.github/ISSUE_TEMPLATE/release-checklist.md
@@ -18,16 +18,17 @@ Link to milestone: https://github.com/surge-synthesizer/surge/milestone/xx
 
 ## Executing the install
 
-- [ ] Update the releases-xt/azure-pipelines so we don't get a tag conflict when the release happens even 
+- [ ] Update the releases-xt/azure-pipelines so we don't get a tag conflict when the release happens even
       if it is just a bump commit
 - [ ] Create git markers in the surge repo
-  - [ ] Unlock your GPG key in case,running `gpg --output x.sig --sign CMakeLists.txt` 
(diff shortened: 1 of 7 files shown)


mkruselj:
Please don't merge, I will update the `setVKBLayout()` method with a cleaner single loop solution that utilizes `std::map`

PR #7058, opened 2023-06-19
OSC improvements: patch loading, OSC output (path of loaded patch)

pkstone:
Enhancements to OSC:

Patch changing via OSC is working, including full path specification, patch incr/decr, category incr/decr, and random patch change.

OSC output has been added. The first (and only, so far) message reports the path for patch changes.

All patch and tuning file paths have been made consistent in excluding file extension from OSC messages. They are added contextually in OSCListener.

### DIFF
--- a/src/common/SurgeSynthesizer.cpp
+++ b/src/common/SurgeSynthesizer.cpp
@@ -3945,11 +3945,15 @@ void loadPatchInBackgroundThread(SurgeSynthesizer *sy)
     {
         Patch p = synth->storage.patch_list[synth->patchid];
         pathstr = path_to_string(p.path);
+        // Trim extension
+        pathstr = pathstr.substr(0, pathstr.find_last_of("."));
         for (auto &it : synth->patchLoadedListeners)
             (it.second)(pathstr);
     }
     if (had_patchid_file)
     {
+        // Trim extension
+        pathstr = pathstr.substr(0, pathstr.find_last_of("."));
         for (auto &it : synth->patchLoadedListeners)
             (it.second)(pathstr);
     }
--- a/src/common/SurgeSynthesizer.cpp
+++ b/src/common/SurgeSynthesizer.cpp
@@ -3940,18 +3940,18 @@ void loadPatchInBackgroundThread(SurgeSynthesizer *sy)
     synth->patchChanged = true;
     synth->halt_engine = false;
 
-    // Notify the patch load listener(s)
+    // Notify the 'patch loaded' listener(s)
     if (patchid >= 0)
     {
         Patch p = synth->storage.patch_list[synth->patchid];
         pathstr = path_to_string(p.path);
-        for (auto &listener : synth->patchLoadedListeners)
-            (listener)(pathstr);
+        for (auto &it : synth->patchLoadedListeners)
+            (it.second)(pathstr);
     }
     if (had_patchid_file)
     {
-        for (auto &listener : synth->patchLoadedListeners)
-            (listener)(pathstr);
+        for (auto &it : synth->patchLoadedListeners)
+            (it.second)(pathstr);
     }
 
     // Now we want to null out the patchLoadThread since everything is done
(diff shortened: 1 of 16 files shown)


src/common/SurgeSynthesizer.cpp
```diff
@@ -3927,14 +3932,32 @@ void loadPatchInBackgroundThread(SurgeSynthesizer *sy)
...
     synth->patchChanged = true;
     synth->halt_engine = false;
 
+    // Notify the 'patch loaded' listener(s)
+    if (patchid >= 0)
+    {
+        Patch p = synth->storage.patch_list[synth->patchid];
+        pathstr = path_to_string(p.path);
+        // Trim extension
+        pathstr = pathstr.substr(0, pathstr.find_last_of("."));
```
baconpaul, line 3949: Turns out doing this in a string based way is a bit fragile. It works on our current OSes but you can use `filesystem::path.replace_extension("")` (like this)[https://en.cppreference.com/w/cpp/filesystem/path/replace_extension] before you convert to a string. (Generally, and mostly thanks to windows, string math on paths is hard because of UTF16 windows file system. We utf8ify early with path_to_string)

src/common/UserDefaults.h
```diff
@@ -163,8 +163,10 @@ enum DefaultKey
     DontShowAudioErrorsAgain,
 
     // OSC (Open Sound Control)
-    StartOSC,
-    OSCPort,
+    StartOSCIn,
+    StartOSCOut,
+    OSCPortIn,
+    OSCPortOut,
```
baconpaul, line 169: this is absolutely fine but if anyone used the old keys they may get an error with the new keys. Nightlies gonna nightly so no change required but if that happens the fix is reset a param or manually edit the xml to restore.

src/common/SurgeSynthesizer.h
```diff
@@ -385,6 +386,19 @@ class alignas(16) SurgeSynthesizer
     void prepareModsourceDoProcess(int scenemask);
     unsigned int saveRaw(void **data);
 
+    // ----  'patch loaded' listener(s) ----
+    // Listeners are notified whenever a patch changes, with the path of the new patch.
+    // Listeners should do any significant work on their own thread; for example, the OSC
+    // patchLoadedListener calls OSCSender::send(), which runs on a juce::MessageManager thread.
+    //
+    // Be sure to delete any added listeners in the desctructor of the class that added them.
+    std::unordered_map<std::string, std::function<void(std::string)>> patchLoadedListeners;
```
baconpaul, line 395: I guess for same reason we could have made this `std::function<void(const fs::path &)>` but lets leave it for now.

src/surge-xt/osc/OSCListener.cpp
```diff
@@ -129,61 +122,93 @@ void OSCListener::oscMessageReceived(const juce::OSCMessage &message)
...
+        {
+            std::string dataStr = getWholeString(message) + ".fxp";
+            {
+                std::lock_guard<std::mutex> mg(synth->patchLoadSpawnMutex);
+                strncpy(synth->patchid_file, dataStr.c_str(), FILENAME_MAX);
+                synth->has_patchid_file = true;
+            }
+            synth->processAudioThreadOpsWhenAudioEngineUnavailable();
+        }
+        else if (address2 == "random")
```
baconpaul, line 145: love it!

baconpaul (commented):
I'm happy to merge this and do a second quick cleanup on the one thing i noted if that works for you. Basically a few places (partly at my guidance) where we went to `std::string` from `fs::path` earlier than needed.

The code as written here will work fine though so 'merge and iterate' works for me. You option - let us know.

Oh and I didn't carefully review all the UI / SurgeGUIEditor bits.

src/common/SurgeSynthesizer.h
pkstone: Is <void(fs::path)> ok? I've been burned using pointers here.

baconpaul: You really don’t want to copy if.  & is wrong but const & is correct

PR #7059, opened 2023-06-19
Hide custom mod menu entry icons when adding a11y submenu

### LINKED ISSUE #6551: Modulations custom menu entry buttons stop working after enabling "Add sub-menus for modulation entries"
So, we are at a crossroad. To move further, we have to choose:

1. Hide the icons when that option is enabled (because a11y users don't care about icons anyways - easier route)
2. ~Make those icons work even if the option is enabled (cooler and potentially still useful for sighted users - more work route)~

mkruselj:
Fixes #6551

### DIFF
--- a/src/surge-xt/gui/widgets/MenuCustomComponents.cpp
+++ b/src/surge-xt/gui/widgets/MenuCustomComponents.cpp
@@ -24,6 +24,7 @@
 #include "SurgeImageStore.h"
 #include "SurgeImage.h"
 #include "UserDefaults.h"
+#include <fmt/format.h>
 #include <StringOps.h>
 
 namespace Surge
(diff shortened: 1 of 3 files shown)


src/surge-xt/gui/widgets/MenuCustomComponents.cpp
```diff
@@ -321,15 +347,17 @@ ModMenuCustomComponent::~ModMenuCustomComponent() noexcept = default;
 void ModMenuCustomComponent::getIdealSize(int &idealWidth, int &idealHeight)
 {
     auto menuText = source + " " + amount;
-    getLookAndFeel().getIdealPopupMenuItemSize(menuText, false, 20, idealWidth, idealHeight);
     auto h = idealHeight - 2 * mg;
+
+    getLookAndFeel().getIdealPopupMenuItemSize(menuText, false, 20, idealWidth, idealHeight);
```
baconpaul, line 352: Why this change in behavior? I don't understand it.

baconpaul (commented):
Just one question otherwise looks great

src/surge-xt/gui/widgets/MenuCustomComponents.cpp
mkruselj: Reverted it!

PR #7065, opened 2023-06-22
Migrate alerts with toggle buttons to surge ui

tejas1993:
[Issue](https://github.com/surge-synthesizer/surge/issues/5663)
Looked up some JUCE forums, they said that the appropriate way to do this was to edit the LookAndFeel. Happy to implement this in a different way too of course! 
Alert With Toggle unticked
alert with toggle ticked

### DIFF
--- a/src/surge-xt/gui/SurgeGUIEditor.cpp
+++ b/src/surge-xt/gui/SurgeGUIEditor.cpp
@@ -8313,13 +8313,6 @@ bool SurgeGUIEditor::promptForOKCancelWithDontAskAgain(
     const ::std::string &title, const std::string &msg, Surge::Storage::DefaultKey dontAskAgainKey,
     std::function<void()> okCallback, std::string ynMessage, AskAgainStates askAgainDef)
 {
-    if (okcWithToggleAlertWindow)
-    {
-        // This is not re-entrant, sorry
-        jassertfalse;
-        return false;
-    }
-
     auto bypassed =
         Surge::Storage::getUserDefaultValue(&(synth->storage), dontAskAgainKey, askAgainDef);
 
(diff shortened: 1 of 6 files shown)


mkruselj:
That looks good! Make sure that colors from SkinColors.cpp are used for checkbox background, border and tick, though - this doesn't seem to be the case here currently.

mkruselj:
Oh also if this is all the dialog types that we have left then we can link this PR with original issue so that it autocloses!

mkruselj:
Also, which skin are those screenshots showing? If it's Surge Classic, these colors don't match the skin colors for the dialog that are already being used by MiniEdit...

tejas1993:
Ah no - that wasn't classic. 

Alert with Toggle classinc

This is what surge classic looks like.

mkruselj:
Yup then it's sold!

So, that's now all dialog types covered? Wahey if true!

tejas1993:
It doesn't appear that this is true yet 😢 
I'll make the changes and see if I can knock out the others in this PR 

For context - 
1. There's a few message boxes -> alerts with no buttons basically: https://github.com/surge-synthesizer/surge/blob/4d4c97dc6a34c3f56e092212602d4b680297f7f6/src/surge-xt/gui/SurgeGUIEditor.cpp#L573 
2. https://github.com/surge-synthesizer/surge/blob/4d4c97dc6a34c3f56e092212602d4b680297f7f6/src/surge-xt/gui/SurgeGUIEditor.cpp#L3677 - which I'm not even sure how I missed. 

That should really be everything. 

Also - I have locked myself out of the alert box by telling it never to ask me again with the toggle. The good news is that it worked. Noob question - is there any way to undo that?

tejas1993:
Okay added the lines to onSkin changed and fixed no. 2 above. Need to stop for the day though - if it's merge-able right now, feel free to merge!
For the message boxes - just want to confirm, we want them migrated over too, yes?

mkruselj:
Yes. Everything should look consistent. :)

PR #7068, opened 2023-06-23
Migrate message boxes to new UI

### LINKED ISSUE #5663: Make juce::AlertWindow look like our miniedits

tejas1993:
Closes #5663.

Previous: 
Current message box

New: 
message box new

Question: Unless I'm missing something, exactly one message box remains: https://github.com/surge-synthesizer/surge/blob/21baada63f332f62b32df3ac2c794b3250e221b8/src/surge-fx/SurgeFXEditor.cpp#L598 
Since it's in surge-fx, I'll have to link the surge-xt project in the surge-fx makefile to use the work we've done so far? That seems like overkill for this, is there something I'm missing?

### DIFF
--- a/src/surge-xt/gui/SurgeGUIEditor.cpp
+++ b/src/surge-xt/gui/SurgeGUIEditor.cpp
@@ -8313,13 +8313,6 @@ bool SurgeGUIEditor::promptForOKCancelWithDontAskAgain(
     const ::std::string &title, const std::string &msg, Surge::Storage::DefaultKey dontAskAgainKey,
     std::function<void()> okCallback, std::string ynMessage, AskAgainStates askAgainDef)
 {
-    if (okcWithToggleAlertWindow)
-    {
-        // This is not re-entrant, sorry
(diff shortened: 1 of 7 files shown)


mkruselj:
We can leave the surge-fx thing alone, since that has quite a different GUI. So this basically means the issue on Surge synth side is done and dusted! 🙂

tejas1993:
Yup! [This issue](https://github.com/surge-synthesizer/surge/issues/5663) can close once this is merged

tejas1993:
Sooooo.... some errors say we can't have forward references to enums, and windows build is refusing to accept that Alert exists if we just include the header file. Hmmm, puzzling. 

I will try to fix/work around.

tejas1993:
There yah go.

PR #7070, opened 2023-06-26
Support MIDI learn over all channels, scene parameters are now learnable separately

### LINKED ISSUE #574: Support all 16 MIDI channels for MIDI learn (+UI/UX implementation suggestion)
This is a kickass feature from u-he plugins. They recognise the MIDI CC's channel when MIDI learn is being executed, so this allows supporting up to 128*16=2048 parameters, and in a leaner way than using NRPN.

They also support 14-bit controllers (so, CCs 0-31 as MSB matched by CCs 32-63 as LSB) for MIDI learn, allowing more precise control over synth's parameters (which can be important for things like oscillator tune or filter cutoff, etc.).

The way it is implemented UI-wise is much better (IMHO) than having submenus on a right-click. There is a "settings" button on the GUI (looks like a gearwheel) which you click and it immediately throws the synth in MIDI learn mode, accenting all controls which can be MIDI learned like so:

image

In case a mistake is made (or you want more options for your MIDI assignments, there is a MIDI table page, which looks like this:

image

And then these are the various options:

image

image

image

That last part is basically absolute ("continuous" vs relative ("encoder") control, depending on what type of hardware device you have (you would use "continuous" for faders or knobs, or, obviously, "encoder" for endless encoders).

### LINKED ISSUE #3582: Don't save MIDI learn to snapshot; don't MIDI learn both scenes; handle automation in Dual/Split etc properly
Hello, guys! 
Happy 2021 and my best wishes!

A strange bug was found.

Any MIDI CC assignment to any chosen parameter inside A or B scene is 'shared' between A and B scene in a very strange way. It's shown at this chosen parameter RMB menu on both A and B scenes but is working only at the scene that's currently active for editing (which is absolute nonsense when using this in dual mode). 

Moreover, this assignment cannot be deleted later. You delete it inside scene A, switch to scene B, and delete it once more. Then switch back to scene A - and ooops! - it's still there. I found no other way to get rid of such assignment than loading an empty CC map from general menu. 

Any internal assignments such as LFO or ADSR or direct parameter modulations from the host work fine. This bug is only with MIDI CC's.

Version/host/desktop info: 

Surge Synthesizer
Plugin Host:	REAPER
System:	Intel CPU, Windows VST3, 64-bit
Build Info:	Built on 2020-12-28 at 18:41:53, using pipeline host 'fv-az74-298'
Version:	1.8.nightly.185c107

**Reproduction Steps:**

1. Load SAW initial preset.
2. Set scene mode to Dual
3. Copy Scene A settings to Scene B
4. Make Scene A active for editing
5. Click RMB at any chosen scene parameter.
6. Set any MIDI CC automation.
7. Switch to  Scene B.
8. Click RMB at the same parameter.
9. See it's automated with the same MIDI CC too.
10. Try to make automation work. See it works only at active scene.
11. Try to delete this automation setting at scene A
12. Do the same at scene B
13. Switch to  Scene A.
14. Click RMB at the chosen parameter.
15. See it's still automated*

*Sometimes it may disappear but after unloading/reloading Surge or the whole DAW project it's still there.

mkruselj:
Closes #574, closes #3582.

### DIFF
--- a/src/common/SurgePatch.cpp
+++ b/src/common/SurgePatch.cpp
@@ -3104,6 +3104,10 @@ void SurgePatch::load_xml(const void *data, int datasize, bool is_preset)
                     ch = TINYXML_SAFE_TO_ELEMENT(ch->NextSibling("ch"));
                 }
             }
+            else
+            {
+                dawExtraState.customcontrol_chan_map.clear();
+            }
         }
     }
 
--- a/src/common/SurgePatch.cpp
+++ b/src/common/SurgePatch.cpp
@@ -1409,7 +1409,6 @@ void SurgePatch::load_xml(const void *data, int datasize, bool is_preset)
         char *temp = (char *)malloc(datasize + 1);
         memcpy(temp, data, datasize);
         *(temp + datasize) = 0;
-        // std::cout << "XML DOC is " << temp << std::endl;
         doc.Parse(temp, nullptr, TIXML_ENCODING_LEGACY);
         free(temp);
     }
@@ -3104,10 +3103,6 @@ void SurgePatch::load_xml(const void *data, int datasize, bool is_preset)
 
                     ch = TINYXML_SAFE_TO_ELEMENT(ch->NextSibling("ch"));
                 }
-            }
-            else
-            {
-                dawExtraState.customcontrol_chan_map.clear();
             }
         }
     }
--- a/src/common/SurgePatch.cpp
+++ b/src/common/SurgePatch.cpp
@@ -3042,6 +3042,30 @@ void SurgePatch::load_xml(const void *data, int datasize, bool is_preset)
                 }
             }
 
+            p = TINYXML_SAFE_TO_ELEMENT(de->FirstChild("midichan_map"));
+
+            if (p)
+            {
+                auto ch = TINYXML_SAFE_TO_ELEMENT(p->FirstChild("ch"));
+
+                while (ch)
+                {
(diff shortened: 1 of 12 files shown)


baconpaul:
Those unit test failures are real. Let me look at them tomorrow if you don't get em.

mkruselj:
> Those unit test failures are real. Let me look at them tomorrow if you don't get em.

Looks like I got 'em! 😁

src/common/SurgePatch.cpp
```diff
@@ -3061,6 +3084,26 @@ void SurgePatch::load_xml(const void *data, int datasize, bool is_preset)
...
+
+                    if (ch->QueryIntAttribute("p", &p) == TIXML_SUCCESS &&
+                        ch->QueryIntAttribute("v", &v) == TIXML_SUCCESS)
+                    {
+                        dawExtraState.customcontrol_chan_map[p] = v;
+                    }
+
+                    ch = TINYXML_SAFE_TO_ELEMENT(ch->NextSibling("ch"));
+                }
+            }
```
baconpaul, line 3106: you need an else clear here also

baconpaul (commented):
Just one small comment. Looks good otherwise!

src/common/SurgePatch.cpp
mkruselj: On it! Thank you 🙂

PR #7088, opened 2023-07-05
Rename Combinator modes to mystery box

mkruselj:
Also a break; was missing so Bonsai routing options were bleeding through the RM volume parameter's context menu

### DIFF
--- a/src/common/SurgeStorage.h
+++ b/src/common/SurgeStorage.h
@@ -189,19 +189,26 @@ enum NoiseColorChannels
     MONO = 1
 };
 
-enum RingModMode
+enum CombinatorMode
 {
-    rmm_ring = 0,
-    rmm_cxor43_0,
-    rmm_cxor43_1,
-    rmm_cxor43_2,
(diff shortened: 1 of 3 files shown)


baconpaul:
Still would be lovely if this was a data structure in a header in src common rather than embedded in the gui! Then I could use the same names in rack

mkruselj:
> Still would be lovely if this was a data structure in a header in src common rather than embedded in the gui! Then I could use the same names in rack

Can do that. SurgeStorage.h, then?

baconpaul:
> > Still would be lovely if this was a data structure in a header in src common rather than embedded in the gui! Then I could use the same names in rack
> 
> Can do that. SurgeStorage.h, then?

I would put it right after the enum for ring mod modes, yeah. And probably add an `n_ringModModes` to the end of the enum also. (Sorry we discussed this on discord but i didn't write it in an issue)

baconpaul:
Awesome looks good to me! Thank you

PR #7094, opened 2023-07-14
Osc enhance

pkstone:
Added OSC output and many new messages are understood and sent.


baconpaul:
DO NOT MERGE this is just for phil and me to chat

src/surge-xt/osc/OpenSoundControl.cpp
```diff
@@ -119,6 +126,7 @@ void OSCListener::oscMessageReceived(const juce::OSCMessage &message)
         if (!message[0].isFloat32())
         {
             // Not a valid data value
+            std::cout << "Invalid data type (not float)." << std::endl;
```
baconpaul, line 129: If we actually want to report this to the user, storage->reportError is the way to go. If we don't then probably strip the print.

baconpaul: Or wrap in a #debug

src/surge-xt/osc/OpenSoundControl.cpp
```diff
@@ -267,5 +297,91 @@ void OSCListener::oscBundleReceived(const juce::OSCBundle &bundle)
...
+                case vt_int:
+                    valStr = std::to_string(p.val.i);
+                    break;
+
+                case vt_bool:
+                    valStr = std::to_string(p.val.b);
+                    break;
+
+                case vt_float:
+                    valStr = std::to_string(p.val.f);
```
baconpaul, line 370: Out of curisoutiy do we need to worry about locales here? `std::to_string` will make pi `3,1415` in france and stuff

baconpaul (approved):

src/surge-xt/osc/OpenSoundControl.cpp
pkstone: hadn't considered any i8n issues on this; I think it does need consideration now that you point it out.

src/surge-xt/osc/OpenSoundControl.cpp
pkstone: Wrapped it in #debug.

src/surge-xt/osc/OpenSoundControl.cpp
baconpaul:
```
inline std::string float_to_clocalestr(float value)
{
    return fmt::format(std::locale::classic(), "{:L}", value);
}
```

there's something which puts into the `C` locale for you from a float but lets do that separately I agree

baconpaul: So I'll merge this one and then we can think about that separately.

PR #7118, opened 2023-07-24
Osc freqnote

pkstone:


src/common/globals.h
```diff
@@ -95,4 +95,6 @@ const int DEFAULT_POLYLIMIT = 16;
 
 const int DEFAULT_OSC_PORT_IN = 53280;
 const int DEFAULT_OSC_PORT_OUT = 53281;
+const float MIDI_MIN_FREQ = 8.1758;
```
baconpaul, line 98: This is already tunings midi note 0 constant. And the other one is that times 1<<7 or what not so if possible I would prefer to not duplicate them

src/surge-xt/osc/OpenSoundControl.cpp
```diff
@@ -112,34 +114,106 @@ void OpenSoundControl::oscMessageReceived(const juce::OSCMessage &message)
...
+        {
+            sendError("Invalid data type for velocity (must be an float between 0 - 127).");
+            return;
+        }
+
+        float32_t frequency = message[0].getFloat32();
+        int velocity = static_cast<int>(message[1].getFloat32());
+
+        // ensure freq. is in MIDI note range
+        if (frequency < MIDI_MIN_FREQ || frequency > MIDI_MAX_FREQ)
```
baconpaul, line 138: So here you could use that constant and the constant * 1<<7

src/surge-xt/osc/OpenSoundControl.cpp
```diff
@@ -346,6 +410,11 @@ void OpenSoundControl::send(std::string addr, std::string msg)
     }
 }
 
+void OpenSoundControl::sendError(std::string errorMsg)
+{
+    OpenSoundControl::send("/error", errorMsg);
```
baconpaul, line 415: What does send error do if you don’t have an outbound osc port? Should it at least log to stdout in that case?

baconpaul (changes requested):
Looks great! Couple of tiny things

src/surge-xt/osc/OpenSoundControl.cpp
pkstone: OK. Are the tunings constants in scope for OpenSoundControl.cpp?

src/surge-xt/osc/OpenSoundControl.cpp
pkstone: It could do. Good idea.

pkstone: Should the stdout messages be wrapped in #if DEBUG?

src/surge-xt/osc/OpenSoundControl.cpp
pkstone: They seem to be

baconpaul:
Oh darn im sorry Phil those midi bounds aren’t right

8*1<<7 is 1024 hz!

The upper bound is going to need a proper definition of nore0*2^(128/12)

I’m sorry! I think we need to fix that again before we merge. Entirely my fault

src/surge-xt/osc/OpenSoundControl.cpp
```diff
@@ -134,13 +134,14 @@ void OpenSoundControl::oscMessageReceived(const juce::OSCMessage &message)
 
         float frequency = message[0].getFloat32();
         int velocity = static_cast<int>(message[1].getFloat32());
+        constexpr float MAX_MIDI_FREQ = 1625.49867722; // 2^(128/12)
```
baconpaul, line 137: This is the pitch. Need to mult by midi note 0 to get the hz

pkstone: (Copied/pasted from Discord): I see 12543.854* referenced as the max MIDI frequency, but 8.175798915643710529 (MIDI_FREQ_0) * 1625.49867722 (your value) = 13289.75. 

*I can get this same value by passing 127 to Pd's 'mtof', which converts MIDI notes to frequencies.

So, I've set MAX_MIDI_FREQ to 12543.854. Let me know if that's ok.

PR #7137, opened 2023-08-04
Add some static asserts for C++20

baconpaul:
in globals.h add ones which require C++20 both syntatically and with the cplusplus macro

in storage.h add one which makes sure our path u8string is a string

Comments on both

No functional change

### DIFF
--- a/src/common/FxPresetAndClipboardManager.cpp
+++ b/src/common/FxPresetAndClipboardManager.cpp
@@ -20,6 +20,7 @@
  * https://github.com/surge-synthesizer/surge
  */
 
+#include <sstream>
 #include "FxPresetAndClipboardManager.h"
 #include "StringOps.h"
 #include "Effect.h"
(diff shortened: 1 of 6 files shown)


src/common/FxPresetAndClipboardManager.cpp
```diff
@@ -195,35 +195,34 @@ bool FxUserPreset::readFromXMLSnapshot(Preset &preset, TiXmlElement *s)
     {
         double fl;
         std::string p = "p";
+        std::string tmp; // Workaround for a g++12 bug
 
-        if (s->QueryDoubleAttribute((p + std::to_string(i)).c_str(), &fl) == TIXML_SUCCESS)
+        tmp = p + std::to_string(i);
+        if (s->QueryDoubleAttribute(p.c_str(), &fl) == TIXML_SUCCESS)
```
baconpaul, line 201: This should be tmp

mx: Yeah; this fix didn't work anyway...resolving comment since I changed the fix.

PR #7151, opened 2023-08-09
Migrate various APIs to fs::path

mx:
DO NOT MERGE this yet. This pull request will end up with several more commits as part of it. I'm opening it early to test Windows CI.

### DIFF
--- a/src/common/FxPresetAndClipboardManager.h
+++ b/src/common/FxPresetAndClipboardManager.h
@@ -37,7 +37,7 @@ struct FxUserPreset
 {
     struct Preset
     {
-        std::string file;
+        fs::path file;
         std::string name;
         int streamingVersion{ff_revision};
         fs::path subPath{};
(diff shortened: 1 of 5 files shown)


src/common/FxPresetAndClipboardManager.h
```diff
@@ -72,7 +72,7 @@ struct FxUserPreset
     bool hasPresetsForSingleType(int type_id);
     bool readFromXMLSnapshot(Preset &p, TiXmlElement *);
 
-    void saveFxIn(SurgeStorage *s, FxStorage *fxdata, const std::string &fn);
+    void saveFxIn(SurgeStorage *s, FxStorage *fxdata, fs::path sp);
```
baconpaul, line 75: Const & again to avoid thencopy

baconpaul (commented):
Looks good! One copy we don’t need

PR #7178, opened 2023-08-12
Deactivate misleading tuning typeins in Tune-After mode

### LINKED ISSUE #6977: Pitch type-ins are a tiny bit borked with "Apply tuning after modulation"
So... you have loaded a tuning with a 3/2 in it, and checked the mentioned option. You type-in "3/2" in the pitch field of an oscillator, it won't play 3/2. 

That's because with apply after modulation checked, putting pitch to 7.00 exactly is remapped to 7 keys up, and typing in 3/2 doesn't land you on 7 steps it lands you on ≈7.02. 

So you probably expected to land on 3/2 since that's what you typed in, and that interval is in your scale. But instead you landed a bit sharp of the 7th degree in your scale. How sharp depends on what the 8th degree is, since this mode interpolates the pitches between the integer degrees.

There are two options for solving this. Either we just document it as it is, basically saying "in this mode, always type in pitch offsets as 100 cent intervals. 

...or make it so whenever the user types in a ratio that's present in the scale, make sure the param ends in the place that actually makes that sound. 

The latter would be more intuitive in some cases. But of course, it'd still only be when playing the root of the scale that you'd actually be guaranteed to hear the interval you typed in.

baconpaul:
In Tune-After mode the typein "9/8" didn't uniformly give you a 9/8 pitch shift; it gave you that many keys shift. This was rather confusing. Some even called it "borked". So just make a clear error that this typein style is only availabel in tune-at-midi mode

Closes #6977

### DIFF
--- a/src/common/Parameter.cpp
+++ b/src/common/Parameter.cpp
@@ -5320,7 +5320,7 @@ bool Parameter::supports_tuning_value_from_string(const std::string &s, std::str
     }
     if (storage->tuningApplicationMode == SurgeStorage::TuningApplicationMode::RETUNE_ALL)
     {
-        errMsg = "No tuning type-ins in tune-after mode";
+        errMsg = "No ratio type-ins in tune-after mode";
         return false;
     }
 
--- a/src/common/Parameter.cpp
+++ b/src/common/Parameter.cpp
@@ -5320,7 +5320,7 @@ bool Parameter::supports_tuning_value_from_string(const std::string &s, std::str
     }
     if (storage->tuningApplicationMode == SurgeStorage::TuningApplicationMode::RETUNE_ALL)
     {
-        errMsg = "No tuning typesin in tune-after mode";
+        errMsg = "No tuning type-ins in tune-after mode";
         return false;
     }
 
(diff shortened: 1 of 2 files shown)


src/common/Parameter.cpp
```diff
@@ -5299,4 +5311,20 @@ float Parameter::calculate_modulation_value_from_string(const std::string &s, st
...
+bool Parameter::supports_tuning_value_from_string(const std::string &s, std::string &errMsg)
+{
+    if (!storage)
+    {
+        // Well that's odd! But let it go.
+        return true;
+    }
+    if (storage->tuningApplicationMode == SurgeStorage::TuningApplicationMode::RETUNE_ALL)
+    {
+        errMsg = "No tuning typesin in tune-after mode";
```
mkruselj, line 5323: Maybe: "Fractional values not allowed in tuning applied after MIDI input mode!". If it fits?

baconpaul: The string above is the longest that fits

baconpaul: Oh but it shouldn’t be types in. Sigh

Andreya-Autumn:
"No ratio typeins in tune-after mode"

PR #7219, opened 2023-09-23
Revert "Skip VST3 Menus in Reason"

mkruselj:
Reverts surge-synthesizer/surge#6766

Test to see if Reason Studios actually fixed it. Surely they wouldn't lie to us!

### DIFF
--- a/src/surge-xt/SurgeSynthEditor.cpp
+++ b/src/surge-xt/SurgeSynthEditor.cpp
@@ -610,10 +610,6 @@ void SurgeSynthEditor::endMacroEdit(long macroNum)
 
 juce::PopupMenu SurgeSynthEditor::hostMenuFor(Parameter *p)
 {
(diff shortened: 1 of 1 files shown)


baconpaul:
I am not sure if the fix is out of beta yet. They gave me beta access to file a bug. So this may be premature

mkruselj:
This is just for @esselfortium to test in latest beta. Pretty sure that beta will be out before 1.3, probably.

mkruselj:
Aaaaaaaand confirmed working. In the current stable version, not beta! So it's out already and we're good.

image

mkruselj:
Confirmed fixed on latest Reason _stable_ release. So it's out of beta and we're good!

image

baconpaul:
Oh awesome!!
and thanks to reason folks if they read this for giving me beta access to confirm and report :)

PR #7237, opened 2023-10-03
Osc work

pkstone:
OSC /param messages now accept natural-ranged values (as well as normalized 0.0 - 1.0 values, as previously)


baconpaul:
Looks great to me! I'll merge once CI finishes.

PR #7244, opened 2023-10-05
Sustain pedal is omni in Channel per octave mode

### LINKED ISSUE #7115: Channel sustain pedal
Yeah <https://github.com/surge-synthesizer/surge/blob/448b4efcc9b1279277b892a18ee7d11038cab889/src/common/SurgeSynthesizer.cpp#L2100> the channel hold only applies to the channel with the sustain pedal. That’s an oddity in all our modes. There’s special mpe treatment elsewhere but you aren’t in mpe mode. I’ll open an issue

In octave per channel mode sustain pedal is channel limited basically 

Discord user conradtao

mkruselj:
Also bypass ch2/3 mode activity if Channel per octave mode is enabled

Closes #7115


baconpaul:
Awesome!

PR #7252, opened 2023-10-14
Exclude `immintrin.h` on RISC-V

moui0:
I am trying to build `surge` on RISC-V, but there is no header `<immintrin.h>` on RISC-V.


baconpaul:
Oh wow is this file even used any more? I thought I had removed it

I’ll look at both yur prs this weekend

baconpaul:
Hi!

Thanks. A more correct fix is to not have this ancient include in this soon-to-be-dead header and make sure the clients include our regular simde mapping, so I'll make that change over in #7255 and once CI passes and that's merged close this one.

It really is just "remove the entire block" and add one header to an upstream client to make sure we get our simde includes.

Thanks so much for your efforts.

baconpaul:
Replaced by #7255 - thanks!

PR #7270, opened 2023-10-23
Support MIDI channels with MTS when not using ch2+3 to play scenes

### LINKED ISSUE #7202: MTS Tuning with Ch2Ch3 mode
Right now SurgeVoice::getPitch does this

```
 keyRetuning = MTS_RetuningInSemitones(storage->oddsound_mts_client, key + mpeBend, 0);
```

the reason for that hardcoded channel 0 is two fold

1. In MPE mode its really what you want. Really.
2. Outside MPE mode surge uses channel 2 and 3 for scene individuation. And so its what you want

But we now have a no-more-ch23-mode-switch. Fine. 

So that 0 should really be

```
(23mode || mpemode || chsplit scene mode) ? 0 : channel
```

but that conditional is a bit tricky. We don't want to call SurgeStorage::getPreference there - it is called on every voice on every block. So need to set up that condition at voice on as a bool and so on.

mkruselj:
Closes #7202

### DIFF
--- a/src/common/UserDefaults.cpp
+++ b/src/common/UserDefaults.cpp
@@ -368,20 +368,22 @@ std::string defaultKeyToString(DefaultKey k)
 */
 
 std::string getUserDefaultValue(SurgeStorage *storage, const DefaultKey &key,
-                                const std::string &valueIfMissing)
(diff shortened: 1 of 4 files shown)


baconpaul:
Looks great

PR #7272, opened 2023-10-23
OSC setup stored in DAW extra state

### LINKED ISSUE #7214: Put OSC port in DAW extra state
Currently they all default back to 53280/53281 on project re-load.

mkruselj:
Reshuffled the OSC settings menu a bit
Minor tweaks to OSC specification HTML

Closes #7214


baconpaul:
Looks correct to me.

PR #7276, opened 2023-10-28
Osc work: added parameter querying and modulation mapping

pkstone:
Individual parameters may now be queried, which causes their current value to be sent to OSC out.

Modulation mapping may now be accomplished via OSC.


baconpaul:
Did a review and happy to merge this. Will ask on discord since i think you said its ready but if i don't here i'll press merge in a bit.

PR #7279, opened 2023-10-30
Prelude update etc...

Andreya-Autumn:
-Put the slew limiter in the prelude. 
-Add a tutorial for good measure. 
-Add a couple Init patches for Duophony and Paraphony.


Andreya-Autumn:
Ready for merge!

PR #7280, opened 2023-10-30
Fix Phaser LFO Waveform names for consistency

mkruselj:


baconpaul:
doesn't the & screw up a menu somewhere? Or am I mis-remembering?

mkruselj:
That was the case with VSTGUI since it used OS menus, but not the case with JUCE :)

PR #7284, opened 2023-11-04
Add the FX Group name to the end of Accessible sliders

### LINKED ISSUE #7238: Labels don't hide in the new angular graph
Know how you're supposed to be able to hide them to do ear training stuff? Doesn't work yet in the new mode. :)

baconpaul:
so rather than getting "channel, mix, channel, mix" you get "channel - Left, Mix - Left, Channel - Right, Mix - Right" etc

The implementation of this is kinda gross, but that's part of the sturcture of the fx. Along the way add a single function go to param index to group index.

Closes #7238

### DIFF
--- a/src/common/SurgeSynthesizer.cpp
+++ b/src/common/SurgeSynthesizer.cpp
@@ -3843,6 +3843,45 @@ void SurgeSynthesizer::getParameterName(long index, char *text) const
         snprintf(text, TXT_SIZE, "-");
 }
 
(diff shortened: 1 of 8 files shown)


mkruselj:
I would say the group label should go first then param name.

baconpaul:
I thought we decided the opposite. If you tab through with the reader it’s a big and slow change to add the group up front

PR #7302, opened 2023-11-12
Fix copy/paste error

Andreya-Autumn:
...which made all the new AW filters load as lowpasses.


baconpaul:
LGTM. Will merge once CI is done. Good catch!

PR #7310, opened 2023-11-18
Refresh Formula overlay after loading LFO preset or undo/redo, if applicable

### LINKED ISSUE #7308: Formula editor window doesn't update on patch/preset load
If you have the Formula editor open and do one of:

Load a formula preset
Switch to a different patch (which has a formula LFO in the same slot)
Undo/Redo the "Apply" formula action

...the editor won't update. Close it and re-open and you'll see that the actual formula does update correctly in each case. Just the UI doesn't. MSEG and Tuning Editor overlays do not have this problem.

mkruselj:
Closes #7308

### DIFF
--- a/src/surge-xt/gui/SurgeGUIEditor.cpp
+++ b/src/surge-xt/gui/SurgeGUIEditor.cpp
@@ -3209,34 +3209,35 @@ juce::PopupMenu SurgeGUIEditor::makeLfoMenu(const juce::Point<int> &where)
                                     .lfo[currentLfoId]
                                     .shape.val.i;
 
-                if (isAnyOverlayPresent(MSEG_EDITOR) || isAnyOverlayPresent(FORMULA_EDITOR))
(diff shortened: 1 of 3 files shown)


baconpaul:
This is clever, and close, but not quite complete.

Rather than doing an || you should do a loop. If formula and mseg are both open this will fail.

So whereas before it was

```
if (isopen(mseg) )
   blah
```

and you changed it to

```
if (isopen(mseg) || isopen (formula))
   which = is mseg ? mseg : formula
```

you should change it to

```
for( const auto &cand : { mseg_enum, formula_enum } )
{
   if (isopen(cand))
      then do it
}
```

so if you have both open and torn out they will simultaneously refresh

baconpaul (changes requested):
See inline comment; we can make it a loop over enum type and work with both open.

baconpaul:
LGTM!

Issue #7309, opened 2023-11-18
FX bypass via OSC 


Andreya-Autumn:
Bypass FX would be a useful thing to add. 
This would need an additional boolean address per FX slot.

It would also be useful to be able to clear an FX slot. 
My first thought was that sending /param/fx/*/1/type = 0 should clear it. This address has type "contextual" in the spec (and I've misplaced my Touchosc somehow lol) so I don't know, perhaps it already does this already? If so, let's just clarify that in the spec, and otherwise let's perhaps make it so?

baconpaul:
the clear action internally is same as set to 'fxt_off' which is 0, yeah

mkruselj:
So yes then we just need the bypass action in the spec.

pkstone:
> the clear action internally is same as set to 'fxt_off' which is 0, yeah

I've confirmed that the OSC message '/param/fx/a/3/type 0' does indeed clear the associated effect slot.

PR #7346, opened 2023-11-29
Osc work -  FX bypass via OSC #7309 

pkstone:
FX bypass (per slot) can now be set by OSC. Changes to the 'deactivate mask' are reported to OSC out. Status of individual bits in the mask can be queried and reported to OSC out.


baconpaul:
Cool let me rebase it around those conflicts for you quickly

PR #7388, opened 2023-12-11
Fix MSVC stdext::checked_array_iterator warnings

mthierman:
Fixes fmt errors in MSVC [deprecating stdext::checked_array_iterator](https://github.com/fmtlib/fmt/issues/3540) which causes compile errors in debug mode


baconpaul:
Great.

PR #7406, opened 2023-12-22
Switch time based param units to ms when value is below 1 s

mkruselj:
Also fix Mod List so that it shows bipolar mod amounts when it should.

### DIFF
--- a/src/common/Parameter.cpp
+++ b/src/common/Parameter.cpp
@@ -28,6 +28,7 @@
 #include <sstream>
 #include <string>
 #include <cstdlib>
+#include <regex>
 #include <algorithm>
 #include <cctype>
 #include <utility>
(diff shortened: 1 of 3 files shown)


src/common/Parameter.cpp
```diff
@@ -1496,7 +1497,6 @@ void Parameter::set_type(int ctrltype)
     case ct_envtime_lfodecay:
         snprintf(displayInfo.maxLabel, DISPLAYINFO_TXT_SIZE, "Forever");
         displayInfo.customFeatures = ParamDisplayFeatures::kHasCustomMaxString;
-        // THERE IS NO BREAK HERE ON PURPOSE so we group to the others
```
baconpaul, line 1499: Please do not remove this comment!

PR #7412, opened 2023-12-27
User-only on Cmd-Click od Cmd-Shift-F10 in patch browser

### LINKED ISSUE #7407: User only patch menu
See convo in accessibility 

basically a keybind which pops just the user patches column makes reader nav way faster

we already do right click for just current category maybe there’s a similar one free or some such

baconpaul:
This change adds a user-only which skips factory and third party patches in the selection browser. Super useful for screen reader users getting to their own patches. Closes #7407


mkruselj:
@baconpaul Maybe it would make sense to just show the regular menu if there are no user patches, even if we explicitly call up the "user only" menu?

baconpaul:
Good idea. Will add and re-push

PR #7458, opened 2024-01-21
add obs workflow for nightly builds

KottV:
Hi,
as discussed, lets try to integrate GH and OBS for the Linux nightly builds


baconpaul:
Ok merged!

now what? 

Exciting!

PR #7479, opened 2024-01-27
Don't build Lua tests if Lua not available

taylordotfish:
Building Surge without Lua using `-DSURGE_SKIP_LUA=ON` results in an error because [UnitTestsLUA.cpp](https://github.com/surge-synthesizer/surge/blob/1ddf4df83c642d9da4b16a5d6483732e502275ad/src/surge-testrunner/UnitTestsLUA.cpp) is still compiled and tries to use Lua. Following other files like [LuaSupport.cpp](https://github.com/surge-synthesizer/surge/blob/1ddf4df83c642d9da4b16a5d6483732e502275ad/src/common/LuaSupport.cpp#L49), this PR adds a preprocessor guard to check for `HAS_LUA` first.

### DIFF
--- a/src/surge-testrunner/UnitTestsLUA.cpp
+++ b/src/surge-testrunner/UnitTestsLUA.cpp
@@ -39,6 +39,8 @@
 
 #include "lua/LuaSources.h"
 
+#if HAS_LUA
+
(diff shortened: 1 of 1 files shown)


baconpaul:
Oh ha thanks!

PR #7486, opened 2024-01-29
Osc work

pkstone:
Surge button appearance when disabled is now 'grayed out'.

Added OSC queries for modulation mapping/muting. Updated OSC spec accordingly.


pkstone:
I should have put more prominently in the comment: the commit changes Surge button appearance when enabled = false.

PR #7487, opened 2024-01-30
Osc work: Current patch query via OSC

pkstone:
Current patch (full path) can now be queried from OSC.

### DIFF
--- a/src/surge-xt/osc/OpenSoundControl.cpp
+++ b/src/surge-xt/osc/OpenSoundControl.cpp
@@ -565,8 +565,15 @@ void OpenSoundControl::oscMessageReceived(const juce::OSCMessage &message)
     {
         if (querying)
         {
-            std::string patchpath = synth->storage.patch_list[synth->patchid].path.u8string();
-            sendPath(patchpath.erase(patchpath.length() - 4));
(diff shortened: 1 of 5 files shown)


src/surge-xt/osc/OpenSoundControl.cpp
```diff
@@ -556,16 +558,18 @@ void OpenSoundControl::oscMessageReceived(const juce::OSCMessage &message)
...
-        std::cout << "Parameter full name:" << p->get_full_name() << std::endl;
-#endif
     }
 
     // Patch changing
     else if (address1 == "patch")
     {
+        if (querying)
+        {
+            std::string patchpath = synth->storage.patch_list[synth->patchid].path.u8string();
```
baconpaul, line 568: Ahh unfortunately there's some cases like in drag and drop where patched can be -1 so you need to check this here I think and return an unknown

so if synth->patched >= 0 && s->pid < s->storage.patch_list.size() then....

pkstone: OK, not sure how to trigger this, but it's handled.

baconpaul: You would find a random fxp and drag and drop it onto your surge

then it would be loaded but not in the patch list

Issue #6834, opened 2023-02-04
Tape kills audio when oversampled


Andreya-Autumn:
Load a Surge in Reaper
Oversample it to 196k 
Latch or something
Load the tape FX
Sound dissappears.

ritschwumm:
.oO(NaN-poisoning?)

Andreya-Autumn:
Yes maybe. 

Notably, this only happens if the oversampling is already in effect when loading the Tape, so initialization-related probably.

baconpaul:
right so at 192 in my debug build pulling in a tape on an empty surge just sends CPU to 100%
let me try a release build to check also

baconpaul:
OK so if you do this at 192 and run the mix the original sound comes back so the FX isn't borked completely
If you do it at 96 you get a big cut but not to 0

I really think this is somewhere in the DSP missing a sample rate reset but i can't see where.

Andreya-Autumn:
Push out of milestone so we can 1.2 sooner?

baconpaul:
Maybe 

I’m curious about the init cycle part thoiugh so let’s leave it float for a bit

mkruselj:
We probably need @jatinchowdhury18's eyes on this one, I guess?

baconpaul:
yes i think we will. i'm not seeing it.

jatinchowdhury18:
I believe we spoke about this issue a little while ago on the Discord... If I remember right, the bug was only presenting itself when using smaller block sizes? I remember having some trouble reproducing this one as well.

baconpaul:
I saw it at 32 yesterday in reaper once I learned how to oversample by track

jatinchowdhury18:
Hmmm, I did some more testing today, and still haven't been able to reproduce. Currently I'm testing with the nightly VST3 in Reaper 6.68. I've tried both "Chain oversampling" and "FX instance oversampling", but maybe there's some other way of doing per-plugin oversampling in Reaper?

Just trying to narrow down what could be the differences.

mkruselj:
Those are the only ways to do plugin oversampling in Reaper AFAIK.

jatinchowdhury18:
Hmm okay... I was also testing on my M1 Mac. Maybe the issue only shows up on Intel? Or only on Windows? I can give those a try tomorrow.

Andreya-Autumn:
Oh hey, Nimbus does it too it seems.

baconpaul:
[...]

PR #7527, opened 2024-02-20
Implementing sampleRateReset() for BBDEnsembleEffect

### LINKED ISSUE #7519: Ensemble effect is sample rate dependent
Specifically, its internal clock frequency is- And it'd be better if it weren't.

jatinchowdhury18:
Resolves #7519.

Re-initializes everything in the effect that is sample-rate dependent. I wasn't noticing the bug originally when testing in Bitwig, since Bitwig was re-instantiating the entire plugin when the sample rate changed.

@Andreya-Autumn if you have a chance to test this branch, that would be very helpful!

@baconpaul thanks for the clue about `sampleRateReset()`.

### DIFF
--- a/src/common/dsp/effects/BBDEnsembleEffect.cpp
+++ b/src/common/dsp/effects/BBDEnsembleEffect.cpp
@@ -96,14 +96,7 @@ BBDEnsembleEffect::BBDEnsembleEffect(SurgeStorage *storage, FxStorage *fxdata, p
     width.set_blocksize(BLOCK_SIZE);
     mix.set_blocksize(BLOCK_SIZE);
 
-    for (int i = 0; i < 2; ++i)
-    {
-        for (int j = 0; j < 3; ++j)
-        {
-            modlfos[i][j].samplerate = storage->samplerate;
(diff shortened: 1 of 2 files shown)


Andreya-Autumn:
On it!

Andreya-Autumn:
Works! Clock rate is consistent now. So now you can oversample Ensemble to 16x if for some reason you want the intermodulation aliasing, but much cleaner saturation! :) 

I got it to DC out one time, by toggling reaper oversample while latched. But I haven't reproduced it even after trying multiple times, and it recovered when un-latching and resetting again. So I'm not too worried about it. 

I wonder if #6834 is a similar issue btw?

Andreya-Autumn:
Oh, and this is maybe likely to have affected the VCV rack community more than Surge, perhaps. Since it's so easy to change samplerates there and people do it a lot. I'm guessing Ensemble isn't used all that much there, but can't hurt to give the forum a heads up when this change makes it there.

baconpaul:
I can do a 2.2.3 rack with this swept no prob

PR #7541, opened 2024-02-29
Invert cli log level logic

eduardoabinader:
It is normally expected the incoming log level, via parameter or configuration, to dictate how the current the set of log msgs are to be shown. If cli in invoked with VERBOSE or TRACE level, all msgs up to that level are shown. Otherwise, if it is said to be BASIC, only msgs with such level should be displayed.


baconpaul:
ha did I really get that backwards? Thanks!

PR #7542, opened 2024-03-01
Cli main small fixes

eduardoabinader:


baconpaul:
Wonderful. This looks good to me. Is it ready to merge? If so I'll squash it and merge it

eduardoabinader:
I've tested and looks fine. I will add a test for cli at some point as well.
Regarding the squash, is that the common standard in Surge? Normally the separate commits are to isolate unrelated stuff.

baconpaul:
Yes we try to have the commits be 'thing with granularity from which you would write the changelog' so this one I would rewrite as 'Improvements to CLI Error, Exit, and Log handling' (which I'm happy to do and will do now).

I often dont work-commits then rebase/squash myself but also happy to do it here also

baconpaul:
Anyway all merged! Thanks for working on this!

eduardoabinader:
Oh I see. Thanks! I will write a proper changelog next time then.
 :D

eduardoabinader:
Just wondering this approach makes a bit harder to work on regression, like doing a bissect.

baconpaul:
so yes, there's definitely a 'goldilocks' approach where commits should be 'just right' and that's definitely hard to know.

For the changes here, 5 commits seems a bit too much. 2 could have been ok (one for the logging/range checking and one for the termination). But odds are we wont' bisect through them anyway. If someone says "the termination behavior is messed up on blimblom linux" we will get here via git blame i bet. 

The way I usually write the changelog is from `git cherry -v (last-release-tag)` so the title of the commit matters most. The sort of platonic ideal, which we meet sometimes, is one open GitHub issue to one PR with one commit. I'd say about 60% of our commits meet that form.

but also its pretty chill. Just '5 was too granular in the log so i squashed it'. no hard and fast rules.

Anyway the first sentence of the CVS manual was "source control is not software engineering" so this is really just "mostly kinda what feels right so we can keep the transaction log reasonable". Happy to shift or adjust if you think appropriate!

PR #7546, opened 2024-03-02
Currently used tempo is saved to patch

### LINKED ISSUE #7529: Saving tempo value in current patch
I'm only using Surge XT _**standalone version**_ (thx for doing it too, besides the plugin ones).
My problem is that when I'm loading a patch for a song, I've always have to set the tempo of the song manualy (on the left side of the virtual keyboard of Surge).
After a fresh start, Surge XT standalone always set the tempo at 120 bpm. This value seems to be hard coded since I could not find in any config file a way to set another default (or start up) value.

I would like that Surge saves the current tempo value in the current loaded patch when I'm saving/overwritting it.
When used as a DAW plugin, Surge XT could simply ignore this parameter of the patch, since Surge should receive the current tempo value from the DAW.

No idea...

**Additional context:**
I'm using Surge XT standalone in Debian 12 (stable with KDE Plasma) on an used Dell Latitude 7200 2-in-1 (only with touch screen). Surge XT is controlled by a hardware MIDI keyboard Nektar Impact GX61. Audio output over the onboard sound card (phone output) of the Dell Latitude.
It's a killer standalone instrument for < 400,- EUR!!!
(and I'm also using Dexed and other software synths on this cheap but powerful hardware)

mkruselj:
It is only loaded in standalone, and only if the "Tempo on patch load" option is set to override. This option only shows in standalone, too.

Closes #7529

### DIFF
--- a/src/common/SurgePatch.cpp
+++ b/src/common/SurgePatch.cpp
@@ -2655,7 +2655,7 @@ void SurgePatch::load_xml(const void *data, int datasize, bool is_preset)
     {
         if (tos->QueryDoubleAttribute("v", &d) != TIXML_SUCCESS)
         {
-            d = 1.0;
+            d = 120.0;
         }
     }
 
--- a/src/common/SurgePatch.cpp
+++ b/src/common/SurgePatch.cpp
@@ -2642,6 +2642,32 @@ void SurgePatch::load_xml(const void *data, int datasize, bool is_preset)
         }
     }
 
+    bool userPrefOverrideTempoOnPatchLoad = Surge::Storage::getUserDefaultValue(
+        storage, Surge::Storage::OverrideTempoOnPatchLoad, true);
+
+    TiXmlElement *tos = TINYXML_SAFE_TO_ELEMENT(patch->FirstChild("tempoOnSave"));
+
+    if (revision < 23)
+    {
+        d = -1.0;
+    }
+    else
+    {
+        if (tos->QueryDoubleAttribute("v", &d) != TIXML_SUCCESS)
+        {
+            d = 1.0;
+        }
+    }
+
+    if (userPrefOverrideTempoOnPatchLoad)
+    {
+        storage->unstreamedTempo = (float)d;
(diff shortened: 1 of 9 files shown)


mkruselj:
Need to work some more on this. Should not clobber the tempo typein in idle() as this makes it impossible to... type anything in! Also it seems when opening the standalone anew, tempo changes typed in manually (after removing the refresh in idle()) don't happen at all until you load any other patch. Hm...

mkruselj:
@baconpaul I think I managed to sort the above two issues, lmk if there's something that looks iffy to you or if you have an idea on how to make it better!

baconpaul:
Will look later today thanks!

src/common/SurgePatch.cpp
```diff
@@ -3571,7 +3597,7 @@ unsigned int SurgePatch::save_xml(void **data) // allocates mem, must be freed b
     }
 
     {
-        TiXmlElement compat("compatability");
+        TiXmlElement compat("compatability"); // TODO: Fix typo for XT2, LOL!
```
baconpaul, line 3600: What is this?

src/common/SurgePatch.cpp
```diff
@@ -2642,6 +2642,32 @@ void SurgePatch::load_xml(const void *data, int datasize, bool is_preset)
...
+
+    if (revision < 23)
+    {
+        d = -1.0;
+    }
+    else
+    {
+        if (tos->QueryDoubleAttribute("v", &d) != TIXML_SUCCESS)
+        {
+            d = 1.0;
```
baconpaul, line 2658: Why 1 as opposed to 120?

baconpaul (commented):
Two small comments but looks correct to me otherwise

src/common/SurgePatch.cpp
mkruselj: It's a typo. The token should have been compat**i**bility, not compat**a**bility :)

PR #7560, opened 2024-03-09
Changed the cli midi behaviour

gitsmol:
- midi is off by default
- specifying a midi device checks for out of bounds index
- requesting all midi devices adds all available devices, which may be 0. 
 
NB both may be enabled at the same time. If the specified device is not present, that will cause an error despite the all-midi flag.

For some reason I needed to set `->default_val("-1")` to get cli11 to actually use that value as the default.

### DIFF
--- a/src/surge-xt/cli/cli-main.cpp
+++ b/src/surge-xt/cli/cli-main.cpp
@@ -211,13 +211,13 @@ void isQuitPressed()
 int main(int argc, char **argv)
 {
     // juce::ConsoleApplication is just such a mess.
-    CLI::App app("surge-xt-CLI : a command line player for surge-xt");
+    CLI::App app("surge-xt-cli - a command line player for Surge XT");
 
     app.set_version_flag("--version", Surge::Build::FullVersionStr);
 
     bool listDevices{false};
     app.add_flag("-l,--list-devices", listDevices,
-                 "List all devices available on this system, then exit");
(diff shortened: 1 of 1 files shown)


baconpaul:
So my only concern here is if you do --midi-device=3 --allmidi you will bind device 3 twice. We should make specifying a non-zero device and all midi a bailout condition

baconpaul:
sorry pressed send too soon

I think the solution is before we do anything check "if midiDevice >= 0 && allMidi LOG("please select individual device or illmidi but not both" exit" type thing once we parse the args and before we start anything

Otherwise fine to merge yup!

mkruselj:
Maybe it would be better to just take the first arg of the two, rather than bail out? And document that.

baconpaul:
The 'first arg' is probably hard. I dont think we get argument order with our CLI parser. But sure if both are specified we can bail out or we can print a warning and allmidi wins is a good solution

gitsmol:
So I went the other way with it :) Hope this also works for you.

baconpaul:
That’s fine! But I think we should still warn if all midi and midi input > 0 that we are ignoring all midi 

basically if we ignore an argument we should make a log message

im in my phone so maybe code does that and I missed it. If so good to merge! But silent ignore == confused user

gitsmol:
Done! The override property is also mention in --help.

PR #7562, opened 2024-03-10
Only use -fno-char8_t if using GCC >= 9

falkTX:
Fixes build with GCC 8 and older.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -89,7 +89,7 @@ if(CMAKE_CXX_COMPILER_ID MATCHES "Clang|GNU")
     $<$<BOOL:${SURGE_SANITIZE}>:-fsanitize=undefined>
 
     # Do *not* use the new, breaking char8_t UTF-8 bits in C++20.
-    $<$<COMPILE_LANGUAGE:CXX>:-fno-char8_t>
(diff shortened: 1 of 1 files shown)


baconpaul:
That’s not the correct check. Clang for instance is not inversion 9. We should do this check on cxx language version greater than 20 not compiler version. Thanks!

baconpaul:
> > We should do this check on cxx language version greater than 20 not compiler version. Thanks!
> 
> that always fails though, because just above the language version is set to 17.

It always fails unless you turn the build to 20 yes. 

If you turn the build to 20 it kicks in and the code compiles

baconpaul:
(basically we did a port to 20 to start using 20 everywhere; then decided to back down to 17 for a couple of reasons, but want the return-to-20 to work for later in the year)

baconpaul:
Closing in favor of the review implementation requested which I placed in #7612

falkTX:
sorry I completely forgot about this, the changes are much appreciated!

baconpaul:
No problem! Figured you were fine otherwise you would have made the change. Just cleaning up ready for a 1.3.2 surge release.

Oh that fix you found in rack with the uninit memory saved us today in another setting too. Will get that into next rack release.

tx

PR #7566, opened 2024-03-11
Broaden search over query

eduardoabinader:
Sometimes query criteria is not written on patch name. Thus, It'd be nice to also search on path name, which holds category info, additionally to patch name. Thus is specially useful for users for filtering Keys, Polys and so on.

### DIFF
--- a/src/common/PatchDB.cpp
+++ b/src/common/PatchDB.cpp
@@ -837,7 +837,28 @@ CREATE TABLE IF NOT EXISTS Favorites (
         }
 
         std::ostringstream searchName;
-        searchName << fs::path(p.path.parent_path().filename() / p.name) << " ";
+        searchName << p.name << " ";
+
+        if (storage)
+        {
+            auto pTmp = p.path.parent_path();
+            std::vector<fs::path> parentFiles;
+            int maxItForSafety{0};
+            while ((pTmp != storage->userPatchesPath) &&
+                   (pTmp != storage->datapath / "patches_factory") &&
+                   (pTmp != storage->datapath / "patches_3rdparty") && !pTmp.empty() &&
+                   (pTmp != pTmp.root_directory()) && maxItForSafety < 10)
+            {
+                parentFiles.push_back(pTmp.filename());
+                pTmp = pTmp.parent_path();
+                maxItForSafety++;
+            }
+
+            for (auto pf : parentFiles)
+            {
+                searchName << pf.u8string() << " ";
+            }
+        }
 
         std::ifstream stream(p.path, std::ios::in | std::ios::binary);
 
--- a/src/common/PatchDB.cpp
+++ b/src/common/PatchDB.cpp
@@ -854,6 +854,11 @@ CREATE TABLE IF NOT EXISTS Favorites (
                 maxItForSafety++;
             }
 
+            if (pTmp == storage->datapath / "patches_3rdparty")
+            {
+                parentFiles.erase(parentFiles.end() - 1);
+            }
+
             for (auto pf : parentFiles)
             {
                 searchName << pf.u8string() << " ";
--- a/src/common/PatchDB.cpp
+++ b/src/common/PatchDB.cpp
@@ -837,7 +837,7 @@ CREATE TABLE IF NOT EXISTS Favorites (
         }
 
         std::ostringstream searchName;
-        searchName << p.name << " ";
+        searchName << fs::path(p.path.parent_path().filename() / p.name) << " ";
 
         std::ifstream stream(p.path, std::ios::in | std::ios::binary);
 


eduardoabinader:
ops, search over unit test failed. Gonna check.

eduardoabinader:
Gotta update the test as well

baconpaul:
LGTM once CI passes

src/common/PatchDB.cpp
```diff
@@ -1484,7 +1484,8 @@ std::string PatchDB::sqlWhereClauseFor(const std::unique_ptr<PatchDBQueryParser:
         }
         break;
     case PatchDBQueryParser::LITERAL:
-        oss << "( p.search_over LIKE '%" << protect(t->content) << "%' )";
+        oss << "( p.path LIKE '%" << protect(t->content) << "%' or";
```
mkruselj, line 1487: Does this search the WHOLE path to the patch? That is probably overkill.

mkruselj: Because you know you can search categories by using the CAT= or CATEGORY= entry? So it's pretty easy to just do CAT=Pad and it will list all pads etc.

baconpaul: Ahh excellent point yes this will make every user patch match “Documents”

eduardoabinader: i think the approach proposed here is a bit  more broaden than CAT=. I know it may be rare, but we may have something named XXX Key, which is not under Keys cat.  By attacking this via search_for, we get this guy. This wont eliminate, CAT=. Furthermore, having the user to type cat= (or SQL like statements) is not so friendly and straightforward, as just only type the wildcard match, as here proposed. Both approaches can coexist.

baconpaul: Right I think the trick is the full path in the default search. Perhaps what we should do is store the path *from the appropriate root* in the sqldb and search in that 

so like strip off the user path and search path

eduardoabinader: yes, something on this direction. Then an update from script is necessary, right?

baconpaul: Yeah lemme jot down some comments. Reading code now.

baconpaul:
There's two ways to do it.

The easy way, which I recommend. No schema change needed just a database rebuild.

The 'search_over` is constructed in code at build time. In PatchDB.cpp around line 840 you see

```
        std::ostringstream searchName;
        searchName << p.name << " ";
```

so what you can do is after that add the sub-path. That is

```
     auto sp = p.path;
     if (sp. filesystem-for-is-inside (storage->patchPath))
       searcName << sp.lexically_relative_to(storage->patchPath) << " ";
     if (sp. filesystem-for-is-inside (storage->userPatchPath))
       searcName << sp.lexically_relative_to(storage->patchPath) << " ";
```

then force a rescan either by blowing away your db or upgrading the schema version to 15 around line 250. This just adds more info to the long char which is searched over. You probably want to change the schema to to make the search over longer. Then you don't need your change you just change the search field.

The alternative which I like less but I wrote up so am keeping here is to add the trimmed path to the DB explicitly. I don't recommend this but wrote it down so am keeping it here in case you think its better

PatchDB.cpp

Around line 250:

1. Change the schema version to 15. This will auto-heal the databases
2. in CreateTable patches under `path varchar(2048)` add `reduced_path varchar(2048)` or some such

In `parsFXPIntoDB` around line 820

1. Create the trimmed path. This is probably something like
```
auto trimmedPath = p.path;
trimemdPath = trimmedPath.lexically_relative(storage->userPatchesDir)
```

but you need to check both the user and main patches dir. You should have a reference to storage though and lexically_relative will do what you want
2. Update the SQL statement and the bind. Should be pretty obvious.
3. Build, re-run your scan, look in the sqlite console if the right thing is there

Then change this PR to be .path to .reduced_path and everything will be fine

eduardoabinader:
@baconpaul , I'd rather go with your first approach, But IMHO, this seems actually overkill to me. My patch really touches on the sqlWhere clause, which is expected for that.

baconpaul:
The problem is we don’t want to search on the full path. Every patch will match document for instance. And the search reduction in sql is way more than your patch. 

So put what we want in the search data is the answer I think here then don’t change the SQL at all. (This is also how tag searches work now)

eduardoabinader:
Actually just adding simply the Immediate parent and the patch name works fine: Keys/AAA, Basses/BBB, Polysynth/AAAA, ....

mkruselj:
That doesn't work if there are more subfolders with different categorizations (which can happen).

baconpaul:
But it’s super close. Also path into stream needs a u8string to avoid being quoted or u16 accidentally 

Lemme fix it up today

eduardoabinader:
@mkruselj , could you please name an example?

baconpaul:
> @mkruselj , could you please name an example?

Save a user patch in "foo/bar/hootie/my patch" and it will make the recursive directories.

eduardoabinader:
Ah ok, now I get this! There's always the possibility of having a personalized folder chosen by the user. I will fix this then.  Thanks.

baconpaul:
@eduardoabinader just pushed an similar approach to this branch. wdyt?

baconpaul:
Great! Will merge once we are through ci

PR #7583, opened 2024-03-18
Channel/octave in MTS-ESP mode

Andreya-Autumn:
- Allow channel per octave mode under MTS-ESP
- Show menu item when MTS-ESP is active
- Re-structure MTS-ESP tuning menu somewhat

### DIFF
--- a/src/common/dsp/SurgeVoice.cpp
+++ b/src/common/dsp/SurgeVoice.cpp
@@ -216,10 +216,9 @@ SurgeVoice::SurgeVoice(SurgeStorage *storage, SurgeSceneStorage *oscene, pdata *
     const bool isCh23Mode = Surge::Storage::getUserDefaultValue(
         storage, Surge::Storage::UseCh2Ch3ToPlayScenesIndividually, true, false);
     const bool isChSplitMode = storage->getPatch().scenemode.val.i == sm_chsplit;
-    const bool ChannelPerOctave = storage->mapChannelToOctave;
 
     state.mtsUseChannelWhenRetuning =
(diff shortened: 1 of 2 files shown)


src/common/dsp/SurgeVoice.cpp
```diff
@@ -215,8 +216,10 @@ SurgeVoice::SurgeVoice(SurgeStorage *storage, SurgeSceneStorage *oscene, pdata *
     const bool isCh23Mode = Surge::Storage::getUserDefaultValue(
         storage, Surge::Storage::UseCh2Ch3ToPlayScenesIndividually, true, false);
     const bool isChSplitMode = storage->getPatch().scenemode.val.i == sm_chsplit;
+    const bool ChannelPerOctave = storage->mapChannelToOctave;
 
-    state.mtsUseChannelWhenRetuning = (mpeEnabled || isChSplitMode || isCh23Mode);
+    state.mtsUseChannelWhenRetuning =
+        (mpeEnabled || isChSplitMode || isCh23Mode || ChannelPerOctave);
```
mkruselj, line 222: You could just directly use storage->mapChannelToOctave here, no need to have it as a variable when it's only used once.

Andreya-Autumn: Right, I thought I'd match what the other parts of the check were doing. But that makes sense. Will change.

Andreya-Autumn:
This all works for me and checks passed. Please take a look in case I did something silly though. ;) As time permits of course.

baconpaul:
Yup

PR #7586, opened 2024-03-24
Update CLAP features

messmerd:
I updated the CLAP features of Surge XT and Surge XT Effects to match the standard plugin features.
This will help hosts categorize the plugins better.

### Changes
- Surge XT:
  - Removed: "hybrid"
  - Added: "stereo"
- Surge XT Effects:
  - Removed: "effect", "multi-effect", "rotary speaker", and "eq"
  - Added: "audio-effect", "filter", "phaser", "equalizer", "phase-vocoder", "granular", "frequency-shifter", "multi-effects", and "stereo"

#### Remaining questions
- I left "free and open source" in, though it doesn't follow CLAP's guidance for non-standard features. Should something be done about this?
- I do not know if Surge is also capable of mono output. If so, the "mono" feature should probably be added.
- There may be some audio effect sub-features I've missed which could be added. Alternatively, maybe it would be better to remove all the sub-features except for "multi-effects"?

Feel free to push changes to this branch if you'd like, or let me know anything that should be changed and I can do it once I get the chance.

### DIFF
--- a/src/surge-fx/CMakeLists.txt
+++ b/src/surge-fx/CMakeLists.txt
@@ -74,6 +74,7 @@ if(SURGE_BUILD_CLAP)
       "audio-effect"
       "filter"
       "phaser"
+      "rotary speaker"
       "equalizer"
       "phase-vocoder"
       "granular"
--- a/src/surge-fx/CMakeLists.txt
+++ b/src/surge-fx/CMakeLists.txt
@@ -70,7 +70,23 @@ target_link_libraries(${PROJECT_NAME} PRIVATE
 if(SURGE_BUILD_CLAP)
   clap_juce_extensions_plugin(TARGET surge-fx
     CLAP_ID "org.surge-synth-team.surge-xt-fx"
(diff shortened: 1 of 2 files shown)


baconpaul:
Cool! Curious why you removed eq and rotary speaker though? Other than that this change is spot on!

messmerd:
> Curious why you removed eq and rotary speaker though?

I removed it because it wasn't a standard feature, though I could add it back in if you'd like

baconpaul:
Yeah if you don’t mind! Since it’s things they do 

the standard list isn’t exhaustive but agree it is good to track as closely as possible. Happy to merge if those two are back

and thanks!

messmerd:
Ok, I added "rotary speaker" back in.
"eq" was replaced with the standard "equalizer".

baconpaul:
Thanks!

mkruselj:
I would remove phase-vocoder because Surge FX don't have one. It has a regular bandpass+env follower vintage style vocoder.

PR #7589, opened 2024-03-26
Osc work: parameter extended options

pkstone:
All parameter extended options now working fully with OSC (both in and out).


pkstone:
Missed a couple of things on the last commit. This should be ready to go

Issue #7563, opened 2024-03-10
Release as gate 


baconpaul:
Basically 

let’s add a deform mode to the aeg and feg envelope release where instead of fading to 0 they just hard gate at level to zero 

this would allow basically attack 0 sustain 1 release deform envelop to just gate a voice and other modulators could modify the level as you saw fit

mkruselj:
I don't think this is needed for FEG, since we can just use an LFO EG directly to filter cutoff. Only AEG matters for this.

baconpaul:
Yeah I agree but the types are super symmetric so i wanted to see where the if had to go. If the code isn't super ugly do AEG only. Otherwise code both and just suppress the menu item on FEG.

PR #7594, opened 2024-04-02
Add Freeze release at sustain level option for AEG

mkruselj:
Addresses #7563. TODO: Analog mode needs impl of this option


baconpaul:
So with 
558402a in place I think this is fine for you to do a final test and then we can merge.

it looks like a git mess, but a squash and merge and rewrite the messages should just do it.

mkruselj:
Yup I'm just testing it locally!

mkruselj:
OK it seems to work fine in analog mode too! Merging.

Issue #7592, opened 2024-03-29
Should OSC messages be sent when MIDI input is received?


botoxparty:
Hey,

I've observed that OSC messages are only output if the controls are changed with the UI. When receiving MIDI input, the updates don't appear to be sent over OSC.

Is that the expected behaviour?

In my use case it would even be cool if it also outputted incoming OSC messages.

e.g. Controller 1 sends OSC message to Surge XT.
Surge XT sends an update to Controller 2 that the parameter was changed.

baconpaul:
Well since you didn’t expect it it’s clearly not expected! But we definitely didn’t think of this case 

so you set up a midi learn control basically and want that to send an osc message? That means the midi automation path needs to fire the parameter listener on the right thread. Doable but just want to make sure that’s the case you are in

also tagging @pkstone who has written most of the osc code.

the code is around here Phil https://github.com/surge-synthesizer/surge/blob/2834ca4c4fb270b1987bc0de1c47b9c0830563e6/src/common/SurgeSynthesizer.cpp#L2337

botoxparty:
Yeah for claritication..

If the Pitch in OSC1 of Scene A has a MIDI CC assigned to it, when it
received changes then it should output them to "/param/a/osc/1/pitch"

pkstone:
I think I get what you're asking for, but I can't promise anything on it for the near future. I would like to make the 'echoing' part of OSC (i.e., outputting Surge state changes to OSC) more consistent and comprehensive, but it's a big job. Things are a bit all over the place right now. For instance, modulation mapping changes always echo, even when changed by incoming OSC messages (this is a side-effect of using an already existing modulation change listenener). Nothing else does that though, yet.

pkstone:
There also starts to arise an issue of how much data pouring out of OSC is acceptable for a given user-case. Do we have to start adding on/off options for it so a perhaps slow client on the other end of the pipe is not overwhelmed?

mkruselj:
My thinking here is: how does host automation behave when receiving MIDI to set MIDI learned parameters? OSC should behave the same.

pkstone:
@botoxparty  I'm not meaning to be discouraging here, this is a good ask, and is something I really want to improve in the future. But it may be a while!  :-)

pkstone:
[...]

PR #7599, opened 2024-04-06
Osc work: fix for #7592

pkstone:
Parameter changes generated within SurgeSynthesizer (e.g. from 'learned' MIDI control input) are now properly echoed to OSC out.  Fixes https://github.com/surge-synthesizer/surge/issues/7592.

Fixed scaling of echoed MIDI pitch bend (is now -1.0 - 1.0, like OSC /pbend message).


baconpaul:
Looks right to me! Merging

PR #7600, opened 2024-04-07
OSC features added: save/load to userPatchPath, param extended options doc messages

gitsmol:
- Added Save/load to userPatchPath
- Added OSC error message when loading invalid path
- Added OSC doc message for param extended options
- Updated OSC specifications html accordingly

### DIFF
--- a/src/surge-xt/osc/OpenSoundControl.cpp
+++ b/src/surge-xt/osc/OpenSoundControl.cpp
@@ -827,14 +827,9 @@ void OpenSoundControl::oscMessageReceived(const juce::OSCMessage &message)
             // Run this on the juce messenger thread
             juce::MessageManager::getInstance()->callAsync([this, message]() {
                 std::string dataStr = getWholeString(message);
-                if (dataStr.empty())
-                    synth->savePatch(false, true);
-                else
-                {
-                    fs::path ppath = synth->storage.userPatchesPath;
-                    ppath += dataStr += ".fxp";
-                    synth->savePatchToPath(ppath);
-                }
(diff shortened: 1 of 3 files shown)


src/surge-xt/osc/OpenSoundControl.cpp
```diff
@@ -24,12 +24,14 @@
...
 #include "SurgeSynthProcessor.h"
 #include "SurgeStorage.h"
+#include <iostream>
 #include <sstream>
 #include <vector>
 #include <algorithm>
 #include <string>
 #include "UnitConversions.h"
 #include "Tunings.h"
+#include "filesystem/ghc-filesystem.h"
```
baconpaul, line 34: this is not the right way to import filesystem. Instead do

```
#include "import/filesystem.h"
```

which will load one of `std::filesystem` or `ghc::filesystem` and set the namespace `fs` to point.

This is why your windows builds (where we don't use GHC) failed.

gitsmol: Much appreciated, Paul. However, `#include "import/filesystem.h"` doesn't compile for me. I see SurgeStorage uses `#include "filesystem/import.h"` though, so probably that's the right header?

baconpaul: yes! sorry about that

gitsmol: Looks it still doesn't compile on windows. I'll investigate later, looks like I should convert Path to string before feeding to strncpy.

PR #7605, opened 2024-04-12
Add a host-side menu modifier method for prettification

mkruselj:
Apply it for FL Studio as it stands (nicer section header rendering)


baconpaul:
Looks great!

PR #7616, opened 2024-04-20
Adding failsafes for OSC saving to user patchdir

gitsmol:
I wrongly assumed that `synth->savePatchtoPath` checks path validity etc. It does not. Added a basic failsafe to the save_user OSC address that checks if path exists, tries to create it if not. If that fails we've tried our best and return an error.


gitsmol:
Whoops, I ran into a clang problem...

PR #7617, opened 2024-04-20
Add failsafe to OSC save patch to user patch dir

gitsmol:
Trying again after some clang formatting fails...


baconpaul:
There are cases where this operation can throw an `fs::filesystem_error` which uncaught here would crash your OSC instance.

You probably want

```
try {
   /// your code here
} catch (const fs::filesystem_error &e)
{
  // report e.what()
}
```

to be safe

PR #7620, opened 2024-04-21
Prettify host context menu for Reaper

mkruselj:
Removes repeated parameter name in ugly square brackets in Reaper case.

This unfortunately had to modify the semantics of SurgeSynthEditor::modifyHostMenu() to work with a copy rather than a direct reference, since I didn't see a way to remove a menu entry when working with juce::PopupMenu::MenuItemIterator.


baconpaul:
oooh lemme make sure this is safe before we merge OK? I don't know the lifetime of all the objects and am not sure the copy preserves them.

mkruselj:
Sure!

PR #7651, opened 2024-05-22
FM2 osc and Freq Shifter param definition fixes

mkruselj:
Extend mode now actually works for FM2 M1/2 Offset.
Increase streaming version (patches with sv < 24 will always load with M1/2 Offset extend disabled to retain their sound).

Frequency Shifter's Range was mistakenly showing 10x smaller values in all the displays.
Fixed this by adding a new ctype (to discern it from FM2 oscillator's M1/2 Offset parameter,
and doing everything necessary to ensure proper value display and conversion from string input etc.

### DIFF
--- a/src/common/Parameter.cpp
+++ b/src/common/Parameter.cpp
@@ -276,6 +276,7 @@ bool Parameter::can_extend_range() const
     case ct_pitch_semi7bp:
     case ct_pitch_semi7bp_absolutable:
     case ct_pitch_extendable_very_low_minval:
+    case ct_freq_fm2_offset:
     case ct_freq_reson_band1:
     case ct_freq_reson_band2:
     case ct_freq_reson_band3:
@@ -415,6 +416,7 @@ bool Parameter::is_bipolar() const
     case ct_percent_bipolar_w_dynamic_unipolar_formatting:
     case ct_noise_color:
     case ct_twist_aux_mix:
+    case ct_freq_fm2_offset:
     case ct_freq_shift:
     case ct_filter_feedback:
     case ct_osc_feedback_negative:
@@ -704,6 +706,7 @@ void Parameter::set_type(int ctrltype)
         val_default.f = 0; // semitones
         moverate = 0.5f;
         break;
+    case ct_freq_fm2_offset:
     case ct_freq_shift:
         valtype = vt_float;
         val_min.f = -10;   // Hz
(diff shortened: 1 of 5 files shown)


mkruselj:
@baconpaul Won't merge until you have a look! I think it's all good but better be safe than sorry.

src/common/Parameter.cpp
```diff
@@ -1496,8 +1499,9 @@ void Parameter::set_type(int ctrltype)
         displayInfo.customFeatures = ParamDisplayFeatures::kAllowsModulationsInNotesAndCents;
         break;
 
-    case ct_freq_shift:
+    case ct_freq_fm2_offset:
         displayType = LinearWithScale;
+    case ct_freq_shift:
```
baconpaul, line 1504: Why did this line move below the assignment?

baconpaul:
It seems freq shift could still be linear just with different coefficients and then we could avoid all the code which does per type formatting no? I will take a closer look when not in phone but don’t understand why we can’t just use linear with range info for this

src/common/Parameter.cpp
```diff
@@ -2814,6 +2821,71 @@ void Parameter::get_display_of_modulation_depth(char *txt, float modulationDepth
 
         break;
     }
+    case ct_freq_shift:
```
baconpaul, line 2824: Why do we need this code and below rather than just a linear scale with extend config above?

mkruselj: It already has the correct extend config (100x) but the parameter base range is +/-10. And so the DSP did a wrong thing where instead of get_extended it manually applied another 10 or 1000 multiply, which created the too wide range and incorrect stringification.

mkruselj: I mean I would love it if there's a way to have no code copypasta I just didn't see it immediately...

src/common/Parameter.cpp
mkruselj: Oh that's just a flub I guess...

src/common/Parameter.cpp
baconpaul: I guess my question is : "How is this freq shift just different than a linear param, so can we delete all this code". It just looks like it's a linear with a scale factor of 10, which we can do without a special case. Or am I misreading?

mkruselj: Right it is linear with a scale factor of 10 and then when extended a scale factor 100x that. So you reckon this would work with the default Linear type?

mkruselj: OK this should be resolved now. There is displayInfo.scale so I just set it to 10 for ct_freq_shift and that seems to work!

PR #7669, opened 2024-06-21
WIP, attempt to add audio inputs to surge-cli

KingaJanicka:
I'm trying to make it so that CLI binds to the audio inputs of the selected audio device, this compiles but gives a segfault when I run it.

I am not particularly strong with C/C++ so opening a draft if anyone knows how to do it


baconpaul:
Subsumed by #7678

PR #7674, opened 2024-06-25
Wtlua 7671

nuoun:
Added error message (split between syntax evaluation error and runtime error so it catches all errors) for wtlua, cleaned up lua stack a bit to fix guarded stack error, updated unittestsLUA


baconpaul:
Code looks good. I’m offline for a couple of hours but if ci passes we should absolutely merge!

PR #7675, opened 2024-06-26
OSC in for Surge FX

pkstone:
Add OSC interfacing and parsing for '/fx/param/<n>' messages (where n is 1 - 12 for the standard parameters).

### DIFF
--- a/src/surge-fx/SurgeFXProcessor.cpp
+++ b/src/surge-fx/SurgeFXProcessor.cpp
@@ -527,6 +527,7 @@ void SurgefxAudioProcessor::getStateInformation(juce::MemoryBlock &destData)
     xml->setAttribute("oscin", oscStartIn);
 
     copyXmlToBinary(*xml, destData);
+    storage->reportError("Wow!", "Error Reporting Testing");
 }
 
 void SurgefxAudioProcessor::setStateInformation(const void *data, int sizeInBytes)
--- a/src/surge-fx/SurgeFXProcessor.cpp
+++ b/src/surge-fx/SurgeFXProcessor.cpp
@@ -470,20 +470,23 @@ void SurgefxAudioProcessor::processBlock(juce::AudioBuffer<float> &buffer,
         }
     }
 
-    processBlockOSC();
+    if (oscReceiving)
+        processBlockOSC();
 }
 
 // Pull incoming OSC events from ring buffer
 void SurgefxAudioProcessor::processBlockOSC()
 {
-    auto messages = oscRingBuf.popall();
-    for (const auto &om : messages)
+    while (true)
     {
-        switch (om.type)
+        auto om = oscRingBuf.pop();
+        if (!om.has_value())
+            break;
+        switch (om->type)
         {
         case SurgefxAudioProcessor::FX_PARAM:
         {
-            fxstorage->p[fx_param_remap[om.p_index]].set_value_f01(om.fval);
+            fxstorage->p[fx_param_remap[om->p_index]].set_value_f01(om->fval);
             resetFxParams(true);
         }
         break;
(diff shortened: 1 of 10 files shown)


baconpaul:
This pr has a bunch of unrelated submodule moves - I’m happy to clean them up but want to check they are not intentional first 

otherwise it looks great!

baconpaul:
Yeah it's clearly accidental. I did a push to revert to main for submodules. Let me re-review rest now.

src/surge-fx/SurgeFXProcessor.cpp
```diff
@@ -411,6 +469,26 @@ void SurgefxAudioProcessor::processBlock(juce::AudioBuffer<float> &buffer,
...
         }
     }
+
+    processBlockOSC();
+}
+
+// Pull incoming OSC events from ring buffer
+void SurgefxAudioProcessor::processBlockOSC()
+{
+    auto messages = oscRingBuf.popall();
```
baconpaul, line 479: So `popall` is an allocating call. We should really do a while pop with a has value.

Also shouldn't this get short circuited with a

```
if (!oscReceiving)
   return;
```

If we add the second we can deal with the first after the fact but to do a vector creation on every block is probably not good! (And we should check if we do this without a short circuit in surge also).

src/surge-fx/SurgeFXProcessor.cpp
```diff
@@ -492,6 +572,11 @@ void SurgefxAudioProcessor::setStateInformation(const void *data, int sizeInByte
             effectNum = xmlState->getIntAttribute("fxt", fxt_delay);
             resetFxType(effectNum, false);
 
+            oscPortIn = xmlState->getIntAttribute("oscpin", 0);
+            oscStartIn = xmlState->getBoolAttribute("oscin", false);
+            // start OSC, if variables merit it
+            oscCheckStartup = true;
```
baconpaul, line 578: Shouldn't this be `oscCheckStartup = oscPortIn > 0 && oscStartIn`?

src/surge-fx/SurgeFXProcessor.h
```diff
@@ -373,6 +406,9 @@ class SurgefxAudioProcessor : public juce::AudioProcessor,
     std::atomic<bool> audioRunning{false};
 
   public:
+    bool oscStartIn = true;
```
baconpaul, line 409: Why does this default to true?

baconpaul (changes requested):
OK fixed the submodules and did a closer review with a couple of bits of feedback. Looks good but would like at least you to take a peek at these three things I flagged. Thanks!

src/surge-fx/SurgeFXProcessor.cpp
pkstone: I used popall() in main Surge as well (it was suggested to me by Surgo, who wrote the RingBuffer). So if I change it here, I'll need to do so in SSP as well.

I believe that nothing will be pushed into the ring buffer is OSC is not receiving, so is that check necessary?

src/surge-fx/SurgeFXProcessor.cpp
pkstone: Those are checked further downstream (in tryLazyOscStartupFromStreamedState(). This code is copied from Surge XT, and was written mostly by Evil, and I haven't modified it much at all.

src/surge-fx/SurgeFXProcessor.h
pkstone: Good catch. Will fix.

src/surge-fx/SurgeFXProcessor.cpp
baconpaul: I looked at what popall does and it will make a vector of size 0 in that case. If we know it will be empty not checking it would be smart. The popall vs repeated pop is less of an urgent issue but we could avoid a potentially unbounded alloc by doin git.

pkstone: OK, sounds like that should be changed both here and in SSP, then.

pkstone: If I commit changes to this PR, will it reintroduce the submodule issues? (thanks for cleaning that up!)

PR #7676, opened 2024-06-28
Support builds on alternative c libraries other than glibc for linux

catarial:


baconpaul:
features.h is not a part of the standard and you included it, breaking all the builds. I think if you remove that and re-push we will see a clean CI.

baconpaul:
Oh my other question. Will __GLIBC__ be defined in, say, a mingw compile? And will that have exec info?

catarial:
> Oh my other question. Will **GLIBC** be defined in, say, a mingw compile? And will that have exec info?

https://stackoverflow.com/questions/6394512/standard-c-library-in-mingw

"MinGW does not build against glibc, it builds against msvcrt. As such, it uses libmsvcrtXX.a instead."

It seems like it doesn't

catarial:
> features.h is not a part of the standard and you included it, breaking all the builds. I think if you remove that and re-push we will see a clean CI.

should I wrap in a `#if LINUX` that header is needed to determine if glibc is being used

PR #7687, opened 2024-07-03
WTSE tearout

nuoun:
* add WTSE overlay tear out
* set WTSE overlay to include VCO number
* change LuaEditors > WavetableEquationEditor from class to struct (since all members are public)
* move +1 frame counter increment for preview graph from WavetableEquationEditor::rerenderFromUIState() to WavetablePreviewComponent
* remove scaling by 10 from WTSE frame slider
* remove redundant parentheses, style changes for Wavetable Script test in UnittestsLUA.cpp
* remove redundant breaks on line 170 and 344 in SurgeGUIEditorOverlays.cpp
* rename overlay tag WT_SCRIPTING_EDITOR to WTSCRIPT_EDITOR in SurgeGUIEditor
* change text width of default script to prevent horizontal scrollbar


baconpaul:
👍

PR #7695, opened 2024-07-09
luajitlib update, initial test

nuoun:
Initial test PR, **not intended** for merging yet.


baconpaul:
Ok the Linux problem looks pretty easy. You need need to compile position independent (so add -fPIC to the compile options)

windows looks like a bad path - maybe resolve to absolute a bit earlier in cmake? 

Oh and Linux docker always passes. I should fix that. I thought I had 

Very close! Good stuff

baconpaul:
I wonder if MSVC is placing it in bin/Release/luajit rather than bin/luajit?

nuoun:
Should at least see build artifacts in the binary output + /src/LuaJIT/src dir then I would think but there are none even though the build-msvc-luajit.bat file does copy the source files there.

nuoun:
call msvcbuild.bat -> "You must open a "Visual Studio Command Prompt" to run this script"
welp

baconpaul:
Ahh right. So you need to have run vcvars.bat so you need to find the compiler.

Just say to yourself three times "cmake solves no problems" then start googling.

The trick is going to be finding the right vcvars.bat to run.  According to the interwebs

```
You need to run the vcvars32.bat or vcvars64.bat instead. As of this writing these were in C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Auxiliary\Build.
```

but of course that is visual studio version dependent.  So that's quite hard to figure out the right thing to do. CMAKE may have enough information to find the path to the visual studio version so you can locate that file though. And then use bittiness to figure 32 or 64.

https://github.com/scikit-build/cmake-FindVcvars/blob/master/FindVcvars.cmake that might help?

nuoun:
Looks more promising than the dirty hacks I was trying so far, I'll try to implement it but I hope it works just as well with execute_process() as  ExternalProject_Add() (altho we might want the latter anyway to compile at build time rather than at CMake configure?)

baconpaul:
> Looks more promising than the dirty hacks I was trying so far, I'll try to implement it but I hope it works just as well with execute_process() as ExternalProject_Add() (altho we might want the latter anyway to compile at build time rather than at CMake configure?)

yeah it will. vcvars.bat can take commands to run iirc. or you can run it and just get the session variables inserted.

nuoun:
0xc0000135 supposedly means dll not found but that's all I got on this last one so far

baconpaul:
Looks exciting.

Let us know when you would like to merge it.

nuoun:
Almost there I hope, just a few more things on my todo list

nuoun:
Can be merged if it passes a final code review

baconpaul:
Have a busy afternoon but will review tonight or tomorrow morning worst case - and thanks!

baconpaul:
Ok all looks great. Squashing and merging!

PR #7703, opened 2024-07-10
Correct WT and Window handling og 16 full range wt files

### LINKED ISSUE #7694: Resolve int16 wt interpretation documentation and comparability
Hi,

According to the specification, if you want to encode 16 bits samples, you need the `flags = 0x0C` while it currently sets `flags = 0x04` which is for 15 bits samples.

Reference: https://github.com/surge-synthesizer/surge/blob/main/resources/data/wavetables/WT%20fileformat.txt#L11

When looking at https://github.com/surge-synthesizer/surge/blob/main/scripts/wt-tool/wt-tool.py#L43

The bug seems to be here: https://github.com/surge-synthesizer/surge/blob/94c8c5b02fb2dadf62dae70a146d67ae487309a1/scripts/wt-tool/wt-tool.py#L43

Cheers,
Alex

baconpaul:
wt files with full-range 16 bit ints (so 16-is-16 or flag 0x8 + 0x4 = 0xC) was broken in a couple of ways, but most importnatly, didn't survive a save/restore roundtrip.

This fixes it in three ways

1. When converting 16-as-16 to float scale accordingly with a different function and leave the ints untouched
2. Modify the window oscillator so if it gets a 16-as-16 it reduces the height of the ints to be consistent with a 15 with an extra shift, making I15 WT vs Window and I16 WT vs Window consistent
3. Make sure it streams properly. No change required but tested.

Update wt-tool to allow int16 and int15 options both

Closes #7694

### DIFF
--- a/src/common/dsp/oscillators/WindowOscillator.h
+++ b/src/common/dsp/oscillators/WindowOscillator.h
@@ -77,7 +77,7 @@ class WindowOscillator : public Oscillator
     void applyFilter();
     template <bool is_init> void update_lagvals();
 
-    void ProcessWindowOscs(bool stereo, bool FM);
+    template <bool FM, bool Full16> void ProcessWindowOscs(bool stereo);
     lag<double> FMdepth[MAX_UNISON];
     lag<float> l_morph;
 
--- a/src/common/dsp/vembertech/basic_dsp.h
+++ b/src/common/dsp/vembertech/basic_dsp.h
@@ -54,6 +54,15 @@ inline void i152float_block(short *s, float *f, int n)
     }
 }
 
+inline void i162float_block(short *s, float *f, int n)
+{
+    const float scale = 1.f / (16384.f * 2);
+    for (int i = 0; i < n; i++)
+    {
+        f[i] = (float)s[i] * scale;
+    }
+}
+
 inline void i16toi15_block(short *s, short *o, int n)
 {
     for (int i = 0; i < n; i++)
--- a/scripts/wt-tool/wt-tool.py
+++ b/scripts/wt-tool/wt-tool.py
@@ -40,7 +40,8 @@ def read_wt_header(fn):
     flags = {}
     flags["is_sample"] = (fl[0] & 0x01 != 0)
     flags["loop_sample"] = (fl[0] & 0x02 != 0)
-    flags["format"] = "int16" if (fl[0] & 0x04 != 0) else "float32"
+
+    flags["format"] = "int16-full-range" if (fl[0] == 0xC) else "int16" if (fl[0] & 0x04 != 0) else "float32"
     flags["samplebytes"] = 2 if (fl[0] & 0x04 != 0) else 4
 
     header["flags"] = flags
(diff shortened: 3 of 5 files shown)


src/common/dsp/vembertech/basic_dsp.h
```diff
@@ -54,6 +54,15 @@ inline void i152float_block(short *s, float *f, int n)
...
+inline void i162float_block(short *s, float *f, int n)
+{
+    const float scale = 1.f / (16384.f * 2);
+    for (int i = 0; i < n; i++)
+    {
+        f[i] = (float)s[i] * scale;
+    }
+}
+
 inline void i16toi15_block(short *s, short *o, int n)
```
abique, line 66: Is this function actually used? Why would you want to reduce precision?

src/common/dsp/vembertech/basic_dsp.h
```diff
@@ -54,6 +54,15 @@ inline void i152float_block(short *s, float *f, int n)
     }
 }
 
+inline void i162float_block(short *s, float *f, int n)
+{
+    const float scale = 1.f / (16384.f * 2);
+    for (int i = 0; i < n; i++)
+    {
+        f[i] = (float)s[i] * scale;
```
abique, line 62: Maybe `std::clamp(x, -1, 1)`?
I think you either need to clamp or assert that the input value is in the expected range.
You could do both as well that may catch programming error.

src/common/dsp/oscillators/WindowOscillator.h
```diff
@@ -77,7 +77,7 @@ class WindowOscillator : public Oscillator
     void applyFilter();
     template <bool is_init> void update_lagvals();
 
-    void ProcessWindowOscs(bool stereo, bool FM);
+    template <bool FM, bool Full16> void ProcessWindowOscs(bool stereo);
```
abique, line 80: Maybe in order to simplify the code, you could convert 16 bits wavetable to float and then have a single wavetable player?

It is a tradeoff, I believe that the win of running the wt on 16 bits is the memory but today I'm not sure how relevant it is versus 1 code path. Especially if you have a GUI.

abique:
I left some comments, but I don't have so much knowledge about surge either.

For the python part, python is a foreign language to me :innocent:

src/common/dsp/vembertech/basic_dsp.h
baconpaul: This is how we did it in the prior commit. Its what caused the bug but also what made the window oscillator work with i16

src/common/dsp/oscillators/WindowOscillator.h
baconpaul: The window oscillator and the wavetable oscillator are really quite different (window convolves with a pre-defined window whereas wavetable does a BLIT chain). Window uses the I16 and wavetable uses the F32 which is why we convert on load. But having a single oscillator would be a breaking change and a feature reduction.

As to why both functions weren't in one oscillator, the pragmatic answer i that the parameter sets are different and parameters used to be relatively non-dynamic. The longer answer is easier for you to get an answer to in your team than it is for me in mine :) (aka : ask claes! )

src/common/dsp/vembertech/basic_dsp.h
baconpaul: So the reason I don't clip it here is that the window oscillator has a wee bit of headroom but yeah i'm not sure where we want the clipping, but i think we don't want to *move* the clipping at this point.

baconpaul:
> I left some comments, but I don't have so much knowledge about surge either.
> 
> For the python part, python is a foreign language to me 😇

Thanks!

PR #7719, opened 2024-07-17
Update .wt file format documentation

mkruselj:
Addresses #7694

### DIFF
--- a/resources/data/wavetables/WT fileformat.txt
+++ b/resources/data/wavetables/WT fileformat.txt
@@ -7,7 +7,7 @@ byte
         0002: is a looped sample
         0004: if not set, wave data is in float32 format
               if set, wave data is in int16 format
-        0008: if not set, int16 data has a maximum peak of +/- 2^15 (legacy)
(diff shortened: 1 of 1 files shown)


resources/data/wavetables/WT fileformat.txt
```diff
@@ -1,14 +1,14 @@
...
 10-11	flags
-        0001: is sample
-        0002: loop sample
-        0004: if clear, wave data is in float32 format
+        0001: is a sample instead of a wavetable
+        0002: is a looped sample
+        0004: if not set, wave data is in float32 format
               if set, wave data is in int16 format
-        0008: if set, int16 data is in range 2^16; if not set it is in range 2^15 (legacy)
+        0008: if not set, int16 data has a maximum peak of +/- 2^15 (legacy)
```
abique, line 10: We're always loading the data as 16 bits, just the bit is set we add +6dB.

mkruselj: OK how to write that spec then, please? :)

baconpaul: If not set, a 0db wavetable has a peak of 2^14 (uses only 15 bits) as opposed to 2^15 (using 16 bits)

PR #7723, opened 2024-07-18
Some Lua cleanup

nuoun:
- Simplify Lua stack operations by replacing lua_pushstring() + lua_settable() with lua_setfield()
- Simplify sandbox whitelist adding by using vector of functions / modules
- Comment out some unused code


baconpaul:
Wow lua setfield is a handy function

baconpaul:
But code looks good to me

nuoun:
Tested some extra patches and can be merged I think

Issue #7359, opened 2023-12-04
Nimbus Improvements: SuperParasites FW and src -> lanczos and so on


baconpaul:
https://github.com/patrickdowling/superparasites/blob/master/supercell/dsp/granular_processor.h

https://community.vcvrack.com/t/surge-xt-2-2-want-to-help-beta-test/21084/38

baconpaul:
I've tagged the issue resolved since everything is done, just we need to toggle the CLOUDS_IS_CLOUDS in libs/eurorack/CMakeLists.txt at the appropriate point.

Probably try it toggled in ShortCircuit first.

baconpaul:
OK I tried this and kammerl and spectral cloud seem like no-ops so some more investigation needed here.

baconpaul:
https://www.kammerl.de/audio/clouds/

and there's the manual for the dynamic names

baconpaul:
PR #8385 gets the fix started but not finished

PR #7726, opened 2024-07-25
Move Nimbus from libsamplerate to our Lanczos downsampler

baconpaul:
1. Makes it ready to move to sst-effects
2. Makes it so we can almost remove libsamplerate (but some twist smidgeof work to do that)
3. Uses less CPU

Addresses #7359

### DIFF
--- a/src/common/dsp/effects/NimbusEffect.cpp
+++ b/src/common/dsp/effects/NimbusEffect.cpp
@@ -21,7 +21,6 @@
  */
 
 #include "NimbusEffect.h"
-#include "samplerate.h"
 #include "DebugHelpers.h"
 #include "fmt/core.h"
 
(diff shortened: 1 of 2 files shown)


mkruselj:
Does this mean we can remove libsamplerate from our list of deps?

baconpaul:
Ha. See the plan I just posted on discord

I wanted to do it as a separate step for bisect reasons but "yes"

baconpaul:
For GitHub folks who find this issue this was the discord comment

Doing a bit of surge / effects / shortcircuit combined work. Moving away from the libsamplerate (SRC) dependency. It's a bit multi step

step one - port Nimbus from SRC to Lanczos. That's done and a PR is in
step two - clean up a couple of tiny other SRC deps (the Twist FM uses the SRC linear mode and some tests do). 
step three - Move Nimbus to sst-effects with some cleverness for the euro-rack depdency in cmake
step four - Add clouds to shortcircuit, but with the parasites firmware turned on!

PR #7727, opened 2024-07-25
Add voice data to formula modulator

nuoun:
- state.voice_limit = poly limit
- state.voice_n = active voice index
- state.voice_order = voiceOrderAtCreate

Just a draft until we decide if this is what we want and if so if this is the best way

I think the first two would be very useful to have at voice level for anything from unison detune to arpeggiator modulators. VoiceOrder was already passed to the voice constructor so easy to add but not sure how useful it will be compared to an actual active voice counter (which I still need to figure out how to do).

### DIFF
--- a/src/common/SurgeStorage.h
+++ b/src/common/SurgeStorage.h
@@ -1344,6 +1344,8 @@ class alignas(16) SurgeStorage
     bool oscReceiving{false};
     bool oscSending{false};
 
+    int voiceCount;
+
     bool getOverrideDataHome(std::string &value);
     void createUserDirectory();
 
--- a/src/common/SurgeStorage.h
+++ b/src/common/SurgeStorage.h
@@ -1344,7 +1344,9 @@ class alignas(16) SurgeStorage
     bool oscReceiving{false};
     bool oscSending{false};
 
-    int voiceCount;
+    int voiceCount; // TODO: use SurgeSynthesizer class to fetch synth->polydisplay directly from
+                    // valueAt() in FormulaModulationHelper.cpp where it's needed and remove
+                    // this and its assignment in SurgeSynthesizer.cpp
 
     bool getOverrideDataHome(std::string &value);
     void createUserDirectory();
--- a/src/common/dsp/SurgeVoice.cpp
+++ b/src/common/dsp/SurgeVoice.cpp
@@ -153,8 +153,8 @@ SurgeVoice::SurgeVoice(SurgeStorage *storage, SurgeSceneStorage *oscene, pdata *
                        int velocity, int channel, int scene_id, float detune,
                        MidiKeyState *keyState, MidiChannelState *mainChannelState,
                        MidiChannelState *voiceChannelState, bool mpeEnabled, int64_t voiceOrder,
-                       int voiceN, int polyLimit, int32_t host_nid, int16_t host_key,
-                       int16_t host_chan, float aegStart, float fegStart)
+                       int32_t host_nid, int16_t host_key, int16_t host_chan, float aegStart,
+                       float fegStart)
 //: fb(storage,oscene)
 {
 #ifdef VOICE_LIFETIME_DEBUG
@@ -191,10 +191,9 @@ SurgeVoice::SurgeVoice(SurgeStorage *storage, SurgeSceneStorage *oscene, pdata *
     state.key = key;
     state.keyRetuningForKey = -1000;
     state.channel = channel;
+
     state.voiceOrderAtCreate = voiceOrder;
-    state.polyLimit = polyLimit;
-    // state.polyLimit = storage->getPatch().polylimit.val.i;
-    state.voiceN = voiceN;
+    state.voiceMax = storage->getPatch().polylimit.val.i;
 
     state.velocity = velocity;
     state.fvel = velocity / 127.f;
--- a/src/common/dsp/SurgeVoice.cpp
+++ b/src/common/dsp/SurgeVoice.cpp
@@ -153,8 +153,8 @@ SurgeVoice::SurgeVoice(SurgeStorage *storage, SurgeSceneStorage *oscene, pdata *
                        int velocity, int channel, int scene_id, float detune,
                        MidiKeyState *keyState, MidiChannelState *mainChannelState,
                        MidiChannelState *voiceChannelState, bool mpeEnabled, int64_t voiceOrder,
-                       int32_t host_nid, int16_t host_key, int16_t host_chan, float aegStart,
-                       float fegStart)
+                       int voiceN, int polyLimit, int32_t host_nid, int16_t host_key,
+                       int16_t host_chan, float aegStart, float fegStart)
 //: fb(storage,oscene)
 {
 #ifdef VOICE_LIFETIME_DEBUG
@@ -192,6 +192,9 @@ SurgeVoice::SurgeVoice(SurgeStorage *storage, SurgeSceneStorage *oscene, pdata *
     state.keyRetuningForKey = -1000;
     state.channel = channel;
     state.voiceOrderAtCreate = voiceOrder;
+    state.polyLimit = polyLimit;
+    // state.polyLimit = storage->getPatch().polylimit.val.i;
+    state.voiceN = voiceN;
 
     state.velocity = velocity;
     state.fvel = velocity / 127.f;
--- a/src/common/dsp/SurgeVoice.cpp
+++ b/src/common/dsp/SurgeVoice.cpp
@@ -193,7 +193,7 @@ SurgeVoice::SurgeVoice(SurgeStorage *storage, SurgeSceneStorage *oscene, pdata *
     state.channel = channel;
 
     state.voiceOrderAtCreate = voiceOrder;
-    state.voiceMax = storage->getPatch().polylimit.val.i;
+    state.polylimit = storage->getPatch().polylimit.val.i;
 
     state.velocity = velocity;
     state.fvel = velocity / 127.f;
--- a/src/common/dsp/modulators/FormulaModulationHelper.cpp
+++ b/src/common/dsp/modulators/FormulaModulationHelper.cpp
@@ -235,7 +235,7 @@ end
             addn("tempo", s.tempo);
             addn("songpos", s.songpos);
             addb("released", s.released);
-            addb("is_display", s.is_display);
+            addb("is_rendering_to_ui", s.is_display);
             addb("clamp_output", true);
 
             auto cres = lua_pcall(s.L, 1, 1, 0);
--- a/src/common/dsp/modulators/FormulaModulationHelper.cpp
+++ b/src/common/dsp/modulators/FormulaModulationHelper.cpp
@@ -533,7 +533,7 @@ void valueAt(int phaseIntPart, float phaseFracPart, SurgeStorage *storage,
         addb("released", s->released);
 
         addn("voice_id", s->voiceOrderAtCreate);
-        addn("voice_polylimit", s->polylimit);
+        addn("poly_limit", s->polylimit);
 
         addn("poly_at", s->polyat);
         addn("mpe_bend", s->mpebend);
@@ -548,7 +548,7 @@ void valueAt(int phaseIntPart, float phaseFracPart, SurgeStorage *storage,
 
     addnil("retrigger_AEG");
     addnil("retrigger_FEG");
-    addb("is_display", s->is_display);
+    addb("is_rendering_to_ui", s->is_display);
 
     if (s->subAnyMacro)
     {
--- a/src/common/dsp/modulators/FormulaModulationHelper.cpp
+++ b/src/common/dsp/modulators/FormulaModulationHelper.cpp
@@ -526,6 +526,9 @@ void valueAt(int phaseIntPart, float phaseFracPart, SurgeStorage *storage,
         addn("rel_velocity", s->releasevelocity);
         addn("channel", s->channel);
         addb("released", s->released);
+        addn("voice_order", s->voice_order);
+        addn("voice_limit", s->voice_limit);
+        addn("voice_n", s->voice_n);
 
         addn("poly_at", s->polyat);
         addn("mpe_bend", s->mpebend);
@@ -891,6 +894,9 @@ void setupEvaluatorStateFrom(EvaluatorState &s, const SurgeVoice *v)
     s.channel = v->state.channel;
     s.velocity = v->state.velocity;
     s.releasevelocity = v->state.releasevelocity;
+    s.voice_order = v->state.voiceOrderAtCreate;
+    s.voice_limit = v->state.polyLimit;
+    s.voice_n = v->state.voiceN + 1;
 
     s.polyat =
         v->storage
--- a/src/common/dsp/modulators/FormulaModulationHelper.cpp
+++ b/src/common/dsp/modulators/FormulaModulationHelper.cpp
@@ -34,6 +34,7 @@ namespace Formula
 {
 
 void setupStorage(SurgeStorage *s) { s->formulaGlobalData = std::make_unique<GlobalData>(); }
+
 bool prepareForEvaluation(SurgeStorage *storage, FormulaModulatorStorage *fs, EvaluatorState &s,
                           bool is_display)
 {
@@ -80,6 +81,11 @@ bool prepareForEvaluation(SurgeStorage *storage, FormulaModulatorStorage *fs, Ev
 
     if (firstTimeThrough)
     {
+        // setup global table
+        lua_newtable(s.L);
+        lua_setglobal(s.L, "shared");
+
+        // setup prelude
         Surge::LuaSupport::loadSurgePrelude(s.L);
         auto reserved0 = std::string(R"FN(
 function surge_reserved_formula_error_stub(m)
@@ -229,6 +235,7 @@ end
             addn("tempo", s.tempo);
             addn("songpos", s.songpos);
             addb("released", s.released);
+            addb("is_display", s.is_display);
             addb("clamp_output", true);
 
             auto cres = lua_pcall(s.L, 1, 1, 0);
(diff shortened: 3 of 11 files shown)


mkruselj:
First two are indeed very good. voice_n in particular will be great to provide i.e. fixed panning offsets for each voice (think vintage Oberheim). I am not sure if voice_order is useful at all... active voice counter seems a lot more useful.

baconpaul:
There’s already a voice counter kicking around somewhere. It’s how we do the last voice calculation on release. Not sure we need to add another one. But also since you can’t write cross voice modulators I’m not quite sure how it would be used in formula. 

Is the voice index stable under pother note releases? Let’s make sure to define these semantics carefully

nuoun:
> First two are indeed very good. voice_n in particular will be great to provide i.e. fixed panning offsets for each voice (think vintage Oberheim). I am not sure if voice_order is useful at all... active voice counter seems a lot more useful.

So I got rid of voice_order. Oberheim style voice panning would be amazing to have as well, is there currently a way to do panning at voice level rather than scene?

> There’s already a voice counter kicking around somewhere. It’s how we do the last voice calculation on release. Not sure we need to add another one. But also since you can’t write cross voice modulators I’m not quite sure how it would be used in formula.

Yes there is the atomic int polydisp in SurgeSynthesizer.cpp, which I'm currently caching and passing directly to SurgeVoice() to set the voice index. Then just updating that value in FormulaModulationHelper.cpp like "cycle" / "intphase" is being done makes it a counter rather than an index.

> Is the voice index stable under pother note releases? Let’s make sure to define these semantics carefully

Will poke around and test a bit more over the weekend as there's probably some edge cases that I missed but so far it seems to work fine for poly play mode in a single scene at least.

mkruselj:
You can of course use the Random modulator to modulate the scene panning. But this is random and not fixed.

nuoun:
> You can of course use the Random modulator to modulate the scene panning. But this is random and not fixed.

Closest I could think of so far is the stereo filter configuration but that's far from ideal. I suppose it would be nice to have actual panning sliders per mixer channel at some point but for individual voices it might need to be done at the oscillator level instead?

Andreya-Autumn:
> Oberheim style voice panning would be amazing to have as well, is there currently a way to do panning at voice level rather than scene?

The pan control is in fact per-voice, so the answer is yes, apply any voice-level modulator to pan. Be advised in wide mode, pan is borked, and you should probably avoid going past ≈73 percent. But you can pan voices independently even in the mono modes!  

I'm not familiar with the Oberheim but I'm guessing voice index determines pan such that the voices spread evenly across the stereo field? In that case, the additions discussed here would let you would let you do that indeed. 

> Will poke around and test a bit more over the weekend as there's probably some edge cases that I missed but so far it seems to work fine for poly play mode in a single scene at least.

Make sure you test with two scenes. Also latch one scene and not the other. Basically throw nonsense at it and see what comes out. The Surge voice manager is... complicated. So let's make sure we actually get what we want.

Andreya-Autumn:
> but for individual voices it might need to be done at the oscillator level instead?

We have that planned for XT2. It's in the issues somewhere. Alongside the "Pan is borked" one I mentioned. :)

nuoun:
> The pan control is in fact per-voice, so the answer is yes, apply any voice-level modulator to pan. Be advised in wide mode, pan is borked, and you should probably avoid going past ≈73 percent. But you can pan voices independently even in the mono modes!

That's good to know, thank you! :)

> I'm not familiar with the Oberheim but I'm guessing voice index determines pan such that the voices spread evenly across the stereo field? In that case, the additions discussed here would let you would let you do that indeed.

Not sure how it's exactly done over different Oberheim models but if I'm not mistaken the essence is that it adds both a fixed panning control for each voice plus a global control to spread voices out over the stereo field, the main difference with Surge would be that it also affects unison voices. Discodsp's OB-Xd VST (which has an older free version iirc) does a pretty good job at implementing this I think and it makes it sound rather lush and wide with the added stereo depth.

> Make sure you test with two scenes. Also latch one scene and not the other. Basically throw nonsense at it and see what comes out. The Surge voice manager is... complicated. So let's make sure we actually get what we want.

Yeah the different scenes might be an issue in the current implementation, luckily the scenes are separated well in SurgeSynthesizer.cpp at least and the complexity there does make me hope we can get away with the smallest possible change here.

Andreya-Autumn:
Sounds like pan per voice + a width control. We have those, but unison voices position (in wide mode) are fixed. So we can't get quite all the way there in XT1. For XT2 we've discussed improving control over unison panning strategy. But yeah, that's for XT2.

mkruselj:
The crucial thing is that every voice is a static offset on an Oberheim. So you have say 8 voice cards and as you play polyphonically they either rotate in round robin (incrementing the voice index) or they advance the voice index only when a new key is played (meaning playing the same key multiple times only takes one voice - we have that as an additional option on Play Mode right click, it's the "piano mode"). That's the voice index thing we'd need to properly emulate an Obie.

baconpaul:
We do not have that voice index (nor do we need it) easily in the synth

nuoun:
Yes actual round robin would need some data structure to keep track of active voices, voiceOrderAtCreate which we do have is the next best thing which I think could do the rotating panning but not something like an arpeggiator, closing this now as it seems there's no further enthusiasm to add this.

baconpaul:
Something like this would be super useful just we need to make sure we know what it is. 

Like a stable voice index is hard because of releases. So do we want which voice of how many sounding we are sorted by creation order? Maybe? 

So I’m enthusiastic for the class of idea just want to make sure to be careful since once we add something it’s hard to remove

mkruselj:
Why close the PR? The first two additions are great!

nuoun:
> Why close the PR? The first two additions are great!

It was just an idea but thinking this over more it's just not something that can be easily added properly I don't think, keeping track of voice releases is not trivial as Paul said and just using voiceOrderAtCreate % maxpoly is rather limited.

Andreya-Autumn:
Spontaneously I think what Paul said sounds good. 

> So do we want which voice of how many sounding we are sorted by creation order? Maybe?

If I understand that right. It means we have two things. Let's call them state.currentVoiceCount and state.VoiceNum
(better names needed). 

Play and hold a single note, they're both 1. 
Play and hold another, currentVoiceCount is 2, voiceNum is still 1 for the first, and 2 for the other. 
Play and hold yet another, you get the pattern. The last voice will have voiceNum 3.
Release the first one, count goes down to 2, the remaining ones become voiceNum 1 and 2. 

That'd make sense to me!

nuoun:
If we have certain modulation (lets say panning) depending on the voice index and have the voice index of still active voices change dynamically after previous voices end, it would make existing voices start doing different things, is that really what we want though? - (we could save the initial voice index inside the state table but that would mean that each voice does not have awareness of it's position relative to the other voices)

nuoun:
Imo, a round robin approach for voice index without the voice stealing makes most sense, so in your example after releasing the first note index 1 becomes available again and would be used again after we roll over when going over the max poly number and start counting from 1 again, and 2 and 3 stay reserved as those voices are still playing.

That way each voice will keep it's own unique voice index, but this requires some data structure which needs to keep track of active voices so each index stays unique, without having a good way to delete active voices after release. My initial thought is a vector that gets renewed by each active voice and deletes the ones that have not gotten a recent update.

Andreya-Autumn:
Well I think that might at least be one thing that we want! 

For pan for example, it does sound super fun to me to have voices move around dynamically to make space for one new ones, or take up space left by ended ones. Can always slew limit them if the shifts are too sudden. And the round-robin thing wouldn't really work here because "how much space is there for each voice" is answered by current voice count not max poly. 

But yeah I see the merits of both.

nuoun:
There's also an issue when a voice ends and a new voice starts during the same sample block, can't really reorder that properly without a way to keep track of the leased out index values I would think (but luckily we do have voiceOrderAtCreate for that which will always keep a unique number)

nuoun:
Still a draft but after sleeping on it and consulting the missus this is the WIP version of what I'd like to suggest here, the parameters that get added to the voice formula editor are:

state.voice_count = updating poly voice counter as used by GUI
state.voice_limit = polylimit
state.voice_order = voiceOrderAtCreate (incremental counter for each new voice gives each a unique ID which will help on the Lua side to keep voices apart)
state.voice_index = basic round robin voice index constructed from voiceOrderAtCreate % polylimit (optional as this and more can also be done on the Lua side)

This is more or less what I added initially but still needs more work, specifically the voice_counter is still an issue as the atomic int polydisplay seems to not be passable as a pointer like all the other values to the voice constructor SurgeVoice() in SurgeSynthesizer.cpp so I'm currently saving and grabbing it using SurgeStorage but would much rather also pass it directly. (May need a `this` pointer from SurgeSynthesizer.cpp passed to SurgeVoice() as polydisplay is public?)

The main change here is to add a "global" table that is shared with every Lua instance (just like the "surge" table for the prelude currently is but this new table should also be cleared every new function environment so it can be garbage collected). Having this table available from the initial Lua state setup allows all the individual formula modulators (and WTSE) over all voices to share parameters in the same table so more complex stuff as figuring out the relative position of modulators/voices becomes possible.

can split this up in 2 smaller PRs, one for the global table and one for the new parameters if that helps with reviewing but am of course interested in the thoughts on this so far, 

(Also still have some stack issues with the clear loop of the global table it seems, wip also)

nuoun:
So this does the following

- Adds voice_id, voice_max, voice_count, is_display parameters to formula modulator
- Adds voiceCount int to SurgeStorage
- Fakes voice_count of "1" and sets defaults for other params on display calls
- Adds "shared" table to formula modulator
- Sets entries of "shared" table to nil on new setSurgeFunctionEnvironment()
- Adds "shared" table to formula debugger
- Adds lua_isfunction to formula debugger
- Does a bit of cleanup

I still need to do some testing on other platforms but if we can reach consensus whether this is what we want, it passes review and finally when we agree on the final var names this can be merged I hope!

baconpaul:
Sorry I’m unclear is this ready for review now? Or still a draft? Fine either way of course but if it is ready and you are waiting and I think it’s a draft that’s a bummer :)

mkruselj:
It seems we're not in Draft anymore, Alice.

baconpaul:
Yeah OK I rebased it properly and pushed. Now it's 8 files changed rather than 50! Phew.

Let me let CI run again but I'll look at the code now

src/common/dsp/SurgeVoice.cpp
```diff
@@ -191,7 +191,9 @@ SurgeVoice::SurgeVoice(SurgeStorage *storage, SurgeSceneStorage *oscene, pdata *
     state.key = key;
     state.keyRetuningForKey = -1000;
     state.channel = channel;
+
     state.voiceOrderAtCreate = voiceOrder;
+    state.voiceMax = storage->getPatch().polylimit.val.i;
```
baconpaul, line 196: Is there any reason we call this 'voiceMax' vs 'polylimit'? We call it 'polylimit' everywhere else in the code so the new name is a potential source of confusion.

src/common/SurgeStorage.h
```diff
@@ -1344,6 +1344,8 @@ class alignas(16) SurgeStorage
     bool oscReceiving{false};
     bool oscSending{false};
 
+    int voiceCount;
+
```
baconpaul, line 1348: I'm kind of bummed we have to put voiceCount onto SurgeStorage. Is there some reason it can't live on SurgeSynthesizer instead, where it more naturally lives? Storage is more 'static' state than 'runtime' state.

If we move it to Synth, what changes? Do you not get access to it at setup time or something?

If the problem is SLFO don't have a Synth reference, we can fix that probably.

This might be what we have to do though; but if it is what we have to do then (1) we should probably make it `std::atomic<int32_t>` and (2) we should add a comment that its somewhat dissatisfying and why we added i there.

src/common/dsp/modulators/FormulaModulationHelper.cpp
```diff
@@ -492,33 +506,24 @@ void valueAt(int phaseIntPart, float phaseFracPart, SurgeStorage *storage,
...
-        addn("release", s->r);
-    }
-    if (true /* s->subLfoParams */)
-    {
-        addn("rate", s->rate);
-        addn("amplitude", s->amp);
-        addn("startphase", s->phase);
-        addn("deform", s->deform);
-    }
+    addn("rate", s->rate);
```
baconpaul, line 516: Thanks for cleaning up this vestigial stuff!

src/common/dsp/modulators/FormulaModulationHelper.cpp
```diff
@@ -810,13 +820,26 @@ std::vector<DebugRow> createDebugDataOfModState(const EvaluatorState &es)
         }
     };
 
-    rec(0, false);
-    lua_pop(es.L, -1);
+    std::vector<std::string> tablesList = {es.stateName, "shared"};
```
baconpaul, line 823: So `es.stateName` exists so each modulator gets a unique state. "shared" allows them to be shared. But I just want to confirm - that's shared within the LUA state, so it's not a cross-instance global right?

Also since we use the string "shared" in several places, I wonder if we want a constexpr char* for it which we use rather than a quoted string.

Very nice moving this to a loop here by the way. Diff confused me at first but I like this approach way better.

src/common/dsp/modulators/FormulaModulationHelper.cpp
```diff
@@ -810,13 +820,26 @@ std::vector<DebugRow> createDebugDataOfModState(const EvaluatorState &es)
...
+    for (const auto &t : tablesList)
+    {
+        lua_getglobal(es.L, t.c_str());
+        if (!lua_istable(es.L, -1))
+        {
+            lua_pop(es.L, -1);
+            rows.emplace_back(0, "Error", "Not a Table");
+            return rows;
```
baconpaul, line 831: Now this is a loop we may not want to do a return here since there's more recursion to be done. Probably we want to put the rec into an else so we can at least do the other table.

but honestly if one isn't there the other probably isnt' either so feel free to ignore this

baconpaul (changes requested):

nuoun:
Thanks for the review (and cleaning up the mess!), I'll try to go over your comments one by one.

src/common/dsp/SurgeVoice.cpp
nuoun: Good point, the undecided name in the Lua state table was guiding me here but makes more sense to keep it consistent.

src/common/SurgeStorage.h
nuoun: Was not too happy with it either but was the only way I got it working so far, was having trouble passing that atomic int as a pointer and doing it static it won't update with voice changes since it only gets passed to SurgeVoice() on creation.

Tried adding a `this` pointer from SurgeSynthesizer to pass to SurgeVoice() and from there to valueAt() in FormulaModulationHelper.cpp where it is needed but could not get that working, ideally we would use it like SurgeGUIEditor.cpp [here](https://github.com/surge-synthesizer/surge/blob/af631b506566184f438f5dfeb24c7acb7d8e6ed7/src/surge-xt/gui/SurgeGUIEditor.cpp#L797) but adding the SurgeSynth class reference seemed beyond my C++ skills and this seemed like a smaller change at that point. So yeah, I'm all for leaving SurgeStorage as is if there is a better way.

src/common/dsp/modulators/FormulaModulationHelper.cpp
nuoun: "shared" is the name of an empty table that gets added [here](https://github.com/nuoun/surge/blob/1ee52c6a8317ff7f31cbc4d59e72bff41a432571/src/common/dsp/modulators/FormulaModulationHelper.cpp#L84) the same way the "surge" table is added that is used to load in the prelude. Since this gets added on the Lua state just once using that `if (firstTimeThrough)` block and before the sandbox environment and evaluation state gets added it is accessible for every Lua Formulator instance in voice(s) and scenes (IIRC only the unittests use their own Lua state).

I wanted to call it "global" initially but while that is not a keyword it has a bunch of uses in Lua so figured this might be less confusing. Both setglobal() and getglobal() use `const char *name` so I suppose we could use a constexpr char* for that vector too I suppose.

src/common/dsp/modulators/FormulaModulationHelper.cpp
nuoun: Maybe a `continue;` would be nicer then instead of that nested return?

baconpaul: Yes that sounds correct to me.

src/common/dsp/modulators/FormulaModulationHelper.cpp
baconpaul: Yeah I have no problem with shared. I just meant lets add

```
static constexpr const char* surgeTableName{"surge"};
static constexpr const char* sharedTableName("shared"};
```

somewhere and not use string literals. Was a minor comment to that end.

src/common/dsp/SurgeVoice.cpp
baconpaul: Yeah cool, so polylimit it is!

src/common/SurgeStorage.h
baconpaul: OK for now lets add it then and put a comment and open an issue to ponder a better spot.

It's not entirely wrong. But it's also not entirely right. But I'm not sure where else it could go right now.

nuoun:
Ok if this passes and there are no other changes needed then I think we're good to go, only thing that remains is to decide on the final parameter names inside the Formula state table, currently they're named:

- voice_count
- voice_polylimit
- voice_id
- is_display
- shared for the new global table name

Then when this is merged I will open an issue to remove voiceCount from SurgeStorage, ponder about a new tutorial patch to showcase the new features and have a look at what can be added to the manual.

PR #7728, opened 2024-07-25
Remove libsamplerate dependency

baconpaul:
After porting Nimbus ot lanczos in 4b3ffe6 we can now remove the libsamplerate dependency altogether as long as we fix up the twist fm (which we did with a simple application of another lanczos ds) and remove src code from the unit tests (where it was explicitly tested)


baconpaul:
Please let CI run to completion here since there are unit tests involved

PR #7738, opened 2024-08-05
Update tempo after patch load and calc ppqPos

dhatadi:
I found this while trying out a patch with a synced MSEG. It was using the default 120bpm.

* Update tempo after patch load if desired and ensure we calculate time_data.ppqPos while processing.


dhatadi:
I remembered to clang format this time! :)

baconpaul:
Thanks.,

PR #7745, opened 2024-08-08
Fix testrunner warnings

nuoun:


baconpaul:
Oh @nuoun I missed this. DO you want to merge it?

nuoun:
Yes please! (its just to suppress the warnings during the unit tests by that following for loop because Formula and WTSE don't use the same tables yet)

baconpaul:
Great! Feel free to bug one of us if we leave a ready-to-go PR dangling!

PR #7755, opened 2024-08-14
WTSE overview class refactor

nuoun:
- Changes WTSE overview class constructor to use similar style as FormulaModulatorEditor
- Renames WavetableEquationEditor class to WavetableScriptEditor
- Renames defaultWavetableFormula string to defaultWavetableScript
- Adds preliminary Prelude to WTSE
- Adds changes to WTSE UI
- Adds WavetableScriptEditState struct to DAWExtraStateStorage and SurgePatch load_xml()/save_xml()
- Increments XML file version to v25


nuoun:
Big PR with changes to SurgePatch that could use a thorough review when you can find the time, no rush so please take your time.

src/common/SurgeStorage.h
```diff
@@ -921,6 +922,11 @@ struct DAWExtraStateStorage
             bool hasCustomEditor = false;
         } oscExtraEditState[n_scenes][n_lfos];
 
+        struct WavetableScriptEditState
+        {
+            int codeOrPrelude{0};
```
baconpaul, line 927: Maybe a comment here briefly about what 'codeOrPrelude' means would be useful?

src/surge-xt/gui/overlays/LuaEditors.cpp
```diff
@@ -1010,7 +1021,455 @@ struct WavetablePreviewComponent : juce::Component
...
-WavetableEquationEditor::WavetableEquationEditor(SurgeGUIEditor *ed, SurgeStorage *s,
+struct WavetableScriptControlArea : public juce::Component,
+                                    public Surge::GUI::SkinConsumingComponent,
+                                    public Surge::GUI::IComponentTagValue::Listener,
+                                    public juce::ComboBox::Listener,
+                                    public juce::TextEditor::Listener
+{
+    enum tags
+    {
+        tag_select_tab = 0x597500,
```
baconpaul, line 1032: We probably want to use a different value here than other control areas just in case. Like just choose some other random starting number. In case messages do cross (which they should not)

src/surge-xt/gui/overlays/LuaEditors.cpp
```diff
@@ -1010,7 +1021,455 @@ struct WavetablePreviewComponent : juce::Component
...
+    std::unique_ptr<Surge::Widgets::MultiSwitchSelfDraw> codeS, applyS, generateS;
+    std::unique_ptr<juce::ComboBox> resolutionB;
+    std::unique_ptr<juce::TextEditor> framesT;
+
+    void paint(juce::Graphics &g) override { g.fillAll(skin->getColor(Colors::MSEGEditor::Panel)); }
+
+    void onSkinChanged() override { rebuild(); }
+
+    JUCE_DECLARE_NON_COPYABLE_WITH_LEAK_DETECTOR(WavetableScriptControlArea);
+};
```
baconpaul, line 1263: I didn't read this all that carefully but I am assuming it is ok since it looks copied from other similar with appropriate changes. But we should add an item to your work list before we ship 1.4 to do an accessibility check here (which I can help with).

src/surge-xt/gui/overlays/LuaEditors.cpp
```diff
@@ -1010,7 +1021,455 @@ struct WavetablePreviewComponent : juce::Component
...
+    if (controlArea->applyS->isEnabled())
+    {
+        return std::make_pair("Close Wavetable Script Editor",
+                              "Do you really want to close the wavetable editor? Any "
+                              "changes that were not applied will be lost!");
+    }
+    return std::nullopt;
+}
+
+/*
```
baconpaul, line 1471: No reason to leave this commented out code around is there? We have git history if we need it.

src/surge-xt/gui/overlays/LuaEditors.h
```diff
@@ -122,20 +167,17 @@ struct FormulaModulatorEditor : public CodeEditorContainerWithApply, public Refr
...
-
-    JUCE_DECLARE_NON_COPYABLE_WITH_LEAK_DETECTOR(FormulaModulatorEditor);
+    JUCE_DECLARE_NON_COPYABLE_WITH_LEAK_DETECTOR(WavetableScriptEditor);
 };
 
-struct WavetablePreviewComponent;
-
-struct WavetableEquationEditor : public CodeEditorContainerWithApply,
+/*
+struct WavetableScriptEditor : public CodeEditorContainerWithApply,
```
baconpaul, line 174: Same on commented out code here

src/surge-xt/gui/overlays/LuaEditors.h
```diff
@@ -88,32 +88,77 @@ struct FormulaModulatorEditor : public CodeEditorContainerWithApply, public Refr
...
     int lfo_id, scene;
+    int32_t updateDebuggerCounter{0};
 
-    void onSkinChanged() override;
-    void setApplyEnabled(bool b) override;
+    DAWExtraStateStorage::EditorState::FormulaEditState &getEditState();
+
+    bool shouldRepaintOnParamChange(const SurgePatch &patch, Parameter *p) override
+    {
+        return false;
```
baconpaul, line 118: This is fine for now but we should add an item to a list which is like "what happens if the wt editor is torn out and you change oscillator from wavetable to twist" or some such. We should make that a conscious decision at least.

baconpaul (changes requested):
This all looks great. Really just some minor cosmetic changes

I think we should also create an issue with a checklist for what we want to do to get to minimal viable WTSE. I added a few items here which would go on that list.

But I'm definitely happy to merge this

src/surge-xt/gui/overlays/LuaEditors.cpp
baconpaul: ha sorry i should have checked.

src/surge-xt/gui/overlays/LuaEditors.cpp
nuoun: I will add this to the WTSE issue.

src/surge-xt/gui/overlays/LuaEditors.cpp
nuoun: No worries, I'm just glad I did it right :)

src/surge-xt/gui/overlays/LuaEditors.h
nuoun: Did a little test and I think this works fine, changing oscillator type does not affect the torn out window and the WT generation still works, ie. the generated wavetable is ready to go after switching back to the WT oscillator.

nuoun:
Good to hear, I think the old WTSE issue is still open so we can add start referring to that issue and I shall create a little roadmap / list of issues there.

And this last commit should take care of the unused code and add a description for the struct in SurgeStorage.

PR #7758, opened 2024-08-14
Formula scene/play mode variables

nuoun:
- Adds scene_mode, poly_mode, split_point
- Renames is_rendering_to_ui -> is_rendering_to_UI


mkruselj:
I don't think capitalizing UI iim variable name is needed nor necessary. I would suggest reverting that, please.

nuoun:
Added a little check that converts to channel number for the channel split mode. I think this is good to go and I hope this will be the last variable that gets added for now as there are already a lot to keep track of.

nuoun:
One last rename: poly_mode should be play_mode user facing.

Issue #7777, opened 2024-08-29
Wavetable morph interpolation is not continuous


blancoberg:
**Bug Description:**
The interpolation between frames is not continuous. This is especially apparent on noisy wavetables. 

**Surge XT Version**
All versions

* Version:
* Plugin Type: [VST2, VST3, AU]
* Bitness: [32-bit or 64-bit]

**Reproduction Steps:**
1. Load the preset included in this ticket
2. Increase LFO1 rate to make the glitching more appearant. 

**Expected Behavior:**
The signal should be continuous and have a lot less noise

[wt_interpolation_bug.zip](https://github.com/user-attachments/files/16796824/wt_interpolation_bug.zip)

mkruselj:
Also fixed now, so closing.

PR #7778, opened 2024-08-29
wavetable morph interpolation discontinuity

blancoberg:
This addresses the following:

**Continuous frame interpolation**
The interpolation between frames weren't continuous, which could introduce lots of noise in some cases. It was especially apparent in noisy/phase shifting wavetables.  

**Seamless loop in "one shot"-mode**
So it turns out that you can actually loop "one shots". In one-shot-mode the unison voices parameter acts as a loop counter.
Anything between 1-6 loops the sample that amount of times. Anything over 7 will loop forever. 
The loop was not seamless though, so that has been adressed.

Addresses #7777

### DIFF
--- a/src/common/dsp/oscillators/WavetableOscillator.cpp
+++ b/src/common/dsp/oscillators/WavetableOscillator.cpp
@@ -62,10 +62,12 @@ void WavetableOscillator::init(float pitch, bool is_display, bool nonzero_init_d
 
     n_unison = limit_range(oscdata->p[wt_unison_voices].val.i, 1, MAX_UNISON);
 
+    isOneShot = 0;
     if (oscdata->wt.flags & wtf_is_sample)
     {
         sampleloop = n_unison;
         n_unison = 1;
+        isOneShot = 1;
     }
 
     if (is_display)
@@ -89,7 +91,8 @@ void WavetableOscillator::init(float pitch, bool is_display, bool nonzero_init_d
 
     float intpart;
     shape *= ((float)oscdata->wt.n_tables - 1.f + nointerp) * 0.99999f;
-    tableipol = modff(shape, &intpart);
+    tableipol = shape;
+    modff(shape, &intpart);
     tableid = limit_range((int)intpart, 0, std::max((int)oscdata->wt.n_tables - 2 + nointerp, 0));
     last_tableipol = tableipol;
     last_tableid = tableid;
@@ -194,7 +197,7 @@ void WavetableOscillator::convolute(int voice, bool FM, bool stereo)
         if (oscdata->wt.flags & wtf_is_sample)
         {
             tableid++;
-            if (tableid > oscdata->wt.n_tables - 3 + nointerp)
+            if (tableid > oscdata->wt.n_tables - 4)
             {
                 if (sampleloop < 7)
                     sampleloop--;
@@ -205,11 +208,14 @@ void WavetableOscillator::convolute(int voice, bool FM, bool stereo)
                 }
                 else
                 {
-                    tableid = oscdata->wt.n_tables - 2 + nointerp;
+                    tableid = oscdata->wt.n_tables - 1;
                     oscstate[voice] = 100000000000.f; // rather large number
                     return;
                 }
             }
+
+            tableipol = tableid;
+            last_tableipol = tableid;
         }
 
         int ts = oscdata->wt.size;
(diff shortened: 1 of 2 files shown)


src/common/dsp/oscillators/WavetableOscillator.cpp
```diff
@@ -438,29 +445,9 @@ void WavetableOscillator::process_block(float pitch0, float drift, bool stereo,
         float shape = l_shape.v;
         float intpart;
         shape *= ((float)oscdata->wt.n_tables - 1.f + nointerp) * 0.99999f;
-        tableipol = modff(shape, &intpart);
+        tableipol = shape;
+        modff(shape, &intpart);
         tableid = limit_range((int)intpart, 0, (int)oscdata->wt.n_tables - 2 + nointerp);
-
-        if (tableid > last_tableid)
```
baconpaul, line 444: Why was this code removed?

src/common/dsp/oscillators/WavetableOscillator.cpp
```diff
@@ -62,10 +62,12 @@ void WavetableOscillator::init(float pitch, bool is_display, bool nonzero_init_d
 
     n_unison = limit_range(oscdata->p[wt_unison_voices].val.i, 1, MAX_UNISON);
 
+    isOneShot = 0;
```
baconpaul, line 65: Since this is a bool you can use the constant true and false for better readability here

src/common/dsp/oscillators/WavetableOscillator.cpp
```diff
@@ -62,10 +62,12 @@ void WavetableOscillator::init(float pitch, bool is_display, bool nonzero_init_d
...
     if (oscdata->wt.flags & wtf_is_sample)
     {
         sampleloop = n_unison;
         n_unison = 1;
+        isOneShot = 1;
```
baconpaul, line 70: And here

baconpaul:
So that test failure looks real. Also added a couple of comments. 

Once those are resolved we should add a 1.4 issue to put this behind a flag or add a changelog entry

src/common/dsp/oscillators/WavetableOscillator.cpp
blancoberg: isOneShot is actually not used anymore. I added it because it was used as a check in convolute(), but I realised later that it wasn't needed. I forgot to remove it. 
Should I remove it or just change it to a bool?

src/common/dsp/oscillators/WavetableOscillator.cpp
blancoberg: This is the logic for the old morph interpolation. It has to be removed for the new one to work.

mkruselj: In that case please comment it out instead of removing it, since this code will have to remain here in order for the option switch to work.

blancoberg: > In that case please comment it out instead of removing it, since this code will have to remain here in order for the option switch to work.

ok, but you you told me the opposite about my last pr that I should just remove instead of commenting, since there is a git history :)

src/common/dsp/oscillators/WavetableOscillator.cpp
mkruselj: If it is completely unused it can be removed.

baconpaul:
Ok will have to look into that test

you don’t have a Linux box right? If not I can look after the weekend but only have Mac with me right now

blancoberg:
> Ok will have to look into that test
> 
> you don’t have a Linux box right? If not I can look after the weekend but only have Mac with me right now

Only pc/windows here :/

baconpaul:
Ok!
We can get to this early next eeek when I have Linux booted then (or maybe I’ll try it under Asan on my Mac if I have a chance)

wonder what it could be

blancoberg:
> Ok! We can get to this early next eeek when I have Linux booted then (or maybe I’ll try it under Asan on my Mac if I have a chance)
> 
> wonder what it could be

So I just installed linux/ubuntu on virtualbox and it builds without any issues. All audio in virtualbox is glitchy though, so I can't really do any sound tests

baconpaul:
Does the test pass when you run it? That’s really what we need to figure out

blancoberg:
How do i do that?

baconpaul:
Sorry 

built the target surge-testrunner 
Then run the executable that results from the base of the source directory

blancoberg:
===============================================================================
All tests passed (5862500 assertions in 109 test cases)

But honestly I don't know what I'm doing so you should probably do a test yourself to be sure :)

mkruselj:
And it worked!

baconpaul:
That’s frustrating - I should have tried that a week ago :( 

sorry about that

Let me merge this then 

evil - any thoughts on the menu for the ifs?

mkruselj:
@baconpaul Similar to Sine osc, call the option Legacy mode, disabled by default for new patches, enabled for old patches.

baconpaul:
> @baconpaul Similar to Sine osc, call the option Legacy mode, disabled by default for new patches, enabled for old patches.

And stick it as deform on which param? Morph probably right?

mkruselj:
Yup!

Issue #7784, opened 2024-09-12
Add wavetable fixes deform menu thingy


baconpaul:
Now that @blancoberg has fixed up the wavetable oscillator I need to do the patch version / deform / right mouse dance to make old patches work the old way. I'll do this next week just wanted to drop in a reminder issue.

blancoberg:
Is this the switch implementation?

baconpaul:
Yes just a reminder to myself to do it

mkruselj:
Don't worry, Paul's got it. If he's writing the if he can just as easily copy the required diff to make it work :)

baconpaul:
Yeah I'll add a switch and retrieve it at the top of process and then you can indeed just plumb it through helpful. thanks

baconpaul:
OK guys sorry i've gotten a bit unexpectedly super busy

I found some time tonight though to add the option and its in PR #7785

That sets up and version the option properly. A few things

1. It doesn't do anything with it other than extract it in the WT oscillator at the top of process block. Theres an if 0 knocked out process block print to check it
2. I don't know what we want as the names for the various states so I chose long things

I'm happy to merge and then let you proceeed or also happy for one of you to pick it up from there and use it as a starting point. I'll be up early in the AM then traveling / offline a lot of the day though so unless I hear otherwise I'll leave it hanging out ask an open PR for you to start with or merge at your leisure.

mkruselj:
It should just be an on off toggle to enable or disable all fixes at once. Or rather, it should work in reverse, call the option Legacy Mode, if it's enabled use everything how it was, if it's disabled use the fixes.

baconpaul:
Yeah that’s just a menu construction change. The rest of the code would be the same. Still suggest we use named values in Case there’s 1.5 changes but one menu with a bool is an easy twist of this pr

baconpaul:
Hold on i have 10 minutes. I'll change it to a binary menu display and then merge

baconpaul:
ok merged the binary option setup 

over to you! (or me but then next week :) )

blancoberg:
@baconpaul Would it be possible to use this switch as a selector for multiple types of interpolation modes? 
So instead of having a on/off switch, we could have a menu item named "interpolation mode" with a submenu containing the modes. ( which today would be legacy and continuous ).

Because I know atleast one other type of interpolation mode that would be a useful complement to the existing one

mkruselj:
[...]

Issue #7696, opened 2024-07-09
Wavetable "morph" smoothing lag


blancoberg:
**Bug Description:**
The wavetable morph parameter has a ~15ms smoothing lag.
Compare to other parameters like pitch, the morph parameter lerps to the new value instead of setting in somewhat directly. 

**Surge XT Version**
* Version: 1.3.main.f4c97fc
* Plugin Type: All
* Bitness: 64-bit

**Reproduction Steps:**
* Select the wavetable oscillator
* Switch to Mono ST mode 
*  Connect the note modulator to the "Morph" parameter at 100%. 
* Play legato notes with c4 and c5 in order to switch between the first and last frame of the wavetable. It should be audible that the wavetable frame position does not switch instantly. It drags behind the note change. 
* If you can't hear it, try to disable keytracking and just listen to the glide between the frames. 

Removing this line in "WaveTableOscillator.cpp":
``` 
float shape = l_shape.v;
``` 
And replacing it with this:
``` 
float shape = limit_range(localcopy[id_shape].f, 0.f, 1.f); 
``` 
Fixes the issue. ( This was suggested by @baconpaul  for testing )

**Expected Behavior:**
The morph parameter should change at the same speed as any other parameter, with as little lag as possible. 

**Computer Information (please complete the following!):**
 - OS: Windows 10
 - Host: Reaper
 - Version: 

**Additional Information:**

blancoberg:
this test is a wavetable that plays a sine wave at 2x speed until it arrives to the last frame where it runs at 1x with any interpolation in between the frames.


How to use the test preset:
When playing a legato note between C4 and C5, it should stay on C4. The note should not have time to switch before arriving to the last frame. 

Demonstration:
https://youtu.be/FxMLi7NjKuc

[Morph lag test.zip](https://github.com/user-attachments/files/16158362/Morph.lag.test.zip)

blancoberg:
I don't know if this issue is related, but I suspect it is. 

Other than having a smoothing lag, there is an inconsistency when modulating the "morph" parameter. If you for example modulate the morph parameter with the modwheel, Its expected that the morph value should start on the value that the modwheel has on note start. 

However, sometimes it seems like it take a whole block for it to set the value, causing a glitch because its essentially scrubbing through the entire wavetable. 

expected result:
![image](https://github.com/user-attachments/assets/53a789d5-ac0e-4171-bc90-bb72eb4fe737)

[...]

PR #7787, opened 2024-09-17
morph switch

blancoberg:
Implemented switch for handling legacy vs new interpolation

I decided to move the actual calculations of the interpolations outside of convolute, to make it more clear if we for some reason decide to add more interpolation modes. 

I then have a function pointer that is set on each block that points to the active interpolation function. I though that would be more optimal than having several if-statements inside of convolute(). 

I have also included the morph interpolation that I talked about yesterday. It is only used in cases where the morph value has changed more than 5 frames in a block. It works pretty well to reduce noise in those cases. 
A nice side effect is that it also reduces the noise when changing value with GUI, as those stepped changes will now morph instead of scrub. 

Addresses #7784 #7777 #7696


baconpaul:
Oh that function pointer solution is very clever. Nice.

Let me merge this and then we can start asking nightly testers to look out for goobits.

PR #7789, opened 2024-09-18
wavetable morph adjust

blancoberg:
* Removed the "morph" interpolation method as you could clearly hear when it was switching over on higher frequency modulations. Now it only uses the new continuous mode.

* Changed the way the new continuous mode reacts to the "extend range" flip. Before it would just remove the interpolation between each in between frame in a block, which made almost no difference in the overall sound. So now it instead removes the inbetween frames entirely, basically switching directly from the "last frame" to "current frame" which is probably closer to how the legacy method works. 

* Fixed a minor mistake in the switch implementation which affected the attack and the waveform display.

### DIFF
--- a/src/common/dsp/oscillators/WavetableOscillator.cpp
+++ b/src/common/dsp/oscillators/WavetableOscillator.cpp
@@ -99,7 +99,7 @@ void WavetableOscillator::init(float pitch, bool is_display, bool nonzero_init_d
     }
 
     shape *= ((float)oscdata->wt.n_tables - 1.f + nointerp) * 0.99999f;
-    tableipol = deformType == XT_134_EARLIER ? shape : modff(shape, &intpart);
+    tableipol = deformType == XT_134_EARLIER ? modff(shape, &intpart) : shape;
     tableid = limit_range((int)intpart, 0, std::max((int)oscdata->wt.n_tables - 2 + nointerp, 0));
     last_tableipol = tableipol;
     last_tableid = tableid;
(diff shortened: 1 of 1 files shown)


mkruselj:
I wonder if we can make the new quantized mode (continuous mode disabled) work like the PPG did? Meaning, it would only switch to a new wavetable frame after the current single cycle has finished, instead of immediately.

blancoberg:
> I wonder if we can make the new quantized mode (continuous mode disabled) work like the PPG did? Meaning, it would only switch to a new wavetable frame after the current single cycle has finished, instead of immediately. This would be a very nice improvement IMO.

it could work, but you would have to store a separate morph value for each voice in that case, otherwise they would all switch simultaneously on different phases. I guess the reason for switching at the ends is for reducing noise right? it does require all the frames to start and end at the same value otherwise it wouldn't work that great

blancoberg:
this would be super useful for me personally IF we could also have a mode for the pitch slider where that is also only allowed to change when a cycle is finished.

With those two, it would be possible to build super smooth legato transitions. Easier said that done though, as I'm pretty sure switching pitch in the middle of a block will screw everything up

mkruselj:
> I guess the reason for switching at the ends is for reducing noise right?

Yes, because PPG did not do any sort of interpolation between frames.

baconpaul:
folks i'm a bit behind. where are we on this. Do we want to merge it or are there outstanding changes?
Thanks!

baconpaul:
(Just want to make sure you aren't waiting on me! If you are waiting on other things that's fine!)

blancoberg:
I think you can merge. I'm not really sure about that "wait until cycle is finished" bit, as that will defeat the purpose of all the other changes I'v made to make the timing better between pitch & morph.

either way its a big fix so that will have to be in a separate pr

PR #7799, opened 2024-09-25
new 3D wavetable display

blancoberg:
The 3d wavetable display has been rebuilt.
All params now affect the entire wavetable graphics, which is pretty neat

I also "fixed" a bug where the display wasn't updating when loading external wavetables. Well it updated on the first file, but stopped after that because there is only a check for "current_id", which I presume does not change when loading external wts. 
It is a string comparison on "paint", which is not optimal, but maybe it could work until there is some kind of event dispatch implemented there instead.

### DIFF
--- a/src/surge-xt/gui/widgets/OscillatorWaveformDisplay.cpp
+++ b/src/surge-xt/gui/widgets/OscillatorWaveformDisplay.cpp
@@ -1295,6 +1295,8 @@ struct WaveTable3DEditor : public juce::Component,
     float samplesCached[128][128];
     float morphValue = 0.0;
 
+    int currentBreakingPoint = -1;
+
     union ParamCached
     {
         int i;
@@ -1302,7 +1304,7 @@ struct WaveTable3DEditor : public juce::Component,
     };
 
     ParamCached paramCached[7];
-    float zoomFactor = 1.0;
+    float zoomFactor = -1.0;
     bool hasResized = false;
 
     /* Values exported from
@@ -1332,7 +1334,6 @@ struct WaveTable3DEditor : public juce::Component,
         : parent(pD), storage(s), oscdata(osc), sge(ed)
     {
         clearSampleCache();
-        resized();
     }
 
     void cacheParams()
@@ -1384,12 +1385,12 @@ struct WaveTable3DEditor : public juce::Component,
             w = getWidth();
             h = getHeight();
 
-            wf = (float)w * 2.f;
-            hf = (float)h * 2.f;
+            wf = (float)w * zoomFactor * 2;
+            hf = (float)h * zoomFactor * 2;
 
             backingImage = nullptr;
-            backingImage = std::make_unique<juce::Image>(
-                juce::Image::PixelFormat::ARGB, (int)wf * zoomFactor, (int)(hf)*zoomFactor, true);
+            backingImage = std::make_unique<juce::Image>(juce::Image::PixelFormat::ARGB, (int)wf,
+                                                         (int)(hf), true);
         }
     }
 
@@ -1527,8 +1528,8 @@ struct WaveTable3DEditor : public juce::Component,
 
         osc->processSamplesForDisplay(samples, rendered_samples, processForReal);
 
-        float xScaled = wf * scale * zoomFactor;
-        float YScaled = hf * scale * zoomFactor;
+        float xScaled = wf * scale;
+        float YScaled = hf * scale;
 
         float skewCalc = skew;
 
@@ -1982,6 +1983,8 @@ void OscillatorWaveformDisplay::showCustomEditor()
         auto b = getLocalBounds().withTrimmedBottom(wtbheight);
         customEditor->setBounds(b);
         addAndMakeVisible(*customEditor);
+        customEditor->resized(); // Resize needs to happen after it has been added to the scene.
+                                 // Otherwise bounds will return 0
         repaint();
 
         customEditorAccOverlay->setTitle("Close Custom Editor");
--- a/src/surge-xt/gui/widgets/OscillatorWaveformDisplay.cpp
+++ b/src/surge-xt/gui/widgets/OscillatorWaveformDisplay.cpp
@@ -317,9 +317,13 @@ void OscillatorWaveformDisplay::paint(juce::Graphics &g)
     {
         // It's a bit unsatisfactory to put this here but we don't really get notified
         // once the wavetable change is done other than through repaint
-        if (oscdata->wt.current_id != lastWavetableId ||
-            oscdata->wt.current_filename != lastWavetableFilename) // more unsatisfactory!)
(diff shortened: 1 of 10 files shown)


blancoberg:
To my defence, I was getting no warnings when compiling. even the test passed. 
This should be solved by adding override after the declarations in wavetableosc and windowosc right?

mvf:
> This should be solved by adding override after the declarations in wavetableosc and windowosc right?

Yes, and preferably removing the `virtual` from the declarations that `override`.

mkruselj:
String comparison on paint is not acceptable IMO.

The way to do this is to call repaint() on the wavetable display component after the WT loading code IMO.

blancoberg:
> String comparison on paint is not acceptable IMO.
> 
> The way to do this is to call repaint() on the wavetable display component after the WT loading code IMO.

The problem is that there is no way of knowing when a new wavetable is loaded, as there are no events for that. 
Thats also why there already are similar checks for new wavetable ids in paint.

baconpaul:
> > String comparison on paint is not acceptable IMO.
> > The way to do this is to call repaint() on the wavetable display component after the WT loading code IMO.
> 
> The problem is that there is no way of knowing when a new wavetable is loaded, as there are no events for that. Thats also why there already are similar checks for new wavetable ids in paint.

Yeah I was about to say exactly that. the comment i wrote above the check was

```
        // It's a bit unsatisfactory to put this here but we don't really get notified
         // once the wavetable change is done other than through repaint
```

your addition fixes the case where you have drag and drop wavetables which are not in the library and i think its fine.

src/surge-xt/gui/widgets/OscillatorWaveformDisplay.cpp
```diff
@@ -1267,260 +1268,276 @@ struct WaveTable3DEditor : public juce::Component,
...
+        /*
+            Draw the base layer
+        */
+
+        if (paramsHaveChanged() || !backingImage)
         {
+
             backingImage =
-                std::make_unique<juce::Image>(juce::Image::PixelFormat::ARGB, wxf, hxf, true);
+                std::make_unique<juce::Image>(juce::Image::PixelFormat::ARGB, w * 2, h * 2, true);
```
baconpaul, line 1417: This is more the line that gives me concern. Why do we have to recreate the backing image on every param change? Can't we simply clear it if it exists? This is quite a bit of allocation.

Also w and h are virtual pixel sizes. Is the *2 for retina? The prior code was path based so did fine at 400% on a retina display. Will this now be pixelated in that case?

baconpaul:
> So correct me if I'm wrong, but couldn't we do this. There's this code in SurgeGUIEdtior:
> 
> image
> 
> So what if we add a method to OscillatorWaveformDisplay that is something like `void OWD::forceWTRepaint()`, which would simply set an internal bool to true so that it forces a repaint in the next idle cycle (then sets the bool back to false) regardless of lastWavetableId not changing, then call `repaint()`?
> 
> Of course, this would only need to be called if we're loading an external WT.

I think the problem was "not all paths set refresh display properly" but i dont remember. I just remember writing that comment when I couldn't get it to work!

src/surge-xt/gui/widgets/OscillatorWaveformDisplay.cpp
blancoberg: ah you are absolutely right. That was just how the code was before. It got nulled and recreated on resize, so I just kept it that way as I thought there was a good reason for that. 
I increased the resolution because it was pixelated on my computer atleast, but I can try and reduce it back to see if that is enough. I doubt it though.

blancoberg: image
difference between 1x vs 2x

blancoberg: is that zoom factor accessible? I could just read that in the resize function och recreate the image only when its changed

mkruselj:
Another possible option would be this, no?

mkruselj:
OK here's my final diff, totally works I just tried it. No string comparison in paint().

[diff.txt](https://github.com/user-attachments/files/17133454/diff.txt)

src/surge-xt/gui/widgets/OscillatorWaveformDisplay.cpp
baconpaul: Yeah my concern wasn't that 2x was too high. It was that 2x was to low! The w and h in this case are in unzoomed pixels 

and yes getZoomFactor * 2 will be a good answer

baconpaul: oh the units on getzoomfactor may be 100, 125, 150 not 1.0, 1.25, 1.50 just so you know. I never remember until I either get a tiny UI or an out of memory. But just a note.

mkruselj:
> have you tried to load several external wavetables in a row? cause that was the problem before

Yes, it works perfectly!

baconpaul:
> Another possible option would be this, no?
> 
> image

This one would rebuild on every repaint for a dragged wavetable. That's not the way. (I didn't read your final diff tho)

mkruselj:
Yep I think my diff nails it, way better than the earlier proposal.

baconpaul:
> Yep I think my diff nails it, way better than the earlier proposal.

back at my laptop and yeah it looks like it

So what do we want to merge? I think merging your diff with this PR and then cleaning up the alloc and resolution and go?

awesome stuff. also its great to have progress on surge still plowing forward with such cool stuff while i'm off in short circuit land. makes me happy!

blancoberg:
Yup I will add  @mkruselj's diff and fix the resolution/resize/image instantiation.

I do have some other fixes I want to do, but I think I'll do them in another pr. Stuff like horizontal skew still isnt implemented, so it doesnt affect the display yet. Still working on how to decode that formula

blancoberg:
so I don't think resized() is ever called inside of wavetable3ddisplay. only once at initiation.  This is also true for the alias editor. 
this is a comment from juce::Component : 
```
This is called when the position relative to its parent changes, not when
its absolute position on the screen changes (so it won't be called for
all child components when a parent component is moved).
```

I think we need to add a call from elsewhere

baconpaul:
The trick is the component is resized when we change zoom. Changing zoom is just changing the rendering transform and we use vectors everywhere except for here 

there’s a set zoom factor call in the editor class which can flag onto the display and cause a rebuild. That’s probably a reasonable approach

blancoberg:
wait with pushing this. There is a bug with the resizing that needs to be fixed.

blancoberg:
the resize bug is resolved if that wasn't clear

baconpaul:
Great let’s merge it! 

I’ll also try and move us to juce 8 this weekend and will test this when I do 

awesome srjff

PR #7852, opened 2024-11-17
Remove null pointer checks when the pointer had already been used. Th…

Dave-Lowndes:
Remove null pointer checks when the pointer had already been used. The code would have caused a crash before the null pointer test if it'd ever occur, so presumably it doesn't.

### DIFF
--- a/src/common/PatchDB.cpp
+++ b/src/common/PatchDB.cpp
@@ -839,7 +839,6 @@ CREATE TABLE IF NOT EXISTS Favorites (
         std::ostringstream searchName;
         searchName << p.name << " ";
 
-        if (storage)
         {
             auto pTmp = p.path.parent_path();
             std::vector<fs::path> parentFiles;
--- a/src/common/PatchDB.cpp
+++ b/src/common/PatchDB.cpp
@@ -803,7 +803,10 @@ CREATE TABLE IF NOT EXISTS Favorites (
         }
         catch (const SQL::Exception &e)
         {
-            storage->reportError(e.what(), "PatchDB - Load Check");
+            if (storage)
+            {
+                storage->reportError(e.what(), "PatchDB - Load Check");
+            }
             return;
         }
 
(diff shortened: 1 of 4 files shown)


baconpaul:
I want to peek at each of these in a touch more detail I merge on this one.

thanks for all of these!

if you would like I would love to see what you find when you run across the shortcircuit-xt codebase we have here too!

Dave-Lowndes:
@baconpaul I've had a brief look at the shortcircuit-xt code and have added comments/issues to a couple. Many of the reported issues are common with this codebase, so it's currently difficult to see the wood for the trees.

baconpaul:
yeah i wonder if the tool integrates with clion yet. 

I'd be happy to look at an html doc or whatever though to get an idea of the sort of things it is showing!

I merged a few other of your changes already this morning by the wya.

Dave-Lowndes:
Yes, it appears that you can use it from CLion https://pvs-studio.com/en/docs/manual/0052/
How can I send you a file output, I'm not familiar with what scxt may be!

mkruselj:
SCXT is just a shorthand for the other repo, Shortcircuit XT.

src/common/PatchDB.cpp
```diff
@@ -839,7 +839,6 @@ CREATE TABLE IF NOT EXISTS Favorites (
         std::ostringstream searchName;
         searchName << p.name << " ";
 
-        if (storage)
```
baconpaul, line 842: The static analyzer is sort of correct here, that storage is used above, but it is only used in an exception path. The correct fix I think is not to remove this guard, but to guard the report errors above. I'll make that change on the branch.

Dave-Lowndes: @baconpaul IIRC there are many places where it's used unguarded and I couldn't think why it should only not be null only in exception paths - which was why I assumed it probably never is null in practice!

baconpaul: In this function it's only used prior in the exception path. so i flipped it to check the exception paths

Those exceptions are pretty rare so I think if storage was null those cases would crash.

I sort of wish there was an easy non-nullable pointer type. I guess that's a & but they are clumsy to mutate.

Dave-Lowndes: The constructor on L413 would crash were it ever passed a nullptr for storage.
Reworking using a ref for storage seems to build fine for me.

baconpaul: Yeah i think you are correct.

Since synth : storage : ui are all 1:1 in surge right now, we really should have set everything up as a & or const&. That's more what we are doing in short circuit. Many of the 2004 bones are still on display in the raw pointer idiom here.

Dave-Lowndes:
@baconpaul 
Can you tell me the best place to save a CSV dump of the analysis warnings?
Should I create an issue and add it there, or something else?

baconpaul:
Oh yeah that would be perfect. Just attach a CSV dump to an issue as a zip file.

I keep meaning to get around to applying for an open source license for the tool but just been a really nutty winter.

PR #7862, opened 2024-11-21
Remove some unnecessary dynamic vectors/strings

Dave-Lowndes:
Eliminate dynamic vector & string storage where there's a simpler alternative.

Code tweak in prepareForEvaluation to use logical OR rather than bitwise and const the initial bool variable.

### DIFF
--- a/src/common/LuaSupport.cpp
+++ b/src/common/LuaSupport.cpp
@@ -181,38 +181,35 @@ bool Surge::LuaSupport::setSurgeFunctionEnvironment(lua_State *L)
     lua_setfield(L, eidx, sharedTableName);
 
     // add whitelisted functions and modules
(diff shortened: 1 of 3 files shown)


baconpaul:
So this change is in the first half correct but I'm not sure I like it. Naming that white list and table set  makes the code quite a bit clearer i thought.

The second change to a constexpr initializer list is, I think, a c++ 20 thing. At least it doesn't compile in ci with clang for mac.

mkruselj:
That whitelist should've been behind a `// clang-format off` (even if it's a valid clang-format that we use), for further clarity I think.

Dave-Lowndes:
How's this one looking now?

baconpaul:
Great!

PR #7890, opened 2024-12-01
Lua code editors indentation

blancoberg:
Some updates to the code editor. 

Added the following:
* Its now possible to indent lines even if they are empty.
* tab does not indent the entire line anymore, instead it adds a tab ( spaces) at the caret position
* String declarations auto closes. 
* Auto indentation for function, while, if statements.
* On enter/return the new line will keep the indentation from the previous line
* Some preparation for a possible search function


baconpaul:
Oh @mkruselj we should not have merged this. As well as adding features it removes all my fixes from yesterday afternoon 

let me clean it up

Issue #4539, opened 2021-05-11
Lua for wavetables


baconpaul:
So if you give me

s (power of two samples)
n (number of tables)
f(n) -> float[s] (function to generate table n)

I can easily make a wavetable

and that sounds like a pretty easy thing to do with our lua evaluator

slightly different ui and storage strategy that formula modulator but most of what we have will carry over (which is why i named the gui classes LuaBlah not FormulaBlah).

mkruselj:
For inspiration in what sort of functionality to add:

[Hive Wavetables.zip](https://github.com/surge-synthesizer/surge/files/6480916/Hive.Wavetables.zip)

SpotlightKid:
The Waldorf Microwave (I/II) also had an interesting way of generating Wavetables on the fly. You had to write a Wavtable program in a assembly-like language and then convert that into a Sysex-File, which you then sent to the synth.

I wrote a Python re-implementation of that converter ages ago. The actual calculation of the wavetable program happens in the hardware, though:

https://github.com/SpotlightKid/upaw2/blob/master/upaw.description

baconpaul:
Gonna bounce this out of 1.2 again and deactivate it in the nightly

nuoun:
So the wavetable script editor has been re-enabled in the Nightly builds for a while and I've been slowly working on it, so far #7674, #7680, #7687, #7755, #7764 and #7767 have been merged which refactor the code, make UI changes, add window tearout, implement the prelude for WTSE and some more and things are slowly coming together. 
I had a much longer writeup ready which I could still post if there is interest in some of the concepts and ideas and possible side projects for the future but I'll try to keep this short and limit this post to priority here which is to try and figure out what is needed to get this component ready for release.

So I will add a (likely incomplete) list of issues that still need work in the next post, but a lot of this is to just test and periodically ask if stuff makes sense or what needs improving and I would of course much appreciate feedback on this. Paul brought up two things on a previous PR which I'll mention here:

1. Window tear out, this currently works different for the WTSE, in formula the editor switches to the next LFO when a different LFO that has formula active gets selected and the WTSE does not do this even when another WT oscillator gets selected. I don't think this is a big issue, maybe a QoL thing at best and I'm not sure how this is handled in the GUI code right now. Other than that, the torn out window just stays open when the VCO gets changed and I think that is in line with how other torn out windows behave.

[...]

PR #7892, opened 2024-12-02
Several WT script editor tweaks

mkruselj:
Trim trailing whitespace on script apply (also in Formula editor)
No vertical axis, grid, nor frame gap in filmstrip
More visible frame number in filmstrip (also aligned top right)
Checkerboard background of odd frames in filmstrip
Add frame number to single view (aligned top right)
Add menu entry to open WT script editor when right-clicking the oscillator display

Addresses one item in #4539

### DIFF
--- a/src/surge-xt/gui/overlays/LuaEditors.cpp
+++ b/src/surge-xt/gui/overlays/LuaEditors.cpp
@@ -35,7 +35,6 @@
 #include "widgets/MenuCustomComponents.h"
 #include <fmt/core.h>
 #include "widgets/OscillatorWaveformDisplay.h"
-#include <regex>
 
 namespace Surge
 {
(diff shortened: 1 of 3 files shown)


blancoberg:
> Oh irritating. I added that me u item yesterday but it seems @blancoberg removed it in [ab8acad](https://github.com/surge-synthesizer/surge/commit/ab8acad00e0a1179ba706f8a3189b031926d2afa) and that merge also removed some of my other lua changes to the stack debugger
> 
> did I merge that without review? I’m sorry if so we should not have merged the changes to the files other than the editor

Oh, sorry about that Paul. I have no idea how that happened other than I suck at git

baconpaul:
It’s no problem I can put it back.

justnits kinda funny that evil merged the change to remove it that menu item then coded the change to add that menu item. That’s how i noticed.   But that pop matters so I need to put it back

PR #7896, opened 2024-12-03
Cache WTSE Evaluation

baconpaul:
to do minimal calculation


baconpaul:
it does a few things

Make sure init and generate are called the smallest possible time
makes getFrame basically free the second time due to cache
invalidates only when resoultion etc change
changes the ui so dragging frame count or resolution just activtes apply button and doesn't recalc

Issue #7895, opened 2024-12-03
lfo display optimization


blancoberg:
**Bug Description:**
The lfo display is in need of some optimizations as all interactions with it is very laggy. It also causes lag to the rest of the UI. One example being the overlays, which starts to lag when you drag them over the bounds of the lfo display. 

A first step to remedy this could be to add image caching to the waveform drawing. That way the waveform only needs to be drawn once each time a parameter is changed. This should solve the lag with the overlays atleast. 

**Reproduction Steps:**
1. Open formula code editor
2. drag the window around the top of the ui and there is no lag.
3. drag the window over the bounds of the lfo display and will lag significantly.

or 
1. Choose formula
2. Drag the sliders so the waveform updates
3. lag

**Expected Behavior:**
No lag :)

nuoun:
So it's calling Formula's process() function thousands of times to update the display just by leaving the LFO area with the mouse even when nothing changed. Seems to me it can just not do that.

Can test this by running Surge from the terminal to see stdout with this Formula script:

```
function init(state)
    print("init")
    state.count = 0
    return state
end

function process(state)
    state.output = state.phase * 2 - 1
    print(state.count)
    state.count = state.count + 1

    return state
end
```

baconpaul:
@blancoberg does merging #7901 mean we can close this?

blancoberg:
@baconpaul I was thinking we can take a look at the actual drawing call for the waveform as well. I don't know if anything can be done there but it could atleast be worth a try

baconpaul:
yeah ok in the waveform case i did quite a bit of work on it when we ported to juce - mostly involving path downsampling and range adjustment. but always good to look!

blancoberg:
@baconpaul here is an idea; add a setBlockRate() to LFOModulationSource, and use that to lower the resolution of the lfo display.
right now it has to process over 8000 samples for a display that is a 10th of that width in pixels.  In theory we could lower that to 1000 by setting the blockrate to 256 and still have enough samples to draw each pixel on the screen. 

One downside to this approach is that this assumes the code in the formula takes blockrate into account, otherwise the output will be off. But at the same time, that could be a positive thing, as it will tell you if you have done something wrong.

baconpaul:
Is block size a variable or a constexpr? I worry about the compute cost of taking it out of the compiler optimization path in the audio thread. 

[...]

PR #7901, opened 2024-12-04
lfo display caching

blancoberg:
* caches the lfo display
* only redraws when parameters are changed, lfo switched or lua code is applied

addresses partially #7895

### DIFF
--- a/src/surge-xt/gui/widgets/LFOAndStepDisplay.cpp
+++ b/src/surge-xt/gui/widgets/LFOAndStepDisplay.cpp
@@ -452,33 +452,17 @@ bool LFOAndStepDisplay::paramsHasChanged()
 {
     bool hasChanged = false;
 
-    if (paramsFromLastDrawCall[lfodata->delay.param_id_in_scene].i != lfodata->delay.val.i)
-        hasChanged = true;
-    if (paramsFromLastDrawCall[lfodata->attack.param_id_in_scene].i != lfodata->attack.val.i)
-        hasChanged = true;
-    if (paramsFromLastDrawCall[lfodata->hold.param_id_in_scene].i != lfodata->hold.val.i)
-        hasChanged = true;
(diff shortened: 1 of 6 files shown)


src/surge-xt/gui/widgets/LFOAndStepDisplay.cpp
```diff
@@ -414,12 +417,112 @@ void LFOAndStepDisplay::paint(juce::Graphics &g)
...
     }
 
     paintTypeSelector(g);
 }
 
+bool LFOAndStepDisplay::paramsHasChanged()
+{
+    bool hasChanged = false;
+
+    if (paramsFromLastDrawCall[lfodata->delay.param_id_in_scene].i != lfodata->delay.val.i)
```
baconpaul, line 455: Happy to merge this, but the lfo data has params which are contiguous so you could also make this code

```
   auto *p = &lfodata->delay; // look in the definition of LFOStorage for which param is first
  while (p <= &lfodata->shape) // and last
   {
      if (paramsFromLastDrawCa[[p->paramid].i != p->val.i;
         hasChanged = true;
   }
   ```
   
   type thing. If you want to make this change now I can wait but I'm also happy to merge.

Thanks for doing this!

blancoberg: That is much cleaner.  I was looking for a way of doing that but Couldn't figure out how.  I'll fix it

blancoberg: image

Not sure how to interpret the order here. Does this mean that rate is first and release last? Or is there some other list that sets the order of these?

baconpaul: rate is first and release is last yes.

blancoberg:
nope those are not included. 
any hints on where I can catch those events?

PR #7911, opened 2024-12-08
code editor search

blancoberg:
Code editor search

* Ctrl + f to open
* Esc to close 
* Searches for texts, case insensitive
* Enter to get the next result, or shift + enter for the previous. 
* Selected text gets added to the search textbox


mkruselj:
Would be nice to have options for whole word and case sensitive search somehow, perhaps!

blancoberg:
> Would be nice to have options for whole word and case sensitive search somehow, perhaps!
That could be useful indeed. 
A switch button for case sensitivity, buttons for next/previous match and a close button.

PR #7922, opened 2024-12-14
Fix CodeEditor indent bug

blancoberg:
* fixed infinite indentation bug
* ALT + mousewheel scroll at 4x speed

### DIFF
--- a/src/surge-xt/gui/overlays/LuaEditors.cpp
+++ b/src/surge-xt/gui/overlays/LuaEditors.cpp
@@ -42,47 +42,177 @@ namespace Overlays
 {
 
 /*
-    ---------------------------------------
-    Search box
-    ---------------------------------------
+    TextfieldPopup
+    Basic class that can be used for creating other textfield popups like the search
 */
+TextfieldButton::TextfieldButton(juce::String &svg) : juce::Component()
+{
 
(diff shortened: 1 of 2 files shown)


src/surge-xt/gui/overlays/LuaEditors.cpp
```diff
@@ -791,15 +791,27 @@ void SurgeCodeEditorComponent::caretPositionMoved()
     }
 }
 
+void SurgeCodeEditorComponent::mouseWheelMove(const juce::MouseEvent &e,
+                                              const juce::MouseWheelDetails &wheel)
+{
+    auto modifiedMouseWheel = juce::MouseWheelDetails();
```
mkruselj, line 797: Unused local, can be removed.

src/surge-xt/gui/overlays/LuaEditors.cpp
```diff
@@ -791,15 +791,27 @@ void SurgeCodeEditorComponent::caretPositionMoved()
...
+    juce::MouseWheelDetails w(wheel);
+    if (e.mods.isAltDown())
```
mkruselj, line 799: After trying this on Windows, no need to hide this behind Alt. Just multiply it 4x altogether.

mkruselj (changes requested):

mkruselj (approved):

PR #7924, opened 2024-12-16
yoshih/submodule sst-jucegui

shih1:
Submodule https://github.com/surge-synthesizer/sst-jucegui into `surge`. 

Testing
- clean build and ran synth + fx, twiddled a few knobs, switched menus

### DIFF
--- a/src/common/CMakeLists.txt
+++ b/src/common/CMakeLists.txt
@@ -33,6 +33,7 @@ surge_add_lib_subdirectory(sst/sst-cpputils)
 surge_add_lib_subdirectory(sst/sst-plugininfra)
 surge_add_lib_subdirectory(sst/sst-filters)
 surge_add_lib_subdirectory(sst/sst-waveshapers)
+surge_add_lib_subdirectory(sst/sst-jucegui)
 target_compile_definitions(sst-waveshapers INTERFACE SURGE_XT_1X_WST=1)
 surge_add_lib_subdirectory(sst/sst-effects)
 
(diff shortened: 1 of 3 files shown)


src/surge-fx/CMakeLists.txt
```diff
@@ -67,7 +67,7 @@ target_sources(${PROJECT_NAME} PRIVATE
 target_link_libraries(${PROJECT_NAME} PRIVATE
   surge::surge-common
   surge-fx-binary
-  surge-juce
```
shih1, line 70: i'm not entirely sure this is ok, it works but I don't fully understand the implications of this change. 

CMake is not my strongest language.

baconpaul:
My guess is the failure is unrelated to this and is related to Ubuntu latest moving to 24 nd me needing to update the prepare for juce action 

but the sst jucegui should have an include on use juce I bet.

baconpaul:
yeah definitely the case that the failure isn't yours

but the inclusion in lib.cmake should be wrapped in a  `if (NOT SURGE_SKIP_JUCE_FOR_RACK)` block

i'll try and get the pipelines fixed then will re-run this if i do over in #7925

src/common/CMakeLists.txt
```diff
@@ -33,6 +33,7 @@ surge_add_lib_subdirectory(sst/sst-cpputils)
 surge_add_lib_subdirectory(sst/sst-plugininfra)
 surge_add_lib_subdirectory(sst/sst-filters)
 surge_add_lib_subdirectory(sst/sst-waveshapers)
+surge_add_lib_subdirectory(sst/sst-jucegui)
```
baconpaul, line 36: This inclusion requires a `if (NOT SURGE_SKIP_JUCE_FOR_RACK)` / `endif()` block surrounding it. Thanks!

baconpaul (changes requested):

src/common/CMakeLists.txt
shih1: i moved it down into the existing block

PR #7928, opened 2024-12-17
surge-fx: replace AccSlider with sst/jucegui/Knob.h

shih1:
This PR switches from `juce::Slider` with `sst-jucegui` knobs. 

it is a follow up on https://github.com/surge-synthesizer/surge/pull/7924

Extra changes
- update sst-jucegui

- [x] **Testing**
  - [x] Change modify knobs and switch FX
  - [x] tested in reaper, live
  - [x] tested clap, au, vst3
  - [x] test accessibility


baconpaul:
I converted to draft to avoid accidental merge. Have fun!

src/surge-fx/KnobSource.h
```diff
@@ -36,4 +40,298 @@ struct ConcreteCM : sst::jucegui::data::ContinuousModulatable
     float getModulationValuePM1() const override { return mv; }
     void setModulationValuePM1(const float &f) override { mv = f; }
     bool isModulationBipolar() const override { return isBipolar(); } // sure why not
+};
+
+struct StyleSheetBuiltInImp : public sst::jucegui::style::StyleSheet
```
baconpaul, line 45: Ooooh we shouldn't need all this code. You should be able to start with the Dark and customize colors from there.

baconpaul: https://github.com/surge-synthesizer/shortcircuit-xt/blob/main/src-ui/theme/ThemeApplier.cpp here's how SC does it

baconpaul (commented):
You shouldn't need to implement a stylesheet, just start with one of the built ins and customize

You also shouldn't need a style sheet just for the knobs. Instead give them a custom class if you need differentiated knobs, or just use the knob properties from the parent.

src/surge-fx/KnobSource.h
shih1: example makes sense to me. Will update tmrw.

src/surge-fx/SurgeFXEditor.cpp
```diff
@@ -119,8 +119,20 @@ struct Picker : public juce::Component
...
+      sst::jucegui::style::StyleConsumer(sst::jucegui::components::Knob::Styles::styleClass),
+      styleSheet(sst::jucegui::style::StyleSheet::getBuiltInStyleSheet(
+          sst::jucegui::style::StyleSheet::DARK)),
+      processor(p)
 {
+    sst::jucegui::style::StyleSheet::initializeStyleSheets([]() {});
+
+    setTransform(juce::AffineTransform().scaled(1.0));
+
+    // @ BP this compiles but causes a segfault on launc
```
shih1, line 132: @baconpaul this is what I think is supposed to happen but JUCE segfaults on `isShowing()`. I reverted to L153 instead. 

I think once we resolve this issue i can move on to updating the param values.

shih1: I dug into it more. Its segfaulting here https://github.com/surge-synthesizer/sst-jucegui/blob/main/src/sst/jucegui/style/StyleAndSettingsConsumer.cpp#L49

src/surge-fx/KnobSource.h
```diff
@@ -11,7 +11,8 @@
 
 struct ConcreteCM : sst::jucegui::data::ContinuousModulatable
 {
-    ConcreteCM() {};
+    ConcreteCM(SurgefxAudioProcessor &p, int i) : processor(p), id(i) {}
```
baconpaul: If you make the base class here just continuous as opposed to continuous modiularanle you will get a better result

src/surge-fx/SurgeFXEditor.cpp
baconpaul: Well that should be in the if but also should not be null 

did you make a non component the consumer?

shih1: I made the editor a consumer, which is a component. My limited understanding is that the component is null when setting the style which doesn't make sense to me either. 

If I made a non component the consumer, i think it should have failed to compile? 

In my std::cout testing, it failed to even enter the isShowing(). 
I thought it might have been an issue with https://github.com/juce-framework/JUCE/blob/master/modules/juce_gui_basics/components/juce_Component.cpp#L342 But testing showed that it faulted on the entry of `isShowing()` from `sst-jucegui`.

shih1:
Screenshot 2025-01-01 at 8 32 08 PM

src/surge-fx/SurgeFXEditor.cpp
```diff
@@ -353,7 +378,7 @@ void SurgefxAudioProcessorEditor::paramsChangedCallback()
         {
             if (i < n_fx_params)
             {
-                fxParamSliders[i].setValue(fv[i], juce::NotificationType::dontSendNotification);
+                sources.at(i)->setValueFromGUI(fv[i]);
```
baconpaul, line 381: Should this be setValueFromModel?

baconpaul (commented):
This all looks good.

I had one question about an API point

And the thing I would test also is: when you do external VST automation of a param in reaper, does the knob repaint get called properly?

But with both those things I am very happy to merge this change!

src/surge-fx/SurgeFXEditor.cpp
shih1: yes

shih1:
did my final pass through and testing. lgtm :ship:

shih1:
jk mighta found a bug or ableton being annoying

PR #7793, opened 2024-09-21
XCode 16 and LuaJit no dedup workaround

baconpaul:

PR #7929, opened 2024-12-17
Update luajitlib

nuoun:
- Updates LuaJIT submodule to head
- Adds amalgamated build for MSVC LuaJIT/LuaJIT#1289
- Remove Xcode no dedup workaround #7793
- Updates README.md


nuoun:
Hmm, don't think `static amalg` is doing what it is supposed to be doing (should be `amalg static` instead but that only writes a 30kb .lib file) so I'm just going to roll that back

baconpaul:
ready to merge this?

nuoun:
Yes sorry, go ahead and merge please. I saw something shiny but it does not seem to work :(

PR #7932, opened 2024-12-21
code editor: find & replace

blancoberg:
* Find & Replace ( ctrl + h) ( Enter : replace one and command + enter : replace all )
* Some refactoring of TextfieldPopup
* Resets UndoManager on lua editor init. ( removed the empty undo state that made it possible to undo to a blank doc)
* Keyboard shortcuts for popups now works inside the popups as well.

_update_
* Fixed a bug where it indents even if the caret is positioned before the keyword

_update 2_
* changed popup font/size to same as editor. 
* added find, replace and go to line to context menu

### DIFF
--- a/src/surge-xt/gui/overlays/LuaEditors.cpp
+++ b/src/surge-xt/gui/overlays/LuaEditors.cpp
@@ -45,9 +45,9 @@ namespace Overlays
     TextfieldPopup
     Base class that can be used for creating other textfield popups like the search
 */
-TextfieldButton::TextfieldButton(juce::String &svg, int r) : juce::Component()
+TextfieldButton::TextfieldButton(juce::String &svg, int rowToAddTo) : juce::Component()
 {
-    row = r;
+    row = rowToAddTo;
     xml = juce::XmlDocument::parse(svg);
 
     svgGraphics = juce::Drawable::createFromSVG(*xml);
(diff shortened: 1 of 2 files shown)


mkruselj:
Great stuff! 👍

baconpaul:
joel - is this ready to merge?

src/surge-xt/gui/overlays/LuaEditors.cpp
```diff
@@ -45,8 +45,9 @@ namespace Overlays
     TextfieldPopup
     Base class that can be used for creating other textfield popups like the search
 */
-TextfieldButton::TextfieldButton(juce::String &svg) : juce::Component()
+TextfieldButton::TextfieldButton(juce::String &svg, int r) : juce::Component()
```
mkruselj, line 48: I would name this second argument more nicely so that it it's clearer, rather than guesswork.

mkruselj:
Looks fine to me apart from the one comment I left. Also those inlined SVGs are kinda ugly, I would suggest moving them to `resources/classic-skin-svgs` (even if they are not going to be exposed to the skin engine necessarily) then adapting the code to load them explicitly.

blancoberg:
> joel - is this ready to merge?

I have one more commit where I've added find,replace and goto to the context menu. I'll also fix @mkruselj comment on the function parameter

PR #7949, opened 2025-01-03
Code Editor scrolling fix

blancoberg:
It turns out that scrollBy() sucks, as it scrolls way past its maximum limit. So went with another approach

### DIFF
--- a/src/surge-xt/gui/overlays/LuaEditors.cpp
+++ b/src/surge-xt/gui/overlays/LuaEditors.cpp
@@ -1118,20 +1118,28 @@ void SurgeCodeEditorComponent::mouseWheelMove(const juce::MouseEvent &e,
     juce::MouseWheelDetails w(wheel);
     w.deltaY *= 4;
 
+    enum
+    {
+        verticalScrollbar,
+        horizontalScrollbar
+    };
+
     // makes it possible to mouse wheel scroll and select text at the same time
     if (e.mods.isShiftDown())
     {
-        auto scrollbar = dynamic_cast<juce::ScrollBar *>(getChildren()[1]);
-        auto pos = scrollbar->getCurrentRange().getStart();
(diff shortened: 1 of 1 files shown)


src/surge-xt/gui/overlays/LuaEditors.cpp
```diff
@@ -1123,11 +1123,19 @@ void SurgeCodeEditorComponent::mouseWheelMove(const juce::MouseEvent &e,
...
+
+        auto width = scrollbar->getCurrentRangeSize();
+        auto maxScroll = std::max((double)0, scrollbar->getMaximumRangeLimit() - width);
+
+        scrollToColumn(std::min((double)maxScroll, pos - w.deltaY * 10));
     }
     else
     {
-        scrollBy(-w.deltaY * 10);
+        auto scrollbar = dynamic_cast<juce::ScrollBar *>(getChildren()[0]);
```
baconpaul, line 1134: Why the [0] vs [1] here?
should we check this dynamic cast for null in case the child list changes in the future?

blancoberg: its the scrollbars. There is no other way of retrieving them from juce:CodeEditorComponent since they are private. Atleast not what I could find. 
I guess we could add a check there in case it gets updated

blancoberg: 0 = horizontal scrollbar, 1 = vertical scrollbar

mkruselj: Maybe worth adding an enum for those magic numbers so that they get a name?

blancoberg: I've added an enum inside that function and added null checks

PR #7951, opened 2025-01-03
surge-fx: split ParameterPanel into its own component

shih1:
Surge-fx has a generic editor layout for each fx. This means that the `GraphicEQ` and the `Reverb` have the same UI. 

In preparation to create more specific UI's for each FX, this PR is a _mostly_ a code reorg but some other changes are being shipped with
* Remove `StyleConsumer` - realized that its actually not needed
* add `accessibleOrderWeakRefs` to the ParameterPanel and `addAndMakeVisibleRecordOrder()` back. This should have been maintained for accessibility. I do think these functions & dataStruct should belong in a separate class but... is outside the scope of this PR. 

The primary motivation is a simplification of `Editor::resized()` (in the spirit of keeping the original layout, its not exactly like this but the logic flow is as described) 
```
void SurgefxAudioProcessorEditor::resized()
{
    auto bounds = getLocalBounds();
[...]
    auto topArea = bounds.removeFromTop(topAreaHeight);
    auto bottomArea = bounds.removeFromBottom(bottomAreaHeight);

    picker->setBounds(topArea);
    defaultKnobLayout->setBounds(bounds); 
    fxNameLabel->setBounds(bottomArea);
}
```

This will make it easier to switch between `ParameterPanel` and `customLayout` later.


mkruselj:
@shih1 Looks like we're missing files to be clang-formatted over here.

shih1:
> @shih1 Looks like we're missing files to be clang-formatted over here.
i was missing a new line at the end of my new *.cpp file. my linter doesn't fix that but it should be resolved now

shih1:
gonna state here i want baconpaul to review this after vacation

mkruselj:
@baconpaul Are we good to merge this?

shih1:
cool beans i think i'm unblocked to start updating the UI on a few FX now

mkruselj:
At last!

PR #7961, opened 2025-01-14
Add option to apply MPE Timbre value as unipolar in the engine

### LINKED ISSUE #7960: MPE TImbre should be switchable between Bipolar and Unipolar
MPE Timbre is Bipolar: this is not appropriate for all devices (e.g the Expressive E Osmose has a 0 value for the CC until you enter the aftertouch region, at which point it proceeds from 0 to 127).

I'd expect to have a way to select unipolar or bipolar for that modulation source (with a pop up menu, similar to the random source).

You can select CC ranges to transmit, so the Osmose could be set to use values 64-128 in the aftertouch region. This works, but it also halves the resolution of that dimension.,

**Additional context:**
I understand that MPE Timbre is commonly also referred to as Y, with the centre of the control surface being the default playing area, but that's not the case for Osmose. I have no real context around Surge's source, but attempted to hack together a change and it did indeed add the UI element, but meant that the Mod matrix entires were incorrectly assigned to the new destinations. Similarly I didn't know what we have access to in `src/common/SurgeSynthesizer.cpp` so I just unconditionally change it to a unipolar value. I assume that someone more familiar with the codebase than me would be able to make this change relatively straightforwardly.

mkruselj:
Closes #7960

### DIFF
--- a/src/common/SurgeSynthesizer.cpp
+++ b/src/common/SurgeSynthesizer.cpp
@@ -3498,7 +3498,7 @@ bool SurgeSynthesizer::isBipolarModulation(modsources tms) const
     }
     if (tms == ms_keytrack || tms == ms_lowest_key || tms == ms_highest_key ||
         tms == ms_latest_key || tms == ms_pitchbend || tms == ms_random_bipolar ||
-        tms == ms_alternate_bipolar || (tms == ms_timbre && !mpeTimbreIsUnipolar))
+        tms == ms_alternate_bipolar || tms == ms_timbre)
         return true;
     else
         return false;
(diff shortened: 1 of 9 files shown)


src/common/SurgeSynthesizer.cpp
```diff
@@ -3496,7 +3498,7 @@ bool SurgeSynthesizer::isBipolarModulation(modsources tms) const
     }
     if (tms == ms_keytrack || tms == ms_lowest_key || tms == ms_highest_key ||
         tms == ms_latest_key || tms == ms_pitchbend || tms == ms_random_bipolar ||
-        tms == ms_alternate_bipolar || tms == ms_timbre)
+        tms == ms_alternate_bipolar || (tms == ms_timbre && !mpeTimbreIsUnipolar))
```
baconpaul, line 3501: So this is the thing I'm not sure about.

TIMBRE is a combination of what we get from the keyboard and the note expression. The note expression is bipolar. So the modulation would still be potentially bipolar. Also for a given patch on a different device, it would be. I sort of feel like we should display the modulator as bipolar still..

This is just a ui thing, but thoughts welcome

mkruselj: Hmm that's a valid point. Will remove.

PR #7975, opened 2025-02-13
Backwards-compatible option to use mathematically correct tilt noise.

mx:
Briefly tested with the Oscilloscope to make sure the tilt was right and the noise levels were the same between the two modes.

### DIFF
--- a/src/surge-xt/gui/SurgeGUIEditorValueCallbacks.cpp
+++ b/src/surge-xt/gui/SurgeGUIEditorValueCallbacks.cpp
@@ -2478,7 +2478,7 @@ int32_t SurgeGUIEditor::controlModifierClicked(Surge::GUI::IComponentTagValue *c
                                 .filterblock_configuration.val.i == fc_wide)
                         {
                             contextMenu.addSeparator();
-                            auto dt = p->deform_type;
+                            auto dt = p->deform_type & 0x1;
 
                             Surge::Widgets::MenuCenteredBoldLabel::addToMenuAsSectionHeader(
                                 contextMenu, "NOISE GENERATOR MODE");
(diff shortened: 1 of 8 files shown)


mx:
Only tested the max level, just checked with -12 dB and that one is wrong (it's -24 when set to -12); I'll have to fix that before this is ready to go.

src/surge-xt/gui/SurgeGUIEditorValueCallbacks.cpp
```diff
@@ -2478,27 +2478,49 @@ int32_t SurgeGUIEditor::controlModifierClicked(Surge::GUI::IComponentTagValue *c
...
                         }
+                        {
+                            contextMenu.addSeparator();
+                            auto dt = (p->deform_type & 0x2) >> 1;
 
+                            Surge::Widgets::MenuCenteredBoldLabel::addToMenuAsSectionHeader(
+                                contextMenu, "NOISE GENERATOR COLOR TYPE");
+
+                            contextMenu.addItem(
+                                "Legacy (Pre-XT 1.4)", true, dt == NoiseColorValue::LEGACY,
```
mkruselj, line 2509: In Sine oscillator we used this verbiage: `Legacy (<1.6.2)`. So over here it would probably make sense to be consistent and say `Legacy (<XT 1.4)`.

src/surge-xt/gui/SurgeGUIEditorValueCallbacks.cpp
```diff
@@ -2478,27 +2478,49 @@ int32_t SurgeGUIEditor::controlModifierClicked(Surge::GUI::IComponentTagValue *c
...
                                     synth->storage.getPatch().isDirty = true;
                                     frame->repaint();
                                 });
                         }
+                        {
+                            contextMenu.addSeparator();
+                            auto dt = (p->deform_type & 0x2) >> 1;
 
+                            Surge::Widgets::MenuCenteredBoldLabel::addToMenuAsSectionHeader(
+                                contextMenu, "NOISE GENERATOR COLOR TYPE");
```
mkruselj, line 2506: I would remove the word COLOR here.

src/surge-xt/gui/SurgeGUIEditorValueCallbacks.cpp
```diff
@@ -2478,27 +2478,49 @@ int32_t SurgeGUIEditor::controlModifierClicked(Surge::GUI::IComponentTagValue *c
...
+                            contextMenu.addItem(
+                                "Legacy (Pre-XT 1.4)", true, dt == NoiseColorValue::LEGACY,
+                                [this, p]() {
+                                    undoManager()->pushParameterChange(p->id, p, p->val);
+                                    update_deform_type_bit(p, NoiseColorValue::LEGACY, 1);
+                                    synth->storage.getPatch().isDirty = true;
+                                    frame->repaint();
+                                });
+                            contextMenu.addItem(
+                                "Tilt", true, dt == NoiseColorValue::TILT, [this, p]() {
```
mkruselj, line 2517: Actually "Tilt Filter" might be a bit clearer, I think.

Andreya-Autumn: "Slope" is what I called it in SC.

src/surge-xt/gui/SurgeGUIEditorValueCallbacks.cpp
mx: Done.

src/surge-xt/gui/SurgeGUIEditorValueCallbacks.cpp
mx: Done.

src/surge-xt/gui/SurgeGUIEditorValueCallbacks.cpp
mx: Done.

Issue #7604, opened 2024-04-10
The virtual keyboard causes the notes E and F to be triggered


susquetzal:
**Bug Description:**
When the virtual keyboard is enabled, sometimes the notes E and F play and remain triggered until I press stop in the DAW transport.

**Surge XT Version**
* Version: Surge XT 1.3.1.4b53827
* Plugin Type: VST3, CLAP, standalone
* Bitness: 64-bit

**Reproduction Steps:**
1. Open Surge XT with the virtual keyboard on
2. Click on the Surge window to make sure it's fully focused
3. Play any note with the computer keyboard
4. E and F will also be triggered, at the same octave as Surge is currently set to

**Expected Behavior:**
No unwanted notes should play.

**Screenshots:**
![Screenshot_20240410_021951](https://github.com/surge-synthesizer/surge/assets/138405054/843dbf62-c4ef-433c-a73f-c0a4d406d4fb)

**Computer Information (please complete the following!):**
 - OS: Arch Linux, KDE Plasma 6.0.3, X11
 - Host: Renoise 3.4.3, Reaper 7.14, standalone

mkruselj:
This is a Renoise thing, we've had a report on this already. Disable Renoise's virtual keyboard or Surge's, you shouldn't have both enabled at the same time.

baconpaul:
Do you have a non-standard keyboard map or some such?

I wonder if our extended keyboard handling is picking up a key code on arch with your setup.

mkruselj:
Def sounds like Arch related issue.

erroreyes:
I've been able to reproduce this in the period of 5 minutes of testing. It doesn't happen all the time, but often enough to see it as a problem.

In standalone with Virtual Keyboard on, I switch focus to some random window and then back to Surge XT; 1/3 of the time nothing happens and I can play it with computer keyboard; another 1/3 the F in the current octave is triggered and is stuck when I play the computer keyboard, sometime another F an octave higher is also triggered and held; and the last 1/3 an F and sometimes an F an octave up play and get stuck as soon as I switch focus back to SurgeXT.

This is only tested with v1.3.1.4b53827 standalone, Kubuntu 23.10, Plasma 5.27.8, KWin.

baconpaul:
OK great. So sounds like 'something' is wrong with our keyboard handling on linux.

i wonder if its the extended key keyboard mask.

Will add to list!

LoipesMas:
[...]

PR #7976, opened 2025-02-13
fix the virtual keyboard causes the E and F to be triggered (#7604)

DevelopMan303:
It seems when the octave up/down shortcut is pressed along with a note, this issue appears. It also shows up on Windows. 
My solution is to stop all the pressed notes after changing the octave, as key repeat will trigger the changed note. With this, it worked on on Ubuntu. 
(BTW: my first pull request :-) )


mkruselj:
That is a venerable attempt, but... I am not sure I would consider turning all notes off if you only change octaves a good UX... no other virtual keyboard I know of works like that...

baconpaul:
But the fact that this fixes it tells us a lot 

Namely a key is getting misinterpreted as a note and as an octave shift 

I wonder if our key map is doing both and the answer is if you do an octave shift don’t do a key action and vice versa

DevelopMan303:
Thanks for your feedback. I agree, this may be a lite simplistic approach. I may find an optimization.  
For me it looks there is an issue with the octave shift. Likely, the NoteOff is provided to the octave-shifted note.

mkruselj:
If you press a normal note and hold it, but then shift the octave, the key you are holding is not going to be triggering the same note you're holding. That's why when you release that key, it doesn't send the correct note off.

PR #7978, opened 2025-02-14
Turn on LTO for release builds.

mx:
Tested locally with g++, it sets -flto=auto and -fno-fat-lto-objects flags. Should work with the other compilers supported by cmake.


baconpaul:
Great - merged

let’s just watch to make sure a nightly happens. If it doesn’t we can figure it out from the actions logs

PR #8015, opened 2025-04-24
Luaeditor addons

blancoberg:
* Editor auto close brackets ()[]""{} -> only auto close when next character is a blank space, new line or closing character.
* Backspacing steps between tabs instead of one character at a time
* Double click only select one variable name at a time. ( e.g only velocity or state in "state.velocity" )
* Saves state for the code editor. Caret position, selection, search/replace/goto line. ( saved on editor focus lost ) 
* Formula editor state is cleared for a specific formula if a formula preset is loaded.
* Formula editor state is cleared for all (both scenes) if a synth preset is loaded. 
* Wt editor state is cleared on preset load/initialise 
* Lfo display cache is cleared when switching between output indexes in formula and when loading a formula preset.


mkruselj:
Awesome additions!!!

PR #8017, opened 2025-04-25
modulation buttons

blancoberg:
* Reverted to 26 in surgestorage, as it wasnt suppose to increase in my previous PR.

* Fixes glitchy behaviour when mouse-dragging a modulation button ( now with 0.7 opacity while dragging )

modsource-glitch
modsource-glitch-fix

* adds state to modulation buttons that saves its current modulation index
* button state is cleared when loading new patch/initialise


blancoberg:
I'm not sure if this is the best approach for reading the mod button state values, but it does seem to work.  ( I'v put it in setModList() )

blancoberg:
did @baconpaul take a look on the state saving part? and also, did I do the revert correctly?

mkruselj:
Mmm you reverted the version number for sure, but if you compared against version <27 elsewhere in code, I don't see that modified back to comparing against version <26... so that would need to be fixed. 

No matter, that can be a separate PR.

PR #8020, opened 2025-04-28
lua editor multi line comment

blancoberg:
Added highlighting for multi line comments in lua editor


blancoberg:
No idea how to fix this error. I probably missed some important steps when adding the new files?

blancoberg:
I forgot to move it. what folder do you think it should be in?

mkruselj:
`src/common` should suffice IMO!

blancoberg:
moved to common!

baconpaul:
src/common is juce free and has no ui code. This absolutely doesn't belong there. It belongs somewhere in src/surge-xt

baconpaul:
(I personally would have left it in overlays)

mkruselj:
Tokenizer is not UI code tho. But sure, moving it to `src/surge-xt` is fine too (actually, yeah, it's even better). To me it doesn't make sense in `overlays`.

baconpaul:
A differently factored tokenizer is not
The juce code tokenizer (1) only works on juce strings and (2) returns juce colors 
So it has to go in juce ui land

blancoberg:
The original luaTokeniser was located in the same folder as the juce::CodeEditorComponent ( juce gui extra ), thats why I put it in overlay.

So where should I put it? src/surge-xt, common or overlays?

baconpaul:
It has to be in surge xt or a subdirectory thereof. Items in common don’t have acccess to the surge gui library 

I personally think it was fine in overlays but technically it can’t be in common.

mkruselj:
Yeah `src/surge-xt` is good enough!

blancoberg:
For some reason I couldnt get it to work in /util/. The file couldnt be accessed. So surge-xt will have to do for now

mkruselj:
It should work the same way as LockFreeStack as far as cmake is concerned...

Then when importing the header where needed, you use the <SurgeLuaTokeniser> form rather than "SurgeLuaTokeniser.h" form.

PR #8024, opened 2025-05-04
Removes macros subscriptions for Formula

### LINKED ISSUE #8023: Remove Macro subscription for Formula
As per discussion on Discord, removing the requirement for the macro parameters inside Formula seemed to raise no objections. The idea is that this allows them to be used inside the init() block, will make working with them easier as well as make Formula's parameters more uniform since subscriptions are no longer being used for anything else.

Unfortunately, this won't be a completely clean removal, since the way subscriptions are set now as a boolean inside a nested table (subscriptions.macros), removing the parenting tables will break existing patches so we do need to keep them.

So this won't be a breaking change but will make the opt-in code redundant in existing patches and it's good to have some eyes on this and make sure this does not have unintended consequences.

The following third party patches could be cleaned up once this is merged:
/patches_3rdparty/A.Liv/Basses/Dynamo Square.fxp
/patches_3rdparty/A.Liv/Basses/Move Like A Bucket Truck.fxp
/patches_3rdparty/A.Liv/Basses/Sub Wrap Example.fxp
/patches_3rdparty/John Valentine/FX/DTMF MF4 Telephone Dialler.fxp
/patches_3rdparty/John Valentine/Percussion/Tongue Drum.fxp

Tagging @Andreya-Autumn and @j5v as this may be relevant to you.

nuoun:
- Closes #8023
- Adds macros to init() and removes subscription checks
- Keeps subscriptions.macros tables to not break existing patches
- Updates Formula tutorials 6 and 12, plus Init Duophonic patches
- Cleans up obsolete lfo_envelope and voice subscriptions from testrunner


src/common/dsp/modulators/FormulaModulationHelper.cpp
```diff
@@ -194,15 +194,9 @@ end
...
-        // add subscription hooks
-        lua_createtable(s.L, 0, 5);                   // subscriptions
-        lua_createtable(s.L, n_customcontrollers, 0); // macros
-        for (int i = 0; i < n_customcontrollers; ++i)
-        {
-            lua_pushnumber(s.L, i + 1);
-            lua_pushboolean(s.L, false);
-            lua_settable(s.L, -3);
-        }
+        // Legacy tables for deprecated macro subscriptions
```
mkruselj, line 197: Maybe add a `TODO: Remove in XT2` comment here?

nuoun:
Can be merged I hope

PR #8026, opened 2025-05-06
A11y announcement of tuning being stored in patch

### LINKED ISSUE #8022: Make TTS Announcement When Loading Patch with Embedded Tuning

At present, the only way for a screen reader to determine whether a patch has an embedded tuning which may be loaded is to open the tuning menu and see if the option exists. I'm proposing to append a message to the tts announcement that exists when accessibility options are set.

I'd like to append a short message, such as, ", tuning available), to the message containing the name of the patch which is announced through the tts when the appropriate accessibility option is enabled.

Discussion on Discord has yielded the possible alternative of adding the message to the tuning button itself; however, I feel that my method would be better because it more closely mimics the sighted experience. Given that screen reader users can't get all the information they need at a glance, putting the info on the tuning button wouldn't have the same effect as it's easily skippable.

**Additional context:**

mkruselj:
Also remove leading path from a11y readout of category name

Fix leading whitespace of A.Liv/Grant Me... patch

Closes #8022


src/surge-xt/gui/widgets/PatchSelector.cpp
```diff
@@ -639,9 +639,9 @@ void PatchSelector::showClassicMenu(bool single_category, bool userOnly)
         // get just the category name and not the path leading to it
         std::string menuName = storage->patch_category[rightMouseCategory].name;
 
-        if (menuName.find_last_of(PATH_SEPARATOR) != std::string::npos)
+        if (menuName.find_last_of("\\/") != std::string::npos)
```
baconpaul, line 642: This is odd why are you testing for both slashes

mkruselj: I noticed in FXP there can be either slash used depending on how the user typed the category string.

I didn't see if this is being normalized anywhere...?

PR #8032, opened 2025-05-21
Fix popup menu scaling

pawelulita:
This is a draft PR with a problem statement and a possible solution.

---

Currently, the scaling of the popup menus seems to be determined once. Setting the zoom level doesn't change the scale factor of the popup menus. The solution I have relies on creating a dummy component that follows the scale set by the user with a factor of 0.75 and using this component as the target component for the popup menus. It's not very clean, but seems to work pretty well. However, there's a caveat. **It seems impossible to close a popup menu by clicking the button again.**

The main technical point here is that JUCE uses `Component::getApproximateScaleFactorForComponent()` on the target component to determine what the scale of the popup menu should be. It seems to boil down to taking whatever we set using `Component::setTransform()` as the scale.

With this in mind, I tried the following approaches:

1. Using `setZoomFactor()` on `juceEditor` in `SurgeGUIEditor::setZoomFactor()` instead of the current (more manual?) approach to scaling. This called `setTransform()` and the menus scaled. However, I was getting a lot of scaling glitches with the main interface and never got it to work properly.

2. Using `o.withTargetComponent(*frame)` instead of `withTargetComponent(juceEditor)` when creating the popup menus. This works kind of alright, because `frame->setTransform()` is called each time the scale changes. However, the menus were to big compared to the rest of the interface. 

3. I created a dummy component called `menu_scaler` that mirrors the scale set on `frame`, but scales it by 0.75. This seems to hit the right spot.

I've tested this on a Linux machine running X11 and Wayland in Reaper and Bitwig using VST3 and CLAP, and the standalone.

The question is: **Is there a better way?**

### DIFF
--- a/src/surge-xt/gui/SurgeGUIEditor.cpp
+++ b/src/surge-xt/gui/SurgeGUIEditor.cpp
@@ -2424,6 +2424,9 @@ bool SurgeGUIEditor::open(void *parent)
     frame->setBounds(0, 0, currentSkin->getWindowSizeX(), currentSkin->getWindowSizeY());
     frame->setSurgeGUIEditor(this);
 
+    menu_scaler = std::make_unique<juce::Component>();
+    juceEditor->addChildComponent(*menu_scaler);
+
     // Comment this in to always start with focus debugger
     // debugFocus = true;
     // y-frame->debugFocus = true;
@@ -2889,7 +2892,7 @@ void SurgeGUIEditor::toggleMPE()
 juce::PopupMenu::Options SurgeGUIEditor::popupMenuOptions(const juce::Point<int> &where)
 {
     auto o = juce::PopupMenu::Options();
-    o = o.withTargetComponent(juceEditor);
+    o = o.withTargetComponent(*menu_scaler);
     if (where.x > 0 && where.y > 0)
     {
         auto r = juce::Rectangle<int>().withWidth(1).withHeight(1).withPosition(
@@ -3060,6 +3063,12 @@ void SurgeGUIEditor::setZoomFactor(float zf, bool resizeWindow)
     if (frame)
     {
         frame->setTransform(juce::AffineTransform().scaled(zff));
+    }
+
+    if (menu_scaler)
+    {
+        // 0.75 scales the menus so that they match the rest of the UI.
+        menu_scaler->setTransform(juce::AffineTransform().scaled(zff * 0.75));
     }
 
     if (oscWaveform)
--- a/src/surge-xt/gui/SurgeGUIEditor.h
+++ b/src/surge-xt/gui/SurgeGUIEditor.h
@@ -136,6 +136,14 @@ class SurgeGUIEditor : public Surge::GUI::IComponentTagValue::Listener,
 
     std::unique_ptr<Surge::Widgets::MainFrame> frame;
 
+    /*
+     * Dummy component used to make sure the scale of the menus visually matches
+     * the scale of the rest of the interface. Using `juceEditor` results in the
+     * scale being set only once, and using `frame` results in the scale being
+     * too big.
+     */
+    std::unique_ptr<juce::Component> menu_scaler;
+
     std::atomic<int> errorItemCount{0};
     std::deque<std::tuple<std::string, std::string, SurgeStorage::ErrorType>> errorItems;
     std::mutex errorItemsMutex;


baconpaul:
Thanks for tackling this.

The way juce works  roughly is

1. The actual editor component is basically empty. That then gets the transform from the DAW.
2. What is *currently* the editor component becomes a sub-component and it gets the transform

That's why we apply the transform to the frame.

the thing is the HDPI transform scale gets applied directly to the juce Editor subclass, so if you apply a transform there, it loses HDPI. That's why we have the kinda approach today. This is also mostly why your #1 approach is not mergable.

Your #2 approach will apply both the OS scaling and the user scaling which is why things are too big.

Your #3 approach might break accessibility in a bunch of ways - I think the component will need to advertise as non-accessible and non-keyboard-focusable - and we will need to test carefully. Its definitely the best bet

I think the approach is

1. Try #3 but have the factor it applies be the inverse of the zoom factor not a hardcoded 0.75. This basically will then follow the OS HDPI (which is the scale factor on the juceEditor)
2. Make your dummy component set up to be explicitly not accessible and not keyboard focusable
3. Then alas we have to test it on mac, on windows, and on windows with HDPI scaling, and windows with HDPI scaling will probably be screwed, since HDPI there is a nightmare. And then
4. We have to test it with a screen reader and be sure that the focus returns to the right element when dismissing a menu and the positioning of the dummy element doesn't break readers.

This is really not easy but happy for you to think about taking a swing at it!

Thoughts?

baconpaul:
Oh and probably we will have to #if this by OS. For instance on my mac the current code has the menu the correct size independent of zoom factor.

So this will require testing in each OS. 

Or an #if linux around your set of the component and then just leave mac and win alone (which also obviates the need for a screen reader test since that tech isn't available on linux in plugins).

pawelulita:
@baconpaul Thanks for your response! There's a lot of information to take in. I'll try to digest it over the next few days.

At the same time, your second comment made me doubt if we're on the same page about what I mean here. The reason is that I'm pretty sure I have seen the behaviour I'm trying to fix here on my friends Mac a few days ago.

Here's 100% and 125% on the latest version of Surge. See how the menu has the same size on both screenshots.

Screenshot_20250521_202814
Screenshot_20250521_202848

Here's 100% and 125% with my patch.

Screenshot_20250521_203047
Screenshot_20250521_203102

The thing is, the size in the first screenshot is fine for my screen when I use it in Reaper, but it's rather small in Bitwig or in standalone mode. So the initial value isn't necessarily wrong on Linux - sometimes it's apparently correct (as in: roughly the same as the DAW scaling factor). And probably on macOS the initial value is (almost) always set correctly. Could you please confirm whether the menus change size when you adjust the zoom level?

baconpaul:
Yes it is what happens in Mac too

pawelulita:
I see, it's a feature, not a bug. :) Well, I can definitely live with the philosophy that the menus should follow the OS scale (or the DAW scale). However, this is not always my experience on Linux. On top of that, there are some other problems I'm experiencing that seem to be related.

It seems like everything works well in Reaper - the scale of the menus seems to follow what I hardcoded in Reaper, 1.25 in my case. Setting 125% in Surge makes everything comfortable to use.

The problems start in Bitwig, the zoom level set to 125% in the DAW itself. Here, to get roughly the same size of the UI, I have to set the zoom level to 155%, which seems to come from 125% * 1.25 = 156,25%. However, the menus stay significantly smaller. (I don't seem to be able to make a screenshot with the menu open in Bitwig.) This was a direct reason why I started investigating this issue.

It's unclear to me at this point if this is a problem with Bitwig or Surge. There are definitely other plugins that are smaller in Bitwig than in Reaper, for example Airwindows Consolidated. (In this case I can't change the zoom level of the plugin, so everything is kind of too small.) But also plugins that are exactly the same with the same zoom level, like Zebra 2 or RipplerX (JUCE-based).

I'm not sure what this all means, but there's definitely _something_ that doesn't work properly, but I'm not sure what and where, haha.

baconpaul:
Sounds to me like bitwig isn't setting the correct scale in the VST3 api in linux.

I'm travelling for a bit but I can check that out when I'm back with my linux box as opposed to laptop

But if multiple VST3 plugins are different linux bitwig vs reaper, then sounds like bitwig needs to look at their implementation of `IPlugViewContentScaleSupport::setScaleFactor` in the VST3 API perhaps?

If you send a bitwig support note let me know and I can tag it to them

pawelulita:
@baconpaul Haha, no worries! don't see how to reopen it either, but I don't think it matters - what I proposed here is probably not the best approach anyway.

I think at this point I'll try to play with JUCE to see if I can pinpoint the problem. However, some plugins built using [DPF](https://github.com/DISTRHO/DPF/) don't seem to be affected. Could this be a JUCE-specific problem? Also, the VST3 and CLAP versions of Surge behave the same here. I'm not sure what it means, though.

baconpaul:
Hmm ok let’s open an issue to discuss debug then and when I’m back with a Linux box I can peel - 10 days tho

pawelulita:
@baconpaul Thanks again for your help!

Just to clarify, since you and @mkruselj are in favour of the popup menus following the OS/DAW scale, addressing the issue of some plugins in Bitwig on Linux being too small seems to be the right approach. This is mostly because it tries to tackle the underlying problem instead of patching it in one instance. Do you think this is the right approach?

pawelulita:
All of the testing so far I've done on my laptop (14", 1920 x 1200), and I kind of forgot that I still have a desktop. :sweat_smile:

On my desktop (24", 3840 x 2160) running basically the same software, the size of the menus is the same between Bitwig and Reaper. Also, given a certain zoom level, the size of the main window is also the same. Both Bitwig and Reaper have 200% as their own scaling factor. The system scaling is 175% (KDE Plasma on Wayland).

However, the size of the menus (and the main window) is independent of the OS or Bitwig scaling. (In Reaper it follows the DAW.) Could it mean that Bitwig tries to detect the size and the resolution of the screen and computes the scale using that? Or maybe it has 2 modes: low DPI and high DPI?

PR #8036, opened 2025-06-10
Fix corrupted morph param crash and affected patches

nuoun:
Thanks again to Discord user Jo1verse319 for reporting

This fixes a crash that occurs when the 3D wavetable display receives a negative morph param value from a corrupt patch and attempts to fix affected 3rd party patches. To prevent saving these patches to the new XML format and lose legacy options a hex editor was used to edit the values while preserving the exact file size

As a workaround a value like `-1.913359` was edited to become `-0.000000`

CC: @inigokennedy to inform you of these changes.


nuoun:
Bit less of a dirty hack adding whitespace instead of a bogus sign for editing these params, thanks ED for the suggestion

PR #8071, opened 2025-08-17
feat: Add TCP control API for programmatic preset and parameter control

shiehn:
- Add TCP listener on port 7833 for external control
- Implement preset loading by name and file path
- Add parameter setting via TCP commands
- Include Lua/ReaScript client for Reaper integration
- Add MCP JavaScript bridge for AI assistant workflow
- Support for NLP -> MCP JS -> Lua/ReaScript -> TCP pipeline

This enables AI assistants and automation tools to control Surge XT without GUI interaction, supporting preset loading, parameter control, and preset enumeration via simple TCP commands.

### DIFF
--- a/src/common/SurgeSynthesizer.cpp
+++ b/src/common/SurgeSynthesizer.cpp
@@ -24,6 +24,7 @@
 #include <fmt/core.h>
 #include "DSPUtils.h"
 #include <ctime>
+#include "TCPControl/SurgeTCPController.h"
 
 #include "SurgeParamConfig.h"
 
@@ -299,6 +300,9 @@ SurgeSynthesizer::SurgeSynthesizer(PluginLayer *parent, const std::string &suppl
 
     patchid_queue = -1;
     has_patchid_file = false;
+    
+    // Initialize TCP control
+    initializeTCPControl();
 }
 
 SurgeSynthesizer::~SurgeSynthesizer()
@@ -311,6 +315,11 @@ SurgeSynthesizer::~SurgeSynthesizer()
         std::lock_guard<std::mutex> mg(patchLoadSpawnMutex);
         if (patchLoadThread)
             patchLoadThread->join();
+    }
+
+    // Stop TCP control
+    if (tcpController) {
+        tcpController->stop();
     }
 
     stopSound();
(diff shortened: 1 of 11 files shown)


baconpaul:
Hmmm this will take some careful review of course.  But I’m curious why the open sound control integration we already have doesn’t cover these use cases?

may be best to join our discord to discuss as well once I review. I’ll try and look top of week

src/common/SurgeSynthesizer.cpp
```diff
@@ -299,6 +300,9 @@ SurgeSynthesizer::SurgeSynthesizer(PluginLayer *parent, const std::string &suppl
 
     patchid_queue = -1;
     has_patchid_file = false;
+    
+    // Initialize TCP control
+    initializeTCPControl();
```
baconpaul, line 305: Oh a Quick Look it seems the docs talk about an optional build step but this turns it on for all instances always. Is that what you wanted to submit?

baconpaul:
Clever idea though but I’ll definitely want to share a few thoughts to make it maintainable and so on. 

Thanks for joining in the fun!

src/common/SurgeSynthesizer.cpp
shiehn: I'm so sorry, this PR way ment for my fork, not the main.  I will close this PR.  This is specific to a project of mine. Essentially I'm building a MCP style agent/assistant targeting the Reaper DAW.  I was looking into exposing patch management via a TCP port.

shiehn:
Was not intended as a PR into main

mkruselj:
Still, why TCP when you already have the OSC implementation? And Reaper also supports OSC...

shiehn:
> Still, why TCP when you already have the OSC implementation? And Reaper also supports OSC...

Oh, well, If Surge XT allows me to programatically load a patch by name via OSC than TCP is completely unneeded. Which is best-case-senario.  I was trying to load patches via reascript without success and realized Surge has its own internal patch management system.  So I immediately started experimenting with adding a TCP backdoor (so-to-speak).  Thanks for bringing this to my attention!

baconpaul:
Yeah the OSC has commands to do almost everything you want. modify parameters load patches load wavetables

I think the only thing missing is enumerating the patches. but you can do that from the filesystem too!

PR #8084, opened 2025-08-22
Split LFO presets menu into factory and user columns

nuoun:


nuoun:
image

Can be reviewed

mkruselj:
Much nicer!

Best variable name so far: `recurseCat`. 😁

PR #8091, opened 2025-08-24
Add WTS Preset Manager

nuoun:
Just a draft for now.


src/surge-xt/gui/widgets/OscillatorWaveformDisplay.cpp
```diff
@@ -719,89 +806,172 @@ void OscillatorWaveformDisplay::populateMenu(juce::PopupMenu &contextMenu, int s
...
 
-            auto action = [this]() {
-                toggleCustomEditor();
+                this->storage->wavetableScriptPreset->loadPresetFrom(p.path, this->storage, scene,
+                                                                     oscInScene);
 
-                if (!customEditor)
+                auto *os = &this->storage->getPatch().scene[scene].osc[oscInScene];
+                auto wtse = std::make_unique<Surge::Overlays::WavetableScriptEditor>(
+                    sge, this->storage, os, oscInScene, scene, skin);
```
nuoun, line 875: Since the overlay may or may not exist when a WTS preset is loaded and even if it exist it could be from a different oscillator, in order to WavetableScriptEditor::generateWavetable() I'm running the WavetableScriptEditor constructor here on each preset load and pass a boolean so WavetableScriptEditor knows if it should use the overlay UI elements or not. Not 100% sure if this is the best way or if this will have unintended side effects so I would appreciate a good look at this.

mkruselj:
Just a minor point: I think a better extension would be .wtscript.

src/common/SurgeStorage.cpp
```diff
@@ -49,6 +49,7 @@
 #endif
 #include "FxPresetAndClipboardManager.h"
 #include "ModulatorPresetManager.h"
+#include "WavetableScriptPresetManager.h"
```
mkruselj, line 52: Could just name it `WavetableScriptManager` then.

src/common/SurgeStorage.cpp
```diff
@@ -380,6 +381,7 @@ SurgeStorage::SurgeStorage(const SurgeStorage::SurgeStorageConfig &config) : oth
     userFXPath = userDataPath / "FX Presets";
     userMidiMappingsPath = userDataPath / "MIDI Mappings";
     userModulatorSettingsPath = userDataPath / "Modulator Presets";
+    userWavetableScriptPath = userDataPath / "Wavetable Script Presets";
```
mkruselj, line 384: As discussed we will want to merge .wtscripts with normal wavetables, so separate path shouldn't be needed. Can resolve in a subsequent PR, although it chould really be as simple as adding a new supported extension to `SurgeStorage::refresh_wtlistFrom()`.

src/common/SurgeStorage.cpp
```diff
@@ -619,6 +621,15 @@ SurgeStorage::SurgeStorage(const SurgeStorage::SurgeStorageConfig &config) : oth
     {
         reportError(e.what(), "Error Scanning Modulator Presets");
     }
+    try
+    {
+        wavetableScriptPreset = std::make_unique<Surge::Storage::WavetableScriptPreset>();
```
mkruselj, line 626: Again can just shorten this to `wavetableScript`, across the board.

src/common/WavetableScriptPresetManager.cpp
```diff
@@ -0,0 +1,275 @@
...
+#include "tinyxml/tinyxml.h"
+#include "sst/plugininfra/strnatcmp.h"
+
+namespace Surge
+{
+namespace Storage
+{
+
+const static std::string PresetDir = "Wavetable Script Presets";
+const static std::string PresetXtn = ".wtspreset";
```
mkruselj, line 36: Super nitpick: Ext instead of Xtn.

src/common/WavetableScriptPresetManager.cpp
```diff
@@ -0,0 +1,275 @@
...
+        auto fullLocation =
+            (containingPath / location).lexically_normal().replace_extension(PresetXtn);
+
+        // make sure your category isnt "../../../etc/config"
+        auto [_, compIt] = std::mismatch(fullLocation.begin(), fullLocation.end(), comppath.begin(),
+                                         comppath.end());
+        if (compIt != comppath.end())
+        {
+            s->reportError("Your save path is not a directory inside the user presets directory. "
+                           "This usually means you are doing something like trying to use ../"
```
mkruselj, line 68: "This usually means you are trying to use ../ in the preset name." Shorter and to the point.

src/common/WavetableScriptPresetManager.cpp
```diff
@@ -0,0 +1,275 @@
...
+
+    auto frm = wts->FirstChildElement("wavetable_formula");
+    if (frm)
+        s->getPatch().wtsFromXMLElement(&(s->getPatch().scene[scene].osc[oscid]), frm);
+}
+
+/*
+ * Note: Clients rely on this being sorted by category path if you change it
+ */
+std::vector<WavetableScriptPreset::Category> WavetableScriptPreset::getPresets(SurgeStorage *s,
```
mkruselj, line 153: I suppose this method is going to be the only casualty of merging .wtscripts with regular wavetables - since we already have a method in SurgeStorage to scan for WT content.

src/common/SurgePatch.cpp
```diff
@@ -4308,3 +4308,17 @@ void SurgePatch::formulaFromXMLElement(FormulaModulatorStorage *fs, TiXmlElement
         fs->interpreter = (FormulaModulatorStorage::Interpreter)(interp);
     }
 }
+
+void SurgePatch::wtsToXMLElement(OscillatorStorage *osc, TiXmlElement &parent) const
+{
+    parent.SetAttribute(
+        "wavetable_formula",
```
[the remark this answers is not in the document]
nuoun: Makes sense. OscillatorStorage uses wavetable_formula for var names, could rename that too in a future PR.

src/common/WavetableScriptPresetManager.cpp
nuoun: Will also change this in ModulatorPresetManager in a future PR then.

src/common/WavetableScriptPresetManager.cpp
nuoun: Agreed and also can be done in ModulatorPresetManager.

src/surge-xt/gui/widgets/OscillatorWaveformDisplay.cpp
```diff
@@ -636,9 +630,102 @@ void OscillatorWaveformDisplay::populateMenu(juce::PopupMenu &contextMenu, int s
 
     contextMenu.addSeparator();
 
+    createWTLoadMenu(contextMenu);
+
+    createWTExportMenu(contextMenu);
+    contextMenu.addSeparator();
+
+#if INCLUDE_WT_SCRIPTING_EDITOR
```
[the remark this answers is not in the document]
nuoun: We could still hide these menu's for builds without Lua, these #ifs would use HAS_LUA instead then but we're not doing this for Formula and would also need to hide the .wtscript items from the wavetable lists.

src/surge-xt/gui/widgets/OscillatorWaveformDisplay.cpp
mkruselj: I _think_ the right way to go is to use the OscillatorStorage state always, not any of the UI elements.

mkruselj: WT generation should simply be from the storage data (both the script and frames/samples are members there IIRC). UI refresh would still happen in OWD after calling the generate function from SurgeStorage. Apply button in WTSE would also self-repaibt after calling generate from SurgeStorage.

nuoun:
Can be reviewed, just two things before merging:

1. Last chance to change `.wtscript` file and XML structure before it gets out in the wild.

```
<?xml version="1.0" encoding="UTF-8" standalone="yes" ?>
<wtscript>
  <params frames="10" samples="5" />
  <wavetable_script lua="[base64 encoded string" />
</wtscript>
```

2. Are we sure the WT generation [here](https://github.com/nuoun/surge/blob/085c9bdb13e4b12fad5ac31054e0a3f58d06597c/src/common/SurgeStorage.cpp#L1379) runs in the UI thread?

mkruselj:
Maybe `meta` instead of `params` but that's it.

load_wt() is never called from audio thread IIRC.

nuoun:
Some changes per discussion on Discord and new insights:

- Moves .wtscript loading and WT generating out of Storage and back to the initiating threads
- Adds script loading to file drag and drop and OSC Wavetable commands
- Refreshes script editor overlay if active on .wtscript load
- Moves saveWavetableScript() to OscillatorWaveformDisplay and removes WavetableScriptManager
- Moves generateWavetable() from LuaEditors to WavetableScriptEvaluator
- Renames wt.is_dnd_imported to wt.force_display_update
- Makes XML file structure similar to .wt and .wav metadata:
<?xml version="1.0" encoding="UTF-8" standalone="yes" ?>
```
<wtscript>
    <script lua="[base64]" frames="10" samples="5" />
</wtscript>
```
Can be merged I hope so I can do the final OSC testing in the Nightly, for some reason I could not get it to connect with my local build.

mkruselj:
One question: why base64 for the Lua code? One would think that could simply be stored directly as a string containing actual code...

nuoun:
I think the main reason is that the Lua string may contain characters that would need to be escaped in XML

mkruselj:
Fair enough!

baconpaul:
That's exactly it yes

PR #8105, opened 2025-09-05
Editor overlay things

nuoun:
- Add WTS data to Copy Osc and Copy Scene
- Switch active WTS overlay on Osc and Scene select if applicable (new osc type == wavetable)
- Use forceRefresh() for editors on patch / preset load instead of opening and closing
- Sets editor apply buttons on forceRefresh() to fix case where unapplied changes prompts chickenbox dialog box hidden behind opened overlay on patch change
- Close active WTS overlay on patch change if applicable
- Disable WTS Editor keybind action if current osc is not wavetable
- Rename some instances of "Wavetable Editor" to "Wavetable Script Editor"

### DIFF
--- a/src/surge-xt/gui/SurgeGUIEditorOverlays.cpp
+++ b/src/surge-xt/gui/SurgeGUIEditorOverlays.cpp
@@ -294,7 +294,8 @@ std::unique_ptr<Surge::Overlays::OverlayComponent> SurgeGUIEditor::createOverlay
             this, &(this->synth->storage), os, current_osc[current_scene], current_scene,
             currentSkin);
 
-        std::string title = fmt::format("Osc {} Wavetable Editor", current_osc[current_scene] + 1);
+        std::string title =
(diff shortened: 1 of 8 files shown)


src/surge-xt/gui/SurgeGUIEditorOverlays.cpp
```diff
@@ -838,16 +840,38 @@ bool SurgeGUIEditor::updateOverlayContentIfPresent(OverlayTags tag)
         }
         break;
     }
-    case WAVESHAPER_ANALYZER:
+    case FORMULA_EDITOR:
```
nuoun, line 843: MSEG editor could also benefit from using refresh (it's much nicer than closing and opening, especially with torn out overlay) but just calling forceRefresh() does not reset the nodes properly on patch change so that would need fixing first.

mkruselj:
Excellent UX fixes and tweaks here!

PR #8164, opened 2025-10-13
Migrate most of the FX parameter overlay into its own component.

mx:
This moves the code that's laying it out, but generally doesn't move ownership, which is an extremely complex beast in the Surge editor.

### DIFF
--- a/src/common/dsp/Effect.h
+++ b/src/common/dsp/Effect.h
@@ -61,9 +61,6 @@ class alignas(16) Effect
     // this is called after the loading state of an item that has changed
     virtual void updateAfterReload() {}
 
-    virtual Surge::ParamConfig::VUType vu_type(int id) { return Surge::ParamConfig::vut_off; }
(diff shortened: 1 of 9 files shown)


baconpaul (commented):
This generally looks pretty good to me. I didn't run it so we should definitely a/b test vs current to make sure the crazy ordering thing hasn't broken.

Other thought is: the current fx display doesn't implement on skin changed or any such. Does the parent handle that rebuild when we change a skin or do we need a callback?

Finally it looks like the tab order is correct and unchanged and the accessible names will be correct but turning on the focus debugger and making sure would be super useful i bet.

baconpaul:
The code looks reasonable. Let me just run it under ASAN to be quadruple extra sure and if that checks out I'll merge it this morning.

baconpaul:
Yeah cool the only sanitizer problems are ones i fixed subsequent to the start of this branch. So I'm gonna merge this!

PR #8166, opened 2025-10-13
wavetable display - skewH

blancoberg:
SkewH finally added to the wavetable oscillator display. 

It's not the prettiest solution, but it works quite well and should at least in theory be faster than actually using the real skew equation. 
I really tried to implement the equation, but I never got it to match the waveform output, so this will have to do for now.


blancoberg:
skewH

PR #8173, opened 2025-10-14
Correct Tx 2, 4, 6 and 8 waveforms

### LINKED ISSUE #7965: Port the proper TX 2,4,6,8 from six sines
and move current 2,4,6,8 to 'spiky' variants and do patch migration

baconpaul:
When writing Six Sines I realized the forms we had for TX 245 and 8 were too "spiky" and while these matched the manual they didn't matc a scope from a TX which was really just doubled shifted cos etc...

So port that fix back here from six sines and add the 4 legacy even shapes as new modes and adjust streaming.

Keep the alias oscillator having the old shapes only.

Closes #7965

### DIFF
--- a/src/common/Parameter.cpp
+++ b/src/common/Parameter.cpp
@@ -1086,13 +1086,13 @@ void Parameter::set_type(int ctrltype)
         break;
     case ct_sineoscmode:
         val_min.i = 0;
-        val_max.i = 27;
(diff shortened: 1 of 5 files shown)


mkruselj:
Ringmod effect uses these waveforms too, it should have the proper and Legacy set as well.

baconpaul:
yeah that's in the diff.

Issue #6283, opened 2022-06-11
No ASIO available in Windows standalone


planist79:
Hello,
i have noticed that Surge XT 1.0.1 Standalone does not recognize my ASIO driver. All my other audio software does provide the ASIO option. 
Surge only allows 

- Windows Audio
- Windows Audio (Exclusice Mode)
- Windows Audio (Low Latency Mode)
- Direct Sound

So i wonder what i can do?

Spec:
OS: Win 10
Audio Interface: Focusrite Scarlett 18i20

Thank you

baconpaul:
Hi!

Unfortunately ASIO is licensed in a way which means we cannot distribute a binary including it in open source software. Audacity has the same issue. 

You can build surge yourself with ASIO support if you execute a license and do not redistribute the binary, that is detailed https://github.com/surge-synthesizer/surge#building-with-support-for-asio

i would love to give another answer, but ASIO is a proprietary toolkit owned by a single market participant

planist79:
Thank you for the quick reply!

planist79:
Hi, Thanks for clarifying the issue. I am no developer at all, so is there a detailed step-by-step description on how to build my own Surge XT Standalone? When i have a plan win10 installation with only Audop software on it. 

I see in the short Surge Wiki that need to install MS Visual Studio and some components, then download ASIO sdk, and then do what..?? Its just to much for me .. please help ;)

Thank you, Peter

mkruselj:
Yes, please read the readme in our repository, it explains build steps.

planist79:
Hi Mario, thank you for the advice! SAVIhost works great, not i have ASIO drivers for the standalone.

mkruselj:
Reopening this since, miraculously, Steinberg has just dual-licensed ASIO. Moving to 1.4 milestone.

- [ ] Unpin this issue after 1.4 is released

baconpaul:
Fixed.

planist79:
great, thank you!! this is good news.

PR #8185, opened 2025-10-15
With ASIO going dual license GPL3, add it to our codebase

baconpaul:
by placing it in sst-plugininfra and adjusting our cmake


baconpaul:
Addresses #6283

PR #8192, opened 2025-10-16
Dispatch build info as JSON to websites

mthierman:
This is to test dispatching build info to our websites as JSON so we can generate a content collection. Current website update logic is left intact until we can verify the payload in the website pipeline.


baconpaul:
Oh this is great.

If this works we can move it to sst-pluginfra so the rest of the properties can do the same right?

Nifty.

mthierman:
Yeah, although I'm not sure if it should be a bash script somewhere, a CMake target, a github action, or what.

Edit: I suppose this also could be done in python, jq just seemed easier

PR #8202, opened 2025-10-26
Expose SURGE_BUILD_CLAP as an option

mrtnvgr:
Allows user to skip CLAP builds, useful in distro packages :)


baconpaul:
Thanks!

happy to merge  once ci passes- it’s nice and symmetric -  and appreciate the upstream vs some package having a diff somewhere.

out of curiosity which distro wants to turn off clap?

mrtnvgr:
> out of curiosity which distro wants to turn off clap?

Me personally 🌚

Since Renoise does not support CLAP yet,
I want to save disk space / compile time

baconpaul:
> > out of curiosity which distro wants to turn off clap?
> 
> Me personally 🌚
> 
> Since Renoise does not support CLAP yet, I want to save disk space / compile time

Ha great yeah that makes sense

anyway thanks for the contribution!

PR #8216, opened 2025-11-17
Surge Editor: Remove unused private variable

RavenRandomz:
I am new to the Surge XT codebase. I was having trouble with understanding it because I would have to frequently trace where a variable was declared due to the non-obvious abbreviation.

Furthermore, I located a variable which was not being used at all (a unique pointer to a logo). All builds including test builds were functional, so I removed it.

I read that the developer guide encouraged these "campground" commits, so I hope that even though these might be small, perhaps nitpicky to some, that they help the codebase in some form.

If I misunderstand what was meant by this policy, I apologize in advance.
image

### DIFF
--- a/src/surge-xt/SurgeSynthEditor.h
+++ b/src/surge-xt/SurgeSynthEditor.h
@@ -61,6 +61,8 @@ class SurgeSynthEditor : public juce::AudioProcessorEditor,
     void parentHierarchyChanged() override;
     void resetWindowFromSkin();
 
+    void paramsChangedCallback();
+    void setEffectType(int i);
 
     void handleAsyncUpdate() override;
 


mkruselj:
You will see that we use `g` as shorthand for `graphics` basically almost everywhere where we paint anything on the screen, so I would not be making that change. You would have to visit a lot more files if you want that to be thorougly enforced. Same thing with `sge`, really...

baconpaul:
Yeah g to graphics is a non starter. Idiom everywhere that g is the graphics both here and I other juce code bases

And some things like constructor arguments not conflicting with members were on purpose. I see you changed a few of those where you went from editor(e) to editor(Editor). In this context since it’s a pointer it doesn’t really matter but since some things have copy semantics I generally try to not have argument and member name clashes in constructors 

sge to surgeEditor will take me a few years to get used to. But it’s used elsewhere in the code base too, so while it is locally clear it makes things a little inconsistent. I’d rather stick with sge or editor as the name of a variable for that

baconpaul:
Oh and the second clause of the coding style is a bit ambiguous but is meant conservatively. If you don’t like a variable name but we use it 30 other places don’t change it was the intent.

and thanks for diving in!

one other thought. Good ides make most of these navigation issues disappear. Curious if you have your vscode or clion or whatever setup to do symbols?

RavenRandomz:
Thanks for the clarification! I am looking at the codebase through cmake, nvim, and coc. Navigation is pretty easy for going to definitions. Hope Thanksgiving went well for you all. I'm very grateful for SurgeXT ^_^

mkruselj:
With that said, could you modify your PR so that it only removes the unused variable? That's the only thing we'd want to accept from it. Thanks!

RavenRandomz:
> With that said, could you modify your PR so that it only removes the unused variable? That's the only thing we'd want to accept from it. Thanks!

Changed to only have the removal commit! ^_^

Curious, for future reference, is force pushing the only way to achieve this (and remove the other commits)? Or is there a cleaner way?

mkruselj:
You don't have to force push, we squash commits when merging anyways.

RavenRandomz:
Do I need to do anything else or is this all good?

RavenRandomz:
image
It doesn't say which formatting violations are present

mkruselj:
You need to clang-format the modified file with clang-format version 21.

mkruselj:
What I see is you have two green modifications in the diff that are entirely unnecessary. Only the removal of a line should be there.

RavenRandomz:
image
Yay! :partying_face:

Issue #8232, opened 2025-12-11
Add menu item to oscillator drop list (menu) items to allow an indirectly assigned CC to be removed.


Dave-Lowndes:
**Is your feature request related to a problem? Please describe!**
Issue mentioned on discord ["Is this a known issue with Surge XT"](https://discord.com/channels/744319641211633774/1447686905847218328)

This issue arises if you've MIDI learned an oscillator control, then change the oscillator type to one that has a drop list control (menu?) for the equivalent control position (ID/slot?).

The MIDI CC will still alter the new oscillator's control, but there's no straightforward way of clearing it - other than by switching the oscillator type back and removing it.

**Describe the solution you'd like:**
If the oscillator's drop list control had the context menu items to show the CC, and allowed you to clear it, that would be useful.

**Describe alternatives you've considered:**
It was suggested that the oscillator control should just ignore the CCs if they aren't appropriate, but that would break anything that relied on the current behaviour.

baconpaul:
Adding some comments here on how to do it since this is a 'good first issue' issue

So to reproduce the problem

Load init saw
On Osc 1 RMB on "shape" and choose midi learn
Twist a CC on your controller and learn engages.
If you now RMB on "shape" you see an entry "Clear Midi CC 14" (or what not

Now change the oscillators type to String. The position 1 changes to a dropdown menu for exciter
CC14 (or whatever it was) will still change the value
But when  you click the exciter menu the "Clear CC Midi CC 14" menu item isn't there

Your goal is to make that menu item appear on the discrete item. Then test the same thing with FX.

The clear items are created in `SurgeGUIEditor::createMIDILearnMenuEntries` in `SurgeGUIEditorValueCallbacks.cpp`

You can see that's called from `controlModifierClicked' in two places. One for macros, one for channel split, and one for params. You want the third one around like 3175

You can see that call is *only* if the type is float. So around line 3175 you could add

```
                createMIDILearnMenuEntries(contextMenu, param_cc, p->id, control);

            } // end vt_float if statement
            else
            {
                createMIDILearnMenuEntries(contextMenu, param_cc, p->id, control);
            }
```

and then the menu shows up. This isn't quite what we want though. We *only* want the delete items and we want a separator if the delete items are there.

So the way I would do it is

1. Add a new argument to createMIDILearnMenuEntries called `bool deleteOnly = false`
2. In the create if delete only is true only add the delete entries
3. And if delete only is true and you are adding the delete entries then add a separator
[...]

PR #8234, opened 2025-12-18
Add menu item to oscillator drop list (menu) items to allow an indirectly assigned CC to be removed. #8232

DevelopMan303:
Adds the MIDILearnMenuEntries to all vt_int (selection boxes) which can show up in the oscillator section. 
Tested on all the boxes (I found) with by assigning the CC1 (Modulation wheel) with virtual keyboard. 

Note: Does not add the MIDILearnMenuEntries to all vt_int (selection boxes). This requires more adaptations. 

Screenshot3


mkruselj:
As mentioned on Discord, this should really apply to all vt_int parameters, not just oscillator ones.

Issue #8243, opened 2025-12-28
Unicode path error on Windows when importing an SCL (and other files?)


arseniiv:
**Bug Description:**
SCL with Unicode in the file path get rejected.
Additionally I expect KBM import is affected as well, maybe preset file import and so on (no checks yet on that).

**Surge XT Version**
1.4.nightly.59b5323 but was in the latest nightly to date as well, and I presume is old.

**Version:**
* Plugin Type: standalone but I presume others as well
* Bitness: 64-bit

**Reproduction Steps:**
1. Take the example scale (deleting the `.txt` from the name): [₁₂êðö.scl.txt](https://github.com/user-attachments/files/24359659/edo.scl.txt).
2. Either import SCL from the menu or drop it on the UI.
3. Import error, with or without mangling the path depending on the way.

**Expected Behavior:**
No import error.

**Screenshots:**
The example file when dragged on the interface:
<img width="464" height="150" alt="mangled path is shown in the error message" src="https://github.com/user-attachments/assets/4a6b534f-3943-486b-92ca-2dc43a8f2d81" />

When opened from the tuning menu:
<img width="464" height="159" alt="path is shown as is but the error is still there" src="https://github.com/user-attachments/assets/8c0aec5f-605c-4e55-85b4-6604e1415440" />

**Computer Information (please complete the following!):**
 - OS: Windows 10 x64
 - Host: standalone but I presume it shouldn't matter here

**Additional Information:**
As expected, this doesn't depend on if there's Unicode in the file contents. (Checked both.)

Moreover, if the scale is actually malformed, this isn't reported, so the path bug should be happening first.

arseniiv:
This is the same as #8242 but slightly expanded.

mkruselj:
Could've expanded in #8242 then instead of opening a duplicate issue. 🙂

Closing the old one then, but please take note for the future!

arseniiv:
Yep I had a mental bug about how better to copy all of this and bailed lol.

baconpaul:
The problem here is that the tuning library isn't windows wchar aware.

It presents this API

```
inline Scale readSCLFile(std::string fname)
{
    std::ifstream inf;
    inf.open(fname);
    if (!inf.is_open())
    {
        std::string s = "Unable to open file '" + fname + "'";
        throw TuningError(s);
    }

    auto res = readSCLStream(inf);
    res.name = fname;
    return res;
}
```

so

1. That should obviously be const std:: string & but aside from that
[...]

PR #8246, opened 2025-12-30
Windows: Fix .SCL paths

mthierman:
#8243

### DIFF
--- a/src/common/SurgeStorage.cpp
+++ b/src/common/SurgeStorage.cpp
@@ -3142,7 +3142,7 @@ void SurgeStorage::loadTuningFromSCL(const fs::path &p)
 {
     try
     {
-        retuneToScale(Tunings::readSCLFile(p.u8string()));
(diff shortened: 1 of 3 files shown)


baconpaul:
This good to merge @mthierman ? Looks it to me but I didn't follow comment on discord about adding more tests.

Thanks!

mthierman:
It works on my machine, but it already worked on my machine (my Windows is UTF8) and I don't have other platforms to test

I didn't add tests to tuning-library, but I did fix the issues that made this PR not compile. I will go back and add some tests to the lib later. I think we can merge this and see if we get any more bug reports?

Later on we should do the same thing for .kbm, or switch to using `<filesystem>` everywhere

baconpaul:
Yeah agree. 

I'm going to open a surge issue to move to U22/GCC12 in docker but can't do that on vacation. That means with 10.15 we can move to filesystem everywhere.

PR #8248, opened 2026-01-02
Fix .KBM loading issues on Windows

mthierman:
Uses the same strategy as our fix for .SCL #8246 

We accidentally pulled in filesystem with the last PR, this removes it.


src/surge-xt/gui/SurgeGUIEditorMenuStructures.cpp
```diff
@@ -53,7 +53,6 @@
 #include "SurgeSharedBinary.h"
 #endif
 
-#include <filesystem>
```
mthierman: woops

Issue #8251, opened 2026-01-10
Feature request: new 2 octave VKB mapping alternative


henrikvilhelmberglund:
**Is your feature request related to a problem? Please describe!**
Right now on Windows there's a JUCE bug which prevents DAW note input from triggering correctly on mousedown. I have no idea how to fix this so instead I thought I could make the VKB do what I want.

This feature request is for a new mapping which allows for 2 octaves, `Z` to `/` and `Q` to `]`.

I'm not interested in the keyboard shortcuts like changing octaves/velocity so I wanted to implement a 2 octave VKB like the default FL typing keyboard.

**Describe the solution you'd like:**

JUCE is a bit silly and doesn't allow for two keys to have the same note but this is what I would like to have:

```C
        // Row 1: Z row (Lower Octave + extra notes)
        // Keys: Z S X D C V G B H N J M , L . ; /
        // Map:  C C# D D# E F F# G G# A A# B C C# D D# E
        std::vector<std::pair<int, int>> lowerKeys = {
            {'z', 0},  {'s', 1},  {'x', 2},  {'d', 3},  {'c', 4},  {'v', 5},  {'g', 6},
            {'b', 7},  {'h', 8},  {'n', 9},  {'j', 10}, {'m', 11}, 
            {188, 12}, {'l', 13}, {190, 14}, {186, 15}, {191, 16}
        };

        // Row 2: Q row (Upper Octave + extra notes)
        // Keys: Q 2 W 3 E R 5 T 6 Y 7 U I 9 O 0 P [ = ]
        // Map:  C C# D D# E F F# G G# A A# B C C# D D# E F F# G
        std::vector<std::pair<int, int>> upperKeys = {
            {'q', 12}, {'2', 13}, {'w', 14}, {'3', 15}, {'e', 16}, {'r', 17}, {'5', 18},
            {'t', 19}, {'6', 20}, {'y', 21}, {'7', 22}, {'u', 23}, {'i', 24}, {'9', 25},
            {'o', 26}, {'0', 27}, {'p', 28}, {'[', 29}, {'=', 30}, {']', 31}
        };

        // Loop through and apply
        for (auto& k : lowerKeys) keyboard->setKeyPressForNote(juce::KeyPress(k.first), k.second);
        for (auto& k : upperKeys) keyboard->setKeyPressForNote(juce::KeyPress(k.first), k.second);
```

[...]

PR #8257, opened 2026-01-12
feat: add QWERTY 2 Octave VKB layout

henrikvilhelmberglund:
This PR adds a new QWERTY 2 Octave which allows for playing 2 octaves (and a bit more) with the lower octave starting at `Z` and the higher octave starting at `Q`.

JUCE doesn't allow for multiple keys representing the same note so to "fill in the gaps" and be able to play `Z` to `/` the DAW VKB can be set to the same key and octave.

When using keyboard shortcuts there are conflicts with for example Raise/lower octave on `C` and `X`, these should be set to for example `Shift+C` and `Shift+X` when using this layout. Or just disabled since they are probably not needed with this layout.

Also `]` seems pretty hard coded, I didn't see that in the keyboard shortcuts, but the other layouts use this as well.

Feel free to make changes.

Fixes https://github.com/surge-synthesizer/surge/issues/8251

### DIFF
--- a/src/surge-xt/gui/overlays/KeyBindingsOverlay.cpp
+++ b/src/surge-xt/gui/overlays/KeyBindingsOverlay.cpp
@@ -362,11 +362,12 @@ void KeyBindingsOverlay::changeVKBLayout(const std::string layout)
 {
     if (layout == "QWERTY 2 Octave")
     {
-        juce::AlertWindow::AlertWindow::showMessageBoxAsync(juce::AlertWindow::WarningIcon,
-        "QWERTY 2 Octave Layout Warning",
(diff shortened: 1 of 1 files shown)


mkruselj:
I'd argue that even if you have two octaves you might still want to shift octaves up or down (I guess a good place would be tilde/1 above Q).

A nice UX improvement would be to pop up a dialog warning about a conflict if this layout is selected.

henrikvilhelmberglund:
Added AlertWindow that shows when you select the layout.

mkruselj:
OK, that's a good start. Ideally it should actually check if there ARE any conflicts and only pop a dialog if they exist, not always. But this can be done in a separate PR.

PR #8260, opened 2026-01-17
Remove hard coded VKB velocity shortcuts

henrikvilhelmberglund:
This PR removes the hard coded VKB velocity shortcuts which were deemed unnecessary.

The hard coded shortcuts were mostly fine but could cause problems, for example in the 2 Octave QWERTY VKB layout you would set velocity when playing the upper octave, and if you wanted to use the 1-8 keys for some other shortcuts instead you couldn't.

Feel free to make changes.


mkruselj:
Let's modify this so that we only have velocity down/up keybinds and they allow us jumping in steps of 16 (rather than 10%). I don't think it's worth it to have a separate keybinding for each of the different velocity layers, honestly.

henrikvilhelmberglund:
This PR was made only to replicate the current behavior. I think it would be a bit annoying to test different velocities with only a keyboard if there were only 2 keybinds.

But what I would really want ideally is either a velocity slider in the GUI for the VKB (next to the modwheel or something) or being able to set the velocity by clicking on the keys, for example clicking around the bottom of the key: `127`, clicking at the top of the key: `5`. In that case all of the hotkeys wouldn't be needed and the two keybinds would work better.

mkruselj:
The GUI is pretty cramped in that area already, frankly... I am not sure if we can modify the JUCE keyboard widget to modify velocity by clicking on a key, maybe this can be done in mouseDownOnKey() callback.

At any rate I think we'd be good if we just had the +/- keybinds instead of 9. Ain't nobody gonna modify all those as they change the VKB layouts...

henrikvilhelmberglund:
> The GUI is pretty cramped in that area already, frankly...

Really? Imo it's huge
image

Adding a new slider would only remove one key or so. 

> I am not sure if we can modify the JUCE keyboard widget to modify velocity by clicking on a key, maybe this can be done in mouseDownOnKey() callback.

I noticed now that the JUCE MidiKeyboardComponent already does this when just clicking on keys, this value would just need to be saved as the new velocity value (behind a new setting I guess).

mkruselj:
Hmmm yeah actually I'm not sure if that's good UX. It's fine that the VKB is already using different velocities depending on where you click (I totally forgot that it did that!), so I don't think clicking on VKB should affect how the QWERTY keypresses work. Two different worlds, in a way.

So yeah, I maintain that we're good without individual keybinds for specific velocity levels.

henrikvilhelmberglund:
I think the alternative UX is worse.

Switch from `127` to `64` example:
- 2 keybinds: `key key key key`
- Slider: `drag from top to middle`
- VKB click: `click in middle of key`

And this becomes worse when you want even more drastic changes.

For me if the VKB click changed the overall velocity of the VKB I wouldn't need any keybind at all. And it doesn't need to be a default anyway, this is very much a personal preference where some want it and some don't want it.

I like the idea of removing the number keybinds so something more useful can potentially be put there instead but I would also like an alternative way of setting the overall velocity of the VKB. This PR can be focused on changing to 2 keybinds though.

mkruselj:
In practice it's not that bad. Had dealt with plenty of VKBs that only had a single static setting and you had to type in a value manually. So the two inc/dec keybinds are gonna be just fine.

henrikvilhelmberglund:
There are already keys for increasing/decreasing velocity by 10%, so should the numbered keys just be removed instead? I did this for now. Changing by 16 would be 12.6% ish percent instead so there's not a huge difference.

mkruselj:
That was exactly my suggestion, keep the 10% keybinds but remove numbered keybinds.

Issue #8252, opened 2026-01-10
Factory data from built Surge ends up in SurgeXTData folder which can't be used


henrikvilhelmberglund:
**Bug Description:**
Factory data from built Surge ends up in SurgeXTData folder which can't be used. Finding the actual location where this should be was a bit confusing, "Open factory patches folder..." does nothing if the folder doesn't exist. Maybe it would be more useful if `C:\ProgramData\Surge XT` (on windows) was opened instead.

Maybe the folder should have been called `Surge XT` instead to make it a bit more obvious that the contents should go into the `C:\ProgramData\Surge XT` folder.

Or maybe this should just happen automatically after building.

**Surge XT Version**
This information is found on the About screen, which you get to from the bottom right menu.

https://github.com/surge-synthesizer/surge/commit/c1589dc1475fdca3376ab6c748bfb589dcc6c316

64bin vst3

**Reproduction Steps:**
1. Build Surge XT.
2. Click on patches at the top.
3. No patches are shown.
4. Click Open factory patches folder...
5. No folder is shown.

What you really have to do is to copy the contents of the SurgeXTData folder into the Surge XT folder which begs the question of why they have different names.

**Expected Behavior:**
Make it less confusing by either showing the empty Surge XT ProgramData folder or automatically move the factory data into this folder when building for the first time.

**Screenshots:**
N/A

**Computer Information (please complete the following!):**
Win 10 on FL latest (but same everywhere)

**Additional Information:**
N/A

baconpaul:
So o windows we can't install to the writable area without elevated permissions which makes development there kind of a pain yeah. Most folks run the installer once and then it all just works or use the environment variable or setup a link per the search path in the manual.

there is no more difficult part of shipping software to desktops than putting files in places users want. You can read the insane (except for on macOS where it is merely a bit confusing) shenanigans we go to in SurgeStorage for the various waterfalls. But for the dev workflow just set one of the env vars in SurgeStorage or run the installer and you'll be OK.

mthierman:
Yes, if you enable copy-by-default it will fail the build because of the elevated permission requirement

I have a production copy installed on my machine so I don't really want those contents overriden by my dev build either

baconpaul:
Yeah get that 

The env var works if you need that in your dev env tho

henrikvilhelmberglund:
[...]

PR #8261, opened 2026-01-17
feat: show warning message when running Open factory patches without …

henrikvilhelmberglund:
…a factory patch folder.

This PR shows an `AlertWindow` when `Open factory patches` is run without an actual factory patches folder, or when `openFileOrFolder()` fails for other files/folders.

Edit: Looks like this now:
surgefactory2

Feel free to make changes.

fixes https://github.com/surge-synthesizer/surge/issues/8252

### DIFF
--- a/src/surge-xt/gui/SurgeGUIEditorMenuStructures.cpp
+++ b/src/surge-xt/gui/SurgeGUIEditorMenuStructures.cpp
@@ -639,7 +639,15 @@ juce::PopupMenu SurgeGUIEditor::makeTuningMenu(const juce::Point<int> &where, bo
 
         tuningSubMenu.addItem(Surge::GUI::toOSCase("Factory Tuning Library..."), [this]() {
             auto path = this->synth->storage.datapath / "tuning_library";
-            Surge::GUI::openFileOrFolder(path);
+            if (!Surge::GUI::openFileOrFolder(path))
+            {
+                synth->storage.reportError(
+                    "The folder " + path_to_string(path) +
+                        " does not exist. It seems that your Surge factory data folder is "
+                        "missing.\n\n"
+                        "Please run the Surge installer to set up the factory data.",
+                    "Factory Tuning Library Folder Not Found");
+            }
         });
 
         tuningSubMenu.addSeparator();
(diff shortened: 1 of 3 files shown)


baconpaul:
Ahh I don’t love hardcoding patches factory here rather than at the calling site - this way if we want an alert for other reveals this util gets super big

perhaps the trick is to have open return a bool and return false if the path is missing, mark it [[nodiscard]] optionally, and have the client sites pop an error

that would also let the client sites us the surge gui editor display error message which is a more consistent error ui

henrikvilhelmberglund:
I'm in the no premature optimization camp. The only problem right now is that nothing happens when you try to open the factory patches folder (or other folders), this PR handles both cases with

1. A specific warning for missing factory patches folder saying "run the installer"
2. All other folders saying that it wasn't found and showing the path

I guess it could use `patches` instead of `patches_factory` to also show for the other patch data folders though.

It's true that if there are others it gets bigger, but right now there aren't any that I know so it would be a solution for a problem that doesn't exist.

But if you want to do it in another way feel free to change it. I haven't actually seen the Surge GUI editor display error message so I have no idea how it looks like.

baconpaul:
Code reviews to keep stuff in the right place aren’t really premature optimization, they are keeping the code base clean. 

Surge gui edit has a display error message which we use for all other paths and the fact that this didn’t use it is what made me realize the diff was in the wrong place (since this low in the utils stack you don’t have editor access). Which is why I suggested making it the problem of the code attempting to open to handle the non existent case 

but if you don’t want to change it that’s fine! One of us will before we merge. Or we won’t and the pr will stay open until it changes!

henrikvilhelmberglund:
Makes sense, I can probably do something better, marking as draft for now.

baconpaul:
Yeah when I’m not on my phone I’ll add a clearer note than ramble above

if we do it right the point that opens can be if ! …. Error with a lot less code and a lot more locality

henrikvilhelmberglund:
Some Git things happened but should be back in business now.

I changed the function to return a bool with false if the file/folder doesn't exist and added if statements around the `openFileOrFolder()` calls that looked like they might fail.

Not sure if this is more what you had in mind.

I'm not really a C++ programmer so input is appreciated.

baconpaul:
That is exactly what I had in mind - thank you!

baconpaul:
If you are ready for this to merge please toggle back from draft and one of @mkruselj or I can do so. Great stuff and thanks for responding to a review and contributing!

henrikvilhelmberglund:
> If you are ready for this to merge please toggle back from draft and one of @mkruselj or I can do so. Great stuff and thanks for responding to a review and contributing!

Thanks for the help!

PR #8272, opened 2026-01-30
Mark convolution as unimplemented in SurgeFX.

mx:
Selecting it will now do nothing, not even clear the previous FX.

### DIFF
--- a/src/surge-fx/SurgeFXProcessor.cpp
+++ b/src/surge-fx/SurgeFXProcessor.cpp
@@ -84,7 +84,13 @@ SurgefxAudioProcessor::SurgefxAudioProcessor()
     fxType->getTextHandler = [this](float f, int len) -> juce::String {
         auto i = 1 + (int)round(f * (n_fx_types - 2));
         if (i >= 1 && i < n_fx_types)
+        {
+            if (i == fxt_convolution)
+            {
+                return "Convolution (Unimplemented)";
+            }
             return fx_type_names[i];
+        }
         return "";
     };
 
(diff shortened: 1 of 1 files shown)


baconpaul:
Did you see my comment that the VST3 param range needs adjusting also on the issue?

mx:
Yeah I saw it, that's why I wanted you to merge it instead of me. This doesn't remove it from the UI at all, or require us to make an int-to-int map and keep it updated. Instead it just tells the user it's nonfunctional, but lets us implement the functionality in the future if we want to. It's kind of "lazy" in that regard, but less ongoing maintenance.

baconpaul:
yeah ok. i guess we can change the display name the param to Convolution (not available) also or some such.

The change looks smart by the way and we should merge it. Just i think it doesn't quite close the issue opened.

mx:
Hmm I thought I did that with line 90 and it already skipping in the menu, did I miss it somewhere else?

mx:
Just looked in Reaper. It's still outside of the main UI. The parameter view lets you select it as `Convolution (Unimplemented)` as expected. It puts the main UI back to the last selected FX if you do (if you use the slider, it'll always be Floaty Delay). The parameter view will also be for that, though it'll still read as Convolution (Unimplemented).

baconpaul:
Yeah I wonder if we just want to not have it in slider range. 

But let’s merge this anyway

mkruselj:
I'd make it say (N/A) instead of (Unimplemented) for brevity. Another PR.

PR #8290, opened 2026-02-26
Various workflow & installer fixes

mthierman:
- Fix a compression bug for Windows Installer that caused failures due to high memory requirements
- Bump checkout action to v6
- Use ubuntu-slim for Clang Format Check and update fmt to v12


mthierman:
fmt 12.1.0 was throwing a warning that we treat as errors because of WX, only a W1 level warning but WX causes a build failure. Downgraded to 12.0.0 for now.

PR #8291, opened 2026-03-05
Fix favorites virtual folder not respecting patch prev/next

### LINKED ISSUE #6435: Favorites virtual folder doesn't respect patch prev/next
1. Select a favorite patch
2. Press prev/next patch either on UI or via keyboard shortcuts

Result: Jog happens within the folder of that favorite patch, instead of within Favorites folder.

Not sure if there's an easy solution to this, but maybe:

1. If the preset was loaded from the Favorites menu, or from Favorites virtual folder in patch browser, make sure prev-next patch is only going through that virtual folder (of course, obeying the "Previous/Next patch constrained to current category" option)
2. If we're just browsing through the patch browser, search patch or prev/next category/patch, don't go through the Favorites virtual folder if you stumble upon a favorite patch.

y-hbb:
## Summary

Fixes #6435

- When a patch is loaded from the Favorites menu, prev/next patch navigation now cycles through favorites instead of the patch's original category
- Added `favoritesOrdering` vector to `SurgeStorage` to maintain a sorted list of favorite patch indices, populated at startup and kept in sync when favorites are toggled
- Added `patchSelectedFromFavorites` flag to track navigation context, cleared on category jog, random patch, and MIDI program change
- Refactored `optionallyAddFavorites()` to use the shared `favoritesOrdering` vector instead of ad-hoc scan+sort

## Test plan

- [x] Mark 3+ patches from different categories as favorites → load from Favorites menu → press Next/Prev → cycles through favorites only, wraps around
- [x] Single favorite → load from Favorites → press Next/Prev → stays on same patch
- [x] Load from Favorites → unfavorite current patch → press Next/Prev → falls back to normal category navigation
- [x] Load from Favorites → press Next Category → press Next/Prev → navigates within new category (not favorites)
- [x] Load from Favorites → Random Patch → press Next/Prev → navigates within random patch's category
- [x] Add a new favorite while browsing favorites → press Next/Prev → new favorite appears in navigation
- [x] Remove all favorites → open Favorites menu → menu doesn't appear / no crash
- [x] Load from normal category menu → press Next/Prev → normal navigation unchanged
- [x] MIDI program change → press Next/Prev → navigates within new patch's category

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -61,3 +61,4 @@ Vincent Zauhar <v.zauhar@bell.net>
 Alexandr Zyurkalov <Alexander.Zyurkalov@gmail.com>
 簡志瑋 <dppss92132@gmail.com>
 Daniel Hatadi <https://github.com/dhatadi>
+Yashas Hebbarabailu <https://github.com/y-hbb>
--- a/src/surge-xt/gui/SurgeGUIEditor.cpp
+++ b/src/surge-xt/gui/SurgeGUIEditor.cpp
@@ -6185,7 +6185,7 @@ void SurgeGUIEditor::setSpecificPatchAsFavorite(int patchid, bool b)
             favs.push_back(patchid);
             std::sort(favs.begin(), favs.end(), [this](const int &a, const int &b) {
                 return strnatcasecmp(synth->storage.patch_list[a].name.c_str(),
-                                    synth->storage.patch_list[b].name.c_str()) < 0;
+                                     synth->storage.patch_list[b].name.c_str()) < 0;
             });
         }
         else
--- a/src/surge-xt/gui/widgets/PatchSelector.cpp
+++ b/src/surge-xt/gui/widgets/PatchSelector.cpp
@@ -1257,10 +1257,7 @@ bool PatchSelector::populatePatchMenuForCategory(int c, juce::PopupMenu &context
     return amIChecked;
 }
 
-void PatchSelector::loadPatch(int id)
-{
-    loadPatch(id, false);
-}
+void PatchSelector::loadPatch(int id) { loadPatch(id, false); }
 
 void PatchSelector::loadPatch(int id, bool fromFavorites)
 {


mkruselj:
Ahh, clang-format check didn't pass!

baconpaul:
tied up today and indeed need to fix formatting
but i have time to review tomorrow

(@y-hbb just fyi we're super happy taking AI-assisted PRs and changes; I use the tools all the time. As with most other team using these tools well, we've kept human review in the loop for changes still. Also you may want to let your claude know to obey our clang-format!).

baconpaul:
Read the code on my commute and my only thought is did the osc path still go through load patch? I think it does but worth eyeballing 

but happy to merge if we fix format and make that change 

oh and @y-hbb you should add yourself to the AUTHORS file

y-hbb:
Thanks for the review! Re: the OSC path question —

OSC patch commands (`/patch/incr`, `/patch/decr`, `/patch/random`) call `synth->jogPatch()`, `synth->selectRandomPatch()`, and `synth->jogCategory()` directly on `SurgeSynthesizer`, bypassing `PatchSelector::loadPatch()`. So OSC can never *set* `patchSelectedFromFavorites = true` — it can only inherit a `true` from a prior GUI action, or clear it.

- `jogPatch` — reads the flag, navigates favorites if set (preserves user's last context regardless of input method)
- `selectRandomPatch` — clears the flag
- `jogCategory` — clears the flag
- `programChange` (MIDI) — clears the flag

Formatting is also passing now on the latest CI run.

(Claude analysis ^)

y-hbb:
Let me know if I need to test any specific flow. I am able test using standalone app. I am not sure how I will test with MIDI and OSC.

y-hbb:
I was able to test using OSC. Like claude analysis say, OSC commands doesn't obey favorites flag. The flag gets cleared and it goes through the current category of patches.

baconpaul:
Yeah and program change does that same thing

so I’m happy here! Thanks!

y-hbb:
Was able to test MIDI program change as well. Work like OSC. It ignores the favorite flag.

PR #8299, opened 2026-03-14
Fix Formula receiving unmodulated EG params on first block

### LINKED ISSUE #8298: Formula modulator envelope values returns 0 on first process call
**Bug Description:**
When using the envelope with use_envelope set to false, and you modulating the slider with an MSEG. The value does not appear in the first process call.  

I'm guessing it has to do with the execution order? 

**Surge XT Version**
Latest Nightly build

* Version:
VST, Windows 11

**Reproduction Steps:**
1. Modulate the delay param in the envelope of the formula modulator 
2.  use this code 
```lua
function init(state)
    state.use_envelope = false
    state.counter = 0
    state.value = state.delay
    return state
end

function process(state)
    
    if(state.counter<1) then
        state.value = state.delay
        state.counter = state.counter+1
    end
    
    state.output = state.value

    return state
end
```
3. The value of state.delay is zero until the second process call.

nuoun:
Fixes #8298


nuoun:
Can be reviewed.

baconpaul:
Nice i'll look this weekend.

Roughly how did you test it? Thanks!

nuoun:
Using this script for debugging with another LFO modulating the LFO EG Delay param (and also the tutorial patches by ear)

```lua
function init(state)
    state.use_envelope = false
    state.counter = 0
    print("init", state.delay)
    return state
end

function process(state)
    if state.counter == 0 then
        print("first block", state.delay)
    elseif state.counter == 1 then
        print("second block", state.delay)
    end
    state.counter = state.counter + 1
    state.output = state.delay
    return state
end
```

nuoun:
Two more fixes, now Formula always retriggers gates again even when !modsource_doprocess in LFO1 and state.phase is no longer nil in init().

mkruselj:
@baconpaul This one lgtm but I wouldn't want to merge it before you had your eyes on it.

baconpaul:
I don't understand the i==0. Why is 0 special? We either need a heaping big comment or to remove that bit.

nuoun:
Moving this back to here from Discord, so LFO1 is special since it always runs regardless of `bool SurgeSceneStorage::modsource_doprocess[n_modsources]` to process retriggers, this is mentioned in the code's comments (which I can make big and honking) but not in the manual. This seems arbitrary and maybe worth opening a new issue for. Could we have something like `:modsource_doprocess` just for retriggers instead?


image

The bigger issue, I messed up and this block here at [SurgeVoice.cpp#L748](https://github.com/nuoun/surge/blob/1734b555c276e242120eae7f5b15c917c316c25b/src/common/dsp/SurgeVoice.cpp#L748) now comes before Formula, meaning that retriggers come a block late. Might be able to move Formula processing before that, not sure but that would require testing (and does not seem any if the DSP tests catch this). So I hope it does not come down to choosing if modulation comes a block late or if retriggers come a block late.

baconpaul:
> 


> The bigger issue, I messed up and this block here at [SurgeVoice.cpp#L748](https://github.com/nuoun/surge/blob/1734b555c276e242120eae7f5b15c917c316c25b/src/common/dsp/SurgeVoice.cpp#L748) now comes before Formula, meaning that retriggers come a block late. Might be able to move Formula processing before that, not sure but that would require testing (and does not seem any if the DSP tests catch this). So I hope it does not come down to choosing if modulation comes a block late or if retriggers come a block late.

but it will

It always does

There has to be an order or we have to topologically sort per patch and even then we have to have a cycle break

we shouldn’t change which one is late

nuoun:
You're right of course, I messed up testing as well. The proper way test I think is to have Formula modulate a square oscillator's mixer from full to -192 dB, and this now results in an audible pop meaning I moved the problem from not modulating the first block to not outputting the first block. Sorry, I don't see an easy fix and I'm closing this.

PR #8305, opened 2026-03-19
Lua editor double click behaviour

blancoberg:
Triple click ( selects line )
Removed line break character from the end of the line selection. 
Quadruple click:
Now selects the entire text/code. Helps a lot with importing if you are coding in an external editor


mkruselj:
Ctrl+A >>> quadruple click, tho 😛

blancoberg:
> Ctrl+A >>> quadruple click, tho 😛
😑well now we have a less superior alternative as well

PR #8306, opened 2026-03-20
Prelude changes

nuoun:
- Adds math.linear_to_dB, math.dB_to_linear, math.full, math.ones, mod.pb_to_semitones and mod.cosine_fade
- Moves math.clamp / math.limit_range from custom function to Prelude
- Adds WTSE math functions also to Formula


mkruselj:
I would perhaps shorten linear to lin and semitones to st...?

PR #8307, opened 2026-03-20
Rust bindings for Surge XT.

CattoByte:
this PR adds rust bindings to the synthesizer, by means of a submodule. you can also build the tests
the rust bindings themselves can be found at https://github.com/CattoByte/surge-rs. basic usage instructions are found there.

this PR is a draft one because some things still have to be done, like editing the README file, attempting to fix LTO and at the very least getting the bindings themselves to load patches. and lua—that's important.

the only windows machine i have is a 32-bit and i don't even know where it is, so testing on that system would be appreciated.

appended is a video showing the changes in action:
https://github.com/user-attachments/assets/1ad05039-9c6d-43d6-b345-5c6ab4025491

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -67,4 +67,4 @@
 	url = https://github.com/facebook/zstd.git
 [submodule "src/surge-rs/surge-rs"]
 	path = src/surge-rs/surge-rs
-	url = https://github.com/CattoByte/surge-rs
+	url = https://github.com/surge-synthesizer/surge-rs


baconpaul:
Primary comment is: we don’t really have out of main repo sub modules to individuals. If you decide to delete or what not project lacks continuity 

the solution to that is multiple including making surge-rs live in surge-synthesizer project tree which we could do happily or using fetch content when optionally building rust, but struck me on first review

CattoByte:
hi, paul. thanks for reviewing.

i wouldn't mind moving the repository somewhere else; however, i don't think it's a good idea to put it inside the tree.
in short, it's:
- because doing that would make it hard for someone to compile just the rust bindings (aka use `cargo`'s tooling);
- and because it acts a bit like a parasite to the surge tree, disabling things like lto and lua, which you might want for a "full build". it also forces cmake to export its commands.

(having it out of tree doesn't fix the second point, but it gives you a way out.)

basically, the best and fastest way to build `surge-rs` is to compile a second, slightly lobotomized surge.
in-tree support is for anything outside cargo, because it's nice to have everything working under the main tree.

that aside, i think fetching content could work. i'm assuming it's using `cmake`'s own functions. though, still, from my understanding, that wouldn't fix the "delete or whatnot" issue.
what do you think about moving the repository into the `surge-synthesizer` group instead? it would solve the individual's submodule issue while keeping it clean on the rust side of things.

mkruselj:
I don't think any of us would have a problem if you moved surge-rs as a new project under surge-synthesizer, really. In fact, that is probably the best way to go. I say YEA.

baconpaul:
Yup thats done

Just update this PR with the right submodule and we can merge

CattoByte:
all done!

PR #8311, opened 2026-03-22
Refactor WTSE hamburger menu

nuoun:
- Renames MenuButtonFromIcon to MenuButtonFromAsset 
- Refactors to use skin asset SVGs, supporting both single frame with separate hover image and two frame stacked assets
- Sets more descriptive menu title for screen reader
- Cleans up old WTSE menu icon


nuoun:
Looks like I could have handled the file renames a bit better, oh well. Can be reviewed.

https://github.com/user-attachments/assets/f0c78f5b-9ed9-4e47-816b-04aaa0d09bbd

PR #8327, opened 2026-03-30
Minor template tweaks

nuoun:
Sets unused oscillators’ Unison Detune values to a default of 10% across all Audio In templates
Sets all unused oscillators in Init Square to use the Classic oscillator


nuoun:
Change of plans, Rolled back Init Square and Init Saw now also uses Modern for its unused oscillators to showcase the different osc types

Edit: That broke the testrunner, closing.

PR #8333, opened 2026-04-07
Upgrade sst-effects and sst-basic-blocks

baconpaul:
which are required to build on latest macos clang


baconpaul:
I want to test this a smidge more before I merge

PR #8339, opened 2026-04-11
Correctly interpolate MPE-bends between keys under MTS-ESP

Andreya-Autumn:


src/common/dsp/SurgeVoice.cpp
```diff
@@ -62,12 +62,17 @@ float SurgeVoiceState::getPitch(SurgeStorage *storage)
...
             keyRetuningForKey = key;
-            keyRetuning = MTS_RetuningInSemitones(storage->oddsound_mts_client, key + mpeBend,
-                                                  mtsUseChannelWhenRetuning ? channel : -1);
+            if (mpeEnabled)
+            {
+                keyRetuning = MTS_RetuningForBentKey(storage, res);
+            }
+            else
+            {
+                keyRetuning = MTS_RetuningInSemitones(storage->oddsound_mts_client, key + mpeBend,
```
baconpaul: this key + mpeBend really is wrong. I know it's knocked out in ape mode now but still I think we should remove it. The API requires a short. That's the one thing convincing about the claude search to me.

src/common/dsp/SurgeVoice.cpp
```diff
@@ -85,6 +90,19 @@ float SurgeVoiceState::getPitch(SurgeStorage *storage)
     return res;
 }
 
+#ifndef SURGE_SKIP_ODDSOUND_MTS
+float SurgeVoiceState::MTS_RetuningForBentKey(SurgeStorage *storage, float bentKey)
+{
+    auto ik = (int)(bentKey);
```
baconpaul: I think this should be std::floor(bentKey) to avoid any ambiguity. Also `(int)-1.1 = -1` whereas `std::flor(-1.1) == -2` so if we ever accidentally get a negative, this will be wrong.

baconpaul (commented):
Two tiny comments

src/common/dsp/SurgeVoice.cpp
Andreya-Autumn: Oh oops, well this is in the !mpeEnabled case anyway so that's just a copy/paste error I think. Fixing.

src/common/dsp/SurgeVoice.cpp
Andreya-Autumn: Oki got it.

PR #8342, opened 2026-04-12
Convolution FX IR prev/next buttons

nuoun:


nuoun:
Adds Wavetable selector style prev/next buttons to the IR selection box plus a tooltip on hovering a truncated filename:
image

PR #8344, opened 2026-04-12
Fix the min-6-sample delay bug in convolution.

mx:
Somehow deleted the branch on github by accident which closed the PR...

### DIFF
--- a/src/common/dsp/effects/ConvolutionEffect.cpp
+++ b/src/common/dsp/effects/ConvolutionEffect.cpp
@@ -21,10 +21,12 @@
  */
 #include "ConvolutionEffect.h"
 #include <algorithm>
+#include <cmath>
 #include <iterator>
 #include <numeric>
 
 #include <CDSPResampler.h>
+#include "sst/basic-blocks/tables/SincTableProvider.h"
 
 namespace
 {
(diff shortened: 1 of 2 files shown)


mkruselj:
This looks good to me. @Andreya-Autumn, wdyt?

src/common/dsp/effects/ConvolutionEffect.cpp
```diff
@@ -260,12 +265,15 @@ void ConvolutionEffect::process(float *dataL, float *dataR)
     }
 
     // Cut out some annoying non-SSEable moves if we can.
-    if (delayTime_.v != 0)
+    if (delayTime_.v != 0 || delay_latched_)
     {
+        const float delay_samples = std::max(
+            delayTime_.v,
+            static_cast<float>(sst::basic_blocks::tables::SurgeSincTableProvider::FIRipol_N) / 2);
```
baconpaul, line 272: should this have a +1 for write head? Like you read before you write so are you sure you don't pick up a specious old data point on the tail at 6 vs 7

Andreya-Autumn:
I built this now. It sounds good, and I believe the clamp to be correct. One super small optimization I can think of @mx is FIRipol_n is a static constexpr right? So maybe our minimum here can be as well, so the float cast and divide don't need to happen every block. Otherwise seems good to me.

PR #8351, opened 2026-04-18
Refactor snapshot serialization

nuoun:
Draft please don't merge yet.

- Store snapshots as binn frame lists instead of blobs with separate metadata keys
- Unify .wtscript snapshot loading into a shared helper matching load_arbitrary_block_storage
- Add missing bounds checks for FX/oscillator indices

Edit: This is a breaking change on how snapshots are saved in patch and .wtscript files.


nuoun:
This is a breaking change on how WTS snapshots are saved in patch and .wtscript files (which I should have mentioned in the above PR message).

This is done and can be merged.

PR #8362, opened 2026-04-22
Add conflict detection when assigning keyboard bindings

mkruselj:
This was a doozy! The conflict detection itself was easy as it was already done for VKB layouts in a prior commit.

But I constantly ended up losing focus from the keybinds overlay! The Learn button would remain lit after dismissing the conflict notification dialog, but since focus got lost, the next keypress (chorded or not) would not work - they would instead bring focus back to keybind overlay. So then you needed to press it once again.

After a long and tedious ordeal of trying various methods (global mouse listener, global component listener, timer shenanigans, focus change listener...) none of which worked, I broke down and simply went back to the main hub: SGE idle(), and did it there. Not sure if this is something @baconpaul will agree with, but I lost 5 hours attempting to make this work, and this worked!

Thoughts welcome!

### DIFF
--- a/src/surge-xt/gui/SurgeGUIEditor.cpp
+++ b/src/surge-xt/gui/SurgeGUIEditor.cpp
@@ -54,6 +54,7 @@
 #include "overlays/MSEGEditor.h"
 #include "overlays/LuaEditors.h"
 #include "overlays/ModulationEditor.h"
+#include "overlays/KeyBindingsOverlay.h"
 #include "overlays/TypeinParamEditor.h"
 #include "overlays/Oscilloscope.h"
 #include "overlays/OverlayWrapper.h"
@@ -587,6 +588,7 @@ void SurgeGUIEditor::idle()
 #ifndef SURGE_SKIP_ODDSOUND_MTS
     getStorage()->send_tuning_update();
 #endif
+
     if (editor_open && frame && !synth->halt_engine)
     {
         bool patchChanged = false;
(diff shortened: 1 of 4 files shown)


src/surge-xt/gui/overlays/KeyBindingsOverlay.cpp
```diff
@@ -478,36 +478,94 @@ void KeyBindingsOverlay::onSkinChanged()
 
 bool KeyBindingsOverlay::keyPressed(const juce::KeyPress &key)
 {
+    using namespace Surge::GUI;
+
     if (isLearning)
     {
-        auto &binding = editor->keyMapManager->bindings[(Surge::GUI::KeyboardActions)learnAction];
+        auto newBinding = editor->keyMapManager->bindings[(KeyboardActions)learnAction];
```
baconpaul, line 486: Why did you remove a & here?

baconpaul: Oh never mind i see. Add a comment that this is explicitly a copy in case we reject then

src/surge-xt/gui/SurgeGUIEditor.cpp
```diff
@@ -642,6 +644,21 @@ void SurgeGUIEditor::idle()
             }
         }
 
+        if (kboNeedsRefocus)
```
baconpaul: Oh I see what you are doing. And this isn't a great way to do it since if we want it for other screens we get a million of these ifs.

So what I would do instead is

1. Add a `juce::Component::SafePointer<juce::Component *> refocusIfNoAlert{nullptr}` member
2. When you set kboNeedsRefocus true instead set refocusIfNoAlsert = juce::Component::SafePointer(this)
3. Then rather than kbo = lookup you just use that variable
4. And set it to mull where you used to set kboNeedsFocus to false

same logic but general to any overlay

baconpaul (dismissed):

src/surge-xt/gui/SurgeGUIEditor.cpp
mkruselj: I think I get it, sounds good!

mkruselj: I got it! Works brilliantly and it's even cleaner - there's no need for restoreFocus() method in KBO then!

PR #8365, opened 2026-04-23
Fix tuning of Twist oscillator

### LINKED ISSUE #8049: Twist oscillator is out of sync from other oscillator types
**Bug Description:**
When using the Twist oscillator together with any other type of oscillator, the Twist oscillator oscillates slightly faster, such that about every second it completes a full phase shift compared to the other

**Surge XT Version**
Version: Surge XT 1.3.4.f7b97c6
Build Info: 2024-08-11 @ 16:02:24 on 'fv-az292-728/pipeline' with 'MSVC-19.40.33813.0' using JUCE 7.0.12
System Info: Windows 64-bit Standalone on AMD Ryzen 7 5800H with Radeon Graphics         , 15 GB RAM
Sample Rate: 48000 Hz
Processing Block: 32 samples

**Reproduction Steps:**
1. Keep the default saw oscillator in slot 1.
2. Enable another oscillator slot, and set it to Twist.
3. Set 'Square Shape' to -100 for easier comparison.
4. Enable 'Retrigger' on  both oscillators.
5. Play a note. you should be hearing a beating as the the oscillators go in and out of phase with eachother, or you can see it in the waveform oscilloscope.

**Expected Behavior:**
The two oscillators should be playing in sync and producing one unified tone.

**Screenshots:**
https://github.com/user-attachments/assets/3262bb16-4b61-4953-8559-5fafcad39f3f

 - OS: Windows 10
 - Host: Standalone, Reaper
 - Version: Reaper v7.41

**Additional Information:**
I apologize if this is intended behavior, i couldn't find any info about this and found it weird. 
Its quite problematic when trying to combine the twist oscillator with any other types, causing unwanted beating. This seems to affect all different twist modes.
I managed to almost fix this by setting the twist pitch to -0.04 but the pitch control isn't precise enough to fully get rid of it.

mkruselj:
Fixes #8049

### DIFF
--- a/src/common/SurgePatch.cpp
+++ b/src/common/SurgePatch.cpp
@@ -2562,8 +2562,38 @@ void SurgePatch::load_xml(const void *data, int datasize, bool is_preset)
                         }
                     }
                 }
+
+                // adjust tuning by 4.61048 cents
+                // see GitHub issue #8049
+                // TODO: This won't be necessary in XT2.0, so can be removed then!
+                if (o.type.val.i == ot_twist)
+                {
+                    if (o.pitch.absolute)
+                    {
+                        if (o.pitch.extend_range)
+                        {
+                            o.pitch.val.f *= 1.00022222222222222166f;
+                        }
+                        else
+                        {
+                            o.pitch.val.f *= 1.00266666666666666f;
+                        }
+                    }
+                    else
+                    {
+                        if (o.pitch.extend_range)
+                        {
+                            o.pitch.val.f += 0.0038420666666666666f;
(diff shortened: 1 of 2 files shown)


mkruselj:
I've tested this against 4 patches created with 1.3.4 (they had a Modern saw with Retrigger on, and Twist set to saw also with Retrigger on, and all combinations of normal, extended and absolutable). They all sounded identical with this commit, so wahey!

Of course, there will be edge cases when oscillator pitch is set to max, but not much we can do about that one, really. It is what it is.

PR #8366, opened 2026-04-23
Replace _BitScanReverse with C+20 std::bit_width()

### LINKED ISSUE #6726: _BitScanReverse inWindow Oscillator
See #6721 

Can't we be a bit more portable than that?

mkruselj:
Closes #6726

This also fixes a pretty bad bug that's been there since forever: Window oscillator sounded incorrectly on *nixes, because the intrinsic shim for `_BitScanReverse()` used a wrong replacement function. This resulted in mip-map selection ending up basically completely broken, in most cases the most detailed mipmap would have been used even at highest pitches, creating a lot more artifacts than there should be.

Considering it is of utmost importance for a synth to sound the same regardless of the OS used, despite this being a breaking change, both @baconpaul and I agreed it's the right thing to just outright fix this directly, no handling in streaming or any such shenanigans.

Now Window oscillator will sound identical across platforms!

### DIFF
--- a/src/common/dsp/Wavetable.cpp
+++ b/src/common/dsp/Wavetable.cpp
@@ -26,11 +26,10 @@
 #include "SurgeStorage.h"
 
 #include "sst/basic-blocks/mechanics/endian-ops.h"
-namespace mech = sst::basic_blocks::mechanics;
 
-#if WINDOWS
-#include <intrin.h>
-#endif
+#include <bit>
+
+namespace mech = sst::basic_blocks::mechanics;
 
 using namespace std;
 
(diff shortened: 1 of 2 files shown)


src/common/dsp/Wavetable.cpp
```diff
@@ -196,15 +187,7 @@ bool Wavetable::BuildWT(void *wdata, wt_header &wh, bool AppendSilence)
...
-#if WINDOWS
-    unsigned long MSBpos;
-    _BitScanReverse(&MSBpos, size);
-#else
-    unsigned int MSBpos;
-    _BitScanReverse(&MSBpos, size);
-#endif
-
-    size_po2 = MSBpos;
+    size_po2 = std::bit_width(static_cast<unsigned int>(size)) - 1;
```
baconpaul, line 190: Dont we need a comparison of size with zero here also? This can run size_po2 to -1 which the prior code would not if I'm reading correctly. At least clamp size_po2 into range after the calculation.

baconpaul (dismissed):

src/common/dsp/Wavetable.cpp
mkruselj: Makes sense, will do!

mkruselj: Actually... this is about wavetable size. Would we ever get a zero for its size? This would be nasty even with old code since _BitScanReverse is undefined for 0 input, and ctz is as well...

baconpaul: Oh sorry I thought we were guarded before and i misread the diff.

PR #8369, opened 2026-04-25
LFO phase now stops at 359° instead of going full circle

### LINKED ISSUE #6568: LFO phase parameter wraps around.
...meaning the value at the bottom of the slider is always = that at the top. That means when modulating this slider you sometimes gotta be careful to omit the last % from the modulation range or get weird zips if the start and end value are different.

I don't know when I'd want this and I'm somewhat regularly running into scenarios where I really don't. Maybe it should go to 359 degrees instead of 360?

mkruselj:
This is a breaking change, but it's a really small one, and only for basic waveforms (it won't change anything for step seq, envelope, MSEG, or formula).

I think people by and large won't be hugely adversely affected by this change. And if they do, we can easily revert this since the rescaling is done in one spot only.

Fixes #6568

### DIFF
--- a/src/common/dsp/modulators/LFOModulationSource.h
+++ b/src/common/dsp/modulators/LFOModulationSource.h
@@ -127,8 +127,27 @@ class LFOModulationSource : public ModulationSource
   private:
     pdata *localcopy;
     bool phaseInitialized;
+
     void initPhaseFromStartPhase();
     void msegEnvelopePhaseAdjustment();
+
+    inline float startPhaseRescaled() const
+    {
+        // start_phase doubles as shuffle parameter for step sequencer
+        // where it must remain as unscaled, full [0, 1] range
+        // which is also used in extended mode as (x * 2 - 1).
+        // Scale only for basic LFO shapes
+        if (lfo->shape.val.i == lt_stepseq || lfo->shape.val.i == lt_formula ||
+            lfo->shape.val.i == lt_envelope || lfo->shape.val.i == lt_mseg)
+        {
+            return localcopy[startphase].f;
+        }
+
+        // Prevent full-circle: scale to [0, 1) so val_max != val_min
+        constexpr float maxPhase = 1.0f - (1.0f / 360.0f);
+
+        return localcopy[startphase].f * maxPhase;
+    }
 
     float phase, target, noise, noised1, env_phase, priorPhase;
     int unwrappedphase_intpart;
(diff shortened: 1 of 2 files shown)


baconpaul:
So the problem with this as opposed to a clamp is folks who have set a start phase to say exactly 0.5 for a pulse start will have their patches very very broken. Why not clamp instead?

src/common/dsp/modulators/LFOModulationSource.h
```diff
@@ -127,9 +127,28 @@ class LFOModulationSource : public ModulationSource
...
+        // which is also used in extended mode as (x * 2 - 1).
+        // Scale only for basic LFO shapes
+        if (lfo->shape.val.i == lt_stepseq || lfo->shape.val.i == lt_formula ||
+            lfo->shape.val.i == lt_envelope || lfo->shape.val.i == lt_mseg)
+        {
+            return localcopy[startphase].f;
+        }
+
+        // Prevent full-circle: scale to [0, 1) so val_max != val_min
+        constexpr float maxPhase = 1.0f - (1.0f / 360.0f);
```
djtuBIG-MaliceX, line 147: Wouldn't it be better to cap it at the lowest possible increment before 'full circle' rather than the quantum being off by exactly 1 degree?  something something resolution.  (ie: 359.9999999...)

mkruselj: No, that sounds basically like 0°, which is exactly what Andreya wants to avoid.

mkruselj:
> So the problem with this as opposed to a clamp is folks who have set a start phase to say exactly 0.5 for a pulse start will have their patches very very broken. Why not clamp instead?

Fair point, indeed that makes more sense. Thankfully this only needs to be changed in one place now 🙂

nuoun:
There's an extend range option which I think is only used for shuffle in the step sequencer, wonder if we should use that to keep the old behavior.

mkruselj:
Nah I don't think that's necessary.

Andreya-Autumn:
Yeah the clamp is the better fix indeed. Thanks for looking into this. And I agree with hindsight we could've fixed this long ago, the break seems unlikely to actually cause issues.

mkruselj:
One of the rare cases where breaking change actually... isn't breaking all that hard! :D

Andreya-Autumn:
Yeah. The mod mapper shenanigans used to sometimes require modulating phase from 0-99.99% and now it can be 100%. Some existing patches which were set up with the 99 may be slightly affected. But I agree it's probably fine. And the change definitely feels worth it.

PR #8371, opened 2026-04-26
Cleanup and assorted fixes

nuoun:
Some stuff that was still on my list, came up recently with gdb Asan / MSVC debugger or was flagged by Claude:

Convolution:
- Free AudioFormatReader after loading Convolution IR (fixes LeakedObjectDetector tripping on exit)
- Use semicolons for Convolution filepicker extension list

LuaSupport:
- Pass functions to parseStringDefiningMultipleFunctions by `const &`, avoiding a copy
- Pop Lua error messages in `loadSurgePrelude`
- Drop unused `SurgeSharedBinary.h`, `basic_dsp.h`, `luajit.h`, and `lj_arch.h` includes

LuaEditors:
- Fix focus assertion when applying/generating from the Prelude tab
- Initialize `saveCaretStartPositionLock`

FormulaModulationHelper:
- Fix `process()` returning a bare number (instead of a table) being silently replaced by the error fallback after the first call
- Handle `luaL_newstate()` returning null (due to lack of memory) by bailing out of `prepareForEvaluation`
- Avoid making some copies in the debugger iteration loops by using `const auto &`
- Avoid sign mismatch warnings in the debugger loops by switching to `size_t`
- Drop unused `<thread>` and add direct includes

WavetableScriptEvaluator:
- Avoid sign mismatch warnings
- Drop unused `StringOps.h` and add direct includes

### DIFF
--- a/src/common/LuaSupport.cpp
+++ b/src/common/LuaSupport.cpp
@@ -21,13 +21,6 @@
  */
 
 #include "LuaSupport.h"
-
-#include "basic_dsp.h"
-
-#if HAS_JUCE
-#include "SurgeSharedBinary.h"
-#endif
-
 #include "lua/LuaSources.h"
 
 #include <iostream>
@@ -47,7 +40,7 @@ bool Surge::LuaSupport::parseStringDefiningFunction(lua_State *L, const std::str
 }
 
 int Surge::LuaSupport::parseStringDefiningMultipleFunctions(
-    lua_State *L, const std::string &definition, const std::vector<std::string> functions,
+    lua_State *L, const std::string &definition, const std::vector<std::string> &functions,
     std::string &errorMessage)
 {
 #if HAS_LUA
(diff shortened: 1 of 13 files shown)


nuoun:
Looks like `#define HAS_LUA 1` in [SurgeGUIEditor.h](https://github.com/surge-synthesizer/surge/blob/f9111579166e8171f303e325dd997b9ea2cbfb41/src/surge-xt/gui/SurgeGUIEditor.h#L59) hides some preexisting gating gaps with the `HAS_LUA` macro in the ARM64EC build. Will fix those in another PR.

mkruselj:
I thought I removed that define!

nuoun:
Looks like it's just 3 or 4 places in SurgeGUIEditorKeyboardActions.h and OscillatorWaveformDisplay.cpp around TOGGLE_WTS_EDITOR that are missing the HAS_LUA macro. Can do it in this PR too since this is already a mess lol.

nuoun:
Looks like that's all of them.

PR #8379, opened 2026-05-01
Bundle shared resources via CMakeRC instead of JUCE binary data

### LINKED ISSUE #8128: Issue with oscillator parameter name getter in SurgePy
User on Discord reports:

"In surgepy when setting an oscillator type to Window, it seems to be using conditional params for oscillator type Sine. Instead of Morph / Formant / Window, getParamInfo returns Shape / Feedback / Behavior. 

Have attached a script to reproduce - switching osc types and printing correct params works, except when setting osc type to Window. 

I'm using a python 3.13 venv on macOS Sequoia / M1 Max. Have tried both pip install and manual build of surgepy and gets the same result. Have also tried loading a wavetable to the oscillator before refreshing conditional params and no luck either. GUI build from the same source shows correct parameters. (1.4.main.2ad1c476)

Wanted to see if anyone else has this issue and if it's a problem on my end or a bug?"

```python
import surgepy

surge = surgepy.createSurge(48000)
patch = surge.getPatch()

def set_osc_type(patch, ot):
    osc_type = patch['scene'][0]['osc'][0]['type']
    surge.setParamVal(osc_type, ot)

def get_osc_params(patch):
    osc = patch['scene'][0]['osc'][0]
    osc_type = osc['type']
    osc_contextual_params = osc['p']
    print(f"Oscillator Type {surge.getParamDisplay(osc_type)} / {surge.getParamVal(osc_type)}")
    for param in osc_contextual_params:
        print(surge.getParamInfo(param))


# Print initial parameters
# As osc type is classic should return Shape, Width 1, Width 2, Sub Mix, Sync, Unison Detune, Unison Voices
print("---Initial Parameters:------------------")
get_osc_params(patch)


# Change osc type to FM3 and refresh conditional params
set_osc_type(patch, 5)
surge.process()
patch = surge.getPatch()

# As osc type is now FM3 should return M1 Amount, M1 Ratio, M2 Amount, M2 Ratio, M3 Amount, M3 Frequency, Feedback
print("---New Parameters:----------------------")
get_osc_params(patch)


# Change osc type to Window and refresh conditional params
set_osc_type(patch, 7)
surge.process()
patch = surge.getPatch()

# As osc type is now Window should return Morph, Formant, Window, Low Cut, High Cut, Unison Detune, Unison Voices
print("---New Parameters:----------------------")
get_osc_params(patch)
```

baconpaul:
Move configuration.xml, paramdocumentation.xml, oscspecification.html, README_UserArea.txt, windows.wt and memoryWavetable.wt from a juce_add_binary_data target into a CMakeRC archive, accessed through Surge::Storage::getSurgeCommonBinaryResource(). This decouples the shared resources from JUCE so non-JUCE consumers (notably the Python bindings) can load windows.wt at startup. Closes #8128.

Assisted-by: Claude Opus 4.7 <noreply@anthropic.com>

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -154,6 +154,8 @@ endif()
 
 message(STATUS "Building Surge XT using the following JUCE wrappers: ${SURGE_JUCE_FORMATS}")
 
+include(cmake/CmakeRC.cmake)
+
 add_subdirectory(common)
 add_subdirectory(lua)
 add_subdirectory(platform)
--- a/src/cmake/CmakeRC.cmake
+++ b/src/cmake/CmakeRC.cmake
@@ -1,0 +1,644 @@
+# This block is executed when generating an intermediate resource file, not when
+# running in CMake configure mode
+if(_CMRC_GENERATE_MODE)
+    # Read in the digits
+    file(READ "${INPUT_FILE}" bytes HEX)
+    # Format each pair into a character literal. Heuristics seem to favor doing
+    # the conversion in groups of five for fastest conversion
+    string(REGEX REPLACE "(..)(..)(..)(..)(..)" "'\\\\x\\1','\\\\x\\2','\\\\x\\3','\\\\x\\4','\\\\x\\5'," chars "${bytes}")
+    # Since we did this in groups, we have some leftovers to clean up
+    string(LENGTH "${bytes}" n_bytes2)
+    math(EXPR n_bytes "${n_bytes2} / 2")
+    math(EXPR remainder "${n_bytes} % 5") # <-- '5' is the grouping count from above
+    set(cleanup_re "$")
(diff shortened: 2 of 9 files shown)


baconpaul:
@mkruselj if this passes CI can i ask you to build and check it out. You will need a clean build to do so for the python.

It seems right here but would appreciate a second look before we merge on windows. I only tested mac. (But I did add a regtest which CI will run all platforms).

baconpaul:
OK cool if this works for you and also looks good I’m happy for it to be merged @mkruselj

mkruselj:
I'll test it in my evening today!

baconpaul:
Great. Feel free to just press merge if it works! I've reviewed it a few times and tested it here pretty closely. But windows is a bit different so want to make sure we don't blow anything out.

PR #8380, opened 2026-05-02
Extend range option tweaks

nuoun:
- The LFO Phase's extend range is only used for the step sequencer's shuffle so hide the menu option for other LFO types
- Clear the Phase extend option when leaving the step sequencer
- Clamp `val.f` when disabling the extend range option on these params: `ct_pitch_extendable_very_low_minval`, `ct_freq_audible_fm3_extendable`, `ct_freq_reson_band1-3`, `ct_lfoamplitude` and `ct_dly_fb_clippingmodes`

Assisted-by: Claude Opus 4.7

### DIFF
--- a/src/common/Parameter.cpp
+++ b/src/common/Parameter.cpp
@@ -2196,6 +2196,10 @@ void Parameter::set_extend_range(bool er)
             {
                 val.f = val_min.f;
             }
+            else if (val.f > val_max.f)
+            {
+                val.f = val_max.f;
+            }
 
             displayType = LinearWithScale;
             snprintf(displayInfo.unit, DISPLAYINFO_TXT_SIZE, "semitones");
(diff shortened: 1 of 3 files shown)


Andreya-Autumn:
Uhhh, the LFO phase does extend in other modes. It makes it bipolar.

Andreya-Autumn:
Also, bit unsure about this part of the Surge code: Do those clamps affect the value under modulation, or just the param baseline? Automation can push some parameters past min/max which does get used in patches sometimes, so if the clamp is after modulation we shouldn't do it.

nuoun:
> Uhhh, the LFO phase does extend in other modes. It makes it bipolar.

There's no difference between -100% and 100% with extend range and 0% and 100%, it just changes the halfway point and tooltip, (start_phase is always 0 .. 1 except for the step sequencer) so not sure if that makes sense. Maybe it does since it at least sets the slider halfway on double click rotating it 180 degrees.

> Also, bit unsure about this part of the Surge code: Do those clamps affect the value under modulation, or just the param baseline? Automation can push some parameters past min/max which does get used in patches sometimes, so if the clamp is after modulation we shouldn't do it.

Those clamps just reset the regular slider value back to min or max when clearing the extend range option, preventing them from glitching out due to going out of range:

image

Some extend range options scale the min / max values but these just set the min / max values allowing them to go out of bounds after switching back to the not extended mode.

Andreya-Autumn:
Ah right. The clamps make sense then. But the phase thing Im really sure is wrong. 0 == -100 indeed but the in between range allows different behaviors under modulation. I doubt it's used much but if users out there do have patches that use the extended range with modulation those would change. So we should not merge that change. 

Relatedly though, @mkruselj when you did the ≈359.9 degrees clamp a while ago, did you do the extended range as well?

Andreya-Autumn:
Oh ha! You're right actually. I could've sworn the extend range option actually gave you two cycles worth of range on the phase control, but I double checked now and see that it does not! 

Just to double check, was it actually always like that? If it was then yeah I agree, the option can be removed. But maybe do a quick git blame to see if I am tripping or if we have in fact changed that at some point.

nuoun:
I tested in 1.3.4 and the behavior is the same there so seems it's always been like that, the [manual ](https://surge-synthesizer.github.io/manual-xt/#parameters)also suggests this is by design only mentioning the shuffle's param extended range functionality. #8370 did set a min value with the clamp of 0, but as far as I can tell that's been the min value since at least #4741 setting `case ct_lfophaseshuffle: val_min.f = 0;`

But on second thought not entirely sure removing the extend range option is a good idea, I don't think it changes anything for existing patches but the functionality of double click setting it to the half way point in extend range mode is useful. That said it's kinda weird how extend range works but allowing phase to actually go to -1 would be a breaking change and probably not worth it. So, I shall ponder over it a bit more.

mkruselj:
I did not cover the extended mode with 359° clamp, no. This is done at LFO engine level, not at param extend level.

Anyways I think that we're good with removing the extend range on non step seq modes. It seems it just changed value display but not actual LFO behavior, so nothing is actually broken if we remove that functionality.

nuoun:
I do think two cycles of range would be more intuitive here, as well as continuous phase modulation but both of these could perhaps be XT2 goals.

mkruselj:
I'm not really a fan of two cycles of phase, it is not really something you see in other synths so it'd feel weird to most, so I'd vote against that for XT2.

PR #8382, opened 2026-05-06
Add state.lfo_id var to Formula

nuoun:
With `lfo_id`, identical scripts (presets mostly) that depend on the `shared` table can each have their own namespace using the index as a unique key to avoid collision.

Example use:

```lua
function init(state)
    shared[state.lfo_id] = shared[state.lfo_id] or {}
    local ns = shared[state.lfo_id]
    ns.val = 1
    
    return state
end
```

Bit of an unconventional way to get the LFO index in FormulaModulationHelper, getting it from the pointer element but the pointer is stable and it's a lot less involved than having to update all the different callers.

Also adds 3 shared table tests and does some cleanup.


nuoun:
"Nimbus at High Sample Rate" test failing locally at head too so seems unrelated to this PR, seems it also sounds a bit different in the Nightly compared to 1.3.4 here.

```/home/runner/work/surge/surge/src/surge-testrunner/UnitTestsFX.cpp:407: FAILED:
  REQUIRE( maxAmp > 0.1 )
with expansion:
  0.06954f > 0.1
```

Last change was a month ago so not sure why it is only failing now (surge-synthesizer/sst-effects#248)
@baconpaul could you please have a look

Edit: Seems that test only fails half of the time so that would explain why it did not come up earlier

nuoun:
Bumping this now that the Nimbus issue got fixed. (Thanks again Paul!)

PR #8391, opened 2026-05-20
Fix CLI MIDI device discovery on Linux by running a message pump

bwanab:
## Problem

  On Linux, `surge-xt-cli --all-midi-inputs` and `--list-devices` report
  "Found 0 devices!" even when MIDI devices are connected at startup.

  ## Root cause

  JUCE's ALSA MIDI backend discovers devices asynchronously. The first call to
  `MidiInput::getAvailableDevices()` creates an `AlsaMidiHelpers::Client` and
  starts a `SequencerThread`. When the Client subscribes to ALSA system
  announcements, ALSA immediately queues a `SND_SEQ_EVENT_PORT_SUBSCRIBED`
  event. The SequencerThread picks this up, calls `triggerAsyncUpdate()`, and
  posts a callback to the JUCE message thread. Only when that callback fires —
  `notifyPortsChanged()` → `findEndpoints()` — is `cachedEndpoints` populated
  and devices visible to callers.

  The GUI always runs a message loop, so this callback arrives before any user
  interaction. The CLI only ran a message loop when OSC was enabled; otherwise

  ## Fix (CLI only, no JUCE changes)
  
  1. **`primeMidiDeviceDiscovery()`** — a Linux-only helper called before the
     real `getAvailableDevices()`. It makes one throwaway call to create the
     Client and start the SequencerThread, sleeps 150 ms (≥ the SequencerThread's
     100 ms poll timeout), then drains the JUCE message queue via
     `juce::detail::dispatchNextMessageOnSystemQueue(true)`.

  2. **Always run the message loop on Linux** — not just when OSC is active,
     matching what the GUI does and covering runtime MIDI hotplug events.

  3. **`--list-devices` fix** — `listAudioDevices()` constructs an
     `AudioDeviceManager` whose in-class `midiDeviceListConnection` initializer
     creates the ALSA Client. The `MessageManager` must exist before that
     construction, or `triggerAsyncUpdate()` has no queue to post to.

  ## Tested on
  
  Raspberry Pi aarch64, Patchbox OS, ALSA 1.2.8, six MIDI devices connected at
  startup — both `--all-midi-inputs` and `--list-devices` now report all six
  devices correctly.
  
Assisted-By: <Claude Sonnet 4.6>


baconpaul:
Looks good! Squash merging.

PR #8397, opened 2026-05-24
OSC: support multiple Surge instances on the same machine

nuoun:
Running multiple Surge instances on the same machine didn't actually work for OSC, every instance silently thought it was bound to the configured port, but only one received packets. By switching away from the [OSCReceiver](https://docs.juce.com/master/classjuce_1_1OSCReceiver.html) class and constructing our own [DatagramSocket](https://docs.juce.com/master/classjuce_1_1DatagramSocket.html) we can clear the flags that allow port hijacking and get a meaningful return when a port is already in use.

With this PR Surge tries 100 ports after the configured input port, binding to the first that works. For DAW sessions Surge tracks what the last bound port was and tries to restore that first before falling back to scanning the ports inside the range, for the standalone it ignores the last bound port and always scans the range.

UI wise this status line was added:
image

Also sets the default output port to 53270 so it's outside the default range (since by default the output is also localhost).

Tested on Windows and Linux with Reaper (still untested on macOS).

Assisted-By: Claude Opus 4.7

### DIFF
--- a/src/common/SurgePatch.cpp
+++ b/src/common/SurgePatch.cpp
@@ -3451,6 +3451,14 @@ void SurgePatch::load_xml(const void *data, int datasize, bool is_preset)
                 storage->oscPortIn = ival;
             }
 
+            p = TINYXML_SAFE_TO_ELEMENT(de->FirstChild("oscPortInLastBound"));
+
+            if (p && p->QueryIntAttribute("v", &ival) == TIXML_SUCCESS)
+            {
+                dawExtraState.oscPortInLastBound = ival;
+                storage->oscPortInLastBound = ival;
+            }
+
             p = TINYXML_SAFE_TO_ELEMENT(de->FirstChild("oscPortOut"));
 
             if (p && p->QueryIntAttribute("v", &ival) == TIXML_SUCCESS)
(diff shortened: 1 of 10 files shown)


mkruselj:
Shouldn't the output port be at least a 100 apart so that it doesn't clash with inputs? If I'm understanding correctly, once we have more than 10 Surge instances, things are gonna clash.

nuoun:
I changed the default out port below the range in [globals.h](https://github.com/surge-synthesizer/surge/pull/8397/changes#diff-afb76a7c75c8c4b550769ae24e8fd3e2410eb8de40ae875c7dfe7b6d61d7728bL71), but on existing installations it restores it from the XML instead. Not worth a streaming revision imo since there is [this](https://github.com/surge-synthesizer/surge/pull/8397/changes#diff-f75da5c3a10e6faf8ba585a77f975ab3d564ae95b839b25a535a50c610793a13R145) preexisting guard in initOSCIn() that skips the outport when trying to bind.

I tested what cross wiring two instances of Surge does with OSC and no endless param change loop so it should be fine.

mkruselj:
Right but all I'm saying is, if we're trying up to 100 connections, it should've been 53180 then, I'd think?

nuoun:
It increments, ie. tries ports upward, not downward.

nuoun:
Or maybe you think it also tries a range for outbound, which it does not (nor should it).

mkruselj:
Oh, I thought it also did so for out port!

nuoun:
Binding a socket to a port is really just a inbound thing, it's also using UDP so there's no way to even know if there even is a connection established, it basically just broadcasts.

PR #8411, opened 2026-06-07
Allow placing MSEG loop start/end together at first or last node

### LINKED ISSUE #3462: Can't place start and end loop on very last point of MSEG
Very very late in the 18 cycle I realized a small bug with big implication.

You can't have the start and end loop on the last point of the mseg

Reason for this is

loop_end means 'end time of segment' and loop_start means 'start time of segment'
so the 'zero length loop' means 'loop_start == loop_end + 1'
but if loop_end = end of mseg then there is no place for loop_start

fixing this requires us to have a special sentinel for 'loop start at end means the end of prior' which touches a lot of code and we are very late in the cycle. So instead I am constraining the UI so you can only drag the start point up to the start of the last segment and will fix this in 1.9.

mkruselj:
This was previously not possible since it was a big scary refactor that touched a lot of code points.

But it is not impossible with a bit of guided assistance.

Closes #3462

Assisted-By: Claude Sonnet 4.6 <noreply@anthropic.com>

### DIFF
--- a/src/common/SurgePatch.cpp
+++ b/src/common/SurgePatch.cpp
@@ -4576,8 +4576,12 @@ void SurgePatch::msegToXMLElement(MSEGStorage *ms, TiXmlElement &p) const
     p.SetAttribute("endpointMode", ms->endpointMode);
     p.SetAttribute("editMode", ms->editMode);
     p.SetAttribute("loopMode", ms->loopMode);
-    p.SetAttribute("loopStart", ms->loop_start);
-    p.SetAttribute("loopEnd", ms->loop_end);
+
+    // Endpoint-based markers. loop_end == -1 legitimately means "point 0" now,
+    // which collides with the legacy loopStart/loopEnd "-1 == unset" sentinel,
+    // so write under new names and stop writing the ambiguous legacy attributes.
+    p.SetAttribute("loopStartPoint", ms->loop_start);
+    p.SetAttribute("loopEndPoint", ms->loop_end);
 
     p.SetDoubleAttribute("hSnapDefault", ms->hSnapDefault);
     p.SetDoubleAttribute("vSnapDefault", ms->vSnapDefault);
@@ -4626,15 +4630,24 @@ void SurgePatch::msegFromXMLElement(MSEGStorage *ms, TiXmlElement *p, bool resto
         ms->loopMode = (MSEGStorage::LoopMode)v;
     else
         ms->loopMode = MSEGStorage::LoopMode::LOOP;
-    if (p->QueryIntAttribute("loopStart", &v) == TIXML_SUCCESS)
-        ms->loop_start = v;
-    else
-        ms->loop_start = -1;
 
-    if (p->QueryIntAttribute("loopEnd", &v) == TIXML_SUCCESS)
-        ms->loop_end = v;
+    // Prefer the new endpoint-based markers; fall back to legacy segment-based
+    // loopStart/loopEnd, translating the old "-1 == unset" into kLoopPointUnset.
+    // Legacy 0..n-1 values keep the same meaning. Works for both patches and the
+    // revision-less MSEG .modpreset files, since it keys on attribute name.
+    if (p->QueryIntAttribute("loopStartPoint", &v) == TIXML_SUCCESS)
+        ms->loop_start = v;
+    else if (p->QueryIntAttribute("loopStart", &v) == TIXML_SUCCESS)
+        ms->loop_start = (v < 0) ? MSEGStorage::kLoopPointUnset : v;
     else
-        ms->loop_end = -1;
+        ms->loop_start = MSEGStorage::kLoopPointUnset;
+
+    if (p->QueryIntAttribute("loopEndPoint", &v) == TIXML_SUCCESS)
+        ms->loop_end = v;
+    else if (p->QueryIntAttribute("loopEnd", &v) == TIXML_SUCCESS)
+        ms->loop_end = (v < 0) ? MSEGStorage::kLoopPointUnset : v;
+    else
+        ms->loop_end = MSEGStorage::kLoopPointUnset;
 
     double dv;
     if (p->QueryDoubleAttribute("hSnapDefault", &dv) == TIXML_SUCCESS)
--- a/src/common/SurgeStorage.h
+++ b/src/common/SurgeStorage.h
@@ -931,7 +931,12 @@ struct MSEGStorage
                        // at current value and progress to end once
     } loopMode = LOOP;
 
-    int loop_start = -1, loop_end = -1; // -1 signifies the entire MSEG in this context
+    static constexpr int kLoopPointUnset = -13214;
+
+    // loop_start: 0..n_activeSegments (n_activeSegments = starts at very last point)
+    // loop_end:   -1..n_activeSegments-1 (-1 = ends at very first point, point 0)
+    // Either field == kLoopPointUnset means "unset, use full-range default"
+    int loop_start = kLoopPointUnset, loop_end = kLoopPointUnset;
 
     int n_activeSegments = 0;
     std::array<segment, max_msegs> segments;
--- a/src/common/dsp/modulators/MSEGModulationHelper.cpp
+++ b/src/common/dsp/modulators/MSEGModulationHelper.cpp
@@ -549,14 +549,13 @@ void rebuildCache(MSEGStorage *ms)
 {
     forceToConstrainedNormalForm(ms);
 
-    if (ms->loop_start > ms->n_activeSegments - 1)
+    if (ms->loop_start != MSEGStorage::kLoopPointUnset && ms->loop_start > ms->n_activeSegments)
     {
-        ms->loop_start = -1;
+        ms->loop_start = MSEGStorage::kLoopPointUnset;
     }
-
-    if (ms->loop_end > ms->n_activeSegments - 1)
+    if (ms->loop_end != MSEGStorage::kLoopPointUnset && ms->loop_end > ms->n_activeSegments - 1)
     {
-        ms->loop_end = -1;
+        ms->loop_end = MSEGStorage::kLoopPointUnset;
     }
 
     float totald = 0;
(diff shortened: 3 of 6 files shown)


mkruselj:
@baconpaul Won't merge until you glance this one over 🙂

baconpaul:
Ok - won’t be today

baconpaul:
This is a tricky PR and it turns out it's incomplete. I had some worries reading it but I did a full review with opus 4.8 just now. Here's the review comment set

-----

## Review of #8411 (fixes #3462) — second look

The direction here is exactly what was spec'd in this issue (a `kLoopPointUnset` sentinel plus endpoint-based indices `0..n`), and the in-file runtime migration is mostly sound. But the change is **incomplete in two ways that are individually blocking**, and it's **stale relative to `main`**. In particular the "**and fix streaming too**" part of this issue is entirely absent, and the MSEG test suite added just before this PR does not actually cover the change — so CI is likely green while existing patches silently break.

### 🔴 Blocking 1 — no streaming migration; existing patches break + OOB read

The PR touches 3 files, but the MSEG loop fields are referenced in 5. Serialization is untouched (`SurgePatch.cpp:4536-4537` save, `4586-4594` load).

The refactor **repurposes the value `-1`**: it used to mean "unset / full range" for *both* fields; now `loop_end == -1` means "loop ends at point 0". But the shipping build writes `loop_start`/`loop_end` verbatim (default `-1/-1`), and the loader's missing-attribute fallback also sets `-1` (`SurgePatch.cpp:4589,4594`). A looping MSEG whose markers were never dragged stores `-1/-1` — a large fraction of real patches and factory presets.

Loading such a patch in the new build:
- `loop_end = -1` → `durationToLoopEnd = 0` (rebuildCache) → **loop collapses to zero length**, output corrupted.
- `loop_start = -1` is neither `kLoopPointUnset` nor a valid index. The clamp only resets `loop_start > n_activeSegments`; negative legacy values fall through to `ls_idx = -1` → **`segmentStart[-1]`, an out-of-bounds read** (rebuildCache; same pattern in `timeToSegment` and the editor).

Because the *meaning* of `-1` changed, a legacy `-1` can't be distinguished from a legitimate new `loop_end == -1`. This needs a `streamingRevision`-gated migration (remap legacy `-1 → kLoopPointUnset`), threading the revision into `msegFromXMLElement` (currently `const`, no revision param; called at `SurgePatch.cpp:2873`) plus a stream-version bump. None of that is present — this is the "~125 code points / fix streaming" scope, deferred silently.

### 🟠 Blocking 2 — migration incomplete within scope

- **`extendTo` un-migrated** (`MSEGModulationHelper.cpp`): `if (ms->loop_end < 0 || ms->loop_end == ms->n_activeSegments - 1)`. `loop_end < 0` now also catches `-1` (point 0), so extending an MSEG whose loop ends at point 0 wrongly jumps `loop_end` to `n_activeSegments - 2`.
- **`LFOAndStepDisplay.cpp:548`** still uses the old `ms->loop_end >= 0` idiom (safe by luck since the sentinel is very negative, but inconsistent).

### 🟡 Behavior change worth a test

`valueAt`'s gated-release guard changed from `loop_end >= n_activeSegments` to `>= n_activeSegments - 1`. Existing gated MSEGs with `loop_end` explicitly at the last segment now hold `releaseStartValue` instead of the prior path. Plausibly intended, but untested.

### Test coverage — weak and misleading

- The PR adds/updates **zero tests**, even though `UnitTestsMSEG.cpp` (#8409) landed minutes earlier as the characterization pass for this refactor.
- Those tests encode the **old** convention and aren't updated — `makeRamp4()` literally hardcodes `loop_start = -1; loop_end = -1` (the now-invalid sentinel).
- Yet CI likely stays **green**: eval tests build via the default ctor (now `kLoopPointUnset` = full loop, fine); `makeRamp4`-based tests override with in-range indices before asserting; and `makeRamp4`'s own `-1/-1` OOB read produces discarded garbage, not a failed assertion (intra-struct underflow ASAN won't flag).
- Result: nothing exercises the two things that matter — loading a legacy `-1` patch, and the new first/last-point capability round-tripped through stream + evaluated.

### Stale vs. `main`

Branched before #8431 (freehand draw) and #8409. Un-migrated paths a merge won't fix:
- `MSEGEditor.cpp:2178-2182` — `loopStartTime/loopEndTime` save-restore uses `loop_start >= 0 && < n_activeSegments`, which **breaks the new `loop_start == n_activeSegments` (last-point) value** (restore fails after a freehand edit).
- `MSEGEditor.cpp:2245-2252` and `:2369-2376` — old-convention index shifting.

Needs a rebase and re-audit of these.

### What a correct version needs

1. `streamingRevision`-gated migration in `msegFromXMLElement` (legacy `-1 → kLoopPointUnset` for both fields) + version bump; thread the revision in.
2. Normalize/guard negative `loop_start` in `rebuildCache` so `segmentStart[-1]` can never happen.
3. Finish the migration: `extendTo`, `LFOAndStepDisplay.cpp:548`, and (post-rebase) the freehand-draw paths.
4. Update `makeRamp4()` off the hardcoded `-1`; add tests for (a) legacy `-1/-1` patch load, (b) loop at first point and last point evaluated, (c) a stream round-trip.

The core idea is right and works for a fresh session, but "a small bug with big implication" still has the big implication — back-compat — unaddressed, plus a latent OOB.

---

## Suggested fix (on top of this PR)

The one subtlety worth flagging up front: **modulator presets (`.msegs`) carry no stream revision**, only patches do. So a pure `streamingRevision`-gate fixes patches but leaves old MSEG presets broken. Keying the migration off a **new attribute name** instead handles both uniformly and is unambiguous (no pre-existing stream ever wrote the new names).

### 1. Serialize under new attribute names, translate the legacy ones on load

`SurgePatch::msegToXMLElement` (`SurgePatch.cpp` ~4536):

```diff
-    p.SetAttribute("loopStart", ms->loop_start);
-    p.SetAttribute("loopEnd", ms->loop_end);
+    // Endpoint-based markers. loop_end == -1 legitimately means "point 0" now,
+    // which collides with the legacy loopStart/loopEnd "-1 == unset" sentinel,
+    // so write under new names and stop writing the ambiguous legacy attrs.
+    p.SetAttribute("loopStartPoint", ms->loop_start);
+    p.SetAttribute("loopEndPoint", ms->loop_end);
```

`SurgePatch::msegFromXMLElement` (`SurgePatch.cpp` ~4586-4594):

```diff
-    if (p->QueryIntAttribute("loopStart", &v) == TIXML_SUCCESS)
-        ms->loop_start = v;
-    else
-        ms->loop_start = -1;
-
-    if (p->QueryIntAttribute("loopEnd", &v) == TIXML_SUCCESS)
-        ms->loop_end = v;
-    else
-        ms->loop_end = -1;
+    // Prefer the new endpoint-based markers; fall back to legacy segment-based
+    // loopStart/loopEnd, translating the old "-1 == unset" into kLoopPointUnset.
+    // Legacy 0..n-1 values keep the same meaning. Works for both patches and the
+    // revision-less .msegs preset files, since it keys on attribute name.
+    if (p->QueryIntAttribute("loopStartPoint", &v) == TIXML_SUCCESS)
+        ms->loop_start = v;
+    else if (p->QueryIntAttribute("loopStart", &v) == TIXML_SUCCESS)
+        ms->loop_start = (v < 0) ? MSEGStorage::kLoopPointUnset : v;
+    else
+        ms->loop_start = MSEGStorage::kLoopPointUnset;
+
+    if (p->QueryIntAttribute("loopEndPoint", &v) == TIXML_SUCCESS)
+        ms->loop_end = v;
+    else if (p->QueryIntAttribute("loopEnd", &v) == TIXML_SUCCESS)
+        ms->loop_end = (v < 0) ? MSEGStorage::kLoopPointUnset : v;
+    else
+        ms->loop_end = MSEGStorage::kLoopPointUnset;
```

Forward-compat: an older Surge loading a new patch won't find `loopStart`/`loopEnd`, falls to its own `-1` default → full loop. Graceful. (Bumping `ff_revision` 29 → 30 to mark the format change is good hygiene but not required by this approach.)

If you'd rather gate on revision instead: same translation, but guarded by `streamingRevision < 30` in the patch path — and you'd still need a separate story for the version-less `.msegs` presets. That's why the attribute-name approach above is preferable.

### 2. Defensive guard in `rebuildCache` so a stray negative can never index

Belt-and-suspenders next to the existing upper-bound clamps (`MSEGModulationHelper.cpp`, top of `rebuildCache`):

```diff
     if (ms->loop_end != MSEGStorage::kLoopPointUnset && ms->loop_end > ms->n_activeSegments - 1)
     {
         ms->loop_end = MSEGStorage::kLoopPointUnset;
     }
+    // loop_start has no valid negative value; loop_end's only valid negative is
+    // -1 (point 0). Anything else negative is stale data — treat as unset so we
+    // never index segmentStart[-1].
+    if (ms->loop_start != MSEGStorage::kLoopPointUnset && ms->loop_start < 0)
+    {
+        ms->loop_start = MSEGStorage::kLoopPointUnset;
+    }
+    if (ms->loop_end != MSEGStorage::kLoopPointUnset && ms->loop_end < -1)
+    {
+        ms->loop_end = MSEGStorage::kLoopPointUnset;
+    }
```

### 3. Finish the in-scope migration: `extendTo`

`MSEGModulationHelper.cpp`, `extendTo`:

```diff
-    if (ms->loop_end < 0 || ms->loop_end == ms->n_activeSegments - 1)
+    if (ms->loop_end == MSEGStorage::kLoopPointUnset ||
+        ms->loop_end == ms->n_activeSegments - 1)
     {
         fixupLoopEnd = true;
     }
```

(`loop_end < 0` used to mean only "unset"; it now also catches `-1` = point 0 and would wrongly relocate a point-0 loop end on extend.)

`LFOAndStepDisplay.cpp:548` (`ms->loop_end >= 0`) is safe as written but reads on the old idiom — tidy it to the `!= kLoopPointUnset && >= 0` form for consistency when you touch this area.

### 4. Tests to add (none exist for the actual change)

- Update `makeRamp4()` off the hardcoded `-1` (use `kLoopPointUnset`).
- Legacy load: hand-build XML with `loopStart="-1" loopEnd="-1"` → assert both resolve to `kLoopPointUnset` and the MSEG evaluates as a full loop (this is the regression test that fails today).
- Legacy load with explicit `loopStart="1" loopEnd="2"` → unchanged meaning.
- New capability: `loop_start = n_activeSegments` and `loop_end = -1`, `rebuildCache` + `runMSEG`, assert no OOB and correct span.
- Stream round-trip: set the new endpoint values, serialize, deserialize, assert identity.

### 5. Rebase note

Rebase onto current `main` and re-audit the post-branch freehand-draw paths (`MSEGEditor.cpp` ~2178-2182, ~2245-2252, ~2369-2376) — `loopStartTime/loopEndTime` in particular mishandles the new `loop_start == n_activeSegments` (last-point) value.

baconpaul:
looking at that review by the way (1) my instinct that this is tricky does seem spot on and (2) i'm not sure i want to do this before 1.4 vs in 2.

mkruselj:
@baconpaul I rebased the branch and applied fixes to specified places.

As for adding tests, I am not sure how to involve the whole TinyXML machinery to check serialization round-trip and loading legacy patches. It seems we don't do this in any of our unit tests. Instead, I have tested a number of legacy patches and verified that they load exactly as they were in 1.3.4 in this new branch (and of course, saving new patches does properly load the MSEG data, that too).

PR #8415, opened 2026-06-11
Ensure binn is the same size as the decompressed data for safety.

mx:
Also allow us to tack more on to the end in the future.

Paul, before you merge this, can you verify if I should bother with the streaming revision check (and increment it, which I'll add a commit to do)? We're still on nightly, so if it's fine to *not* do it I can just hose people's IR and WT script patches (they'll need the IRs and WT scripts re-imported basically).

### DIFF
--- a/src/common/SurgeStorage.h
+++ b/src/common/SurgeStorage.h
@@ -1371,6 +1371,7 @@ class SurgePatch
     int32_t paramModulationCount{0};
     static constexpr int maxMonophonicParamModulations = 256;
     std::array<MonophonicParamModulation, maxMonophonicParamModulations> monophonicParamModulations;
+    int claimedArbitraryBlockStorageSize{0};
 };
 
 struct Patch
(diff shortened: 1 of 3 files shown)


baconpaul:
This late in the nightly cycle would probably do a streaming revision and it blow out scripts if its not insanely painful - 1.4 encompasses several streaming versions already

src/common/SurgeStorage.h
```diff
@@ -149,7 +149,7 @@ const int FIRoffsetI16 = FIRipolI16_N >> 1;
 // 27 -> 28 (XT 1.4.* nightlies) added corrected TX shapes to Sine oscillator
 // clang-format on
 
-const int ff_revision = 28;
+const int ff_revision = 29;
```
baconpaul, line 153: If you have a moment we normally add a comment about like 28->29 with a brief description

PR #8418, opened 2026-06-13
Refresh MSEG/Formula overlay on scene change

nuoun:
- Fixes MSEG/Formula overlay not refreshing on scene change (like it does on switching modulators)
- Keep torn-out overlay windows on top after an editor rebuild
- Make `OverlayWrapper::doTearOut` re-entrant so reopening an already torn-out overlay (scene/osc/modulator switch) doesn't break re-docking using the minimize button
- Adds missing `break` in `updateOverlayContentIfPresent`

Assisted-by: Claude Opus 4.8

### DIFF
--- a/src/surge-xt/gui/SurgeGUIEditor.cpp
+++ b/src/surge-xt/gui/SurgeGUIEditor.cpp
@@ -2414,6 +2414,15 @@ void SurgeGUIEditor::openOrRecreateEditor()
     sendStructureChangeIn = 120;
 
     frame->repaint();
+
+    // Also bring any torn-out overlay back to the front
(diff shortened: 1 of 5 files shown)


mkruselj:
How about WTSE, wouldn't that overlay also be affected? And ISTR oscilloscope can also be torn out... are those two not exhibiting the same issues being fixed here?

nuoun:
WTSE already behaved correctly so I ported that function to MSEG/Formula, other overlays that are scene specific like Filter Analysis and Waveshaper already refreshed correctly but now they stay on top of the main window after switching scenes when torn out, same for the Oscilloscope so I think that's all of them.

PR #8427, opened 2026-06-19
Disable the deactivated-rate ghost wave for Formula LFO display

nuoun:
When an LFO's rate is deactivated, the waveform display draws a faint "ghost" reference wave by re-running the modulator with rate re-enabled and start_phase forced to 0. For Formula modulators this is not that useful, best case it just makes thousands of extra `process()` calls to draw the graph a second time, worst case it produces output that has no relation to what the script is doing (when the Phase slider is used for other parameters for example).

### DIFF
--- a/scripts/misc/Formula-state-variable-test.lua
+++ b/scripts/misc/Formula-state-variable-test.lua
@@ -49,6 +49,7 @@ function init(state)
     print("tuned_key", state.tuned_key)
     print("use_amplitude", state.use_amplitude)
     print("use_envelope", state.use_envelope)
+    print("use_rate", state.use_rate)
     print("velocity", state.velocity)
     print("voice_count", state.voice_count)
     print("voice_id", state.voice_id)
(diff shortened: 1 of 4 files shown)


mkruselj:
I think the ghost reference should stay simply for consistency against other LFO modes.

nuoun:
Can add a flag to disable it instead like `use_amplitude` does, named `use_rate` or something. Currently it defeats the purpose of disabling the rate parameter for presets when that slider is unused because it produces garbage output for some scripts and adds twice the graph overhead.

mkruselj:
Do you actually feel the extra graph overhead?

nuoun:
It's ~8000 extra calls @ 48Khz sampling rate to render the extra graph so it's noticeable for larger scripts yes, but it caches at least after first parse so it's more annoying while editing.

mkruselj:
Mmmmh.

This is tricky, because it actually is useful in some cases with Formula, but not always... For example in Piano Phase tutorial patch, for V1 Env and V2 Env outputs it is entirely valid. For V1 Tone and V2 Tone it isn't.

Harumph.

nuoun:
I agree it can be useful, so I added a `use_rate` toggle to this PR making it opt-out just like `use_amplitude` and `use_envelope`. Hope that solves it.

mkruselj:
Awesome, thanks! I'll merge this, please update the nightly changelog and manual where appropriate 🙂

PR #8429, opened 2026-06-24
Param clumping: no root name

mkruselj:
This avoids putting all parameter groups into a Root group in hosts like Studio One (and in Reaper it prefixes every group name with Root/ which is ugly)


mkruselj:
NOTE: This temporarily breaks param clumping in CLAP because CJE builds the param tree differently and requires a named root, JUCE's VST3 param tree builder doesn't. Needs a CJE fix upstream.

PR #8431, opened 2026-06-28
MSEG freehand draw (Shift+Ctrl+Alt) 

### LINKED ISSUE #6469: MSEG freehand drawing
msegalt

Example from UVI Falcon.

I don't think this needs to be a new Movement Mode, it's more of an in-place overwrite of the time segment within which you moused through, via keyboard modifier. I suggest Alt/Opt+drag for this new freehand drawing mode.

Relevant link for curve simplification algorithm: [HERE](https://gist.githubusercontent.com/TimSC/0813573d77734bcb6f2cd2cf6cc7aa51/raw/df7bacf4b472b949f90f57b113803c802d638d72/RamerDouglasPeucker.cpp).

mkruselj:
Closes #6469. 

Inspired by the same feature in UVI Falcon. We're trying to fit more curve types, Falcon only does linear piecewise fitting, we're using hold, linear, S-curve and Bezier. 

Hold Shift+Ctrl+Alt to initiate the freehand draw in MSEG. 

Also fixes a long-standing bug where loop start/end vertical lines would be occluded by gridlines, if loop points were snapping to the grid. 

Needs two things fixed subsequently but we can do that in separate PRs.

Assisted-By: Claude Sonnet 4.6 <noreply@anthropic.com>

### DIFF
--- a/src/common/dsp/modulators/MSEGModulationHelper.cpp
+++ b/src/common/dsp/modulators/MSEGModulationHelper.cpp
@@ -467,7 +467,7 @@ void freehandRDP(std::span<const std::pair<float, float>> samples, float epsilon
     for (size_t i = 1; i < samples.size() - 1; ++i)
     {
         float frac = (duration > 1e-6f) ? (samples[i].first - tStart) / duration : 0.f;
-        frac = std::clamp(frac, 0.f, 1.f, frac);
+        frac = std::clamp(frac, 0.f, 1.f);
 
         float predicted;
         switch (fit.type)
--- a/src/common/dsp/modulators/MSEGModulationHelper.cpp
+++ b/src/common/dsp/modulators/MSEGModulationHelper.cpp
@@ -53,14 +53,16 @@ static float evalLinear(float frac, float v0, float v1, float cpv)
 
     if (fabs(V) > 1e-3f)
     {
-        float Q = std::max(1e-5f, std::min(1e6f, (1.f - sqrtf(disc)) / (2.f * V)));
+        float Q = std::clamp((1.f - sqrtf(disc)) / (2.f * V), 1e-5f, 1e6f);
         a = amul * 2.f * logf(Q);
     }
 
     float cpline = frac;
 
     if (fabs(a) > 1e-3f)
+    {
         cpline = (expf(a * frac) - 1.f) / (expf(a) - 1.f);
+    }
 
     // df=0 so bend3 is identity — no additional deform step needed
 
@@ -99,14 +101,16 @@ static float evalSCurve(float frac, float v0, float v1, float cpv)
 
     if (fabs(V) > 1e-3f)
     {
-        float Q = std::max(1e-5f, std::min(1e6f, (1.f - sqrtf(disc)) / (2.f * V)));
+        float Q = std::clamp((1.f - sqrtf(disc)) / (2.f * V), 1e-5f, 1e6f);
         a = amul * 2.f * logf(Q);
     }
 
     float cpline = f2;
 
     if (fabs(a) > 1e-3f)
+    {
         cpline = (expf(a * f2) - 1.f) / (expf(a) - 1.f);
+    }
 
     if (!mirrored)
         cpline *= 0.5f;
@@ -126,7 +130,9 @@ static float evalBezier(float timeAlong, float duration, float v0, float v1, flo
 
     // Guard: walk off exact midpoint to avoid degenerate quadratic
     if (fabs(cpt - duration * 0.5f) < 1e-5f)
+    {
         cpt += 1e-4f;
+    }
 
     // Convert stored on-curve point to actual Bezier control point
     float tp = duration * 0.5f;
@@ -172,8 +178,11 @@ static float maxResidual(std::span<const std::pair<float, float>> samples, float
         frac = std::clamp(frac, 0.f, 1.f);
         float predicted = eval(frac);
         float r = fabs(s.second - predicted);
+
         if (r > maxR)
+        {
             maxR = r;
+        }
     }
 
     return maxR;
@@ -183,7 +192,7 @@ static float maxResidual(std::span<const std::pair<float, float>> samples, float
 // https://en.wikipedia.org/wiki/Golden-section_search#Iterative_algorithm
 template <typename F> static float goldenSearch(F f, float lo, float hi, int iters = 30)
 {
-    constexpr float phi = 0.6180339887f; // 1/golden_ratio
+    constexpr float phi = 0.6180339887f; // 1 / golden_ratio
     float c = hi - phi * (hi - lo);
     float d = lo + phi * (hi - lo);
     float fc = f(c), fd = f(d);
@@ -209,6 +218,45 @@ template <typename F> static float goldenSearch(F f, float lo, float hi, int ite
     }
 
     return (lo + hi) * 0.5f;
+}
+
+static std::pair<float, float>
+fitBezierLeastSquares(std::span<const std::pair<float, float>> samples,
+                      std::span<const float> param, float v0, float v1, float tStart, float tEnd)
+{
+    float duration = tEnd - tStart;
+
+    // In time axis:
+    //   P0.x = 0 (local), P2.x = duration, seg_mid.x = duration/2
+    float segMidT = duration * 0.5f;
+    float numT = 0.f, denT = 0.f;
+
+    // In value axis:
+    //   P0.y = v0, P2.y = v1, seg_mid.y = (v0+v1)/2
+    float segMidV = (v0 + v1) * 0.5f;
+    float numV = 0.f, denV = 0.f;
+
+    for (size_t i = 0; i < samples.size(); ++i)
+    {
+        float u = param[i];
+        float w = 4.f * u * (1.f - u);
+
+        float localT = samples[i].first - tStart;
+        float fixedT =
+            (1.f - u) * (1.f - u) * 0.f + u * u * duration - 2.f * u * (1.f - u) * segMidT;
+        numT += w * (localT - fixedT);
+        denT += w * w;
+
+        float fixedV = (1.f - u) * (1.f - u) * v0 + u * u * v1 - 2.f * u * (1.f - u) * segMidV;
+        numV += w * (samples[i].second - fixedV);
+        denV += w * w;
+    }
+
+    float pxStored = (denT > 1e-8f) ? (numT / denT) : duration * 0.5f;
+    float pyStored = (denV > 1e-8f) ? (numV / denV) : (v0 + v1) * 0.5f;
+
+    return {(duration > 1e-6f) ? std::clamp(pxStored / duration, 0.f, 1.f) : 0.5f,
+            std::clamp(pyStored, -1.f, 1.f)};
 }
 
 // ------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


src/common/dsp/modulators/MSEGModulationHelper.h
```diff
@@ -58,6 +58,28 @@ struct EvaluatorState
...
+    MSEGStorage::segment::Type type;
+    float cpv{0.f};
+    float cpduration{0.5f};
+    float maxResidual{1e10f};
+};
+
+// Fit a single segment spanning samples[start..end] (inclusive endpoints pinned
+// to v0/v1 at absolute times tStart/tEnd).  Returns the best-fitting type and
+// its control-point values. LFO Deform is assumed as 0 throughout.
+FitResult fitSegment(const std::vector<std::pair<float, float>> &samples, size_t start, size_t end,
```
mx: These const vector inputs should be spans. If you do that, you don't need "start" or "end" parameters either.

src/common/dsp/modulators/MSEGModulationHelper.h
```diff
@@ -58,6 +58,28 @@ struct EvaluatorState
 
     void seed(long l) { gen = std::minstd_rand(l); }
 };
+
+struct FitResult
```
mx, line 62: Since you're fitting a segment, why not return a MSEGStorage::segment directly, rather than a new object that contains a strict subset of its fields?

Also if you do that, you can make fitSegment a method of segment, rather than needing these free-floating functions under a helper file.

src/common/dsp/modulators/MSEGModulationHelper.cpp
```diff
@@ -32,6 +32,471 @@ namespace Surge
 namespace MSEG
 {
 
+// ------------------------------------------------------------
+// Internal helpers
+// ------------------------------------------------------------
+
+// Evaluate LINEAR shape at normalized phase frac in [0,1], cpv in [-1,1], df=0.
```
mx: Does this algorithm come from somewhere? Could use a reference, otherwise in the future this is just mystery load-bearing code that nobody understands.

If the intent is to duplicate what's done in valueAt(), then valueAt() should be updated to call these.

src/common/dsp/modulators/MSEGModulationHelper.cpp
```diff
@@ -32,6 +32,471 @@ namespace Surge
...
+    if (fabs(a) > 1e-3f)
+        cpline = (expf(a * frac) - 1.f) / (expf(a) - 1.f);
+
+    // df=0 so bend3 is identity — no additional deform step needed
+
+    return cpline * (v1 - v0) + v0;
+}
+
+// Evaluate SCURVE shape at normalized phase frac in [0,1], cpv in [-1,1], df=0.
+static float evalSCurve(float frac, float v0, float v1, float cpv)
```
mx: Same as above.

src/common/dsp/modulators/MSEGModulationHelper.cpp
```diff
@@ -32,6 +32,471 @@ namespace Surge
...
+    else
+        cpline = 1.f - 0.5f * cpline;
+
+    return cpline * (v1 - v0) + v0;
+}
+
+// Evaluate QUAD_BEZIER at timeAlongSegment, given segment duration and the
+// stored (on-curve) control point (cpduration normalized 0..1, cpv absolute).
+// df=0 so no pow() warp on t.
+static float evalBezier(float timeAlong, float duration, float v0, float v1, float cpdurNorm,
```
mx, line 161: Same as above.

src/common/dsp/modulators/MSEGModulationHelper.cpp
```diff
@@ -32,6 +32,471 @@ namespace Surge
...
+    // df=0: no pow() warp
+
+    return (1.f - bt) * (1.f - bt) * v0 + 2.f * (1.f - bt) * bt * py1 + bt * bt * v1;
+}
+
+// Compute max absolute residual for a set of (time,value) samples against
+// a user-supplied eval function that takes normalized phase [0,1].
+// tStart/tEnd define the segment boundaries.
+template <typename EvalFn>
+static float maxResidual(const std::vector<std::pair<float, float>> &samples, size_t start,
```
mx: Use a span!

src/common/dsp/modulators/MSEGModulationHelper.cpp
```diff
@@ -32,6 +32,471 @@ namespace Surge
...
+                         size_t end, float tStart, float tEnd, EvalFn eval)
+{
+    float duration = tEnd - tStart;
+    float maxR = 0.f;
+
+    for (size_t i = start; i <= end; ++i)
+    {
+        float frac = (duration > 1e-6f) ? (samples[i].first - tStart) / duration : 0.f;
```
mx: There's some magic going on here that makes this not exactly a "maxResidual" as documented by the function header. For example, the introduction of this epsilon on this line. Or frac getting truncated to [0, 1] below. Can you please explain what's going on here?

(For reference, a basic maxResidual() would just be std::ranges::max_element() over a std::ranges::transform()).

src/common/dsp/modulators/MSEGModulationHelper.cpp
```diff
@@ -32,6 +32,471 @@ namespace Surge
...
+        float predicted = eval(frac);
+        float r = fabs(samples[i].second - predicted);
+        if (r > maxR)
+            maxR = r;
+    }
+
+    return maxR;
+}
+
+// 1-D golden section search for minimum of f over [lo, hi].
```
mx, line 235: Can you reference the source for this algorithm so it's not just unknown magic to anyone coming in and reading it?

src/common/dsp/modulators/MSEGModulationHelper.cpp
```diff
@@ -32,6 +32,471 @@ namespace Surge
...
+        }
+    }
+
+    // ---- LINEAR ----
+    {
+        auto costFn = [&](float cpv) {
+            return maxResidual(samples, start, end, tStart, tEnd,
+                               [&](float frac) { return evalLinear(frac, v0, v1, cpv); });
+        };
+        float cpvRange = std::min(1.f, (float)(end - start) / 10.f);
```
mx: Can you explain this choice?

src/common/dsp/modulators/MSEGModulationHelper.cpp
```diff
@@ -32,6 +32,471 @@ namespace Surge
...
+            for (size_t i = 1; i < n; ++i)
+            {
+                float dt = samples[start + i].first - samples[start + i - 1].first;
+                float dv = samples[start + i].second - samples[start + i - 1].second;
+                param[i] = param[i - 1] + sqrtf(dt * dt + dv * dv);
+            }
+
+            float totalLen = param[n - 1];
+
+            if (totalLen > 1e-6f)
```
mx, line 395: Can you explain this choice of epsilon and why it's not so bad if it's not normalized by total length when epsilon falls below this?

src/common/dsp/modulators/MSEGModulationHelper.cpp
```diff
@@ -32,6 +32,471 @@ namespace Surge
...
+            //
+            //   weight_i = 4 * uᵢ * (1 - uᵢ)                (coefficient of P1_stored in B)
+            //   fixed_i  = (1-uᵢ)² * P0.x + uᵢ² * P2.x
+            //            - 2*uᵢ*(1-uᵢ) * seg_mid.x           (everything except P1_stored)
+            //   target_i = sample.x (for time) or sample.v (for value)
+            //
+            // Least squares: px_stored = Σ wᵢ*(targetᵢ - fixedᵢ) / Σ wᵢ²
+
+            // In time axis:
+            //   P0.x = 0 (local), P2.x = duration, seg_mid.x = duration/2
```
mx: Move the least-squares implementation out into its own function, that way you can provide a test that verifies it's correct.

mx (dismissed):

src/common/dsp/modulators/MSEGModulationHelper.cpp
mkruselj: This is somewhat of a simplification of stuff from valueAt because for curve fitting we don't need to take LFO Deform (df) parameter into account at all.

mx: I think it would be better to break the stuff in valueAt() into a function, and then call it with df = 0. That way we aren't repeating the logic (and possibly introducing a bug).

src/common/dsp/modulators/MSEGModulationHelper.cpp
mkruselj: https://en.wikipedia.org/wiki/Golden-section_search#Iterative_algorithm

I'll add a comment.

src/common/dsp/modulators/MSEGModulationHelper.cpp
mkruselj: After some back and forth, I noticed that when doing quicker motions, linear segments would overshoot the control point value and would look too concave/convex vs how the original gesture looked. The solution was to scale down the control point value depending on length of the segment. I'll add a comment.

src/common/dsp/modulators/MSEGModulationHelper.cpp
mkruselj: `frac` is converting abs time position (which is what's coming in as `samples`) to a normalized phase within the segment. Epsilon guard is there to protect against a degenerate 0-length segment case, returning 0 here means we treat everything as the startpoint.

src/common/dsp/modulators/MSEGModulationHelper.cpp
```diff
@@ -32,6 +32,471 @@ namespace Surge
...
+static float maxResidual(const std::vector<std::pair<float, float>> &samples, size_t start,
+                         size_t end, float tStart, float tEnd, EvalFn eval)
+{
+    float duration = tEnd - tStart;
+    float maxR = 0.f;
+
+    for (size_t i = start; i <= end; ++i)
+    {
+        float frac = (duration > 1e-6f) ? (samples[i].first - tStart) / duration : 0.f;
+        frac = std::max(0.f, std::min(1.f, frac));
```
mx: stdlib makes it simpler: frac = std::clamp(frac, 0.f, 1.f)

src/common/dsp/modulators/MSEGModulationHelper.cpp
mkruselj: If we're below that epsilon it means we're dealing with basically a zero length segment. This makes param[i] stay at almost 0, we skip the div by zero of the length normalization, and our parameter `u` of `4 * u * (1 - u)` also ends up basically very close to zero, which results in segment's midpoint the control point, which is as good fallback as any for a basically zero length segment case.

We do need the normalization, though.

src/common/dsp/modulators/MSEGModulationHelper.h
mkruselj: Hmm but it's not a strict subset when it contains an additional `maxResidual` member, which RDP requires (and MSEG segment itself does not)?

src/common/dsp/modulators/MSEGModulationHelper.cpp
mkruselj: This to me sounds like a separate refactor PR. `valueAt()` has a lot of other surrounding context that is completely unimportant for curve fitting (loop handling, inverting deform etc.), so this would have to be done very carefully, and we already have a sizable PR here. Agree?

src/common/dsp/modulators/MSEGModulationHelper.h
mkruselj: It's good separation of concerns, in my book...

src/common/dsp/modulators/MSEGModulationHelper.cpp
mx: I'm not really sure I understand the objection. What I'm suggesting is to take (for example) lines 271-408 in the original file and pull that into its own function, "evaluateSCurve()", with no changes. Then instead of re-writing the whole evaluation function here, you can just call that "evaluateSCurve()". This guarantees you're exactly the same as what Surge's evaluation is already doing and don't need to introduce any tests to ensure their equivalency, you get that for free. Do that for linear (which is the same as s-curve) and quadratic beziere, and you're done and you don't need any of this code here.

mkruselj: OK, I think I got it!

PR #8432, opened 2026-06-28
Add Find and Goto line to the Lua editors' Prelude tabs

nuoun:
- Switches Prelude tab to SurgeCodeEditorComponent and handle keybinds for find and go to line
- Grey out replace in the Prelude's context menu
- Also keeps keyboard focus when switching components/tabs

Assisted-By: Claude Opus 4.8


mkruselj:
IMO it would be a better UX if Replace didn't show at all for prelude tab.

nuoun:
Okay, I trimmed the menu to only show copy, select, find and goto.

mkruselj:
Perfect! Merging.

PR #8434, opened 2026-06-29
OSC: send outbound from the configured output port

### LINKED ISSUE #8097: OSC replys should be sent from OSC input port
When using OSC to control and get feedback from Surge, you have to define an output and an input port.
The input port let you control  Surge XT. The output port let you get some feedback from Surge, but for now those reply messages are sent from a random address. If you use several instances of Surge, it might be confusing, and also if you  want to take advantage of Non-Session Manager protocol.

It would be cooler if the replys would be sent from the output port, so that we can identify who's the emitter.

Thank you.

nuoun:
Closes #8097

Assisted-By: Claude Opus 4.8


mkruselj:
Nice!

PR #8443, opened 2026-07-04
A few updates to the Modern oscillator!

mkruselj:
- We can now disable each of the waveforms to save some CPU (except in Saw case, if Saw is disabled but Pulse isn't then saw still runs because it's needed)
- We can tune the suboscillator two octaves down
- Dynamic deactivation is used to gray out Width along with Pulse slider

### DIFF
--- a/src/common/dsp/oscillators/ModernOscillator.cpp
+++ b/src/common/dsp/oscillators/ModernOscillator.cpp
@@ -308,10 +308,25 @@ void ModernOscillator::init(float pitch, bool is_display, bool nonzero_init_drif
 template <ModernOscillator::mo_multitypes mtype, bool subOctave, bool FM>
 void ModernOscillator::process_sblk(float pitch, float drift, bool stereo, float fmdepthV)
 {
-    float submul = 1;
+    const bool sawActive = !oscdata->p[mo_saw_mix].deactivated;
+    const bool pulseActive = !oscdata->p[mo_pulse_mix].deactivated;
+    const bool thirdActive = !oscdata->p[mo_tri_mix].deactivated;
+    const bool needSaw = sawActive || pulseActive;
+    int numUnisonVoices = n_unison;
+
+    float submul = 1.f;
+
     if (subOctave)
     {
-        submul = 0.5;
+        const bool subdowntwo =
+            oscdata->p[mo_tri_mix].deform_type & ModernOscillator::mo_submask::mo_subdowntwo;
+
+        submul = 0.5f - (0.25f * subdowntwo);
+
+        if (!needSaw)
+        {
+            numUnisonVoices = 1;
+        }
     }
 
     float ud = oscdata->p[mo_unison_detune].get_extended(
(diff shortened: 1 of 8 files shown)


baconpaul:
OK I'm going to merge #8444 (and thanks for the review). But that will invalidate this PR pretty hard

I think we do 3 separate PRs in a row

1: deactivate triangle etc...
2: deactivate sync
3: Add double-sub, and modify menu presets to have deactivation and double sub on for some.

mkruselj:
@baconpaul I've just rebased this PR and redid the changes as necessary. It all works from my testing!

mkruselj:
...except hm I have no idea why are tests failing now :|

mkruselj:
Ah, I do. It's not running init_default_values(), so looks like the newly added deactivated states don't get properly set.

src/common/dsp/oscillators/ModernOscillator.cpp
```diff
@@ -425,9 +442,22 @@ void ModernOscillator::process_sblk(float pitch, float drift, bool stereo, float
                 {
                     double p = (phs[s] - 0.5) * 2;
                     double p3 = p * p * p;
-                    sB[s] = (p3 - p) * dpwOneOverSix;
 
-                    if (subOctave)
+                    if (needSaw)
```
baconpaul: ugh so this is bad. we've played a branch set per sample in the inner loop. Is this actually faster than just doing the calculations? I'm skeptical.

src/common/dsp/oscillators/ModernOscillator.cpp
```diff
@@ -467,20 +493,26 @@ void ModernOscillator::process_sblk(float pitch, float drift, bool stereo, float
...
-                // returns the final component value (1/(4 dsp^2) scale folded in).
-                double sawv = dpwSawComp(pfm, dsp, false);
+                double sawv = 0.0, sqrv = 0.0;
 
-                // Pulse = saw(phase) - saw(phase + width); the second saw turns
-                // around at a different phase so it gates independently.
-                double poff = pfm + pwHalf;
-                poff -= (poff >= 1.0);
-                double sqrv = dpwSawComp(poff, dsp, false) - sawv;
+                if (needSaw)
```
baconpaul: Again a per sample branch.

baconpaul (dismissed):
So I'm really worried about in-loop if clauses vs at the block level.

Also if we merge this I think we should add another preset like 'all waves' to the xml

but really the per-loop if - does this actually improve performance measurably in a real test environment? I'm skeptical that 20 flops > 1 branch even with prediction.

src/common/dsp/oscillators/ModernOscillator.cpp
mkruselj: Removed it.

src/common/dsp/oscillators/ModernOscillator.cpp
mkruselj: Removed it!

mkruselj:
@baconpaul Could be merged now?

PR #8447, opened 2026-07-09
Formula: Add dynamic names for LFO sliders

nuoun:
Makes it possible to optionally set the names of the four main LFO sliders from Lua for Formula modulators. Only works from the `init()` block during display script evals. Custom names are used inside the UI, modulation matrix and slider context menus.

Example use:

```lua
function init(state)
    state.labels = { rate = "Slider A", phase = "Slider B", deform = "Slider C",  amplitude = "Slider D" }
    return state
end
```

Assisted-By: Claude Opus 4.8


mkruselj:
LGTM! Awesome feature.

PR #8455, opened 2026-07-16
Async wavetable script generation

nuoun:
Draft, please don't merge yet.

[wip]
- Runs .wtscript generation on a persistent, storage-owned background worker so the UI does not block (see exceptions below)
- Adds an animated spinner component to show generation status in the the Oscillator Waveform Display and overlay's preview section
- Adds 3 new `WtGenService` regtests for the worker (the old WTS tests exercise the evaluator directly like before to specifically test the evaluator)

UI blocking paths (these use the worker but still hang the calling thread) that remain are:
- Export to file (have the job tail do the file write is possible, but not sure it's worth it)
- OSC (the OSC thread does not have an `idle()` to poll the worker's result so this one is genuinely tricky)

Todo:
- Some UI tweaks (overlay refreshing
- Make tests more robust
- Restore WT load menu from before the icon addition (can be handled in different PR)
- Add an edit icon to indicate a WT is script or not on the OWD display (also can be different PR)

Assisted-By: Claude Opus 4.8.


src/common/dsp/WavetableScriptEvaluator.cpp
```diff
@@ -62,6 +60,36 @@ struct LuaWTEvaluator::Details
     lua_State *L{nullptr};
     int snapshotRef{LUA_NOREF}; // Registry ref for cached wt.snapshot table
 
+    // When pushSnapshotTable is set builds wt.snapshot from this immutable bundle rather than
```
mx: I'm finding this split of changes where "if <this new thing> it works this way, otherwise the old way" to be fraught. I think this would be much more maintainable in the future, and readable now, if it **only** worked the new way, and you updated the parts that called the old way to use the new way. So for this, it should only ever build the snapshot from an immutable bundle -- stuff that currently touches live oscdata should instead copy it in or out. If there's only one way to do things, life is a lot simpler.

mx: Also you never set pushSnapshotTable; that's a function you call.

src/common/dsp/WavetableScriptEvaluator.cpp
```diff
@@ -453,7 +496,9 @@ LuaWTEvaluator::frame_t LuaWTEvaluator::getFrame(size_t frame)
 #endif
 }
 
-bool LuaWTEvaluator::populateWavetable(wt_header &wh, float **wavdata)
+bool LuaWTEvaluator::populateWavetable(wt_header &wh, std::unique_ptr<float[]> &wavdata,
```
mx: Taking a reference to a unique_ptr is very strange and very unclear about what ownership is intended. Can this be changed to have more clear ownership semantics? If it's not supposed to own, you can use a span.

src/common/dsp/WavetableScriptEvaluator.cpp
```diff
@@ -453,7 +496,9 @@ LuaWTEvaluator::frame_t LuaWTEvaluator::getFrame(size_t frame)
...
+                                       std::vector<std::vector<float>> *outFrames,
+                                       const std::function<bool()> &cancel, bool previewOnly)
```
mx: nit: Rename cancel to canceled. cancel() is something you do, canceled() is a state you're in.

src/common/dsp/WavetableScriptEvaluator.cpp
```diff
@@ -572,11 +619,25 @@ LuaWTEvaluator::parseWtscript(const fs::path &filename, SurgeStorage *storage,
         where wtscript_header = { tag "wts1", xmlsize, blobsize }
     */
 
+    // Route parse errors to errorOut when the caller must surface them itself (the OSC thread must
+    // not call reportError). Message-thread callers pass errorOut == nullptr and get the
+    // direct dialog.
+    auto fail = [&](const char *msg) -> std::optional<WtscriptData> {
```
mx: Don't you already have an emit_error method? Why the duplicated functionality here?

src/common/dsp/WavetableScriptEvaluator.h
```diff
@@ -38,6 +42,36 @@ namespace Surge
...
+        int nsamples{0};
+        std::vector<float> data; // flat, nframes * nsamples; data[t * nsamples + i]
+    };
+    std::array<Slot, n_wt_snapshots> slots;
+    uint64_t version{0};
+};
+
+// Copy the live wtSnapshots of osc into a fresh bundle. Caller must hold storage->wtSnapshotMutex
+// (it reads the snapshot sample data, which mutation sites write under that mutex).
+std::shared_ptr<SnapshotBundle> buildSnapshotBundle(const OscillatorStorage &osc);
```
mx: These are effectively constructors, so make these static members of SnapshotBundle.

src/common/SurgeStorage.h
```diff
@@ -1738,6 +1755,24 @@ class alignas(16) SurgeStorage
...
 
     std::mutex waveTableDataMutex;
+
+    // Guards oscdata.wtSnapshots mutations + wtSnapshotsVersion bumps against the
+    // copy-on-mutate SnapshotBundle rebuild. Distinct from waveTableDataMutex; the two are
+    // never held nested except captureWavetableSnapshot (waveTableDataMutex outer,
+    // wtSnapshotMutex inner).
+    std::mutex wtSnapshotMutex;
+
+    // Per-(scene,osc) config-generation token. Bumped INSIDE a waveTableDataMutex region
```
mx: Since all checks of this happen inside of a mutex, why are you using atomics for this?

src/common/dsp/WtGenService.h
```diff
@@ -0,0 +1,211 @@
...
+    // Poll-based Generates (overlay Generate, drop/menu) set this so a newer same-osc Generate
+    // supersedes a still-pending older one, which the worker drops before running. Never set on
+    // block-waiting Generates (export/OSC): a parked waiter needs its own result. Ignored for
+    // Preview (its path supersedes separately).
+    bool supersedable{false};
+
+    // Input, const once submitted. Read by the worker after it claims the job.
+    std::string script;
+    int resolution{2048}, frameCount{10};
+    std::shared_ptr<const SnapshotBundle> snapshot; // immutable, refcount-shared
```
mx: The presence of a shared_ptr in a struct like this makes me nervous; this is designed to be communicated between threads so you have to audit accesses carefully, but a shared_ptr can get spilled out to just about anywhere (that's the point of it).

src/common/dsp/WtGenService.h
```diff
@@ -0,0 +1,211 @@
...
+
+    // Generate: publish wt buffers into ->wt, nullptr == export (build into exportOut).
+    OscillatorStorage *generateTarget{nullptr};
+
+    // Status barrier: release on store in the worker, acquire on load in the poller/waiter.
```
mx: This isn't a barrier at all, and has no release or acquire semantics, this is a confusing comment.

src/common/dsp/WtGenService.h
```diff
@@ -0,0 +1,211 @@
...
+    std::atomic<Status> status{Status::Pending};
+
+    // Block-wait handshake for synchronous submitters (export, OSC). These wait on
+    // done.get_future(), the worker fulfills it right after the status store. Pollers ignore it.
+    // Generates only (never superseded).
+    std::promise<void> done;
```
mx: This is what std::latch is designed for.

src/common/dsp/WtGenService.cpp
```diff
@@ -0,0 +1,369 @@
...
+    }
+
+    queueCV.notify_one();
+
+    return job;
+}
+
+void WtGenService::submitBlocking(const std::shared_ptr<WtGenJob> &job)
+{
+    auto fut = job->done.get_future(); // retrieve before submit so set_value() can't race ahead
```
mx: Whoa, hold up here. Why can't we rely on the future that is the sole property of this job after submission exactly? That seems like a serious problem. Also seems like a good reason to use a latch again.

mx (changes requested):
Two additional notes:
(1) I've only looked at the stuff under src/common/dsp so far, since changes there might flow downstream.
(2) There's enough mutex use here that you should consider using llvm's thread safety annotations. That'll cause improper locking to emit compiler errors.

src/common/dsp/WavetableScriptEvaluator.cpp
nuoun: I made `pushSnapshotTable()` a single unconditional path and fixed the comment. All the code paths including the reg test already used the immutable bundle.

src/common/dsp/WavetableScriptEvaluator.cpp
nuoun: Valid and it does own, `populateWavetable()` dynamically allocates the buffer and hands ownership back to the caller. Changed it to return a PopulatedWavetable struct by value instead.

src/common/dsp/WavetableScriptEvaluator.cpp
nuoun: Good call, I moved `emitError()` out of the Details struct and made both the worker and the parser use it, dropping the lambda and extra deferredError buffers.

src/common/dsp/WavetableScriptEvaluator.h
nuoun: Done, moved both factories onto SnapshotBundle as static members: `SnapshotBundle::build` and `SnapshotBundle::current`

src/common/SurgeStorage.h
nuoun: Not all access is under the mutex, the token is captured lock-free at submit time in two code paths (`makeGenJob`, `makeLiveGenerateJob`) on the message/OSC thread, the only way to drop the atomic would be to also take waveTableDataMutex around those two captures, which trades the lock-free submit for a lock the audio thread also takes, just to read the token.
Updated the comment to describe this better.

src/common/dsp/WtGenService.h
nuoun: This seems genuinely tricky, I audited it and it should be safe because the bundle is `const` once published (built + version-stamped under `wtSnapshotMutex` while still uniquely owned, never mutated after), and no single shared_ptr instance is shared across threads: job->snapshot is written pre-submit and only read by the worker (pollers never touch it), and osc.wtSnapshotBundle is only ever accessed inside current() under the mutex. So the only cross-thread ops are copies/destructions of distinct instances, which the atomic refcount handles.
Added a comment to make this clearer.

src/common/dsp/WtGenService.h
nuoun: Rewrote the comment.

src/common/dsp/WtGenService.h
nuoun: Was going to say std::latch helps a lot here but it's unavailable in our macOS 10.15 target.

`/Users/runner/work/surge/surge/src/common/dsp/WtGenService.cpp:181:15: error: 'wait' is unavailable: introduced in macOS 11.0`

src/common/dsp/WtGenService.cpp
nuoun: Good point, no reason to retrieve the future before submitting, moved it below `submit()` to make that clear.

src/common/dsp/WtGenService.h
```diff
@@ -0,0 +1,290 @@
...
+    bool generateSuperseded(const Ticket &t) const
+    {
+        return t.id < latestGenerate[t.denseIdx].load(std::memory_order_relaxed);
+    }
+
+  private:
+    std::array<std::atomic<int>, n_scenes * n_oscs> busyCount{};     // outstanding jobs
+    std::array<std::atomic<int>, n_scenes * n_oscs> generateCount{}; // outstanding live generates
+    std::array<std::atomic<uint64_t>, n_scenes * n_oscs> latestPreview{};  // newest Preview ID
+    std::array<std::atomic<uint64_t>, n_scenes * n_oscs> latestGenerate{}; // newest supersedable ID
```
mx, line 212: As previously discussed in discord: we can drop the atomic arrays to a uint64 and examine everything under a mutex lock.

src/common/dsp/WtGenService.h
```diff
@@ -0,0 +1,290 @@
...
+    WtGenJobRequest &operator=(const WtGenJobRequest &) = delete;
+};
+
+// Outputs from one generation. Written by the worker, then read by the submitter through the
+// future once it is ready (the future's shared state is the sole synchronization: the caller
+// cannot observe these fields before the worker has finished writing them). Move-only.
+struct WtGenJobResponse
+{
+    bool ok{false};        // populateWavetable succeeded (frames / exportOut are valid)
+    bool published{false}; // live-osc Generate only: true if BuildWT ran into generateTarget->wt
```
mx: I don't understand this comment, what does it mean if BuildWT "ran into" generateTarget->wt?

src/common/dsp/WtGenService.h
```diff
@@ -0,0 +1,290 @@
...
+// polls, and the newest Preview/Generate ID the supersede checks compare against.
+// Both are atomic, so queries are safe from any thread; begin() is the exception and must run
+// under queueLock.
+class WtGenStatus
+{
+  public:
+    // Move-only accounting handle. Its destructor decrements the busy (and, for a live generate,
+    // the generate) count exactly once, so the worker cannot leak a count on any exit path. Also
+    // carries the requestId + denseIdx the worker needs for the supersede tests.
+    class Ticket
```
mx, line 140: We actually have a class in sst-cpputils called `scope_guard` that acts as an arbitrary RAII holder. It doesn't support std::move right now but I can add that (or you can) easily enough. So you don't need to reinvent the wheel with this. You just give it your callback and it executes it on destruction.

mx: Of course if you find this simpler you can keep it.

src/common/dsp/WtGenService.h
mx: I didn't really understand why we needed an array of these anyway (shouldn't we only care about the overall outstanding number of jobs?), but I guess it doesn't matter if you're going to nix the arrays.

src/common/dsp/WtGenService.h
```diff
@@ -0,0 +1,290 @@
...
+    // so no explicit invalidation is needed: a change to script/res/frames/snapshots yields a
+    // different key and misses. Bounded to n_scenes*n_oscs entries, each overwritten on regen.
+    struct PreviewCacheEntry
+    {
+        std::vector<std::vector<float>> frames;
+        int frameCount{0};
+        uint64_t key{0}; // 0 == empty
+    };
+    std::array<PreviewCacheEntry, n_scenes * n_oscs> previewCache;
+    std::mutex previewCacheMutex;
```
mx: Can you explain why we have the previewing cache, in a comment or otherwise? Reason I'm asking is this feels like it seriously complicates things, since now you have workers performing thread communication outside "receive job, return result", there's this side channel that the threads communicate in and that pretty much destroys all the benefits of the simple restricted communication. Which I think you noticed with all the code that is centered around supporting this.

mx: I *think* what you could do to maintain this cache in a way that didn't require all this locking and communication is plop the cache into SurgeStorage, and then have whatever receives the future put it in there, rather than have the thread handle it.

Would that still need a mutex? If so I guess you don't really gain anything doing that.

src/common/dsp/WtGenService.h
```diff
@@ -0,0 +1,290 @@
...
+{
+    WtGenMode mode{WtGenMode::Preview};
+    int scene{0}, osc{0};
+
+    std::string script;
+    int resolution{2048}, frameCount{10};
+
+    // Immutable, refcount-shared snapshot copy captured on the submitting thread (under
+    // wtSnapshotMutex). The const pointee is safe to read from the worker.
+    std::shared_ptr<const SnapshotBundle> snapshot;
```
mx, line 67: I don't quite follow why this needs to be a shared_ptr anymore. Can it be owned by the job request instead, and a reference given to the evaluator? There's no risk of the reference becoming invalidated by the worker thread's lifetime ending first, right?

src/common/dsp/WtGenService.h
```diff
@@ -0,0 +1,290 @@
...
+
+    // Generate: publish target. nullptr == export (the worker builds into the response's
+    // exportOut).
+    OscillatorStorage *generateTarget{nullptr};
```
mx, line 71: Is this needed? To me feels like a separation of concerns issue. The caller should do the publishing, the job loop shouldn't need to have any special case logic for it.

mx (changes requested):
At this point I think most of my comments are centered around questions with me not understanding some requirements of the system. There's also the "can we replace all our atomics with the existing mutexes" question discussed in the discord.

src/common/dsp/WtGenService.h
nuoun: Unlike the publish token I think these are better off staying as atomics because their readers (OWD/WTSE poll timers, the worker's per-frame supersede check) tolerate stale values by design, so a lock would add blocking without adding any guarantee. Suggestions maybe?

src/common/dsp/WtGenService.h
nuoun: "built its output into" I'll fix up that comment.

src/common/dsp/WtGenService.h
nuoun: I'll have a look, would be nice to use `scope_guard` if it helps.

src/common/dsp/WtGenService.h
nuoun: Without the cache every overlay refresh (open/close, switch oscillators, etc.) would be a new job.
It only needs the mutex now because of the OSC handler which still crosses threads. If we accept OSC generates not filling the cache then it all lands on the message thread, the mutex goes away, and the cache can become a plain GUI-owned struct, maybe not even in SurgeStorage.

src/common/dsp/WtGenService.h
nuoun: The shared_ptr isn't for the worker's lifetime, it's the copy-on-mutate share. `SnapshotBundle::current` only rebuilds when the snapshots actually changed: the osc caches its bundle (`osc.wtSnapshotBundle`) and each submit just refcount-shares that same immutable instance while the version is unchanged. Owning it by value in the request would mean deep-copying the snapshot data on every submit. I do wonder if this can be simplified but this seems the minimal form.

src/common/dsp/WtGenService.h
nuoun: It's a fair trade but two downsides to this. First, with caller-publish an abandoned response silently drops the Generate (close the GUI mid-evaluation and the osc never gets the wavetable the user asked for). Second, the token-check + BuildWT transaction under waveTableDataMutex would be repeated in all three consumers (overlay poll, drop/menu tail, OSC handler) instead of living in one place.

src/common/dsp/WtGenService.h
mx: I think OSC not updating the UI is okay. That's all it does right?

src/common/dsp/WtGenService.h
mx: Okay that's the part I missed. Seems fine then.

src/common/dsp/WtGenService.h
mx: Good points all around.

mx (approved):
Amazing, well done.

PR #8463, opened 2026-07-18
Minor Keyboard Tweaks

ironcross32:
This addresses feedback from @mkruselj from Discord regarding the accessible MSEG editor work:
* CTRL+up/down snaps regardless of whether vertical snap is on or off
* bare up/down moves by 0.05 per press
* SHIFT+up/down move by 0.01 per press
* Keyboard navigation now respects the constraints of the move switch being set to draw, accessible announcement if X axis movement is attempted
*

Assisted by Claude Fable 5.

### DIFF
--- a/src/surge-xt/gui/overlays/MSEGEditorAccessibleKeyboard.cpp
+++ b/src/surge-xt/gui/overlays/MSEGEditorAccessibleKeyboard.cpp
@@ -45,6 +45,7 @@ constexpr int kTypein2DChooser = -1;
 
 // matches MSEGCanvas::TimeEdit
 constexpr int kTimeEditShift = 1;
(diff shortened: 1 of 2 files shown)


mkruselj:
Thank you for this!

mkruselj:
BTW @ironcross32 the same snapping behavior should be for left/right arrows. That doesn't seem  like it's implemented in this commit?

ironcross32:
> BTW @ironcross32 the same snapping behavior should be for left/right arrows. That doesn't seem like it's implemented in this commit?

No, I left horizontal snapping alone. I can fix though.

ironcross32:
OK, snapping behavior implemented for CTRL+left/right.

PR #8478, opened 2026-07-21
Show only selected effect's presets in FxMenu

mkruselj:
Previously, if we clicked a slot in the FX grid or the dropdown menu right below it we'd get mostly the same menu (minus some verbiage differences).

Now, if we selected a slot that already has a loaded effect, we will see only that effect's presets, and reduced FUNCTIONS column (FX chain stuff is hidden here since this menu now shows only stuff related to this, currently selected slot).

Also modified how prev-next jog buttons work. Previously they simply traversed all the presets, so they could change which effect was loaded. Now, their jogging is constrained only to the currently selected effect. Much more logical!

Assisted-By: Claude Sonnet 5 <noreply@anthropic.com>

### DIFF
--- a/src/surge-xt/gui/widgets/XMLConfiguredMenus.cpp
+++ b/src/surge-xt/gui/widgets/XMLConfiguredMenus.cpp
@@ -1324,6 +1324,11 @@ void FxMenu::jogBy(int dir)
     }
 
     auto matches = [this, myType](int i) {
+        if (i < 0 || i >= (int)allPresets.size())
+        {
+            return false;
+        }
+
         const auto &p = allPresets[i];
         return !p.isSeparator && !p.isSectionHeader && p.itemType == myType;
     };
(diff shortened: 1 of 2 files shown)


src/surge-xt/gui/widgets/XMLConfiguredMenus.cpp
```diff
@@ -1239,6 +1309,77 @@ void FxMenu::loadByIndex(const std::string &name, int index)
...
+        const auto &p = allPresets[i];
+        return !p.isSeparator && !p.isSectionHeader && p.itemType == myType;
+    };
+
+    auto idx = selectedIdx;
+
+    // If we don't currently have a valid starting point within this effect's own
+    // presets (e.g. nothing loaded yet, or the effect type just changed), jump to
+    // the first matching preset instead of jogging from an unrelated position.
+    if (idx < 0 || idx >= (int)allPresets.size() || !matches(idx))
```
mx: This isn't currently a problem but it's a little dangerous here. The only thing stopping a call to matches(-1) which then would execute allPresets[-1] is the short circuit evaluation at the start of this. That sort of raw memory check really needs a belt-and-suspenders kind of safety; matches itself should check for the validity of its parameter. I'd suggest moving those checks directly into matches, so it can't ever be called with a bad value and do something bad.

mx (changes requested):

src/surge-xt/gui/widgets/XMLConfiguredMenus.cpp
mkruselj: Noted and modified the code accordingly!

mx (approved):

Issue #7734, opened 2024-07-30
MSVC Formula Modulator tests segv but only in debug build


baconpaul:
A consistent release build of the unit tests works, but an MSVC mixed release/debug or debug build will generate a luajit crash / segv in the formula modulator tests.

nuoun:
Backtrace from msvc for x64-debug target (ignore the extra errors, test-runner could not find resource folder)

```
Unhandled exception at 0x00007FF7C6311177 in surge-testrunner.exe: 0xC00000FD: Stack overflow (parameters: 0x0000000000000001, 0x000000682A009000).

call stack:
 	surge-testrunner.exe!__chkstk() Line 109	Unknown
>	surge-testrunner.exe!CATCH2_INTERNAL_TEST_27() Line 466	C++
 	surge-testrunner.exe!Catch::TestInvokerAsFunction::invoke() Line 6156	C++
 	surge-testrunner.exe!Catch::TestCaseHandle::invoke() Line 7265	C++
 	surge-testrunner.exe!Catch::RunContext::invokeActiveTestCase() Line 5500	C++
 	surge-testrunner.exe!Catch::RunContext::runCurrentTest(std::string & redirectedCout, std::string & redirectedCerr) Line 5463	C++
 	surge-testrunner.exe!Catch::RunContext::runTest(const Catch::TestCaseHandle & testCase) Line 5195	C++
 	surge-testrunner.exe!Catch::`anonymous namespace'::TestGroup::execute() Line 1021	C++
 	surge-testrunner.exe!Catch::Session::runInternal() Line 1243	C++
 	surge-testrunner.exe!Catch::Session::run() Line 1174	C++
 	surge-testrunner.exe!Catch::Session::run<char>(int argc, const char * const * argv) Line 5320	C++
 	surge-testrunner.exe!runAllTests(int argc, char * * argv) Line 64	C++
 	surge-testrunner.exe!main(int argc, char * * argv) Line 98	C++
 	[External Code]	

output:
Exception thrown at 0x00007FF82C7FB699 in surge-testrunner.exe: Microsoft C++ exception: Catch::TestFailureException at memory location 0x000000682A0FD150.
Exception thrown at 0x00007FF82C7FB699 in surge-testrunner.exe: Microsoft C++ exception: Catch::TestFailureException at memory location 0x000000682A0FD150.
[...]

PR #8479, opened 2026-07-22
Heap-allocate SurgeStorage in Lua unit tests

nuoun:
Also:
- Drop the MSVC Debug /STACK workaround (See: #7734)
- Fix failed assert in [formula] tests 
- Make both factory wavetable tests skip the user data folder
- Fix "All Factory .wtscript Files Validate" test failing silently

Assisted-By: Claude Opus 4.8.


mkruselj:
LGTM!

PR #8486, opened 2026-07-24
LuaEditors: Control group focus navigation and accessibility

nuoun:
Formula and Wavetable Script editors:

- Alt+. / Alt+, now move keyboard focus between the visible code display and the control area
- Code selection switches announce their options to a screen reader instead of a bare value
- Removed the tab switch editor focus grabs in `showModulatorCode` / `showPreludeCode`, which
  the switch accessibility cells now override
- Explicit focus orders for the Formula control area
- Debugger Init and Step come after Show/Hide in the tab order
- Show/Hide no longer re-initializes the debugger when closing (missing break)
- Wavetable View and Display Mode are no longer reachable by Tab

MultiSwitch:

- Fixed an always-true condition and refactored to a switch case
- Home and End go to the first and last option instead of End just incrementing
- Return selects the focused cell, or steps to the next option and wraps

Assisted-by: Claude Opus 4.8

### DIFF
--- a/src/surge-xt/gui/overlays/LuaEditors.cpp
+++ b/src/surge-xt/gui/overlays/LuaEditors.cpp
@@ -1060,8 +1060,43 @@ void SurgeCodeEditorComponent::mouseDrag(const juce::MouseEvent &event)
     repaint();
 }
 
+// Hand the control group focus keys to the overlay wrapper, which owns that navigation,
+// before the code editor takes them as text input.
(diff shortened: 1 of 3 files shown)


mkruselj:
I'd retain Debug button where it was but adjust the tabbing to make more sense for a11y.

nuoun:
Done, set the tab order instead.

PR #8488, opened 2026-07-24
Formula: Add host transport to state table

nuoun:


mkruselj:
I think we don't need is_recording tbh...

nuoun:
Only `is_playing` then.

PR #8489, opened 2026-07-24
Add export frames options for wavetables

nuoun:
Adds a menu option to export each wavetable frame as its own mono 32-bit float WAV file in a selected folder, available for both scripted and regular wavetables from the Oscillator Waveform Display menu (and only for WT scripts in the editor).
Frames land in a [WT Name] - Frames subfolder (numerically incremented to [WT Name #] if it already exists), so exports never overwrite or mix.

- Also allows for UTF-8 strings and paths so unicode characters decode properly
- Also fixes regular wavetables trying to export through the script evaluator.
- Removes unused `export_wt_wav_portable` overload from WAVFileSupport.cpp

Assisted-by: Claude Opus 4.8


mkruselj:
Suggestion: frames should go in a folder named the same as wavetable, and contain frames named as `<wtname> #n`, n being 1-based. If by some chance there is no wavetable name, use `Exported WT n` for folder name, and increment it if another one already exists.

nuoun:
Oh messed up the formatting on that but yes, it does do that. An example, export Sine HQ wavetable and it creates:

Sine HQ_frames\Sine HQ_001.wav to Sine HQ_016.wav

I don't think a wavetable can not have a name but let me double check.

mkruselj:
Yeah my nitpick is about using "_frames". The folder name should be without any suffixes, and individual WAVs shouldn't have any underlines, just number the frames like `Sine HQ #1`, `Sine HQ #2` etc.

Very old patches can have no WT names, this was one of the early additions when Surge was open-sourced. A bunch of Claes' factory patches don't have WT names, for example.

nuoun:
Ok it gets a bit weird if the scripter sets the name as just a space but it still outputs valid filenames, empty name is handled (sets it to "Wavetable") and I have faith in `juce::File::createLegalFileName()`

nuoun:
That does seem nitpicky tbh, I think _frames in the subfolder is informative, idk about you but I have plenty of other folders in my goto export wav folder, it also adds an underscore with a number if the folder already exists (because deleting existing folders or trying to overwrite potentially hundreds of files does not sound great) so no complete escaping the suffix here.

mkruselj:
I don't like the underscore and lowercase, basically. 😛 Too programmery, ugly for your regular Joe.

I just think we can be just fine without the suffix.

mkruselj:
OK so to round it all up, here's my proposal through some worked examples:

Wavetable Name: Sine FM
Folder Name: Sine FM - Frames
Individual Filenames: Sine FM #n (I would like to avoid repetitiveness of "Frame x" here)

If a folder "Sine FM" already exists

Folder Name: Sine FM 2 - Frames
Individual Filenames: Sine FM 2 #n

If a folder "Sine FM 2" already exists, increment number by one.

If a folder "Sine FM 2 - Frames" already exists, trim " - Frames", increment the number, re-add " - Frames".

If wavetable name is accidentally whitespace only, default to "Exported WT", autoincrement the number if needed.


Make sense?

nuoun:
Good suggestions, I kept the zero padding of the file name numeration (since it's genuinely useful for sorting in lists/file explorer) and otherwise implemented as suggested.

Two more changes: fixed UTF-8 names and paths to allow unicode characters and removed an `export_wt_wav_portable` overload that is dead code.

There's a few more existing issues that came up reviewing WAVFileSupport but I'll tackle those in another PR.

mkruselj:
I'm really not sure zero padding helps, it just creates more visual noise with all those zeroes around, and if your file explorer cannot sort numbers naturally, it's broken IMO 😛

nuoun:
If you want to drag and drop an entire selection without it messing up the order you need the zero padding, so yes it does help.

mkruselj:
I think this depends on app receiving the drop, if it sorts the input drop files or not...

PR #8496, opened 2026-07-26
Alert boxes: Accessibility and fixes

nuoun:
- Tab now stays inside the dialog instead of escaping into the UI behind it
- Focus lands on the message, so screen readers announce the prompt
- "Don't ask me again" checkbox no longer covers the OK/Cancel buttons and eats their clicks
- Fixed a crash when the Confirm Patch Change dialog opens the Confirm Patch Loading dialog through a callback
- Fixed a rare condition where ticking "don't ask me again" and then cancelling would permanently disable the patch/category jog keybinds

Assisted-by: Claude Opus 5


mkruselj:
Good fixes!!

PR #8513, opened 2026-07-30
Highlight tickboxes on focus and hover

nuoun:
image

Assisted-by: Claude Opus 5


mkruselj:
Hrm I wouldn't use blue, I would have just used .brighter() instead...

nuoun:
Seems a bit odd to highlight differently than any other UI element, then again the contrast sucks in the classic theme.
image

mkruselj:
Yeh different skins do it differently. I use Royal Surge and I'd expect the checkboxes to just go a bit brighter on hover...

nuoun:
Ok took the numfield hover color since that one seems to work better accross skins and switched to highlighting the tickbox interior only, contrast still kind of sucks with the classic theme but better than nothing.

mkruselj:
Well Classic theme is an eyesore IMO anyways, I'm glad it's not gonna be there in XT2 😁

PR #8518, opened 2026-08-10
Add Formula presets

nuoun:
- 28 new Formula presets
- Shared table tutorial
- Init Arpeggiator template
- Formula presets subfolder structure
- Set names, some slider labels and a bit of cleanup for existing Formula presets


mkruselj:
Awesome work!

Issue #304, opened 2019-01-15
Feature Request: Independent pan per OSC


baconpaul:
I noticed that each OSC doesn't have a pan, so one of my favorite tricks (to widen and narrow pan on a control or gesture between a pair of detuned oscillators) isn't available. Would be great if it was!

sense-amr:
yes this would be an awesome addition too :) just to add sonic capabilities especially considering that if added they could presumably be automated :)

sense-amr:
heheh hold yourself back bacon paul !

baconpaul:
Honestly the biggest problem with implementing this is that there is no place to put the control in the UI!

sense-amr:
i found a place for it :) 
![surge_placeforpan](https://user-images.githubusercontent.com/1504019/51185185-8ad44a00-192a-11e9-8cc0-0e1b82a2b0a3.jpg)

baconpaul:
To make it modulatable easily we need it to be one of the regular sliders (kinda). Also to have the UI be consistent. And when you are in wavetable mode that space gets used for something else doesn't it? But I think you're mostly joking. And anyway after 1.6 we have an idea how to make the layout way more flexible using an SVG file.

But first: zoom.

sense-amr:
Hmm no just checking through all the options for Waves and wavetable that area is always clear and available .. i would think its not a bad place for a pan ..

baconpaul:
Yeah I was thinking maybe a little round knob below the solo and filter routing buttons could work also. But then we need a modulatable knob widget. Which we also don't have (but could write - and it might be useful in other places). The problem with putting it where you say is there's not a natural modulation gesture which fits the rest of the modulation approaches.

But this is for after 1.6 anyway! We can ask @kurasu @cyanit and @itsmedavep for opinions if we add this of where the control could go.

baconpaul:
Link this to #1117

baconpaul:
Link #1150 here also.

PR #8522, opened 2026-08-13
Parse type-ins independently of the active locale

### LINKED ISSUE #7574: Locale issue with typeins
Miętus on Discord reports that our typeins show him decimal numbers with dot as decimal separator, but this text entry doesn't work properly (it truncates the decimal part). They have to use comma instead, which works, but also means they manualyl have to replace the dot every time typein shows up. They are using Polish locale on Windows.

For me, Croatian locale on Windows, it's the other way around. Dot works just fine but comma is not parsed and truncates the decimal part to 0.

manuaudio:
Fixes #7574.

Type-ins run through `std::atof`, which follows `LC_NUMERIC`. In a comma-decimal locale `0.5` stops at the dot and silently becomes `0`; in a dot-decimal locale `0,5` does the same. Either way the fractional part disappears with no error, so you get a value you didn't type. That's how it was reported on Discord for Polish and Croatian layouts.

We already keep patch files locale independent with `float_to_clocalestr`, so this adds the matching `clocalestr_to_double` beside it. It takes either separator and parses through a classic-locale stream, and the type-in paths in `Parameter.cpp` use it.

One spot needed care: the `extend_range` branch relied on `std::stof` throwing to flag bad input, so a straight swap would have turned invalid input into `0`. It uses a checked overload and keeps the old behaviour. Both of its error sentinels (`val_min.i - 1` and `- 2`) land in the same out-of-range branch and both read as "below minimum", so they collapse to one without any visible change.

### Testing

New case in `UnitTestsPARAM.cpp` covering both separators under German, Polish and Croatian, skipping any locale that isn't installed, plus a `ct_decibel` type-in round trip through `Parameter`.

I checked it fails without the fix rather than just passing with it. Backing out the separator handling gives:

```
UnitTestsPARAM.cpp:141: FAILED:
  REQUIRE( p.val.f == Approx(-6.5f).margin(0.001) )
with expansion:
  -6.0f == Approx( -6.5 )
with messages:
  locale de_DE.UTF-8
  typein -6,5
```

which is the reported bug. With the fix, 44 assertions pass. Full test runner is green: 127 cases, 5687316 assertions. `clang-format` clean.

Worth flagging: a thousands separator like `1,234.5` now reads as `1.234` where `atof` gave `1`. Both are wrong, and I didn't think Surge type-ins were a place anyone types thousands separators, but say the word if you'd rather I reject that shape outright.

I don't have a Windows box in a comma-decimal locale, so this is verified through the unit tests and `std::locale::global` rather than against the original reporter's setup.

Assisted-By: Claude Code (Claude Fable 5)

### DIFF
--- a/src/common/UnitConversions.h
+++ b/src/common/UnitConversions.h
@@ -28,6 +28,7 @@
 #include <string>
 #include <cstring>
 #include <fmt/core.h>
+#include "sst/basic-blocks/mechanics/string-ops.h"
 #include <fmt/format.h>
 
 static float env_phasemulti = 1000 / 44100.f;
@@ -74,38 +75,17 @@ inline std::string float_to_clocalestr(float value)
  */
 inline double clocalestr_to_double(const char *s, bool *parsed)
 {
-    if (parsed)
-    {
-        *parsed = false;
-    }
-
-    if (!s)
-    {
-        return 0.0;
-    }
-
-    std::string t{s};
-    std::replace(t.begin(), t.end(), ',', '.');
-
-    // strtod is LC_NUMERIC-dependent, so parse through a classic-locale stream
-    std::istringstream is{t};
-    is.imbue(std::locale::classic());
-
-    double res{0.0};
-    is >> res;
-
-    if (is.fail())
-    {
-        // match atof, which returns 0 for input that isn't a number
-        return 0.0;
-    }
+    // sst-basic-blocks owns the parsing now; this just adapts the shape the
+    // type-in call sites here expect - a double plus a did-it-parse flag, with
+    // zero for input that isn't a number, which is what std::atof used to give.
+    const auto v = sst::basic_blocks::mechanics::parseNumber(s);
 
     if (parsed)
     {
-        *parsed = true;
+        *parsed = v.has_value();
     }
 
-    return res;
+    return v.value_or(0.0);
 }
 
 inline double clocalestr_to_double(const char *s) { return clocalestr_to_double(s, nullptr); }
--- a/src/common/UnitConversions.h
+++ b/src/common/UnitConversions.h
@@ -24,6 +24,8 @@
 #include <stdio.h>
 #include <algorithm>
 #include <locale>
+#include <sstream>
+#include <string>
 #include <cstring>
 #include <fmt/core.h>
 #include <fmt/format.h>
@@ -61,6 +63,56 @@ inline std::string get_notename(int i_value, int i_offset)
 inline std::string float_to_clocalestr(float value)
 {
     return fmt::format(std::locale::classic(), "{:L}", value);
+}
+
+/*
+ * Parse a number a user typed in, independent of the active locale. Accepts
+ * either '.' or ',' as the decimal separator, since what a user types depends
+ * on their keyboard and locale rather than on ours. std::atof follows
+ * LC_NUMERIC, so in a comma-decimal locale it stops at the '.' in "0.5" and
+ * silently yields 0, and in a dot-decimal locale it does the same to "0,5".
+ */
+inline double clocalestr_to_double(const char *s, bool *parsed)
+{
+    if (parsed)
+    {
+        *parsed = false;
+    }
+
+    if (!s)
+    {
+        return 0.0;
+    }
+
+    std::string t{s};
+    std::replace(t.begin(), t.end(), ',', '.');
+
+    // strtod is LC_NUMERIC-dependent, so parse through a classic-locale stream
+    std::istringstream is{t};
+    is.imbue(std::locale::classic());
+
+    double res{0.0};
+    is >> res;
+
+    if (is.fail())
+    {
+        // match atof, which returns 0 for input that isn't a number
+        return 0.0;
+    }
+
+    if (parsed)
+    {
+        *parsed = true;
+    }
+
+    return res;
+}
+
+inline double clocalestr_to_double(const char *s) { return clocalestr_to_double(s, nullptr); }
+
+inline double clocalestr_to_double(const std::string &s)
+{
+    return clocalestr_to_double(s.c_str(), nullptr);
 }
 
 #endif // SURGE_SRC_COMMON_UNITCONVERSIONS_H
--- a/AUTHORS
+++ b/AUTHORS
@@ -64,3 +64,4 @@ Alexandr Zyurkalov <Alexander.Zyurkalov@gmail.com>
 Daniel Hatadi <https://github.com/dhatadi>
 Yashas Hebbarabailu <https://github.com/y-hbb>
 Bill Allen <https://github.com/bwanab>
+Manu Arimaka <https://github.com/manuaudio>
--- a/src/common/Parameter.cpp
+++ b/src/common/Parameter.cpp
@@ -4789,7 +4789,7 @@ bool Parameter::set_value_from_string_onto(const std::string &s, pdata &ontoThis
                     ++strip;
                 }
 
-                ni = (std::atof(strip) * 8) - 1;
+                ni = (clocalestr_to_double(strip) * 8) - 1;
                 factor = 8;
                 offset = 1;
 
@@ -4833,20 +4833,18 @@ bool Parameter::set_value_from_string_onto(const std::string &s, pdata &ontoThis
             }
             else if (extend_range)
             {
-                try
+                bool parsed{false};
+                auto sv = clocalestr_to_double(s.c_str(), &parsed);
+
+                if (parsed)
                 {
-                    ni = std::round(std::stof(s) * 100);
+                    ni = std::round(sv * 100);
                     factor = 100;
                 }
-                catch (const std::invalid_argument &)
+                else
                 {
                     // set value of ni out of range on invalid input
                     ni = val_min.i - 1;
-                }
-                catch (const std::out_of_range &)
-                {
-                    // likewise for out of range input
-                    ni = val_min.i - 2;
                 }
             }
         }
@@ -4918,7 +4916,7 @@ bool Parameter::set_value_from_string_onto(const std::string &s, pdata &ontoThis
         return true;
     }
 
-    auto nv = std::atof(s.c_str());
+    auto nv = clocalestr_to_double(s.c_str());
 
     switch (displayType)
     {
@@ -5037,7 +5035,7 @@ bool Parameter::set_value_from_string_onto(const std::string &s, pdata &ontoThis
 
             if (std::regex_search(s, m, r))
             {
-                nv = std::atof(m[1].str().c_str()) * .001f;
+                nv = clocalestr_to_double(m[1].str().c_str()) * .001f;
             }
         }
 
@@ -5178,8 +5176,8 @@ bool Parameter::set_value_from_string_onto(const std::string &s, pdata &ontoThis
 
             if ((slp = strchr(strip, '/')) != nullptr)
             {
-                float num = std::atof(strip);
-                float den = std::atof(slp + 1);
+                float num = clocalestr_to_double(strip);
+                float den = clocalestr_to_double(slp + 1);
 
                 if (den == 0)
                 {
@@ -5192,7 +5190,7 @@ bool Parameter::set_value_from_string_onto(const std::string &s, pdata &ontoThis
             }
             else
             {
-                nv = std::atof(strip);
+                nv = clocalestr_to_double(strip);
             }
 
             if (extend_range)
@@ -5268,7 +5266,7 @@ float Parameter::calculate_modulation_value_from_string(const std::string &s, st
     errMsg = "Input is out of bounds!";
     valid = true;
 
-    float mv = std::atof(s.c_str());
+    float mv = clocalestr_to_double(s.c_str());
 
     switch (displayType)
     {
@@ -5401,7 +5399,7 @@ float Parameter::calculate_modulation_value_from_string(const std::string &s, st
         {
             if (s[0] == 'N' || s[0] == 'C' || s[0] == 'n' || s[0] == 'c')
             {
-                auto mv = (float)std::atof(s.c_str() + 1);
+                auto mv = (float)clocalestr_to_double(s.c_str() + 1);
 
                 if (s[0] == 'C' || s[0] == 'c')
                 {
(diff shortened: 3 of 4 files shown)


mkruselj:
Thanks for the PR! Please add yourself to AUTHORS file.

As for thousands separator, we really don't know how are people in India using Surge (IIRC that's one of countries using a thousands separator). My gut tells me this should be supported somehow but it may be overkill/unwieldy/complicated to do so...

mkruselj:
Also, the writeup about the PR looks AI generated. Please add Assisted-By tag to the PR description, naming the model that was used, if so.

src/common/UnitConversions.h
```diff
@@ -63,4 +65,54 @@ inline std::string float_to_clocalestr(float value)
...
+        *parsed = false;
+    }
+
+    if (!s)
+    {
+        return 0.0;
+    }
+
+    std::string t{s};
+    std::replace(t.begin(), t.end(), ',', '.');
```
baconpaul: Oooh why replace vs just parsing in the local locale?

manuaudio:
Added myself to AUTHORS, and yes — Assisted-By tag is in the description now. I used Claude Code to find the call sites and draft the patch; the analysis, the both-ways test check and the review of it are mine. Happy to rewrite anything that reads oddly.

On parsing in the local locale instead: that alone doesn't fix the report. The two people who hit this are on Polish and Croatian Windows, and they see opposite halves of it — one gets a dot shown in the typein but has to enter a comma, the other gets the reverse. So whichever separator we pick for parsing, we break whoever types the other one, and there's no error to tell them why the fraction vanished. Accepting both is what makes it work for both of them without having to know what the display did.

The other reason not to lean on the local locale is that it isn't stable ground for us — we already parse patch data with the classic locale via float_to_clocalestr, so type-ins going through LC_NUMERIC while streaming doesn't is part of what got us here.

On thousands separators: agreed it's not obviously right either way. What I have now treats "1,234.5" as 1.234, and atof gave 1 — both wrong, just differently. If you'd rather, I can reject anything with a separator in a thousands position outright so it errors instead of quietly reading as something else. Say which you prefer and I'll push it.

mkruselj:
Well, it's also a problem because some languages don't group digits in 3, but 4 (India, Japan...)

I don't really have an answer other than "woof, it's tricky". So maybe we just leave it as is?

manuaudio:
Works for me — let's leave it as is. Good point about 4-digit grouping too; that kills any "just strip the group separator" shortcut, since you can't tell a group separator from a decimal one without knowing the convention the person is typing in, and we don't.

So the recorded behaviour before merge: with a thousands separator in play, `1,234.5` and `1.234,5` both come out as `1.234`, and `1,234` does too. `atof` gave `1` for those in the locale where a user would actually type them. Both are wrong, just differently — nothing that worked before gets worse, and every plain number with a single decimal separator, which is all a Surge type-in normally sees, is now right in both locale families.

Happy to revisit if someone actually reports typing grouped numbers into a type-in.

src/common/UnitConversions.h
manuaudio: Because parsing in the local locale still only accepts one separator, and the two reporters on #7574 hit opposite halves of it — one sees a dot in the type-in but has to enter a comma, the other the reverse. Whichever separator we pick, the other person's fraction silently vanishes with no error. Accepting both is what covers them both without having to know what the display did.

It also keeps type-ins consistent with how we already stream patch data, which goes through the classic locale via `float_to_clocalestr` right above this. Type-ins following `LC_NUMERIC` while streaming doesn't is part of what produced the bug.

(Fuller version in the conversation above, didn't want to bury it in the thread.)

manuaudio:
One thing I turned up while answering Paul's question, and I'd rather flag it than sit on it.

This PR only converts the `Parameter.cpp` type-ins. The same `std::atof` is still in **thirteen other calls across seven files**, all free-text entry, all the same failure:

- `SurgeSynthProcessor.h` — both `getValueForText` overrides (macro and bypass), so the host's own parameter entry box
- `MSEGEditor.cpp` and `LFOAndStepDisplay.cpp` — the MSEG and step-sequencer type-ins, including their `n/d` fraction forms
- `TuningOverlays.cpp` — even-division span and the KBM reference frequency
- `SurgeGUIEditorMenuStructures.cpp` — the "enter a new frequency for A440" remap prompt
- `SurgeGUIEditorValueCallbacks.cpp` — the macro percentage entry
- `SurgeGUIEditor.cpp` — font size

The standalone tempo box is *not* affected — `setInputRestrictions(3, "0123456789")` means no separator can be typed there at all.

Each is the same one-line swap to `clocalestr_to_double`. I deliberately left them out to keep this PR to the reported bug rather than growing it under review. Happy either way: say the word and I'll add them here, or I'll open a follow-up once this lands.

baconpaul:
Right I agree we should add a definitive function and fix all call sites 

seems std::numpunct which is a facet of std::locale has a thousands_separator which we can use to knock out , or . Properly before we use the locale plus imbue right?

we should also probably put these functions in basic blocks and have a string and string view version

thank you for tackling this. Will be good to get it right

manuaudio:
Yes to a definitive function, and `numpunct` is the right instrument — but I probed the facets before answering, because a couple of things bite:

```
locale C            decimal='.' thousands=',' grouping=[]
locale ""           decimal='.' thousands=',' grouping=[3]
locale de_DE.UTF-8  decimal=',' thousands='.' grouping=[3]
locale pl_PL.UTF-8  decimal=',' thousands=' ' grouping=[3]
locale en_IN        decimal='.' thousands=',' grouping=[3 2]
locale bogus_XX     THREW: collate_byname<char>::collate_byname failed to construct
```

Three things fall out:

1. **Don't hardcode "knock out , or ."** — Polish groups with a **space**, not a dot. You have to knock out the actual `thousands_sep()` from the facet. Marijan's India worry is also answered for free: `grouping()` reports `[3 2]` there, so the facet already knows about lakh/crore and we never have to guess.
2. **Constructing a named locale throws.** `std::locale("bogus_XX")` raises, and `std::locale("")` reads the environment, so a user with a junk `LC_ALL` takes a throw on a type-in. Needs a try/catch and wants caching — building a locale per keystroke isn't free.
3. **Which locale we query is the actual design decision.** On Windows `""` maps to the regional settings, so this genuinely fixes the two reporters. On macOS/Linux under a plugin host `LC_*` is very often unset, `""` collapses to classic, and a comma-typing user is still broken. So I'd read the facet when `grouping()` is non-empty and it disambiguates, and keep the accept-either-separator fallback for when it can't.

Also worth knowing: `atof` and `std::stof` follow the **C** locale, while `imbue`/`numpunct` follow the **C++** locale, and they move independently — `setlocale(LC_NUMERIC, "de_DE")` alone flips `std::stof("0.5")` to `0` with no C++-side change at all.

On "put these in basic" — agreed, and it's worse than this PR suggests. `ParamMetadata::valueFromString` in sst-basic-blocks parses with `std::stof` in about sixteen places, so every sst-based plugin has the same bug, not just Surge.

If that split sounds right, I'll put the function in sst-basic-blocks, convert `valueFromString` there, then follow up here with the Surge call sites and a submodule bump. Say the word and I'll start on the basic-blocks side.

One flag: this mechanism would also handle the thousands separator properly, which Marijan and I settled as "leave it as is" a couple of hours ago in this thread. Happy to go either way, but you two may want to agree before I build it.

baconpaul:
YEah I think a basic blocks function which handles properly makes loads of sense then update PMD and surge both.

When we generate basic blocks we can generate a good test suite too - there's a catch2 test suite in the submodule.

Again thanks for tackling this. And hey, isn't claude great!

baconpaul:
Think this one is waiting on set-basic-blocks#305 where I added a comment. So let's park this until that merges and this adapts.

Great stuff though. Thanks!

manuaudio:
sst-basic-blocks#305 has merged, so this can adapt — but there's a wrinkle worth your call before I push anything.

Surge pins `sst-basic-blocks` at `312702f`, which is **24 commits** behind the merge. Only two of those are mine (#304 and #305); the rest are other people's, including the ModMatrix RT-safety fixes and the one-pole/delay-line work. So converting this PR to call `mechanics::parseNumber` means bumping the submodule across all 24.

That felt like your call rather than mine — the last bump was yours in #8333, and I'd rather not fold two dozen unrelated commits into a locale fix.

Two ways I'm happy to go:

1. **You bump the submodule** on your normal cadence, and I follow immediately with a small PR that deletes `clocalestr_to_double` from `UnitConversions.h` and points the type-in sites at `mechanics::parseNumber`. Clean diff, nothing unrelated in it.
2. **I include the bump here**, in which case I'd want to run the full runner against the new submodule before pushing, and this PR stops being just a locale fix.

I'd lean towards 1, but say which and I'll do it.

manuaudio:
Adapted, now that #305 is in.

`UnitConversions.h` no longer carries its own copy of the parsing — `clocalestr_to_double` keeps its name and shape, since the type-in call sites want a double plus a did-it-parse flag, but the body is a call into `mechanics::parseNumber` and the duplicated classic-locale stream handling is gone. Net −28/+8 there.

The submodule bump to `3ee58b0` also brings in the `ParamMetaData` half of #305, so the seventeen `std::stof` sites behind the shared parameter machinery are covered by the same change rather than by a second copy here.

Full runner green at 5671332 assertions in 127 test cases, `[param]` at 45 assertions in 2 cases, `clang-format` clean.

Worth noting the behaviour is now slightly better than what this PR originally proposed: because the shared function reads the `numpunct` facet, a grouped input like `1.234,5` resolves rather than truncating, which is the thousands-separator case you and Marijan discussed on here early on.

baconpaul:
Feel free to bump. Surge doesn’t touch that code which was changed

manuaudio:
Already bumped — the branch pins `3ee58b0`, which is basic-blocks main head as of now, so there's nothing further to move it to.

Agreed on the risk being nil for surge: the `ParamMetaData` half of that PR sits behind machinery surge doesn't use, and surge's own type-ins go through `Parameter.cpp`, which is what this PR touches. All twelve CI jobs are green on it — including the MSVC ARM64/ARM64EC and Windows runners, which are where I'd expect locale handling to misbehave if it were going to.

baconpaul:
Surge actually does use ParamMetadata for some of the fx which are ported to set-effects just fyi. But via parameter.cpp

manuaudio:
You're right and I was wrong — thanks for the correction. I went and looked rather than take my own word for it, and the path is more direct than I'd assumed:

- `Parameter.cpp:4721` → `basicBlocksParamMetaData->valueFromString(s, errMsg)`
- `Parameter.cpp:5254` → `basicBlocksParamMetaData->modulationNaturalFromString(s, val.f, errMsg)`

Those are precisely the two functions #305 rewired, and the ported effects reach them through `SurgeSSTFXAdapter` — Bonsai, Delay and the rest.

So the bump is more consequential than I said, and in the useful direction: type-ins on the ported FX pick up the locale fix straight from basic-blocks rather than needing anything here. It also means the seventeen converted sites aren't dormant in surge the way I claimed — they're live for those effects.

Evidence that it holds: full runner green at 5671332 assertions in 127 test cases, which includes the nine `[fx]` cases, and all twelve CI jobs pass including the MSVC ARM64/ARM64EC ones.

manuaudio:
Good to know — that sharpens the read on the bump rather than changing the answer.

#305 converted `ParamMetaData`'s seventeen `std::stof` sites to the locale-independent parser. So if Surge reaches `ParamMetaData` through `Parameter.cpp` for the ported FX, those type-ins pick up the same accept-either-separator behaviour this PR adds to the rest of them — same direction, and arguably the point.

But it does mean the bump isn't inert for Surge, so it's worth running the full runner over it rather than treating it as a no-op pin change. I'd rather find out here than have someone find it in an FX type-in later.

baconpaul:
seems your response agents are doubled up! CI runs the full test and the change passed review so i think we are fine. But sure fire up delay or reverb2 - those are ported effects - and give it a whirl in the ui

manuaudio:
Apologies for the pile-up of comments — consolidating.

I did build the standalone from merged main (`21a59873`, submodule at `3ee58b0`) and it launches and runs fine. Where I have to be straight with you: I can't finish the visual check. This machine has no usable display session — screen capture comes back blank — so I can get the app running but I can't type into a Delay field and watch what happens, and I'm not going to report a blank screenshot as if it were confirmation.

What is covered rather than assumed: the ported-FX type-in path you pointed me at, `Parameter.cpp:4721` → `basicBlocksParamMetaData->valueFromString`, is exercised by the `[param]` cases, and the full runner is green at 5671332 assertions in 127 test cases with the bumped submodule. So the parsing is tested; what's untested is the last hop from the text field into that call.

If you or Marijan get a spare minute in front of a screen, Delay feedback or Reverb2 size with `0,5` typed in would close that gap properly. Happy to be told it's fine as-is too — you've already said CI covers it.

baconpaul:
I bet, if you are creative, you can use the accessibility api on windows or mac to find the components and trigger them. If you want to burn some tokens and have some fun. There's some python scripts for accessibility test on macOS in scripts. 

Not urgent, and if the human side of manu is reading this sorry to address you more like an agent but if you re just sitting in a loop on gh refresh it may be fun to try.

Zero concerns either way though.

manuaudio:
Gave it a proper go, and the answer is a clean no — but a better-diagnosed one than "the screenshot was blank".

Found `scripts/pyauto-tests` and `sxttest.py`, which is exactly the right harness — `fx-setup.py` already does the "pick a slot, open the FX Type menu, walk to a submenu entry" dance I'd need, so driving to Delay and then into a parameter field is a small edit of that rather than new work.

Where it stops: **accessibility itself is fine here** — `UI elements enabled` is `true`, so it isn't a TCC problem. The blocker is that the Surge process comes up with **zero windows**. It launches, stays alive, prints nothing, reports `visible: true`, and `count of windows` is `0` however long I wait or whether I activate it first. There's a console session for the user, but this process can't attach to the window server, so there's simply no UI tree for AX to walk and nothing to press.

So the harness is right, the permissions are right, and the machine is wrong. On a box with a real logged-in desktop session, `fx-setup.py` plus a type-in step should do exactly what you described.

And no apology needed on the agent front — every commit here carries an `Assisted-By: Claude Code` trailer for precisely that reason, so you weren't misreading anything. Manu directs and reviews it; the loop does the legwork.

PR #8523, opened 2026-08-13
Do not assume XML attributes are present when reading a patch

manuaudio:
`TiXmlElement::Attribute` returns null for an attribute that isn't there. Most of the readers already account for that — `FxUserPreset::readFromXMLSnapshot` bails when `name` is missing, and the `meta` and `label` reads check before using what they got — but a few build a `std::string` straight from the result.

### The one that matters

`SurgePatch.cpp`, reading the patch tag list:

```cpp
std::string tagName = tag->Attribute("tag");
```

A `<tag/>` element carrying no `tag` attribute constructs a `std::string` from `nullptr`. Patches are shared, downloaded and hand edited, so opening someone else's patch is enough to reach this. A malformed entry is now skipped and the patch keeps its remaining tags, rather than one bad entry taking the whole load down.

### The same shape elsewhere

`SurgeStorage.cpp` reads `help_url` and `id` the same way in three places while parsing the configuration. An entry can legitimately carry its `group` or `id` without carrying a URL, so those are read as optional now too.

### Testing

Two cases added to `TEST_CASE("XML Direct", "[io]")`, which already exists with the comment *"we want to make sure it doesn't nuke surge with garbage"* — this is a piece of garbage it didn't yet cover. One case feeds a bare `<tag/>`, the other mixes named and unnamed tags to check a malformed entry doesn't cost the patch its good ones.

I checked they fail without the fix rather than only passing with it. Reverting the two sources and keeping the tests gives:

```
UnitTestsIO.cpp:908: FAILED:
due to a fatal error condition:
  SIGSEGV - Segmentation violation signal
```

With the fix, `[io]` is green at 5463 assertions in 10 test cases, and the full runner is green at **5702164 assertions in 126 test cases**. `clang-format` clean.

Related: same class of thing as surge-synthesizer/sst-plugininfra#84, which was the user defaults reader.

---

Assisted-By: Claude Code (Claude Opus 5) — I used it to sweep the call sites and draft the patch; the analysis, the both-ways check and the review of it are mine.

### DIFF
--- a/src/common/SurgePatch.cpp
+++ b/src/common/SurgePatch.cpp
@@ -1814,8 +1814,16 @@ void SurgePatch::load_xml(const void *data, int datasize, bool is_preset)
 
             while (tag)
             {
-                std::string tagName = tag->Attribute("tag");
-                tags.emplace_back(tagName);
+                // Attribute returns null when it isn't there, and patches arrive
+                // from other people, so a <tag/> without one would build a
+                // std::string from nullptr rather than simply having no name.
+                const char *tagName = tag->Attribute("tag");
+
+                if (tagName)
+                {
+                    tags.emplace_back(tagName);
+                }
+
                 tag = TINYXML_SAFE_TO_ELEMENT(tag->NextSiblingElement("tag"));
             }
         }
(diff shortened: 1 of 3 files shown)


mkruselj:
Patches are not really hand-edited, if by hand-editing you mean people take an FXP open it with a text editor and modify stuff in it. This is not what a regular user would be doing.

baconpaul:
thats true but we should never have an unchecked return @mkruselj . Dont' let people do things which crash!

baconpaul:
we should merge this once ci passes. good change!

PR #8524, opened 2026-08-13
Bound the extraoscdata indices and count taken from a patch

manuaudio:
The `extraoscdata` block reads three numbers out of the patch and uses all three without checking them against what they index. Patches get shared, downloaded and hand edited, so opening someone else's patch is enough to reach every one of these.

### `osc` and `scene`

They index `osc[n_oscs]` and `scene[n_scenes]`, and both come from the file:

```cpp
int sos = std::atoi(lkid->Attribute("osc"));
int ssc = std::atoi(lkid->Attribute("scene"));
...
auto ec = &(scene[ssc].osc[sos].extraConfig);
```

A patch naming an oscillator or scene we don't have writes outside the patch object. A negative index does the same in the other direction.

### `extra_n`

This is the one I'd look at first. It's the loop bound for writes into `ExtraConfigurationData::data`, which is a fixed `max_config` floats:

```cpp
ec->nData = ti;                                  // straight from the file

for (int qq = 0; qq < ec->nData; ++qq)
{
    if (lkid->QueryDoubleAttribute(attr, &tf) == TIXML_SUCCESS)
        ec->data[qq] = tf;
    else
        ec->data[qq] = 0.f;                      // writes even with no attribute
}
```

Because the `else` branch writes too, `extra_n` on its own decides how far past the end of that array we go — no `extra_data_N` attributes are needed at all. A negative value was also stored unchanged, leaving `nData` negative in the patch.

### The change

Entries whose `osc` or `scene` is outside range are skipped, and `extra_n` is held inside the array it indexes.

### Testing

Three cases added to `TEST_CASE("XML Direct", "[io]")`, which already exists to check `load_xml` copes with garbage.

Worth mentioning for anyone extending that test case: `load_xml` returns before `extraoscdata` if the patch has no `<parameters>` element, so a test patch needs one or it silently never reaches this code. My first attempt at these tests passed against the unfixed build for exactly that reason.

Once they actually reach it, the unfixed build gives:

```
UnitTestsIO.cpp:933: FAILED:
  REQUIRE( ec.nData >= 0 )
with expansion:
  -5 >= 0

UnitTestsIO.cpp:936: FAILED:
due to a fatal error condition:
  SIGSEGV - Segmentation violation signal
```

With the fix, `[io]` is green at 5468 assertions in 10 test cases and the full runner is green at **5702780 assertions in 126 test cases**. `clang-format` clean.

### A note on how I'm reporting this

There's no `SECURITY.md` and private vulnerability reporting is turned off for the repo, so there was no private channel to use and I've written this as an ordinary bug report. It is a bounded, attacker-influenced write reachable from a shared file, so you may want to treat it with more urgency than the other two I've sent today, and it might be worth enabling private reporting on the repo for next time.

Sits alongside #8523, which was the null-attribute reads in the same loader. They don't overlap.

---

Assisted-By: Claude Code (Claude Opus 5) — I used it to trace the loader and draft the patch; the analysis, the both-ways check and the review of it are mine.

### DIFF
--- a/src/surge-testrunner/UnitTestsIO.cpp
+++ b/src/surge-testrunner/UnitTestsIO.cpp
@@ -939,8 +939,7 @@ TEST_CASE("XML Direct", "[io]")
         surge->storage.getPatch().load_xml(test.c_str(), test.size(), false);
 
         auto &ec = surge->storage.getPatch().scene[0].osc[0].extraConfig;
-        REQUIRE(ec.nData >= 0);
-        REQUIRE(ec.nData <= (int)OscillatorStorage::ExtraConfigurationData::max_config);
+        REQUIRE(ec.nData == (int)OscillatorStorage::ExtraConfigurationData::max_config);
     }
 
     SECTION("extraoscdata with a negative extra_n")
@@ -952,7 +951,40 @@ TEST_CASE("XML Direct", "[io]")
         surge->storage.getPatch().load_xml(test.c_str(), test.size(), false);
 
         auto &ec = surge->storage.getPatch().scene[0].osc[0].extraConfig;
-        REQUIRE(ec.nData >= 0);
+        REQUIRE(ec.nData == 0);
+    }
+
+    SECTION("msegs with out of range scene and index")
+    {
+        auto surge = Surge::Headless::createSurge(44100);
+        std::string test{"<patch><parameters/><msegs>"
+                         "<mseg scene=\"99\" i=\"99\"/><mseg scene=\"-1\" i=\"-1\"/>"
+                         "</msegs></patch>"};
+        surge->storage.getPatch().load_xml(test.c_str(), test.size(), false);
+        SUCCEED("an out of range mseg scene or index did not write outside the patch");
+    }
+
+    SECTION("formulae with out of range scene and index")
+    {
+        auto surge = Surge::Headless::createSurge(44100);
+        std::string test{"<patch><parameters/><formulae>"
+                         "<formula scene=\"99\" i=\"99\"/><formula scene=\"-1\" i=\"-1\"/>"
+                         "</formulae></patch>"};
+        surge->storage.getPatch().load_xml(test.c_str(), test.size(), false);
+        SUCCEED("an out of range formula scene or index did not write outside the patch");
+    }
+
+    SECTION("lfo bank labels with out of range indices or no value")
+    {
+        // three indices and the label text all come from the file
+        auto surge = Surge::Headless::createSurge(44100);
+        std::string test{"<patch><parameters/><lfobanklabels>"
+                         "<label scene=\"99\" lfo=\"99\" idx=\"99\" v=\"x\"/>"
+                         "<label scene=\"-1\" lfo=\"-1\" idx=\"-1\" v=\"x\"/>"
+                         "<label scene=\"0\" lfo=\"0\" idx=\"0\"/>"
+                         "</lfobanklabels></patch>"};
+        surge->storage.getPatch().load_xml(test.c_str(), test.size(), false);
+        SUCCEED("out of range label indices and a missing value were both discarded");
     }
 
     SECTION("extraoscdata with out of range scene and osc")
--- a/src/surge-testrunner/UnitTestsIO.cpp
+++ b/src/surge-testrunner/UnitTestsIO.cpp
@@ -926,4 +926,45 @@ TEST_CASE("XML Direct", "[io]")
         surge->storage.getPatch().load_xml(test.c_str(), test.size(), false);
         REQUIRE(surge->storage.getPatch().tags.size() == 2);
     }
+
+    SECTION("extraoscdata with an oversized extra_n")
+    {
+        // extra_n is the loop bound for writes into a fixed max_config array and
+        // it arrives from the file, so a patch could ask us to write far past the
+        // end of it. Patches get shared and downloaded, so opening one is enough.
+        auto surge = Surge::Headless::createSurge(44100);
+        std::string test{"<patch><parameters/><extraoscdata>"
+                         "<od osc=\"0\" scene=\"0\" extra_n=\"100000\"/>"
+                         "</extraoscdata></patch>"};
+        surge->storage.getPatch().load_xml(test.c_str(), test.size(), false);
+
+        auto &ec = surge->storage.getPatch().scene[0].osc[0].extraConfig;
+        REQUIRE(ec.nData >= 0);
+        REQUIRE(ec.nData <= (int)OscillatorStorage::ExtraConfigurationData::max_config);
+    }
+
+    SECTION("extraoscdata with a negative extra_n")
+    {
+        auto surge = Surge::Headless::createSurge(44100);
+        std::string test{"<patch><parameters/><extraoscdata>"
+                         "<od osc=\"0\" scene=\"0\" extra_n=\"-5\"/>"
+                         "</extraoscdata></patch>"};
+        surge->storage.getPatch().load_xml(test.c_str(), test.size(), false);
+
+        auto &ec = surge->storage.getPatch().scene[0].osc[0].extraConfig;
+        REQUIRE(ec.nData >= 0);
+    }
+
+    SECTION("extraoscdata with out of range scene and osc")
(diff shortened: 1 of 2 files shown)


mkruselj:
I'm really not sure this is a necessary addition. There are thousands of patches floating about and they are all fine. Surge itself doesn't write malformed patches, so this could only be happening by manually tweaking the FXP file, which is not a realistic threat IMO.

manuaudio:
That's fair on ordinary patches, and I agree Surge doesn't produce these — nothing in the factory set, and nothing Surge writes, can hit it. This isn't a bug anyone is running into today.

Where I'd differ is on what an `.fxp` is. They're files people download and open — passed around on forums, in Discord, in patch packs. Opening one you didn't write is a completely normal thing to do, which makes it untrusted input in the same way a downloaded image is, independently of whether anyone is currently being malicious with it. "Only by manually tweaking the file" is the part I'd read as the risk rather than the reassurance, because tweaking it is not difficult.

The reason I'd treat it as different from an ordinary malformed-input crash: `extra_n` isn't just read, it's the loop bound for writes into a fixed 64-float array, and the `else` branch writes even when no matching `extra_data_N` attribute exists. So one attribute decides how far past the end of that array we write. That's an out-of-bounds write with a length taken from the file, not a read that faults.

On cost: it's +54/−1, two range checks in a patch-load path that isn't hot, and no behaviour change for any patch that is valid today. Cheap insurance even at low probability.

All that said, it's your call and I won't push it. If the objection is scope rather than the change, I'm happy to drop the `osc`/`scene` half and keep only the `extra_n` clamp, which is the part that writes past the end of the array — that would take it down to a few lines. And if you'd rather not carry any of it, just say and I'll close it.

mkruselj:
Surge won't open an FXP that belongs to a different synth at all. This is a non issue IMO.

baconpaul:
I disagree @mkruselj - we are days or weeks or months from open source models looking for any vulnerability in software (or are 4 weeks past that happening). We should harden everything even if it isn’t in normal workflow.

I welcome these changes and think they are important.

manuaudio:
Thanks both. And to be fair to @mkruselj, his read on the normal workflow was right — nothing anyone does with Surge today produces one of these, and no existing patch is affected.

Since you'd rather have the whole thing, I'll leave it as written rather than trimming it down to just the `extra_n` clamp.

One practical note so it doesn't sit: CI hasn't been triggered on this branch — `gh pr checks` reports nothing on it, which looks like the same first-contributor workflow approval that #8523 needed before its run started. Worth a click when you get a moment, since you'll presumably want it green before merging.

If you do want the hardening principle pushed further through `load_xml`, I'm happy to take a pass at the rest of it the same way. I've got five PRs sitting in your queue already though, so I'd rather land those before adding to the pile — say the word whenever it's useful.

baconpaul:
Yeah my view on threat on this has changed in last few months. Basically normal workflow as defense is not something I really think we should rely on as much. I think it is very unlikely someone crafts an fxp that does something nefarious - and in normal workflow they wouldn’t - but I’d sooner make it so they can’t if someone is willing to carry that torch!

we do need to review these carefully though. Appreciate that these come with clear documentation and tests. Will spend some time tomorrow reading your prs while Claude generates mine. lol

src/surge-testrunner/UnitTestsIO.cpp
```diff
@@ -904,4 +904,45 @@ TEST_CASE("XML Direct", "[io]")
...
+    SECTION("extraoscdata with a negative extra_n")
+    {
+        auto surge = Surge::Headless::createSurge(44100);
+        std::string test{"<patch><parameters/><extraoscdata>"
+                         "<od osc=\"0\" scene=\"0\" extra_n=\"-5\"/>"
+                         "</extraoscdata></patch>"};
+        surge->storage.getPatch().load_xml(test.c_str(), test.size(), false);
+
+        auto &ec = surge->storage.getPatch().scene[0].osc[0].extraConfig;
+        REQUIRE(ec.nData >= 0);
```
baconpaul: Shouldn’t this be ==

nuoun:
Hello and welcome @manuaudio 

Thanks for looking into this, the two in this PR should definitely be fixed, can you also have a look at these three others please:

https://github.com/surge-synthesizer/surge/blob/241c131f90c6cc641c39393a2d5d11da93474af2/src/common/SurgePatch.cpp#L2939
https://github.com/surge-synthesizer/surge/blob/241c131f90c6cc641c39393a2d5d11da93474af2/src/common/SurgePatch.cpp#L2991
https://github.com/surge-synthesizer/surge/blob/241c131f90c6cc641c39393a2d5d11da93474af2/src/common/SurgePatch.cpp#L3096

There are some established pattern used elsewhere, and probably useful to use if possible to help maintainability, for clamping I would recommend our included `limit_range` function rather than `std::min` / `std::max`, you can maybe use this as an example:

https://github.com/surge-synthesizer/surge/blob/241c131f90c6cc641c39393a2d5d11da93474af2/src/common/SurgePatch.cpp#L2735

And for discarding:

https://github.com/surge-synthesizer/surge/blob/241c131f90c6cc641c39393a2d5d11da93474af2/src/common/SurgePatch.cpp#L2874

And also

<https://github.com/surge-synthesizer/surge/blob/241c131f90c6cc641c39393a2d5d11da93474af2/src/common/SurgePatch.cpp#L3020>

src/surge-testrunner/UnitTestsIO.cpp
manuaudio: Yes — better assertion. `<=` would have passed even if the clamp had produced a zero, which is exactly the failure it's meant to catch. Changed to pin it at `max_config`, and the negative case now pins `== 0` rather than `>= 0` for the same reason.

manuaudio:
Thanks for the welcome @nuoun, and for the pointers — both taken.

**House helpers.** Switched to `limit_range` for the clamp and `within_range` for the range checks, so the extraoscdata reader now reads like the stepsequence one just above it rather than inventing its own comparisons.

**The three you listed** — all three were real, and worse than I'd expected in one case:

- `msegs[sc][mi]` and `formulamods[sc][mi]` both take `scene` and `i` from the file and index straight in. Out of range entries are now skipped, following the pattern from the stepsequence reader.
- `LFOBankLabel[scene][lfo][idx]` indexes with three unchecked numbers **and** passes `lb->Attribute("v")` to `strxcpy` without checking it, so a `<label>` with no `v` was a null source as well as a potentially out of range destination. Both are handled now.

I checked these fail without the guards rather than only passing with them — reverting the source and keeping the tests gives a `SIGSEGV` on the msegs case, which is the first one the runner reaches.

Three cases added covering the new readers, and the whole thing is rebased on main now that #8523 has landed. `[io]` green at 5459 assertions in 10 test cases, `clang-format` clean.

If there are more of these you know about, I'm happy to keep going — though I've a few PRs in the queue already, so I'd rather not pile on until those clear.

baconpaul:
yeah if you happen to want to have claude look around for other trusted-xml points in the codebase we welcome it but also appreciate the cadence constraint

I'm doing same on short circuit now after our convo. I can send you the work list there if interested but don't want to post it up public right now.

manuaudio:
Both good, thanks.

On the wider sweep — yes, I'd like to. I'll start once this batch clears rather than adding to what you've already got open. `load_xml` is the obvious next target since it's the largest surface, and I'd work it the same way as here: one PR, the house helpers, and a case per reader that actually fails without the guard rather than just passing with it.

On the shortcircuit list — very interested, and understood about not posting it publicly. Let me come back to you on the channel; I still owe you an answer on the Discord question from the other thread, so I'll sort both together.

baconpaul:
Great. I'm out at a rehearsal for rest of night tonight (I'm US eastern time) but if you end up joining discord or figuring a way to dm me on another channel i'll post the short circuit list to gist and share it with you. Although i have fixed most of the memory ones so a lot of those are more 'what should this do' at this point. still would be interesting for you I bet!

manuaudio:
Easiest channel is email — manu@arimaka.com. Send the shortcircuit gist there whenever suits and I'll pick through it.

No rush on it either; you've got four of mine still open and those matter more than starting a new pile.

PR #8525, opened 2026-08-14
Refuse a WAV whose data chunk cannot be sized

manuaudio:
`datasamples` is computed as `cs * 8 / bitsPerSample / numChannels`, and **both divisors come out of the fmt chunk**:

```cpp
datasamples = cs * 8 / bitsPerSample / numChannels;
```

The format check above it validates `audioFormat` and `bitsPerSample`, but never the channel count — so a file declaring zero channels divides by zero. `bitsPerSample` also had no starting value, and since that check lives *inside* the fmt branch, a file carrying a data chunk with no fmt chunk at all divides by whatever was on the stack.

### What actually happens

This one surprised me, so it's worth spelling out:

- **On arm64 the divide doesn't trap — it yields zero.** The loader sizes the data to no samples and then **returns `true`**, reporting success for a file it could not read.
- **On x86_64, where integer division by zero does trap**, the same file takes the process down instead.

So the symptom differs by platform but the defect is the same on both. WAVs get downloaded and shared, so the fmt chunk isn't something the importer can take on trust.

### The change

Both fields start at zero, and a file we cannot size is refused with an error rather than sized to nothing.

### Testing

New case in `TEST_CASE("WAV with a zero channel count", "[io]")` writing a minimal RIFF/WAVE with `audioFormat=3`, `bitsPerSample=32` (so it passes the existing format check) and `numChannels=0`.

The interesting part is the assertion. On this machine the unfixed build doesn't crash — arm64 returns zero from the divide — so a crash test would have quietly passed against broken code, and did on my first attempt. The assertion that actually catches it is the return value:

```
UnitTestsIO.cpp:133: FAILED:
  REQUIRE( !loaded )
with expansion:
  false
```

Unfixed, `load_wt_wav_portable` returns `true` for a file it sized to zero samples. With the fix it refuses the file. `[io]` green at 5455 assertions in 11 test cases, `clang-format` clean.

If you'd rather see this as a clamp than a refusal — treating a zero-channel file as mono, say — happy to switch it; refusing seemed more honest than guessing at what the file meant.

---

Assisted-By: Claude Code (Claude Opus 5) — I used it to trace the importer and draft the patch; the analysis, the both-ways check and the review of it are mine.

### DIFF
--- a/src/common/WAVFileSupport.cpp
+++ b/src/common/WAVFileSupport.cpp
@@ -80,9 +80,7 @@ bool SurgeStorage::load_wt_wav_portable(std::string fn, Wavetable *wt, std::stri
         return false;
     }
 
-    // Both of these have to hold. A RIFF container which is not a WAVE - an AVI or a
-    // WebP, say - shares the outer header, and with an && here it walked on into the
-    // chunk loop and could load as a wavetable.
+    // an AVI shares the outer RIFF header, so check the form type too
     if (!four_chars(riff, 'R', 'I', 'F', 'F') || !four_chars(wav, 'W', 'A', 'V', 'E'))
     {
         std::ostringstream oss;
@@ -140,8 +138,7 @@ bool SurgeStorage::load_wt_wav_portable(std::string fn, Wavetable *wt, std::stri
         std::cout << "\'  sz = " << cs << std::endl;
 #endif
 
-        // cs is a 32 bit quantity read from the file, so a corrupt or hostile chunk can
-        // claim a size we cannot represent as an int. Stop rather than malloc it.
+        // chunk size comes from the file; refuse one we cannot represent as an int
         if (cs < 0)
         {
             break;
@@ -159,9 +156,7 @@ bool SurgeStorage::load_wt_wav_portable(std::string fn, Wavetable *wt, std::stri
             break;
         }
 
-        // Each branch below reads a fixed number of bytes out of the chunk, but the size
-        // it was allocated with came from the file. Nothing guarantees the two agree, so
-        // check the chunk is big enough before reading the fields out of it.
+        // each branch reads fixed bytes out of a file-declared size; check it fits
         if (four_chars(chunkType, 'f', 'm', 't', ' '))
         {
             if (cs < 16)
@@ -274,8 +269,7 @@ bool SurgeStorage::load_wt_wav_portable(std::string fn, Wavetable *wt, std::stri
 
             dp += 4;
 
-            // Each cue point is six 32 bit words, and the count is read from the file,
-            // so trust the chunk's own size over the count it declares.
+            // trust the chunk size over the cue count it declares
             const int cuesInChunk = cs >= 4 ? (cs - 4) / 24 : 0;
 
             if (numCues < 0 || numCues > cuesInChunk)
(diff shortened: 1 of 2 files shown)


baconpaul:
Good change
But a riff file with no fmt chunk should error out earlier as an invalid wav file - wonder if we should detect that condition as separate? 
it’s an unlikely condition and we are now protected but it might be useful

also I guess we should think about bail out it like you rename mp3 a wav - do we do appropriate header fail bails upstream?

would rather do these as one commit so I think let’s poke and see if there’s other edges we missed

thanks!

manuaudio:
Good call on both counts — I went looking and there was more here than I expected, so I've pushed it onto this branch as you asked rather than opening anything new.

**On "does an mp3 renamed to .wav bail out upstream?"** — an mp3 does, but only by luck. The header check is

```cpp
if (!four_chars(riff, 'R','I','F','F') && !four_chars(wav, 'W','A','V','E'))
```

which needs `||`. As written a file is only rejected when it fails *both* halves, so anything that is a RIFF container but not a WAVE — an AVI, a WebP — sails straight into the chunk loop. I built one: a RIFF file with form type `AVI ` carrying a valid `fmt`/`clm`/`data` set loads as a wavetable today. mp3 escapes because it matches neither.

**On the missing fmt chunk** — you were right that we're protected, and it turns out we were protected before this branch too, so that one is a diagnostic fix rather than a crash fix. I've made it say what's actually wrong instead of blaming the zero bits-per-sample the absent chunk left behind, and added the same bail at the end of the parse for a WAVE with no fmt at all.

**The edges we did miss.** Every branch of the chunk reader pulls a fixed number of bytes out of a chunk whose size was declared by the file, and the two are never checked against each other. An ASan build turns up five heap-buffer-overflows in `pl_int`/`pl_short`, all from files a few bytes long:

| chunk | reads | overflows when |
|---|---|---|
| `fmt ` | 16 bytes | `cs < 16` (`cs == 0` is enough) |
| `cue ` | 24 bytes × its own declared count | count exceeds what the chunk holds |
| `smpl` | 36 bytes, then 24 more | `cs < 36`, or `cs < 60` with a loop record |
| `srge`/`srgo` | 8 bytes | `cs < 8` |
| `wtmd` | to the first null | the chunk contains no null at all |

`wtmd` is the nastiest of them: `metadata = std::string(data)` on a buffer that is exactly the bytes the chunk declared, so it runs off the end hunting a terminator and whatever it finds ends up in the metadata string. Now bounded by the chunk.

Also added a `cs < 0` bail before the `malloc`, since the size is a 32-bit quantity from the file and can land negative in an `int`.

**Verification.** Sanitizer build, each case run on its own so one abort can't mask another. Before: six of the eight new sections trip ASan or fail. After: all eight clean. The two fmt-chunk sections pass either way, which is what tells me that part is cosmetic — I'd rather flag that than let it read as a sixth fix. Full suite is green at 5,664,906 assertions across 128 cases, and "All Factory Wavetables Are Loadable" holds at 1839 assertions, so all 919 factory tables still load through the new guards.

One unrelated thing I noticed while I had UBSan on, in case it's useful: `sst-basic-blocks/mechanics/endian-ops.h:102/111` reports misaligned `int16_t*`/`int32_t*` loads and stores during wavetable loading. Pre-existing and a different repo, so I've left it alone — happy to look if you want it.

baconpaul:
ok this looks good but i want to drive it by hand in a few cases as well. will do so this weekend. thanks!

mkruselj:
@baconpaul Reminder for you to check this one out by hand as you mentioned above.

PR #8528, opened 2026-08-15
Bound the patch and wavetable readers by the data they were given

manuaudio:
Two commits, kept separate. Both are the same shape: a reader taking a size from the file and using it before checking it against the buffer it actually has. Everything here is reachable from ordinary input — `.wt` and `.fxp` files get downloaded, and `loadRaw` is also what restores DAW session state.

---

## 1. `load_wt_wt` sizes an allocation from an unchecked header

```cpp
ds = sizeof(float) * mech::endian_read_int16LE(wh.n_tables) *
     mech::endian_read_int32LE(wh.n_samples);

const std::unique_ptr<char[]> data{new char[ds]};
```

`BuildWT` applies `max_wtable_size` (4096) and `max_subtables` (512) to those same counts — but thirty lines later, after the buffer exists. A header claiming the maximum of each works out to roughly **281 TB**, so `new[]` throws `std::bad_alloc` first, and nothing on the path from `load_wt` catches it.

```
due to unexpected exception with message:
  std::bad_alloc
```

Applying the same bounds before computing `ds` only changes *how* the file is refused — it was going to fail `BuildWT` anyway — so the error the user sees is unchanged. The message moved into a lambda so both the early bail and the `BuildWT` failure report it.

`load_wt_wt_mem` is unaffected; it already bounds `ds` by the size of the buffer it was handed.

---

## 2. `load_patch` reads past the end three different ways

**The header itself.** `patch_header` is 32 bytes; the only check was `datasize <= 4`. And because the reads endian-swap in place through a cast-away `const`, this writes out of bounds as well as reading:

```
heap-buffer-overflow ... WRITE of size 4
  #0 SurgePatch::load_patch(void const*, int, bool) SurgePatch.cpp:1191
  #1 SurgeSynthesizer::loadRaw(void const*, int, bool)
located 0 bytes after 8-byte region
```

**The embedded wavetable header.** Guarded with `if (wth > end)`, which only asks whether it *starts* past the end. A header landing exactly on `end` passes and is read:

```
  #0 Wavetable::BuildWT(void*, wt_header&, bool) Wavetable.cpp:172
  #1 SurgePatch::load_patch(...) SurgePatch.cpp:1206
located 10 bytes after 32-byte region
```

**The frames.** `BuildWT` sizes its copy from the counts inside that header, not from `wtsize`, so a header claiming 4096 samples reads 16k out of whatever remains:

```
READ of size 16384
  #1 sst::basic_blocks::mechanics::endian_copyblock32LE endian-ops.h:111
  #2 Wavetable::BuildWT(void*, wt_header&, bool) Wavetable.cpp:257
located 0 bytes after 60-byte region
```

Worth noting that bounding by `wtsize` does **not** fix that third one — the test sets `wtsize` to exactly the bytes remaining and `BuildWT` still overruns, because it trusts its own header over the chunk size. The check has to use the same product `BuildWT` will use.

I looked at reusing `load_wt_wt_mem` here, since it already performs all three of these checks. It can't be dropped in as-is: it requires the `vawt` tag, and `save_patch` does `memset(wth[sc][osc].tag, 0, 4)`, so every existing patch would be rejected. Happy to consolidate them properly if you'd rather have one implementation — it would need the tag check made optional.

---

### One change that is defensive rather than a fix

I also bounded `xmlsize` by the remaining bytes before stepping `dr` over it. No sanitizer report came out of that one and I could not produce a case that misbehaves — `load_xml` copes. But `dr` is advanced by a count straight from the file, and forming a pointer outside the allocation is undefined regardless of what is checked afterwards. Say the word and I'll drop it.

### Verification

Every case above was proven in both directions — reverting only the source change and rebuilding brings each report back:

| case | before | after |
|---|---|---|
| `.wt` float / int16 header at max counts | `std::bad_alloc` | refused, load error |
| patch shorter than its header | OOB write | refused |
| wt header exactly at `end` | OOB read | refused |
| wt frames not present | 16k OOB read | refused |
| huge `xmlsize` | *no report* | *no report* |

- `All Patches Are Loadable` green under ASan — every factory patch still loads through the new guards.
- `All Factory Wavetables Are Loadable` unchanged at **1839 assertions**, so all 919 factory tables still load.
- `[io]` green, no sanitizer output.

### DIFF
--- a/src/common/SurgePatch.cpp
+++ b/src/common/SurgePatch.cpp
@@ -1176,11 +1176,30 @@ void SurgePatch::load_patch(const void *data, int datasize, bool preset)
     assert(data);
     void *end = (char *)data + datasize;
     patch_header *ph = (patch_header *)data;
-    ph->xmlsize = mech::endian_read_int32LE(ph->xmlsize);
 
     if (!memcmp(ph->tag, "sub3", 4))
     {
+        // Reading the rest of the header needs the rest of the header to be there.
+        // datasize only had to clear four bytes to get this far, and these reads
+        // endian swap in place, so a truncated patch both reads and writes past the
+        // end of the caller's buffer.
+        if (datasize < (int)sizeof(patch_header))
+        {
+            return;
+        }
+
+        ph->xmlsize = mech::endian_read_int32LE(ph->xmlsize);
+
         char *dr = (char *)data + sizeof(patch_header);
+
+        // xmlsize is a count from the file and dr is stepped forward by it, so bound
+        // it to what the buffer actually holds rather than forming a pointer outside
+        // the allocation.
+        if (ph->xmlsize > (unsigned int)(datasize - (int)sizeof(patch_header)))
+        {
+            return;
+        }
+
         load_xml(dr, ph->xmlsize, preset);
         dr += ph->xmlsize;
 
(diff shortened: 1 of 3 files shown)


manuaudio:
Heads up — I pushed a second commit and retitled this. It started as just the `.wt` allocation, then the same question applied to the patch reader turned up three more overruns in `load_patch`, so they are here as a separate commit rather than a separate PR. Retitled and rewrote the description to match. Happy to split them if you'd rather review them apart.

baconpaul:
can we do a comment density pass here also (and on all the other prs you have open) then i can find time this afternoon to do final review on rest. thanks.

manuaudio:
Done across all of them — surge#8525, #8528, #8529, #8530, sst-effects#253, and tuning-library#93 earlier. Every multi-line block is now one line; the detail stays in the commit messages. Ready for review whenever suits.

baconpaul:
This one looks good. There's a regtest which is 'all patches load' which doesn't always run in CI I dont think. Can we confirm it runs locally with this change? If so I'll merge.

manuaudio:
Confirmed locally on the PR branch (c808445, clean tree), macOS arm64 Release:

```
surge-testrunner "All Patches Are Loadable"   ->  All tests passed (1 assertion in 1 test case)
surge-testrunner "[io]"                       ->  All tests passed (5431 assertions in 12 test cases)
```

One caveat worth stating, since the headline number is misleading: `All Patches Are Loadable` reports only 1 assertion because the sole `REQUIRE` is on the synth handle — the loop body has none, so it would look exactly this green if the patch list were empty. I temporarily instrumented it to check that it isn't: `patch_list.size() = 3561`, all 3561 loaded without a throw. That instrumentation was reverted before the run above; the tree was clean for the numbers I'm quoting.

The stderr noise under `[io]` is this PR's own new test tripping the wavetable guard on purpose (the "32767 frames / 2147483647 samples per frame" message) — that's the case being covered, not a failure.

PR #8529, opened 2026-08-15
Do not size the arbdata buffer from the size the frame claims

manuaudio:
### What

`load_arbitrary_block_storage` asks zstd how large the block decompresses to, and allocates that straight away:

```cpp
auto decompressedSize = ZSTD_getFrameContentSize(data, remainder);
...
sst::cpputils::DynArray<std::uint8_t> decompressed(decompressedSize);
```

That number is a **claim in the frame header**, not a measurement, and it is independent of how many bytes were actually supplied. Sixteen bytes of hand-written zstd header can declare 17.6 TB.

### The failure mode is a process kill, not an exception

`DynArray` constructs its elements, so the pages are touched rather than merely reserved. The allocation does not fail cleanly — the OS kills the process:

```
NEUTERED exit=137 (137 = SIGKILL) output_lines=0
```

No `bad_alloc`, no output, nothing to catch. A DAW loading such a patch loses the session, and patches arrive both as `.fxp` files and as restored session state.

Both call sites bound the *compressed* input they hand in — `(char *)end - dr`, or `claimedArbitraryBlockStorageSize` after checking it against the remaining bytes — but neither bounds what the frame claims to expand to. `streamingRevision` and `absSize` both come from the patch XML, so which of the two paths runs is attacker-chosen as well.

### The ceiling

Picking a number here is a judgement call, so I measured instead of guessing. Instrumenting `load_arbitrary_block_storage` and loading the whole factory set: **128 patches carry arbitrary block storage, and the largest decompresses to 3,649,380 bytes** — about 3.5 MB.

256 MB leaves roughly two orders of magnitude of headroom over anything the factory content produces, while keeping a corrupt header from taking the process down. It is a `static constexpr` with a comment explaining where the number came from — very happy to move it if you would rather it were tighter, or drawn from something else.

### Verification

- Against the unfixed loader the new test dies with **SIGKILL**, so the proof here is the exit code rather than a failed assertion — reverting only `SurgePatch.cpp` and rebuilding reproduces it every time.
- With the fix it returns 0 and reports a corrupt patch, which is what the surrounding code already does for the other "we can't read this" cases.
- `All Patches Are Loadable` green, `[io]` green at 5444 assertions across 11 cases.

This is independent of #8528 — different function, different failure mode — so I kept it on its own branch rather than widening that one again.

### DIFF
--- a/src/common/SurgePatch.cpp
+++ b/src/common/SurgePatch.cpp
@@ -1542,6 +1542,22 @@ unsigned int SurgePatch::load_arbitrary_block_storage(const void *data, std::siz
         // Same for the rest of these "return 0"s.
         return 0;
     }
+    // decompressedSize is what the zstd frame header claims, not anything measured, so
+    // a few bytes of crafted header can ask for terabytes. DynArray constructs its
+    // elements, so the pages are touched rather than merely reserved and the process is
+    // killed outright - there is no bad_alloc to catch. The largest block the factory
+    // content produces is about 3.6 MB, so this ceiling is roughly two orders of
+    // magnitude of headroom over real patches.
+    static constexpr std::uint64_t maxArbitraryBlockStorageSize{256 * 1024 * 1024};
+
+    if (decompressedSize > maxArbitraryBlockStorageSize)
+    {
+        std::cerr << "Arbitrary block storage claims " << decompressedSize
+                  << " bytes, which is beyond what a patch should carry; possibly corrupted patch."
+                  << std::endl;
+        return 0;
+    }
+
     sst::cpputils::DynArray<std::uint8_t> decompressed(decompressedSize);
     decompressedSize = ZSTD_decompress(decompressed.data(), decompressedSize, data, remainder);
     if (ZSTD_isError(decompressedSize))
--- a/src/common/SurgePatch.cpp
+++ b/src/common/SurgePatch.cpp
@@ -1542,12 +1542,7 @@ unsigned int SurgePatch::load_arbitrary_block_storage(const void *data, std::siz
         // Same for the rest of these "return 0"s.
         return 0;
     }
-    // decompressedSize is what the zstd frame header claims, not anything measured, so
-    // a few bytes of crafted header can ask for terabytes. DynArray constructs its
-    // elements, so the pages are touched rather than merely reserved and the process is
-    // killed outright - there is no bad_alloc to catch. The largest block the factory
-    // content produces is about 3.6 MB, so this ceiling is roughly two orders of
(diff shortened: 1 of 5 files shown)


mkruselj:
@mx You may want to have a say in this one!

src/common/SurgePatch.cpp
```diff
@@ -1542,6 +1542,17 @@ unsigned int SurgePatch::load_arbitrary_block_storage(const void *data, std::siz
...
     }
+    // cap the declared size; real patches top out around 3.6 MB
+    static constexpr std::uint64_t maxArbitraryBlockStorageSize{256 * 1024 * 1024};
+
+    if (decompressedSize > maxArbitraryBlockStorageSize)
+    {
+        std::cerr << "Arbitrary block storage claims " << decompressedSize
+                  << " bytes, which is beyond what a patch should carry; possibly corrupted patch."
+                  << std::endl;
+        return 0;
```
baconpaul, line 1595: so if we are doing a read check for valid, we should do a write check for the same. That is, if there ever was a patch which had > 256 * 1024 * 1024 (which maybe there could be - convolutions can be big) that should fail to write with the same constant that would fail to read.

This constant scanning the factory presets is probably not representative of what users could use also. I think we want to be careful here.

baconpaul (dismissed):

manuaudio:
Both done in eeeeafe0.

The cap is now one file-scope constant used by `save_arbitrary_block_storage` and `load_arbitrary_block_storage`, and the write path bails with the same threshold, so a block we would refuse to load is a block we refuse to save.

On the constant itself — you're right that scanning the factory presets was the wrong justification, and I've dropped that claim. It says what ships today, not what a user can build, and convolutions are exactly the counterexample. The number isn't a measurement; it's a bound on a length read out of a file before that length is used to size an allocation. What makes it safe is the symmetry you asked for: if 256 MB ever is too low for real content, the save now fails loudly when the patch is made, instead of writing a file that quietly won't load later. Raising it is a one-line change in one place.

No test on the write side — it'd need a >256 MB binn map, which isn't a reasonable allocation for the unit tests. Happy to pick a different number if you'd rather.

mkruselj:
@manuaudio Care to resolve merge conflicts on both your open PRs?

baconpaul:
I'm still not sure this max is the right max either. What's the largest size of a convolution? 256 meg seems like a lot but we should block saves if we are going to fail reads.

I'll take a look this weekend and will try and resolve conflict also then.

baconpaul:
OK this is good to merge when CI completes. I’m turning in for day but will merge in morning if it is still open.

PR #8530, opened 2026-08-15
Bound wavetable samples before converting them to int

manuaudio:
### What

```cpp
s[i] = (short)(int)limit_range((int)((float)f[i] * 16384.f), -16384, 16383);
```

The clamp runs on the *result* of the conversion, so it cannot prevent anything — `(int)` has already been applied by the time `limit_range` sees the value. `f` is the sample data of a wavetable read off disk, so it can hold an infinity or a value outside int's range, and converting either to `int` is undefined.

Same shape as the velocity narrowing in ee714d30: a clamp written to bound a conversion, sitting after it instead of before.

Reaching it takes no cleverness — four bytes of `0x7F800000` in the sample data of a `.wt`:

```
basic_dsp.h:44:45: runtime error: -inf is outside the range of
representable values of type 'int'
    #0 float2i15_block(float*, short*, int) basic_dsp.h:44
    #1 Wavetable::BuildWT(void*, wt_header&, bool) Wavetable.cpp:259
    #2 SurgeStorage::load_wt_wt(...) SurgeStorage.cpp:1762
```

### The change

Clamp while the value is still a float. `float2i15_block` has exactly one caller, `Wavetable::BuildWT`, so this is on the load path rather than the audio thread.

**Scope, after review:** this covers infinities and out-of-range finite values. It does **not** cover NaN — `limit_range` is `std::clamp`, which returns a NaN unchanged, so that input still converts and still trips UBSan. That is a deliberate choice per review discussion rather than an oversight; see the comments.

### Nothing moves for finite input

Comparing both forms across -4.0 to 4.0 in steps of 0.0000131, plus the boundary values, at `-O0` and `-O2`:

```
finite checked=610701 diffs=0
```

`[golden]` unchanged at **2322** assertions, `[dsp]` at **856931**, `All Factory Wavetables Are Loadable` at **1839**, `[io]` green.

### How it was found

Not by reading — by mutating real factory `.wt` files and loading them through `load_wt_wt` under UBSan.

I calibrated that harness first: against `upstream/main` it reproduces the `.wt` allocation problem from #8528 as `AddressSanitizer: allocation-size-too-big`, and with that fix applied it goes quiet. An earlier version of the same harness found nothing *even on unfixed code*, so a clean run from it would have meant nothing at all.

### DIFF
--- a/src/common/dsp/vembertech/basic_dsp.h
+++ b/src/common/dsp/vembertech/basic_dsp.h
@@ -41,9 +41,7 @@ inline void float2i15_block(float *f, short *s, int n)
 {
     for (int i = 0; i < n; i++)
     {
-        const float v = f[i] * 16384.f;
-
-        s[i] = (short)(int)limit_range(v, -16384.f, 16383.f);
+        s[i] = (short)(int)limit_range((int)((float)f[i] * 16384.f), -16384, 16383);
     }
 }
 
--- a/src/common/dsp/vembertech/basic_dsp.h
+++ b/src/common/dsp/vembertech/basic_dsp.h
@@ -41,16 +41,9 @@ inline void float2i15_block(float *f, short *s, int n)
 {
     for (int i = 0; i < n; i++)
     {
-        // f can be the sample data of a wavetable read off disk, so it can hold a NaN
-        // or an infinity, and converting either of those to int is undefined. The
-        // limit_range cannot help where it was: it ran on the result of the conversion,
-        // so the conversion had already happened. Bound the value while it is still a
-        // float, and say the NaN case explicitly since a NaN compares false against
-        // every bound. For finite values this produces what it always did.
         const float v = f[i] * 16384.f;
-        const float c = (v != v) ? 0.f : limit_range(v, -16384.f, 16383.f);
 
-        s[i] = (short)(int)c;
+        s[i] = (short)(int)limit_range(v, -16384.f, 16383.f);
     }
 }
 
--- a/src/common/dsp/vembertech/basic_dsp.h
+++ b/src/common/dsp/vembertech/basic_dsp.h
@@ -41,7 +41,16 @@ inline void float2i15_block(float *f, short *s, int n)
 {
     for (int i = 0; i < n; i++)
     {
-        s[i] = (short)(int)limit_range((int)((float)f[i] * 16384.f), -16384, 16383);
+        // f can be the sample data of a wavetable read off disk, so it can hold a NaN
+        // or an infinity, and converting either of those to int is undefined. The
+        // limit_range cannot help where it was: it ran on the result of the conversion,
+        // so the conversion had already happened. Bound the value while it is still a
+        // float, and say the NaN case explicitly since a NaN compares false against
+        // every bound. For finite values this produces what it always did.
+        const float v = f[i] * 16384.f;
+        const float c = (v != v) ? 0.f : limit_range(v, -16384.f, 16383.f);
+
+        s[i] = (short)(int)c;
     }
 }
 
--- a/src/surge-testrunner/UnitTestsIO.cpp
+++ b/src/surge-testrunner/UnitTestsIO.cpp
@@ -85,10 +85,7 @@ TEST_CASE("We Can Read Wavetables", "[io]")
 
 TEST_CASE("A wavetable of non finite samples loads", "[io]")
 {
-    // float2i15_block converted each sample to int and clamped the result, so a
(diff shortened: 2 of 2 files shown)


baconpaul:
really dont think we need to pay for nan protection here especially with the pretty odd idiom v != v ? 0 : limit

i think just changing this to float f = f * 16...; res = short int limit; and stripping the comment and excluding nan check is fine

if someone loads a wavetable with a nan i'm fine for surge to make bad sounds

manuaudio:
Done — dropped the NaN branch and the comment, so it is just the clamp in float now.

One thing you should know before this goes in, because it changes what the PR actually accomplishes rather than being an argument for the idiom. `limit_range` is `std::clamp`, and `std::clamp` returns a NaN unchanged — a NaN compares false against both bounds, so it falls through to `return v`. The conversion therefore still happens for that input.

Measured on the non-finite wavetable test, one report each:

| | report |
|---|---|
| `upstream/main` | `inf is outside the range of representable values of type 'int'` |
| this branch now | `nan is outside the range of representable values of type 'int'` |
| previous version | none |

So the change as it now stands does fix the infinity and the out-of-range-finite cases, which is most of it, and leaves NaN converting.

Entirely your call whether that matters — you already said bad sounds from a NaN wavetable are acceptable, and I am not trying to relitigate that. The only reason I am flagging it is that the UBSan report does not go away, so if the goal was a clean sanitizer run over `[io]` this does not get you there.

If you do want it gone without the `v != v` idiom, ordering the comparisons so a NaN lands on a bound does it for the same two compares and no extra branch:

```cpp
float v = f[i] * 16384.f;
if(!(v > -16384.f))   // false for NaN, so NaN takes this
    v = -16384.f;
if(v > 16383.f)
    v = 16383.f;
s[i] = (short)(int)v;
```

Happy either way — say the word and I will push that, or leave it exactly as it is.

I have updated the PR description so it no longer claims the NaN case is handled.

mkruselj:
There's not gonna be a wavetable with a NaN ever I think. Nobody is going to write WAV bytes manually, and all wavetable editors in existence already do the right thing and produce valid WAV files without surprises.

manuaudio:
Fair enough — leaving it exactly as it is.

One small thing to decide, since it is about the test rather than the fix: the file that test builds contains NaN samples alongside the infinities, so a local UBSan build will still print one report from it even though the test passes. Your CI does not run sanitizers, so nothing goes red. Happy to drop the NaN bytes from the test file so it only covers what the code now handles — or leave them in as a marker of the known case. Either is a one line change, just say which you prefer.

mkruselj:
I would say we could remove NaNs from the test. Realistically I think we don't have to expect that they will ever show up, I firmly believe that.

manuaudio:
Done — NaN samples dropped from the test, infinities kept since those are what the clamp handles. Swapped the NaN word for another finite sample so the file still mixes both.

Checked it still earns its keep: against upstream `basic_dsp.h` the test produces one UBSan report, and none with the change applied. So `[io]` is now clean under a sanitizer build too.

baconpaul:
Agree completely with @mkruselj here we can safely assume all floats we read from a structurally valid wave are std::is_finite

baconpaul:
So this is now just a test for a case we don't have and which if it occurred wouldn't recover from. I think I'm going to close this one as not needed, again because of the assumption that properly formatted wav and wt files don't hit nans and if they do there's not an expectation of recovery.

manuaudio:
Closing was the right call on what was actually on the branch, but the reason it looked like "just a test" is my mistake, and I want to correct the record rather than leave it.

The fix was still supposed to be there. Commit 7245ffc, whose message says "Drop the NaN samples from the test, per review", also reverted `src/common/dsp/vembertech/basic_dsp.h` back to upstream. So from that push onward the PR was a test asserting the behaviour of unmodified code — and that same commit message claims "against upstream basic_dsp.h it produces one UBSan report, and none with the change applied", which was not true of the tree I pushed. That claim was wrong and I should have caught it in the diff.

For the record, the clamp you and @mkruselj had signed off on is intact in 2a9b955 and is exactly the form you asked for:

```cpp
const float v = f[i] * 16384.f;

s[i] = (short)(int)limit_range(v, -16384.f, 16383.f);
```

The objection on record was only ever to the NaN branch, never to the infinity/out-of-range clamp — the UB is the out-of-range `float`->`int` conversion, which survives on structurally valid files.

Entirely your call whether that's still worth having. If you'd like it, say the word and I'll push the restored clamp with the test trimmed to the infinity case; if you'd rather leave it closed on the "structurally valid wavs don't hit this" reasoning, that's fine too and I won't reopen it.

PR #8534, opened 2026-08-31
Initialize timedata struct

### LINKED ISSUE #8533: Uninitialised timedata::ppqPos can put NaN into storage.songpos and crash on the first note with a freerun LFO (transport stopped)
`struct timedata`'s `ppqPos` and `tempo` are declared with no initialisers. With the host transport
**stopped**, `processBlockPlayhead()` deliberately does not assign `ppqPos` (that guard came from
#1493, so freerun could evolve without a transport), so the indeterminate value survives and is
accumulated onto every block. `resetStateFromTimeData()` then copies it into `storage.songpos`
**unguarded**, one line *before* the `tempo > 0` guard. If that value happens to be non-finite, a
patch with a **freerun** LFO routed to a filter cutoff crashes the host on the first note, inside
`SurgeStorage::note_to_pitch_ignoring_tuning`.

This is the same stack @baconpaul diagnosed in **#5566**, reached **without a formula modulator**,
so the 2021 `isfinite` fix on the formula-modulator output does not cover it.

I have a core dump and have read the live values out of it. Evidence below.

### Environment

| | |
|---|---|
| Surge XT | **1.3.4** (binary version string `1.3.4.f7b97c6`; tag `release_xt_1.3.4` = `f7b97c682ade`) |
| Package | Ubuntu `surge-xt` 1.3.4, plugin at `/usr/lib/clap/Surge XT.clap` |
| Plugin type | **CLAP** |
| Host | REAPER (Linux x86_64), audio via PipeWire's JACK layer, 96 kHz |
| OS | Ubuntu 26.04, kernel 7.0.0-29, x86_64 |
| Transport | **STOPPED throughout** — the rig is played live over MIDI and the transport is never rolled |
| Instances | 9 Surge XT instances in one REAPER process (told apart in the core by their distinct `oscPortIn`) |

Observed on Linux. Nothing in the code path below is platform-specific, but Linux is all I tested.

### The crash

Kernel log:

```
data-loop.0[2345772]: segfault at ffffffffc67b12e0 ip 00007d8d555cf25e sp 00007d8dc295d988
error 5 in Surge XT.clap[9cf25e,7d8d54c00000+1a23000]
```

`error 5` = bits 0 and 2 = a **user-mode read** protection fault. The faulting thread is the
SCHED_FIFO audio thread. Stack, as symbolised by `systemd-coredump` (demangled):

```
#0  SurgeStorage::note_to_pitch_ignoring_tuning(float)
#1  sst::filters::FilterCoefficientMaker<SurgeStorage>::Coeff_SVF(float, float, bool)
#2  SurgeVoice::SetQFB(QuadFilterChainState*, int)
#3  SurgeVoice::process_block(QuadFilterChainState&, int)
#4  SurgeSynthesizer::process()
#5  SurgeSynthProcessor::clap_direct_process(clap_process const*)
#6  ClapJuceWrapper::process(clap_process const*)
```

### Proof that the input was NaN, from the core

Registers on the faulting thread:

```
rdi    0x1c679a460            <- SurgeStorage* this (valid; its table_two_to_the reads 1, 1.00006, ...)
rax    0xffffffff80000000     <- pow2idx  == INT_MIN
rdx    0xffffffff80000000     <- e        == INT_MIN
xmm0   v4_float = {0xffffffff, 0, 0, 0}   <- NaN
xmm1   v4_float = {0xffffffff, 0, 0, 0}   <- NaN
eflags 0x10247  [ CF PF ZF IF RF ]
```

`CF`+`PF`+`ZF` set together is the x86 `COMISS` **unordered** result — the hardware signature of a
NaN compare. Both `(int)` conversions produced the x86 "integer indefinite" value `INT_MIN`, which
is what `cvttss2si` yields for NaN.

The faulting instruction is `mulss 0x16e80(%rdi,%rax,4),%xmm1`, and the fault address reproduces
exactly from the registers:

```
rdi + 0x16e80 + rax*4
  = 0x1c679a460 + 0x16e80 + (-2147483648 * 4)
  = 0xffffffffc67b12e0     == the kernel's reported fault address
```

`0x16e80` is the offset of `table_two_to_the` in `SurgeStorage`, so this is
`table_two_to_the[INT_MIN]` — a read 8 GiB below the table.

### Which filter, and which modulator

The voice's two `FilterCoefficientMaker`s, decoded from the core:

| | `C[0..3]` | `FirstRun` | `provider` |
|---|---|---|---|
| `CM[0]` | `0.0175202, 0.736222, 0.00102987, 0.617915` | `false` | the crashing `SurgeStorage*` |
| `CM[1]` | all zero | **`true`** | the crashing `SurgeStorage*` |

`CM[1].FirstRun == true` means `FromDirect` — the last statement of `Coeff_SVF` — had never run on
it: **filter unit 2 faulted inside its first-ever coefficient computation**, while filter unit 1
succeeded with a finite cutoff. Since `keytrack` and `fenv` are shared between `cutoffA` and
`cutoffB` in `SetQFB`, and `cutoffA` was finite, the NaN had to arrive through a modulator reaching
**filter 2's cutoff only**.

The voice's LFOs, decoded from the core (`output_multi[0]`, i.e. what `get_output` returns):

| LFO | trigmode (from the patch XML) | routed to | `output_multi[0]` |
|---|---|---|---|
| VLFO 1 | keytrigger | filter 2 cutoff | `0.004643627908080816` |
| **VLFO 2** | **freerun** | **filter 2 cutoff** | **`nan`** |
| VLFO 3 | keytrigger | filter 1 cutoff | `0.7247098088264465` |

**The only free-running LFO in the patch is the only one that is NaN.** The voice's state was
otherwise clean (`pitch = 83.0`, `fvel = 0.5039370059967041`, both finite), ruling out velocity and
keytracking as the source.

### Proof that the source is uninitialised memory, not the host

Nine Surge instances in one process, same host, same instant:

```
oscPortIn   ppqPos                    tempo   tsig   temposyncratio   songpos
53282       31178.19200528361         120.0   4/4    1.0              31178.19200528361
53283       3.0001761422722924e+184   120.0   4/4    1.0              3.0001761422722924e+184
53284       31178.24000528362         120.0   4/4    1.0              31178.238671950287
53285       31178.208005283614        120.0   4/4    1.0              31178.20667195028
53286       31178.24000528362         120.0   4/4    1.0              31178.238671950287
53287       31178.24000528362         120.0   4/4    1.0              31178.238671950287
53288       nan                       120.0   4/4    1.0              nan       <-- crashed
53289       nan                       120.0   4/4    1.0              nan       <-- also armed
53304       31178.24000528362         120.0   4/4    1.0              31178.238671950287
```

Every instance agrees on everything the host supplies — `tempo`, `timeSig`, `temposyncratio`. They
**disagree only on `ppqPos`**, the one field never initialised and, with the transport stopped,
never assigned: six benign, one at `3.0e+184`, two at `NaN`. A host-transport problem would make
all nine identical; divergence in exactly the uninitialised field is an uninitialised read.

### The code path, at `f7b97c682ade`

**1. The declaration** — [`src/common/SurgeSynthesizer.h#L41-L45`](https://github.com/surge-synthesizer/surge/blob/f7b97c682ade/src/common/SurgeSynthesizer.h#L41-L45)

```cpp
struct timedata
{
    double ppqPos, tempo;
    int timeSigNumerator = 4, timeSigDenominator = 4;
};
```

The two `int`s have default member initialisers. The two `double`s do not.

**2. `ppqPos` is assigned only while the transport is running** —
[`src/surge-xt/SurgeSynthProcessor.cpp#L739-L758`](https://github.com/surge-synthesizer/surge/blob/f7b97c682ade/src/surge-xt/SurgeSynthProcessor.cpp#L739-L758)

```cpp
        playhead->getCurrentPosition(cp);
        surge->time_data.tempo = cp.bpm;

        // isRecording should always imply isPlaying but better safe than sorry
        if (cp.isPlaying || cp.isRecording)
        {
            surge->time_data.ppqPos = cp.ppqPosition;
        }
```

`tempo` is assigned unconditionally; `ppqPos` is not. With the transport stopped it is never
written from the host at all. The `else` branch (standalone / no playhead,
[`#L759-L777`](https://github.com/surge-synthesizer/surge/blob/f7b97c682ade/src/surge-xt/SurgeSynthProcessor.cpp#L759-L777))
does not assign it either.

**3. …and then accumulated onto every block**, in both wrappers —
[`#L686-L687`](https://github.com/surge-synthesizer/surge/blob/f7b97c682ade/src/surge-xt/SurgeSynthProcessor.cpp#L686-L687)
in `processBlock` and
[`#L1156-L1157`](https://github.com/surge-synthesizer/surge/blob/f7b97c682ade/src/surge-xt/SurgeSynthProcessor.cpp#L1156-L1157)
in `clap_direct_process`:

```cpp
            surge->time_data.ppqPos +=
                (double)BLOCK_SIZE * surge->time_data.tempo / (60. * surge->storage.samplerate);
```

Both functions call `processBlockPlayhead()` first (L599 and L1077), so this is not CLAP-specific.
NaN is sticky under `+=`, so once it is NaN it stays NaN for the life of the instance. (This also
explains the six benign instances' `~31178.2` — a zero-or-otherwise-benign start value plus the
accumulation since the instance was created.)

**4. The guard that is there, and the one that is not** —
[`src/common/SurgeSynthesizer.cpp#L4265-L4283`](https://github.com/surge-synthesizer/surge/blob/f7b97c682ade/src/common/SurgeSynthesizer.cpp#L4265-L4283)

```cpp
void SurgeSynthesizer::resetStateFromTimeData()
{
    if (time_data.timeSigNumerator < 1)
        time_data.timeSigNumerator = 4;
    if (time_data.timeSigDenominator < 1)
        time_data.timeSigDenominator = 4;
    storage.songpos = time_data.ppqPos;          // <-- unguarded
    if (time_data.tempo > 0)                     // <-- tempo IS guarded, one line later
    {
        storage.temposyncratio = time_data.tempo / 120.f;
        storage.temposyncratio_inv = 1.f / storage.temposyncratio;
    }
    else
    {
        storage.temposyncratio = 1.f;
        storage.temposyncratio_inv = 1.f;
    }
}
```

Both time-signature fields and `tempo` are sanity-checked; `ppqPos` is copied straight through.
`SurgeStorage`'s constructor *does* initialise `songpos = 0`
([`SurgeStorage.cpp#L157`](https://github.com/surge-synthesizer/surge/blob/f7b97c682ade/src/common/SurgeStorage.cpp#L157)) —
that good value is then overwritten by the bad one. `SurgeSynthesizer::processControl()` makes the
same assignment at
[`#L4304-L4306`](https://github.com/surge-synthesizer/surge/blob/f7b97c682ade/src/common/SurgeSynthesizer.cpp#L4304-L4306),
with **no guard on `tempo` there either**.

**5. The consumer** — `LFOModulationSource::attackFrom()`, i.e. note attack, not per block —
[`src/common/dsp/modulators/LFOModulationSource.cpp#L283-L310`](https://github.com/surge-synthesizer/surge/blob/f7b97c682ade/src/common/dsp/modulators/LFOModulationSource.cpp#L283-L310)

```cpp
        case lm_freerun:
        {
            double ipart;

            // Get our rate from the parameter and temposync
            float lrate = pow(2.0, (double)localcopy[rate].f);
            if (lfo->rate.temposync)
                lrate *= storage->temposyncratio;

            ...
            double timePassed = storage->songpos * storage->temposyncratio_inv * 0.5;

            ...
            float totalPhase = startPhase + timePassed * lrate;

            phase = (float)modf(totalPhase, &ipart);
```

`songpos` NaN ⟹ `timePassed` NaN ⟹ `totalPhase` NaN ⟹ `phase` NaN. `process_block`'s
normalisation `if (phase >= 1 || phase < 0)`
([`#L692`](https://github.com/surge-synthesizer/surge/blob/f7b97c682ade/src/common/dsp/modulators/LFOModulationSource.cpp#L692))
is **false for NaN**, so the wrap is skipped and the shape evaluation carries NaN into
`output_multi[0]` — which is exactly what the core shows.

**6. The landing** — the NaN modulates filter 2's cutoff, so `SetQFB` computes a NaN `cutoffB`
([`SurgeVoice.cpp#L1466-L1467`](https://github.com/surge-synthesizer/surge/blob/f7b97c682ade/src/common/dsp/SurgeVoice.cpp#L1466-L1467),
inside `SetQFB` at L1382), `MakeCoeffs` applies no finiteness check, `fut_lp12` + `st_Standard`
dispatches to `Coeff_SVF`, whose first statement is
`double f = 440.f * provider->note_to_pitch_ignoring_tuning(Freq);`, and
[`SurgeStorage.cpp#L2478-L2491`](https://github.com/surge-synthesizer/surge/blob/f7b97c682ade/src/common/SurgeStorage.cpp#L2478-L2491):

```cpp
float SurgeStorage::note_to_pitch_ignoring_tuning(float x)
{
    x = limit_range(x + 256, 1.e-4f, tuning_table_size - (float)1.e-4);
    ...
    int e = (int)x;
    ...
    int pow2idx = (int)pow2pos;
    ...
        (1 - pow2frac) * table_two_to_the[pow2idx] + pow2frac * table_two_to_the[pow2idx + 1];
```

`limit_range` is `std::clamp`
([`basic_dsp.h#L29-L33`](https://github.com/surge-synthesizer/surge/blob/f7b97c682ade/src/common/dsp/vembertech/basic_dsp.h#L29-L33)),
and `std::clamp` returns NaN unchanged (`v < lo` and `hi < v` are both false), so `(int)NaN` =
`INT_MIN` indexes the table 8 GiB low.

This is not confined to `Coeff_SVF`: `note_to_omega_ignoring_tuning`
([`#L2518-L2530`](https://github.com/surge-synthesizer/surge/blob/f7b97c682ade/src/common/SurgeStorage.cpp#L2518-L2530)),
used by the other coefficient paths, has the same unmasked `[e]` index after the same `limit_range`.

### Reproduction

**This is probabilistic per instantiation, and that is inherent to the bug** — it depends on what
the memory holding `ppqPos` contains at construction. Fresh zero-filled pages give a benign `0.0`;
recycled heap gives whatever was there. In our nine-instance process, one came up at `3.0e+184` and
two at `NaN`. A single instance may need many instantiations, in a busy process, to land on a bad
value.

1. Any host, with **the transport stopped** and never started. (The standalone build is exposed
   too: its `else` branch at L759-L777 never assigns `ppqPos` either.)
2. Load Surge XT and select the factory patch **`Leads/Tok`**, which ships with 1.3.4. Its VLFO 2 is
   `trigmode = 0` (`lm_freerun`) and is routed to `a_filter2_cutoff`; the other five voice LFOs are
   `trigmode = 1` (`lm_keytrigger`); `a_filter2_type = 1` (`fut_lp12`) with `a_filter2_subtype = 0`
   (`st_Standard`), which dispatches to `Coeff_SVF`. Any factory or user patch with a freerun LFO
   routed to a filter cutoff will do — `Tok` is simply the one we hit, and the file our instance
   loaded is byte-identical (same MD5) to the shipped `patches_factory/Leads/Tok.fxp`.
3. Play one note.
4. If that instance's `ppqPos` came up non-finite, the host dies immediately.

**A deterministic reproduction, without waiting on chance:** in a debug build, set
`time_data.ppqPos = std::numeric_limits<double>::quiet_NaN();` once after construction (or break on
`resetStateFromTimeData` and poke it), keep the transport stopped, and play one note into `Tok`.
That exercises exactly the path above. A unit test could equally call `resetStateFromTimeData()`
with a NaN `ppqPos` and assert `std::isfinite(storage.songpos)`.

### Proposed fix

**Minimal, and it matches the struct's own precedent:**

```diff
 struct timedata
 {
-    double ppqPos, tempo;
+    double ppqPos{0.0}, tempo{120.0};
     int timeSigNumerator = 4, timeSigDenominator = 4;
 };
```

Two initialisers, alongside the two already there. `120.0` is the value consistent with
`SurgeStorage`'s constructor setting `temposyncratio = 1.f`
([`SurgeStorage.cpp#L150`](https://github.com/surge-synthesizer/surge/blob/f7b97c682ade/src/common/SurgeStorage.cpp#L150)),
since `temposyncratio = tempo / 120.f`.

**This does not undo #1493.** With the transport stopped, `ppqPos` still evolves — it now starts at
a defined `0.0` and accumulates by the per-block `+=`, which is exactly what that change wanted.

There is precedent a few lines away for treating an uninitialised temposync field as a bug worth
fixing —
[`SurgeStorage.cpp#L152-L157`](https://github.com/surge-synthesizer/surge/blob/f7b97c682ade/src/common/SurgeStorage.cpp#L152-L157):

```cpp
    // Use this as a sentinel, since it was not initialized prior to 1.6.5
    // this was the value at least Windows and Mac had
    // https://github.com/surge-synthesizer/surge/issues/1444
    temposyncratio_inv = 0.0f;

    songpos = 0;
```

**A defensive second, if wanted:** guard `songpos` the way `tempo` already is —

```diff
-    storage.songpos = time_data.ppqPos;
+    storage.songpos = std::isfinite(time_data.ppqPos) ? time_data.ppqPos : 0.0;
```

(and the same in `processControl()` at `SurgeSynthesizer.cpp:4304`).

**I am deliberately *not* proposing a change to `limit_range`,** because @baconpaul ruled that out
when diagnosing this same stack in #5566
([comment](https://github.com/surge-synthesizer/surge/issues/5566#issuecomment-986843286)):

> But more importantly, our limit_range assumes a finite input so with a nan all our checks blow up.
> The particular stack in this case happens in SurgeVoice where we use the pitch to interpolate
> across a 2^x interpolation array in the filter blah blah and it would be fine if we were
> accidentally out of bounds as long as we are finite
>
> I don't want to make limit_range check std::isfinite. We use that everywhere. So the fix I propose
> - since we have made sure all the rest of the modulators never make nan and we can't guarantee
> that with lua - is to just check isfinite on the outbound of the formula modulator, set a flag,
> and voila.

That fix was scoped to the formula modulator on the reasoning that *"we have made sure all the rest
of the modulators never make nan"*. **Here a non-formula modulator does make one** — not because the
modulator is wrong, but because the `songpos` it reads is. Fixing it at the source keeps that
invariant true and leaves `limit_range` alone.

### Related

- **#5566** — same stack, via a Lua formula modulator; fixed 2021 at the formula output only.
- **#2837** — *"Make sure time_data and Storage are consistent"*, which already names the failure
  mode: *"the random ppqpos value was snapped for freerun LFO phase"*. That fix addressed an
  ordering race; here the same value is garbage at construction.
- **#1491 / #1493** — the origin of the `if (cp.isPlaying || cp.isRecording)` guard.
- **#5812** — *"Freerun doesn't freerun if transport stopped and loop mode enabled"*: same feature
  area and transport condition, reported as wrong behaviour rather than a crash.
- **#5534 / #5536** — introduced the per-block `ppqPos +=` accumulation.
- **#2830**, **#7738**, and **#4291** (which already flags this function's bounds handling).
- **#1444** — the issue linked from the `temposyncratio_inv` comment quoted above.

Searching issues and PRs for `songpos`, `ppqPos`, `uninitialized timedata`,
`note_to_pitch_ignoring_tuning` and NaN crashes, I found none covering an uninitialised `timedata`,
a NaN `songpos`, or a crash from this path.

### Still present on `main`

Checked against `main` today:

- `struct timedata` gains only `bool isPlaying = false;` — **`double ppqPos, tempo;` are still
  uninitialised**.
- `resetStateFromTimeData()` gains only `storage.isPlaying = time_data.isPlaying;` — **the
  `songpos` assignment is still unguarded**.
- `note_to_pitch_ignoring_tuning` is **byte-identical** to 1.3.4.
- `LFOModulationSource.cpp`'s freerun block is **byte-identical** to 1.3.4.

`release_xt_1.3.4` is also the newest release tag (14 `release_xt_*` tags), so there is no released
version that fixes it.

### What I can provide

- A core dump of the crashed host (10.9 GB compressed, 34 GB extracted). It is a full DAW process
  dump, so it needs care before sharing, but I can extract any specific object or field on request.
- The decoded offsets used above, so the numbers can be re-derived independently: within
  `SurgeStorage`, `table_two_to_the` at `+0x16E80`, `vu_falloff` at `+0x18E04`, `temposyncratio` at
  `+0x18E08`, `songpos` at `+0x18E18`, `oscPortIn` at `+0x54F44`; `timedata` sits immediately before
  `SurgeStorage storage alignas(16)`, so `ppqPos` is at `storage - 0x20`. (Offsets validated by
  `vu_falloff == exp(-2*pi*60/48000)` and by all nine instances' `oscPortIn` forming the expected
  dense port set.)
- Any further register or memory read from the core that would help.

Happy to open a PR with the two initialisers if that is welcome.

mkruselj:
Fixes #8533

### DIFF
--- a/src/common/SurgeSynthesizer.h
+++ b/src/common/SurgeSynthesizer.h
@@ -40,9 +40,9 @@ struct QuadFilterChainState;
 
 struct timedata
 {
-    double ppqPos, tempo;
-    int timeSigNumerator = 4, timeSigDenominator = 4;
-    bool isPlaying = false;
+    double ppqPos{0.0}, tempo{120.0};
+    int timeSigNumerator{4}, timeSigDenominator{4};
+    bool isPlaying{false};
 };
 
 struct parametermeta


baconpaul:
oh thanks!

PR #8535, opened 2026-09-01
Persist MPE enabled state to user defaults

vicwomg:
Saves the MPE enabled toggle in user defaults so that MPE mode persists across application restarts and patch loading.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,0 +1,1 @@
+SURGE_IOS_DEV_TEAM=<your_dev_team_id_here>
--- a/.gitignore
+++ b/.gitignore
@@ -88,6 +88,7 @@ build-arm/
 build_lv2
 buildiwyu/
 buildpy/
+build_ios/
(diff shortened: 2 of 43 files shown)


vicwomg:
closing, did not mean to add the rest of my fork

mkruselj:
This should not be necessary, Surge responds to the RPN message to enable MPE remotely. So this just needs to be sent by the host.

vicwomg:
> This should not be necessary, Surge responds to the RPN message to enable MPE remotely. So this just needs to be sent by the host.

That's true, but that usually occurs when the controller is first connected. If the controller is plugged in prior to launching Surge XT the RPN is never synchronized, right?

mkruselj:
Depends on the controller or host I guess... for example in Ableton Live you can enable MPE mode per plugin and this sends RPN on every plugin instantiation.

Issue #7570, opened 2024-03-13
BLIT oscillators are a bit late


Andreya-Autumn:
Or should that be "a blit late"? :P

Anyway: 
![Old vs  new?](https://github.com/surge-synthesizer/surge/assets/105538426/994b8479-01f7-4696-8a4b-e504634f670f)

Classic seems to start late by a varying amount somewhere between 40 and 80 samples, wavetable ditto by about 80-160 samples. 

S&H Noise is interesting. With Correlation at max or min it behaves exactly like Classic. But with correlation at default it seems like it's accurate! As accurate as the MIDI block processing will let it that is. Maybe there's a clue there somehow?

blancoberg:
I stumbled upon this bug as well. To me it looks like instead of starting on a random phase, it instead delays the signal for 0-1 cycle. So the delay varies depending on what note you are playing.
On lower notes the delay can be quite substantial

mkruselj:
That's a plausible explanation!

baconpaul:
Yes that’s exactly what they do and what the test above shows.

the fix to in init start them at start phase - 1 then run them up to zero but when I did that they clicked like crazy at low attacks.

lluisestape-upc:
I had a dig at this, and I think the numbers in the original report pin the mechanism down exactly.

`oscstate` is measured in cycles, not samples — `process_block` subtracts `a = BLOCK_SIZE_OS * pitchmult` per block, and `pitchmult` is the reciprocal of the period. So in the non-retrigger branch of `ClassicOscillator::init`:

```cpp
double st = 0.5 * drand * storage->note_to_pitch_inv_tuningctr(detune);
oscstate[i] = st;
```

`oscstate` starts uniform in [0, 0.5) cycles. With `oscbuffer`/`dcbuffer` freshly memset, `last_level = 0` and `dc_uni = 0`, there's nothing to emit until the first `convolute()` fires — so the output is silent for `st * period` samples. That isn't a random start phase, it's a random start *delay*.

The reported numbers line up with that. At MIDI 60 / 44.1k the period is ~168 samples, so Classic should be silent for 0-84 samples; `WavetableOscillator::init` uses `oscstate[i] = drand` with no 0.5 factor, giving 0-168. That matches the "40 to 80" and "80 to 160" in the first post, including the 2x ratio between them.

[...]

PR #8537, opened 2026-09-02
Sanitize non-ASCII characters in the build hostname

lluisestape-upc:
### Problem

Surge does not build on Windows/MSVC when the machine's hostname contains non-ASCII characters.

`cmake/versiontools.cmake` reads the hostname with `cmake_host_system_information(... QUERY FQDN)` and `version.cpp.in` embeds it verbatim into a narrow C string literal:

```c
const char* Build::BuildHost = "@SURGE_BUILD_FQDN@";
```

On a machine named `ThinkPadLluís` the generated `version.cpp` contains a byte MSVC cannot interpret in that literal, and the build fails during `surge-common`:

```
version.cpp(26,35): error C2001: newline in constant
version.cpp(27,4): error C2143: syntax error: missing ';' before 'const'
version.cpp(1,1): error C2220: the following warning is treated as an error
```

Hostnames with accents, umlauts or CJK characters are perfectly ordinary, so this is reachable by anyone building on Windows with a localized machine name.

### Fix

Replace anything outside printable ASCII with an underscore, immediately after the FQDN is queried. `SURGE_BUILD_FQDN` only feeds the `BuildHost` display string, so degrading it is harmless — the above host reports as `ThinkPadLlu_s`.

### Testing

Verified locally on Windows 11 / MSVC 19.44 with VS 2022:

- before: `surge-testrunner` fails to build with the errors above
- after: `version.cpp` generates a valid literal and `surge-testrunner` builds and runs clean

Found while investigating #7570.

### DIFF
--- a/cmake/versiontools.cmake
+++ b/cmake/versiontools.cmake
@@ -55,6 +55,11 @@ endif ()
 
 cmake_host_system_information(RESULT SURGE_BUILD_FQDN QUERY FQDN)
 
+# The FQDN is embedded verbatim into a C string literal in version.cpp.in.
+# Hostnames legitimately contain non-ASCII characters, which MSVC rejects
+# in a narrow literal, so replace anything outside printable ASCII.
+string(REGEX REPLACE "[^ -~]" "_" SURGE_BUILD_FQDN "${SURGE_BUILD_FQDN}")
+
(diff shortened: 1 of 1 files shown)


baconpaul:
Thanks!

PR #8541, opened 2026-09-03
Fix pitch ramp in String oscillator when changing scene tuning

### LINKED ISSUE #6985: changing pitch of string oscilator produces a small pitch ramp on note on
If you set an osc to string, and adjust its pitch using either scene or osc controls, the osc pitch seems to be initialized with the untransposed pitch and has to ramp to the correct one. Eg. Raising the osc pitch by 3 octaves sounds very different from transposing the incoming midi data. In my testing, no other osc types exhibited this behavior.

mkruselj:
`switch_toggled()` (runs at voice construction, right before the initial `osc[i]->init(usep, ...)` call):

```cpp
float pb = modsources[ms_pitchbend]->get_output(0);
if (pb > 0) pb *= ...
else pb *= ...
state.scenepbpitch = pb;                          // <-- ONLY pitch bend
state.pitch = state.pkey + state.scenepbpitch;
```
`process_block()` (runs every block afterward):
```cpp
state.scenepbpitch = pb + localcopy[pitch_id].f * (scene->pitch.extend_range ? 12.f : 1.f) +
                     (octaveSize * localcopy[octave_id].i);   // <-- scene Pitch AND scene Octave included
state.pitch = state.pkey + state.scenepbpitch;
```
`switch_toggled()`'s version of `scenepbpitch` silently drops the scene Pitch and Octave contribution entirely. Since `switch_toggled()` is what seeds the initial `osc[i]->init(usep, ...)` call for a brand-new voice (comment at the top of the constructor: "`switch_toggled()` comes last since it creates and activates the oscillators"), the String oscillator's `tap[0].startValue(pitchmult_inv)` gets seeded from a pitch that's missing the scene octave offset. The very next `process_block()` recomputes `scenepbpitch` correctly (now including octave), passes the correct pitch into `StringOscillator::process_block_internal()`, which calls `tap[0].newValue(pitchmult_inv)` - and the lag smoother glides from the wrong (octave-less) seed up to the correct pitch.

Scene octave = 0 → the missing term is 0 → no discrepancy → no audible ramp.
Scene octave = 3 → seed is 3 octaves flat, corrects itself over the smoother's time constant → audible ramp.

Per-osc octave/pitch never shows this because that offset (`octaveSize * scene->osc[i].octave.val.i`) is added identically in both `switch_toggled()` (line ~556) and `process_block()` (lines ~1080 etc.) — no mismatch possible there.

Closes #6985!

Assisted-By: Claude Sonnet 5 <noreply@anthropic.com>

### DIFF
--- a/src/common/dsp/SurgeVoice.cpp
+++ b/src/common/dsp/SurgeVoice.cpp
@@ -516,7 +516,8 @@ void SurgeVoice::switch_toggled()
 
     // scenepbpitch is pitch state but without state.pkey, so that it can be used to add
     // scene pitch/octave, pitch bend and associated modulations to non-keytracked oscillators
-    state.scenepbpitch = pb;
+    state.scenepbpitch = pb + localcopy[pitch_id].f * (scene->pitch.extend_range ? 12.f : 1.f) +
+                         (octaveSize * localcopy[octave_id].i);
     state.pitch = state.pkey + state.scenepbpitch;
 
     modsources[ms_keytrack]->set_output(0, (state.pitch - (float)scene->keytrack_root.val.i) *
@@ -1702,11 +1703,10 @@ void SurgeVoice::retriggerOSCWithIndependentAttacks()
         {
             // This matches the override in ::process_block
             float ktrkroot = 60;
-            auto usep = noteShiftFromPitchParam((scene->osc[i].keytrack.val.b
-                                                     ? state.getPitch(storage)
-                                                     : ktrkroot + state.scenepbpitch) +
-                                                    octaveSize * scene->osc[i].octave.val.i,
-                                                0);
+            auto usep = noteShiftFromPitchParam(
+                (scene->osc[i].keytrack.val.b ? state.pitch : ktrkroot + state.scenepbpitch) +
+                    octaveSize * scene->osc[i].octave.val.i,
+                0);
 
             /*
              * This is awfully special case but it's the best solution
--- a/src/common/dsp/SurgeVoice.cpp
+++ b/src/common/dsp/SurgeVoice.cpp
@@ -1703,10 +1703,11 @@ void SurgeVoice::retriggerOSCWithIndependentAttacks()
         {
             // This matches the override in ::process_block
             float ktrkroot = 60;
-            auto usep = noteShiftFromPitchParam(
-                (scene->osc[i].keytrack.val.b ? state.pitch : ktrkroot + state.scenepbpitch) +
-                    octaveSize * scene->osc[i].octave.val.i,
-                0);
+            auto usep = noteShiftFromPitchParam((scene->osc[i].keytrack.val.b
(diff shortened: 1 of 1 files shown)


baconpaul:
Will get to this one this weekend. Want to read some surrounding code

baconpaul:
That retrigger case I am pretty sure is wrong. Moving from .getPitch to .pitch actually uses the wrong key/note in the retriever case.

src/common/dsp/SurgeVoice.cpp
```diff
@@ -1702,11 +1703,10 @@ void SurgeVoice::retriggerOSCWithIndependentAttacks()
...
             // This matches the override in ::process_block
             float ktrkroot = 60;
-            auto usep = noteShiftFromPitchParam((scene->osc[i].keytrack.val.b
-                                                     ? state.getPitch(storage)
-                                                     : ktrkroot + state.scenepbpitch) +
-                                                    octaveSize * scene->osc[i].octave.val.i,
-                                                0);
+            auto usep = noteShiftFromPitchParam(
+                (scene->osc[i].keytrack.val.b ? state.pitch : ktrkroot + state.scenepbpitch) +
+                    octaveSize * scene->osc[i].octave.val.i,
```
baconpaul:
```
(scene->osc[i].keytrack.val.b ? state.getPitch(storage) : ktrkroot) + state.scenepbpitch +
      octaveSize * scene->osc[i].octave.val.i
```

I believe is the right actual statement here. state.pitch has not been updated to the new voice at this point, but state.getPitch which uses key will have been. The original put the +scenepbpitch in the wrong spot but fixing the ternary I think is the right thing here rather than .pitch -> .getPitch

We should investigate this before we merge.

src/common/dsp/SurgeVoice.cpp
```diff
@@ -516,7 +516,8 @@ void SurgeVoice::switch_toggled()
 
     // scenepbpitch is pitch state but without state.pkey, so that it can be used to add
     // scene pitch/octave, pitch bend and associated modulations to non-keytracked oscillators
-    state.scenepbpitch = pb;
+    state.scenepbpitch = pb + localcopy[pitch_id].f * (scene->pitch.extend_range ? 12.f : 1.f) +
+                         (octaveSize * localcopy[octave_id].i);
```
baconpaul, line 520: This is correct

baconpaul (dismissed):

src/common/dsp/SurgeVoice.cpp
mkruselj: Claude seems to agree and I'm building with that change to check it out.

Claude is also asking if the calculation in `switch_toggled()` might be incorrect in the same way, since it uses `state.pitch` instead of `state.getPitch(storage)`?

mkruselj: Oh and I just built with your change and the fix still works, all good!

PR #8543, opened 2026-09-03
Start BLIT oscillators mid-cycle instead of delaying their first impulse

### LINKED ISSUE #7570: BLIT oscillators are a bit late

lluisestape-upc:
Fixes #7570.

### The problem

With retrigger off, the three `AbstractBlitOscillator` subclasses seeded `oscstate` with a positive random value:

```cpp
double st = 0.5 * drand * storage->note_to_pitch_inv_tuningctr(detune);
oscstate[i] = st;
```

`oscstate` is remaining phase space, not phase. With the buffers freshly memset and `last_level` and `dc_uni` at zero, there is nothing to emit until the first `convolute` fires, so the voice is simply silent for `oscstate * period` samples. That is a random *delay*, not a random start phase.

Measured onset delay before this change, in oversampled samples, 300 inits per point:

| Oscillator | MIDI 24 | MIDI 60 | MIDI 96 |
|---|---|---|---|
| Classic | 665.6 | 82.8 | 10.0 |
| Wavetable | 1263.1 | 167.5 | 21.7 |
| S&H Noise | 655.4 | 81.5 | 9.9 |

It halves per octave, as a fixed fraction of a cycle should. At MIDI 24 that is 7.5 ms off the front of the attack on average, and up to about 15 ms worst case.

### The change

Each oscillator starts mid-cycle instead, but the right way to do that differs by oscillator, so the three are not the same patch:

- **Classic** has a four segment cycle summing to `2 * t`, with the waveform tracked as `last_level` plus a `dc_uni` ramp. `init` picks a uniform point in the cycle and replays that bookkeeping up to the segment holding it. State 0 sets the level absolutely rather than incrementally, so replaying from zero lands on the right answer. It seeds `last_level`, `dc_uni`, `dcbuffer`, `state`, `pwidth`/`pwidth2` and `oscstate`. The `dc_uni` slope is computed with `mech::rcp(t)` rather than `/ t`, to match `convolute` exactly. The value is otherwise carried forward as `olddc` on the next call, so a mismatch would only differ over the first partial segment, but keeping them identical avoids the confusion.
- **Wavetable** emits pure steps with no DC ramp, so it just starts at a random table index with `oscstate` at zero. The first `convolute` then fires immediately and steps to the correct level *through the existing sinc convolution*, so the onset is band limited and no integrator state needs priming at all.
- **S&H Noise** holds a random value by construction, so there is no level to reconstruct. It draws one, and starts partway through the segment holding it.

Classic and S&H both need the shared integrator seeded with the level their voices start at. That depends on the stereo flag, which `init` does not receive, so it happens on the first `process_block` behind the existing `first_run` flag, which until now was written but never read.

Worth a second look when reviewing: the two classes did not treat that flag the same way. `ClassicOscillator` already cleared it at the end of `process_block`, but `SampleAndHoldOscillator` only ever set it in the constructor, so this change clears it there as soon as it has been used.

Each oscillator makes the same number of RNG draws as before, so the rest of the random sequence is unchanged.

### Testing

The `Oscillator Onset` regtest from #7571 now asserts rather than just computing a number. It covers all eight oscillator types at MIDI 24, 60 and 96 with retrigger both on and off, and requires a worst case onset of at most 16 oversampled samples.

- It **fails on master**, at `Classic at note 24 with retrigger off`, the first case it reaches.
- It passes with this change, and every oscillator lands at 1 to 2 samples with retrigger off, matching or beating the retrigger-on path.

The whole test runner passes on both platforms I have:

| | test cases | assertions |
|---|---|---|
| Windows, MSVC, Release | 108 | 5,660,060 |
| Linux (WSL2), Clang, Debug + ASan/UBSan | 108 | 5,680,582 |

ASan reports nothing. UBSan reports one pre-existing issue unrelated to this change: a misaligned `int16_t` load in `sst-basic-blocks`, reached from `Wavetable::BuildWT` while parsing patch wavetables. Happy to open that separately if it is not already known.

### What I would like eyes on

Starting at an arbitrary phase means starting at an arbitrary level, so a step at note-on is unavoidable. The question is whether it is bigger than what the waveform does anyway.

At the bare oscillator, MIDI 36, worst of 16 note-ons. The jump at sample 0 is measured from silence, since the buffers start empty:

| Oscillator | retrigger on | retrigger off | its own largest step while running |
|---|---|---|---|
| Classic | 0.0007 | **1.24** | 0.45 |
| Wavetable | 0.0000 | 0.0013 | 0.0044 |
| S&H Noise | 0.0003 | 0.45 | 0.45 |

Only Classic exceeds its own normal behaviour, by about 2.7x. Wavetable is effectively perfect because its opening step goes through `convolute` and comes out band limited. S&H matches its own step size by construction. It is a staircase, so starting on a stair is in character.

Classic is the one that seeds `osc_out` directly, so its opening step is a raw discontinuity rather than a band limited one.

At the synth output it looks better. Init patch, Classic, fastest amp attack and full sustain so nothing masks the onset, worst of 16 note-ons:

| | largest step in the first 200 samples | largest step while running |
|---|---|---|
| retrigger on | 0.10 | 0.22 |
| retrigger off | 0.20 | 0.22 |

So by the time it reaches the output the onset step no longer exceeds the waveform's own edges.

It is still visible, though. Captured in a DAW, a note-on with retrigger off starts with a single sample step to an arbitrary level, where retrigger on ramps up from zero. It did not sound like anything in play testing, at any tempo or register I tried, but I would rather flag it than quietly file it as fixed.

If you do want it gone, the direction I would try is routing Classic's opening level through the sinc convolution the way Wavetable does, though the DC ramp makes that less straightforward.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -65,3 +65,4 @@ Daniel Hatadi <https://github.com/dhatadi>
 Yashas Hebbarabailu <https://github.com/y-hbb>
 Bill Allen <https://github.com/bwanab>
 Manu Arimaka <https://github.com/manuaudio>
+Lluís Estapé <https://github.com/lluisestape-upc>
--- a/src/common/dsp/oscillators/ClassicOscillator.cpp
+++ b/src/common/dsp/oscillators/ClassicOscillator.cpp
@@ -256,7 +256,25 @@ void ClassicOscillator::init(float pitch, bool is_display, bool nonzero_init_dri
             */
             double detune = oscdata->p[co_unison_detune].get_extended(localcopy[id_detune].f) *
                             (detune_bias * float(i) + detune_offset);
-            float t = storage->note_to_pitch_inv_tuningctr(detune);
+
+            /*
+            ** Mirror ::convolute's non-absolute t, sync included, so a synced voice is seeded
+            ** from the period it will actually run at. The absolute branch is deliberately
+            ** left alone: its formula is samplerate dependent and known to be odd (see the
+            ** comment in ::convolute), so seeding from the unsynced period there keeps the
+            ** existing behaviour rather than baking the oddity into a second place.
+            */
+            float t;
+
+            if (oscdata->p[co_unison_detune].absolute)
+            {
+                t = storage->note_to_pitch_inv_tuningctr(detune);
+            }
+            else
+            {
+                float sync = min((float)l_sync.v, (12 + 72 + 72) - pitch);
+                t = storage->note_to_pitch_inv_tuningctr(detune + sync);
+            }
 
             float pw = limit_range(l_pw.v, 0.001f, 0.999f);
             float pw2 = 2.f * l_pw2.v;
--- a/src/common/dsp/oscillators/ClassicOscillator.cpp
+++ b/src/common/dsp/oscillators/ClassicOscillator.cpp
@@ -232,21 +232,103 @@ void ClassicOscillator::init(float pitch, bool is_display, bool nonzero_init_dri
             oscstate[i] = 0.f;
             syncstate[i] = 0.f;
             last_level[i] = 0.f;
+            start_level[i] = 0.f;
+            dc_uni[i] = 0.f;
+            state[i] = 0;
+            pwidth[i] = limit_range(l_pw.v, 0.001f, 0.999f);
+            pwidth2[i] = 2.f * l_pw2.v;
         }
         else
         {
-            double drand = (double)storage->rand_01();
+            /*
+            ** Start the voice at a random point inside the cycle, carrying the state it
+            ** would have had if it had been running, rather than delaying its first impulse
+            ** by a random amount. Seeding oscstate with a positive value and nothing else
+            ** leaves the buffers empty, so the voice is simply silent until the first
+            ** ::convolute fires - a delay of up to a full cycle. See issue #7570.
+            **
+            ** ::convolute walks a four segment cycle whose durations sum to 2 * t, tracking
+            ** the waveform as last_level (the level at the end of the current segment) and
+            ** dc_uni (the slope across it). So: pick a uniform point in the cycle, replay
+            ** that bookkeeping up to the segment holding it, and hand the voice over ready
+            ** for the next ::convolute. State 0 sets the level absolutely rather than
+            ** incrementally, so replaying from a zero level lands on the right answer.
+            */
             double detune = oscdata->p[co_unison_detune].get_extended(localcopy[id_detune].f) *
                             (detune_bias * float(i) + detune_offset);
-            double st = 0.5 * drand * storage->note_to_pitch_inv_tuningctr(detune);
-            oscstate[i] = st;
-            syncstate[i] = st;
-            last_level[i] = 0.f;
+            float t = storage->note_to_pitch_inv_tuningctr(detune);
+
+            float pw = limit_range(l_pw.v, 0.001f, 0.999f);
+            float pw2 = 2.f * l_pw2.v;
+            float wf = l_shape.v;
+            float sub = l_sub.v;
+
+            // These mirror the rate computation at the end of ::convolute
+            float seg[4] = {t * pw * pw2, t * (1.f - pw) * (2.f - pw2), t * pw * (2.f - pw2),
+                            t * (1.f - pw) * pw2};
+            float cycle = seg[0] + seg[1] + seg[2] + seg[3];
+
+            // One draw per voice, as before, so the rest of the random sequence is unchanged
+            float phase = storage->rand_01() * cycle;
+
+            // mech::rcp, not a division, to match how ::convolute computes this
+            float dcu = (1.f + wf) * (1.f - sub) * mech::rcp(t);
+            float level = 0.f, lvl_start = 0.f, acc = 0.f;
+            int s = 0;
+
+            for (int step = 0; step < 4; ++step)
+            {
+                float g;
+
+                switch (s)
+                {
+                case 0:
+                {
+                    float tg =
+                        ((1 + wf) * 0.5f + (1 - pw) * (-wf)) * (1 - sub) + 0.5f * sub * (2.f - pw2);
+                    g = tg - level;
+                    break;
+                }
+                case 1:
+                    g = wf * (1.f - sub) - sub;
+                    break;
+                case 2:
+                    g = 1.f - sub;
+                    break;
+                default:
+                    g = wf * (1.f - sub) + sub;
+                    break;
+                }
+
+                lvl_start = level + g;
+                level = lvl_start - seg[s] * dcu;
+
+                if (acc + seg[s] > phase)
+                {
+                    break;
+                }
+
+                acc += seg[s];
+                s = (s + 1) & 3;
+            }
+
+            float elapsed = phase - acc;
+            float frac = (seg[s] > 0.f) ? (elapsed / seg[s]) : 0.f;
+
+            oscstate[i] = seg[s] - elapsed;
+            syncstate[i] = oscstate[i];
+            last_level[i] = level;
+
+            // The level runs linearly from just after the impulse to last_level
+            start_level[i] = lvl_start + (level - lvl_start) * frac;
+
+            dc_uni[i] = dcu;
+            dcbuffer[bufpos + FIRoffset] += dcu;
+            state[i] = (s + 1) & 3;
+            pwidth[i] = pw;
+            pwidth2[i] = pw2;
         }
 
-        dc_uni[i] = 0.f;
-        state[i] = 0.f;
-        pwidth[i] = limit_range(l_pw.v, 0.001f, 0.999f);
         driftLFO[i].init(nonzero_init_drift);
     }
 }
@@ -614,6 +696,34 @@ void ClassicOscillator::process_block(float pitch0, float drift, bool stereo, bo
                                       storage->note_to_pitch_inv(pitch));
     // This must be a real division, reciprocal approximation is not precise enough
     pitchmult = 1.f / pitchmult_inv;
+
+    /*
+    ** ::init chose a start phase for each voice but could not seed the shared integrator,
+    ** because how much level a voice contributes depends on the stereo flag, which only
+    ** arrives here. Do it on the first block, before anything is generated. See issue #7570.
+    */
+    if (first_run)
+    {
+        float sL = 0.f, sR = 0.f;
+
+        for (int u = 0; u < n_unison; u++)
+        {
+            float v = start_level[u] * out_attenuation;
+
+            if (stereo)
+            {
+                sL += v * panL[u];
+                sR += v * panR[u];
+            }
+            else
+            {
+                sL += v;
+            }
+        }
+
+        osc_out = SIMD_MM(set1_ps)(sL);
+        osc_outR = SIMD_MM(set1_ps)(sR);
+    }
 
     int k, l;
 
--- a/src/common/dsp/oscillators/ClassicOscillator.cpp
+++ b/src/common/dsp/oscillators/ClassicOscillator.cpp
@@ -303,7 +303,10 @@ void ClassicOscillator::init(float pitch, bool is_display, bool nonzero_init_dri
                 lvl_start = level + g;
                 level = lvl_start - seg[s] * dcu;
 
-                if (acc + seg[s] > phase)
+                // >=, not >: rand_01() can return 1.0f, so phase can equal cycle. With >
+                // the loop would fall out with s wrapped back to 0 and a stale start_level;
+                // >= breaks at the last segment with frac == 1, which is the correct end.
+                if (acc + seg[s] >= phase)
                 {
                     break;
                 }
--- a/src/common/dsp/oscillators/ClassicOscillator.h
+++ b/src/common/dsp/oscillators/ClassicOscillator.h
@@ -56,6 +56,10 @@ class ClassicOscillator : public AbstractBlitOscillator
     bool first_run;
     float dc, dc_uni[MAX_UNISON], elapsed_time[MAX_UNISON], last_level[MAX_UNISON],
         pwidth[MAX_UNISON], pwidth2[MAX_UNISON];
+    // Output level each voice starts at, from the phase ::init picked for it. The shared
+    // integrator is seeded from these on the first ::process_block, which is the first
+    // point at which the stereo flag is known.
+    float start_level[MAX_UNISON];
     template <bool is_init> void update_lagvals();
     float pitch;
     lipol_ps li_hpf, li_DC;
--- a/src/common/dsp/oscillators/SampleAndHoldOscillator.cpp
+++ b/src/common/dsp/oscillators/SampleAndHoldOscillator.cpp
@@ -109,21 +109,41 @@ void SampleAndHoldOscillator::init(float pitch, bool is_display, bool nonzero_in
         {
             oscstate[i] = 0;
             syncstate[i] = 0;
+            last_level[i] = 0.0;
+            start_level[i] = 0.f;
+            pwidth[i] = limit_range(l_pw.v, 0.001, 0.999);
         }
         else
         {
-            double drand = (double)storage->rand_01();
+            /*
+            ** Start partway through a held value rather than delaying the first step by a
+            ** random amount, which is what seeding oscstate does and is the bug in #7570.
+            **
+            ** Unlike the Classic oscillator there is no level to reconstruct: the held
+            ** value is random by construction, so any draw from the same range is as
+            ** correct as any other. Pick one, and put the voice partway through the
+            ** segment holding it.
+            **
+            ** Two draws per voice, matching the previous code, which also drew twice and
+            ** discarded the second - so the rest of the random sequence is unchanged.
+            */
             double detune = oscdata->p[shn_unison_detune].get_extended(localcopy[id_detune].f) *
                             (detune_bias * float(i) + detune_offset);
-            double st = drand * storage->note_to_pitch_tuningctr(detune) * 0.5;
-            drand = (double)storage->rand_01();
-            double ot = drand * storage->note_to_pitch_tuningctr(detune);
-            oscstate[i] = st;
-            syncstate[i] = st;
+            float phase = storage->rand_01();
+            float level = storage->rand_01() - 0.5f;
+
+            // l_pw is a lag<double> in this oscillator, unlike the Classic one
+            float pw = (float)limit_range(l_pw.v, 0.001, 0.999);
+            // convolute uses the inverted form; state 0 runs for t * pwidth
+            float seg = storage->note_to_pitch_inv_tuningctr(detune) * pw;
+
+            oscstate[i] = seg * (1.f - phase);
+            syncstate[i] = oscstate[i];
+            last_level[i] = level;
+            start_level[i] = level;
+            pwidth[i] = pw;
         }
         state[i] = 0;
-        last_level[i] = 0.0;
-        pwidth[i] = limit_range(l_pw.v, 0.001, 0.999);
         driftLFO[i].init(nonzero_init_drift);
     }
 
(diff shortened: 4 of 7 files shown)


mkruselj:
@Andreya-Autumn Care to do the honors in checking this one out? 🙂

baconpaul:
Caveat up front: I'm still mid-way through my own review of this. What follows is the result of a separate AI read-through, posted verbatim — I haven't finished checking all of it myself, so please treat it as input to the discussion rather than as my review. I'll follow up with my own conclusions.

---

I applied the branch locally, built the test runner, instrumented `convolute` with call counters, and measured against both `main` and the PR. Here's what I found.

## Verdict

The diagnosis is right and the Classic replay math is correct — I checked each of the four `g`/drop expressions against `convolute`'s switch and the `rate` derivation, and they match exactly. Nicely, the level computation turns out to be **invariant to `t`**: `seg[]` scales with `t`, `dcu` scales with `1/t`, and `frac` is a pure ratio, so `start_level` and `last_level` are right even when `init`'s period estimate isn't. That's a real robustness property and worth keeping.

One genuine regression, one case the fix doesn't reach, and a factoring hazard.

## 1. Wavetable: the mipmap isn't computed on the first convolute (must fix)

`convolute` recomputes `mipmap[voice]`/`mipmap_ofs[voice]` only inside `if (state[voice] == 0)` (WavetableOscillator.cpp:290). `init` leaves them at 0 (line 145). Starting `state[i]` at a random index means that block is skipped, so the voice runs the **un-mipmapped** table until `state` wraps to 0 — i.e. `wt_inc = 1` instead of `1 << 6`, and `size - r` convolutes for the first partial cycle instead of `size >> 6`.

Convolute calls, one unison voice, mean of 20 note-ons, debug build:

| table | note | retrigger | first block | steady state |
|---|---|---|---|---|
| 512 (init patch) | 96 | on | 49 | 48.6 |
| 512 | 96 | off, main | 34.8 | 48.6 |
| 512 | 96 | off, **PR** | **283.5** (251 at mipmap 0) | 48.6 |
| 512 | 120 | off, **PR** | **309.4** | 48.6 |
| 2048 (Sine To Sawtooth HQ) | 96 | off, main | 32.1 | 48.6 |
| 2048 | 96 | off, **PR** | **1083.5** | 48.6 |

Retrigger defaults to off (SurgePatch.cpp:894), so this is the normal path. Scale it: 8 notes × 16 unison × 3 wavetable oscs on a 4096-frame factory table is ~100k extra 12-float convolutions landing in one block. The audio is fine — the sinc convolution still band-limits it — but it's a note-on CPU spike that wasn't there before.

`WavetableOscillator` already declares and sets `first_run` and never reads it, so the fix is available in place: `if (state[voice] == 0 || first_run)`, then clear it. `state & (wtsize - 1)` further down already wraps the seeded index into the mipmapped range, so nothing else needs to change.

## 2. Sync: the fix doesn't reach it

`init` computes `t = note_to_pitch_inv_tuningctr(detune)`, but `convolute` uses `note_to_pitch_inv_tuningctr(detune + sync)` (and a completely different formula under absolute detune). So the seeded first segment is sized from the *un-synced* period.

Blocks until the first impulse, Classic, retrigger off, MIDI 24, worst of 50:

| sync | 0 | +12 | +36 | +60 |
|---|---|---|---|---|
| blocks | 20 | 21 | 20 | 21 |

Identical — at sync +60 the first segment should be 32× shorter. It doesn't blow up (peak stays 0.83, and the total level drop across the segment is `t`-independent), but for up to ~7 ms at MIDI 24 you get an un-synced glide instead of the sync waveform. Main gave silence there, so it's not a regression — just an incomplete fix. Mirroring `convolute`'s `t` (including the `absolute` branch) closes it.

## 3. Clicks and pops

**For the normal note-on path, no — and for a structural reason, not a lucky one.** The old code's first `convolute` already injected the *absolute* level (`g = tg - last_level` with `last_level == 0`) as a step into the leaky integrator, so a step-into-DC-then-decay transient was always present. The PR moves it to sample 0 and randomizes which level it is. On top of that, `SurgeVoice` zeroes `FBP` at construction (SurgeVoice.cpp:262), so `Q->dGain` ramps the amp gain linearly from 0 across the first oversampled block — the sample-0 edge is delivered at ~zero gain.

Measured at MIDI 36, 32 note-ons, `|output[n] - output[n-1]|` for n < 4 vs. the steady-state maximum step and peak:

| Classic params | openStep main | openStep PR | steady step | steady peak |
|---|---|---|---|---|
| defaults | 0.000 | 0.498 | 0.454 | 0.727 |
| Shape +1 | 0.001 | 0.491 | 0.454 | 0.639 |
| Sub 100% | 0.000 | 0.499 | 0.455 | 0.962 |
| W1 2%, W2 2%, Shape +1 | 0.000 | **2.752** | 0.792 | 2.988 |
| (retrigger on) | 0.016–0.031 | unchanged | | |

So at default settings the opening step is at parity with the waveform's own largest edge. The two places it can bite:

- **Extreme Width/Shape.** The Classic oscillator's internal level genuinely swings to ±3 there, and the opening jump lands anywhere in that range — 2.75 at sample 0 against a running step of 0.79. It's *inside* the waveform's own peak (2.99), so it's a DC offset the keytracked HPF sheds over its time constant rather than a broadband click. But main put ~0 there. Worth an ear at a fast attack and full sustain.
- **Mid-note re-init.** `SurgeVoice::switch_toggled()` calls `osc[i]->init()` on a *sounding* voice whenever the oscillator type changes (SurgeVoice.cpp:545–558). That's the one path where the amp EG isn't at zero, so the new opening level arrives at full gain. `retriggerOSCWithIndependentAttacks` only re-inits `ot_string`/`ot_twist`, and legato/portamento don't re-init, so this is the only one.

I also checked the character filter: `osc_out2` is left at 0 while `osc_out` is seeded, but `LastOscOut` comes from the seeded `osc_out`, and for both filtering characters the DC gain is unity, so the first sample lands right. Not a problem.

## 4. Smaller things

- **`>` should be `>=` in the Classic replay loop.** `std::uniform_real_distribution<float>` can return 1.0f, and `phase = rand * cycle` can round up to `cycle` anyway. Then `acc + seg[3] > phase` is false, the loop falls out with `s` wrapped back to 0, and you get `state = 1` with a `start_level` from segment 3 — a cycle of wrong duty before case 0's absolute reset repairs it. Probability ~2^-24, one-character fix. With `>=` the `phase == cycle` case breaks at `s = 3`, `frac = 1`, `oscstate = 0`, `state = 0`, which is correct.
- **`wtf_is_sample`:** `init` does `tableid -= 1` and the compensating `tableid++` lives inside the skipped `state == 0` block, so `tableid` sits at -1 for the first partial frame. It does *not* read out of bounds — `tableid >= oscdata->wt.n_tables` promotes to unsigned (`n_tables` is `unsigned int`), so `process_block` resets it to 0 first. But the frame bookkeeping for a one-shot sample is off for its first partial frame; I confirmed `tableid == -1` after `init` and `0` at the first convolute.
- **`storage->rngGen` is seeded from `system_clock::now()`**, so the new test is non-deterministic. That's fine for a property that should hold at every phase, but it means a rare bad phase (like the one above) will surface as an unreproducible CI flake rather than a finding.

## 5. Factoring

The three-different-fixes-for-three-oscillators argument is right; I wouldn't unify them. Two things I'd change:

**The duplicated state machine is the real hazard.** The `g` expressions and the `-= seg * dcu` drops in `init` are a hand-copy of `convolute`'s switch, and `seg[4]` is a hand-copy of the `state & 1` / `(state + 1) & 2` rate logic. Edit `convolute` later and `init` drifts silently, producing a wrong opening level — and the new test only checks *onset timing*, so nothing catches it. Extracting the per-state `(g, segment length)` into a small helper that both call would make `init` genuinely replay `convolute` instead of mirroring it, and would make it obvious that `t` needs sync folded in.

**The integrator seeding is duplicated verbatim** in `ClassicOscillator::process_block` and `SampleAndHoldOscillator::process_block` — nine identical lines including the pan law. That belongs on `AbstractBlitOscillator` as something like `seedIntegrator(const float *startLevel, bool stereo)`. And the three classes now handle `first_run` three ways: Classic clears at the end of `process_block`, S&H at the top, Wavetable sets it and still never reads it. Pick one.

Minor: the block comments in `init` run to paragraphs and partly restate the commit message. The `See issue #7570` pointer plus a line or two would carry it — though this file is unusually discursive already, so it's not out of place here.

## 6. Tests I'd add

The suite currently covers one axis (onset time, default parameters, stereo, no FM). The things that would actually catch a future break:

1. **Retrigger-off output should be a phase shift of retrigger-on output.** Run both for the same parameters, and after the first cycle assert they agree up to a time shift (cross-correlate, or compare cycle-aligned peak/RMS/zero-crossing count). This is the test that fails when someone edits `convolute`'s state machine and forgets `init` — the single highest-value addition.
2. **Bound the opening step, not just the onset.** Assert `|output[0]|` and the largest step in the first N samples are within the oscillator's own steady-state peak for the same parameters. That's the property the PR argues for in prose.
3. **Convolute-count regression for the wavetable** — first block within ~2x steady state. Catches finding 1 and anything like it.
4. **Parameter sweeps**: Width 1/Width 2/Shape/Sub Mix for Classic, Correlation/Width for S&H, Morph/Skew/Formant/Saturate for Wavetable. `init_default_values()` only exercises one corner.
5. **Sync > 0** (currently fails, per finding 2), **absolute unison detune**, **unison 2–16**, and **nonzero init drift** — the per-voice seeding and pan sum are entirely untested with more than one voice.
6. **`stereo = false`** — the mono branch of the `first_run` sum (`sL += v`, no pan) never runs today. And check `outputR` in the stereo case.
7. **FM mode** (`process_block(..., FM=true, depth)`) — different delay path, and the seeding lands in the same first block.
8. **`wtf_is_sample`, a 1-frame wavetable, and the smallest table Surge ships.**
9. Hoist `createSurge` out of the loops — the test now builds 48 synths where 2 (wavetable / non-wavetable) would do.

By ear: a dense chord with 16-voice unison on all three oscillators using a large factory wavetable at high notes (that's where finding 1 shows); changing oscillator type while holding a note; extreme Width/Shape with the fastest amp attack and full sustain; MIDI 0–12 with retrigger off, where a quarter cycle is ~30 ms. And a run of the sanitizer build over `[dsp]` is cheap insurance given the new table indexing.

mkruselj:
"unusually discursive already". Haha! 😁

lluisestape-upc:
Thanks, this is a useful read. Several of these I would not have caught. I've pushed a follow-up commit that fixes the two clear ones; the rest I've left for us to agree scope on rather than pile into this branch.

## Fixed in the follow-up

**1. Wavetable mipmap skip.** Confirmed and fixed. I gated the mipmap recompute on `state == 0 || first_run` and clear `first_run` at the end of `process_block`. I kept the state advance and the `wtf_is_sample` bookkeeping on `state == 0` only, so `first_run` recomputes the mipmap but does not re-run the sample `tableid` logic. My own before/after, MIDI 96, 20 note-ons, first block vs steady:

| | first block | steady |
|---|---|---|
| retrigger off, before | ~1137 | 48.6 |
| retrigger off, after | 49 | 48.6 |
| retrigger on | 49 | 48.6 |

Back to parity.

**4. `>` to `>=` in the Classic replay loop.** Fixed, with your reasoning in a comment. Verified that `phase == cycle` now breaks at `s = 3`, `frac = 1`, `oscstate = 0`, `state = 0`.

Full test runner still green on both platforms after these.

## Left for us to decide

**2. Sync not reached.** Agreed it's an incomplete case rather than a regression. I'd rather fold the synced and absolute `t` into this PR than leave a known gap, but the `absolute` detune branch is exactly where I'd want your eyes before I touch it, so tell me whether you'd prefer it here or as a separate change.

**`wtf_is_sample` on the retrigger-off path.** Beyond the `tableid == -1` first-frame bookkeeping you flagged, there's a design question underneath it: should a one-shot sample start from a random offset at all when retrigger is off? Randomising the start phase makes sense for the periodic oscillators, but a sample seems like it should start at the top. If you agree, the cleanest fix is to keep samples on the old `oscstate[i] = 0` path, which also sidesteps the `tableid` issue.

**5 and 6. Factoring and tests.** I agree with all of it. The duplicated state machine between `init` and `convolute` is the real hazard, and the "retrigger-off output is a phase shift of retrigger-on output" test is the one that would actually catch a future break. Both are more than a touch-up, so I'd rather do them as a focused follow-up PR once the behaviour here is settled, unless you'd prefer them in-branch.

**3. Clicks.** Thanks for the structural explanation, it matches what I measured at the synth output. I'll ear-check the two cases you flagged, extreme Width/Shape at a fast attack and the `switch_toggled` mid-note re-init, and report back.

mkruselj:
Regarding sample with retrigger off, current behavior should be retained. Although personally I think it makes sense to randomize the startpoint when retrigger is off, it should be within the first frame only.

baconpaul:
Great. The factoring and tests - yeah we can do this as a two-commit. If you want it to be in one PR and we rebase so we retain bisect ability that's also fine. Agree with @mkruselj on wtf_is_sample mostly because 'the definition is weird so lets not change it'

I need to re-read the absolute branch to determine the sync thing. Let me do that now

baconpaul:
Oh gosh I remember that absolute issue from way way back.

I am almost sure that the analysis I put in at that point (1) worked but (2) isn't fully considered. I should revisit the like 415 classic comment.

I propose for this PR we fix sync-absent-absolute and leave sync-with-absolute path unchanged.

lluisestape-upc:
Pushed the sync fix, and I'll take your two-commit option as a separate follow-up PR.

## Sync

Done the way you proposed: `init` now mirrors `convolute`'s non-absolute `t`, sync included, in both Classic and S&H. The absolute branch is untouched, and there's a comment saying why, so whoever revisits it later finds the reasoning in one place instead of two.

Classic, note 72, 30 note-ons, convolutions in the first block against the steady-state count:

| sync | before | after |
|---|---|---|
| 0 | 0.87 / 0.75 | 0.83 / 0.75 |
| 12 | 1.27 / 1.73 | 2.67 / 1.70 |
| 36 | 3.60 / 6.26 | 7.17 / 6.26 |
| 60 | 11.67 / 24.46 | 25.27 / 24.48 |

At sync 60 the first block was running at 48% of the correct rate. It's at parity now. Full test runner still green, 108 cases.

One thing I noticed while in there, not fixed and not new to this PR: S&H `init` sets `syncstate[i] = oscstate[i]`, so a retrigger-off voice starts mid-segment but with its sync cycle boundary landing exactly at that point. It's self-consistent and matches what a voice looks like right after a sync reset, so it isn't wrong, but the sync cycle isn't randomised the way the segment is. Mentioning it in case you'd rather it were.

## wtf_is_sample

Agreed, dropping it. If the definition is odd, that's a separate conversation and not one this PR should start.

## Factoring and tests

I'll do them as a follow-up PR rather than in-branch. Three reasons: this branch already has four rounds of changes on top of a settled narrative, the factoring touches `convolute` and deserves review on its own rather than at the end of a long thread, and the test list you sketched (parameter sweeps, unison, FM) is real work rather than a touch-up. Losing the bisect granularity here costs less than burying a hot-path refactor in a fix PR.

I'll open it once this one lands, so the two commits sit on top of the behaviour rather than moving under it.

## Clicks

Still owe you the ear check on extreme Width/Shape at a fast attack and on the `switch_toggled` mid-note re-init. Reporting back on those separately.

lluisestape-upc:
Follow-up on the two click cases you flagged. 

## Extreme Width and Shape

I swept Shape, Width 1, Width 2 and Sub over 225 combinations at note 48, 40 note-ons each, measuring the sample-0 jump against the oscillator's own largest step while running. Worst case is Shape 0.5, Width 1 0.1, Width 2 0.99, Sub 0: a jump of 1.97 against a running step of 0.535, so 3.7x. Default parameters give 2.8x, so the extremes make it worse but there is no corner where it blows up.

At the synth output with those same parameters, amp attack and decay at minimum and sustain full so nothing masks it, worst of 16 note-ons:

| | largest step in the first 200 samples | largest step while running |
|---|---|---|
| retrigger on | 0.27 | 0.31 |
| retrigger off | 0.42 | 0.61 |

So the onset step still lands under the waveform's own edges even at the worst settings. Same conclusion as with the default patch.

## switch_toggled

This one I think you are right to flag. `spawn_osc` placement-news a fresh object, so `osc_out` starts at zero and `init` clears `oscbuffer`. Before this change a mid-note type switch dropped the oscillator to silence and then faded in through convolute; now it steps straight to the seeded level. Both are discontinuous against the outgoing oscillator's last sample, but the magnitude goes from `|old_level|` to `|old_level - new_level|`, which is up to twice as large. It only applies with retrigger off, and it is an edit-time event rather than something that happens while playing, so I have not tried to fix it here. Say the word if you want it addressed and I will look at seeding from the outgoing level instead of from silence.

Ear check on both still to come, I will report back separately on that.

lluisestape-upc:
Checking in on this one. Everything from the review is addressed as far as I know: mipmap recompute, the `>=` boundary, and sync for the non-absolute path, with the absolute branch left alone per your call. Checks are green and it still merges clean, though main has moved 42 commits since my base, so say the word if you would rather I rebase.

No rush from my side. Mostly want to know whether anything is still open that I should be working on, or whether it is fine as it stands and waiting on a slot.

The factoring and the test work are still queued to go on top once this lands.

mkruselj:
I think actually that you could go ahead with factoring and added tests in separate commits, we can rebase-merge so that it can all be bisected down the road.

lluisestape-upc:
Works for me, pushed both on top as separate commits.

**Factoring.** The four segment cycle lived in three places: the switch in `convolute`, the replay loop in `init`, and the `rate` computation at the end of `convolute` that derives the same four durations a third way from `state & 1` and `(state + 1) & 2`. They are now one `classicSegment` helper that both callers use. Two things fell out of that: the rate block reduces to a single multiply, and the four durations sum to `2 * t` identically once the `pw` and `pw2` terms cancel, so `init` no longer walks the cycle to add them up. It also removes a leftover numerical difference, `init` was computing the level drop as `seg * dc_uni` with `dc_uni` going through `mech::rcp` while `convolute` used the exact product; they agreed to about 4e-4 and now agree exactly.

**Tests.** Four cases, and the part worth flagging is that three of them catch different things and none subsumes another. I only know that because I broke the oscillator on purpose to check each test actually fails, and the first two designs did not.

- The steady-state correlation test, the "retrigger off is a phase shift of retrigger on" one, passed happily on a voice seeded into the wrong segment and on a voice seeded with a level 50% too large. Both sat at 0.999. That is structural rather than a threshold problem: state 0 sets the level absolutely and `rate` is recomputed every `convolute`, so seeding errors wash out within a cycle or two, and a correlation searching over lag is blind to a phase offset anyway. It still earns its place, it just proves the voice settles into the right waveform, not that it starts correctly.
- An onset step bound catches a wrongly seeded level: 1.11 clean, 1.67 with the level 50% too large.
- A first-block rate check catches a wrongly seeded period, which is the sync bug from the previous round: 0.937 clean against 0.365 with the period seeded without sync. It asserts on the mean over 100 note ons rather than the worst one, because with sync active a voice gets reset often enough that individual trials recover inside the first block. The broken version still produced single trials at 0.997 while averaging 0.365, so a worst-case assertion would have flaked instead of failing.

One gap I did not close and did not want to paper over: seeding the voice into the wrong segment is not reliably caught by any of the three. It leaves the correlation at 0.999, and it moves the onset step ratio only by changing the steady-state waveform that ratio is measured against, which is not a detection worth asserting on. It is noted in a comment above the tests.

Full runner green at 112 cases. Still merges clean against main, so no rebase needed from my side unless you want one before the rebase-merge.

PR #8544, opened 2026-09-03
Mono mode: don't treat a pedal-sustained note as a trill anchor

### LINKED ISSUE #6620: Mono mode note-off retrigger misbehavior
**Observed behavior**
1. Press C
2. Press sustain pedal and hold it
3. Release C -> sustain pedal keeps playing it
4. Press any other key staccato while still holding sustain pedal -> it goes back to C after releasing the key you've just played staccato

**Expected behavior**
Surge should only do note-off retrigger if there's another key held down, acting as an anchor for trills. Sustain pedal holding a note is _not_ to be considered an anchor for trills.

mkruselj:
In `RELEASE_IF_OTHERS_HELD` pedal mode, releasing a key whose note the sustain pedal was holding left its keystate set, so that note still counted as a physically-held anchor for note-off retriggering. Playing and releasing another note then snapped the voice back to the pedal-held note instead of staying on the note just played.

Clear keystate when the note is moved to the hold buffer in that mode, so only actually-held keys anchor a trill. `HOLD_ALL_NOTES` is unchanged.

Adds coverage in UnitTestsMIDI "Sustain Pedal And Mono".

Fixes #6620!

### DIFF
--- a/src/common/SurgeSynthesizer.cpp
+++ b/src/common/SurgeSynthesizer.cpp
@@ -1288,6 +1288,7 @@ void SurgeSynthesizer::releaseNote(char channel, char key, char velocity, int32_
     {
         bool sceneNoHold = noHold;
         auto pm = storage.getPatch().scene[sc].polymode.val.i;
+
         if (!sceneNoHold && !mpeEnabled && storage.monoPedalMode == RELEASE_IF_OTHERS_HELD &&
             (pm == pm_mono || pm == pm_mono_fp || pm == pm_mono_st || pm == pm_mono_st_fp))
         {
@@ -1307,10 +1308,20 @@ void SurgeSynthesizer::releaseNote(char channel, char key, char velocity, int32_
         }
 
         if (sceneNoHold)
+        {
             releaseNotePostHoldCheck(sc, channel, key, velocity, host_noteid);
+        }
         else
-            holdbuffer[sc].push_back(HoldBufferItem{
-                channel, key, channel, key, host_noteid}); // hold pedal is down, add to buffer
+        {
+            // hold pedal is down, add to buffer
+            holdbuffer[sc].push_back(HoldBufferItem{channel, key, channel, key, host_noteid});
+
+            // #6620 on reasoning why
+            if (storage.monoPedalMode == RELEASE_IF_OTHERS_HELD)
+            {
+                channelState[channel].keyState[key].keystate = 0;
+            }
+        }
     }
 }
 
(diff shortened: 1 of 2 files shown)


baconpaul:
Had to resolve a conflict. Happy to merge this once CI on the unconflicted branch passes

baconpaul:
Right so the key state zeroing broke sostenuto

The right check was keyIsDown and so i modified for that

Retained the tests

baconpaul:
OK now this is merged and does sostenuto again I think it is right. But please pull and try before a merge. Feel free to merge this if it does what you expect still

mkruselj:
OK I just checked it. Sostenuto should NOT be anchoring the trill, just like sustain pedal isn't, and that's what's happening with current state. Trilling should only ever be allowed between physically held keys. I _think_.

EDIT: Using the brainstorming superpower with Claude, it suggests to keep this behavior.

___

## The choice isn't "anchor vs. different anchor" — it's "anchor vs. no-op"

Mono has one voice, so sostenuto literally cannot do its poly job (keep these notes ringing while you play freely above). The only surviving meaning in mono is "the line falls home to the captured note," which is exactly what the current !sostenutoHolds exemption produces.

And the alternative collapses. If sostenuto participated in `RELEASE_IF_OTHERS_HELD`, releasing the captured key with any other key down routes to `releaseNotePostHoldCheck()`, which clears keystate (:1342). The mono fallback scan at :1416 reads keystate — so the captured note is gone from the fallback pool too. Sostenuto would then only "hold" a note in the one situation where you don't need a pedal at all: nothing else is playing. That's not a milder behavior, it's zero behavior.

## Why #6620's reasoning doesn't transfer

`RELEASE_IF_OTHERS_HELD` exists to undo indiscriminate capture: the sustain pedal grabs whatever you happen to release, so a trill turns sticky and the anchor is an accident. Sostenuto's set is chosen — you stomped at a moment you picked, capturing exactly the keys under your fingers. There's nothing accidental to undo. The menu item even scopes itself by name: "Sustain Pedal Allows Note Off Retrigger." Applying it to sostenuto would be reading past its own label.

Musically it lands as a pedal-point/drone: hold C, stomp, and the mono line always falls back to C until you lift. I think that's the right translation, and it's a genuinely nice thing to have.

## The one sharp edge worth naming

The anchor doesn't get priority — it just joins the ordinary highest/lowest/latest scan. So with `ALWAYS_HIGHEST` and a captured note above your playing range, every release snaps up to the captured note and the keyboard below it becomes unreachable until you lift the pedal. That's real, but it's identical to what sustain already does in `HOLD_ALL_NOTES` (the default), so it's a pre-existing property of mono + held notes, not something sostenuto introduces. I wouldn't special-case it.

## Recommendation

**Keep it as is.** The behavior is the only non-degenerate reading of sostenuto in mono, it's already covered by the test at UnitTestsMIDI.cpp:2161, and the code comments explain the divergence at the point where someone would question it.

The only gap I see is discoverability — nothing tells a user that the mono pedal preference is sustain-only.

___

Well, ok then. Merging!

PR #8547, opened 2026-09-04
Fix stuck note when a context menu eats VKB's key up

### LINKED ISSUE #5480: VKB sticks if you open a menu
Press a key
Open a menu
Voila, no way to release other than down/up again
Tricky....

mkruselj:
The VKB has `wantsKeyboardFocus()` false so that it receives QWERTY input no
matter what else is focused, with key events fed to it through
SurgeSynthEditor's `KeyListener`. But that also means JUCE never calls
`focusLost()` on it, and `focusLost()` is where `MidiKeyboardComponent`
releases the keys it believes are held. Its `keysPressed` bitset is
otherwise only reconciled from `keyStateChanged()`, which is event driven.

So if a context menu enters modal state while a key is held, the key up
goes to the menu instead. The bitset stays marked down and no further event
ever arrives to clear it, so the note hangs until you press and release
that key again.

Poll for it in the editor `idle()` instead. `keyStateChanged()` rescans
every bound key against the real OS key state, so calling it at a moment
when none of them is physically down can only ever emit note offs, never
note ons. That makes the VKB self-healing regardless of who swallowed the
key up, which also covers Alt-tabbing away and hiding the VKB mid-note.

Fixes #5480

### DIFF
--- a/src/surge-xt/SurgeSynthEditor.cpp
+++ b/src/surge-xt/SurgeSynthEditor.cpp
@@ -358,6 +358,11 @@ void SurgeSynthEditor::setVKBLayout(const std::string layout)
  * none of them is physically down it can only ever emit note offs, never note ons. Under that
  * guard it is a no-op when nothing is stale and a release when something is, which makes the
  * VKB self healing regardless of who ate the key up.
+ *
+ * Note that the guard has to ask isKeyCurrentlyDown rather than KeyPress::isCurrentlyDown.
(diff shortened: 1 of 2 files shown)


baconpaul:
hmmm i get it but i'm not sure i love this solution. I want to think about it a bit more.

A better solution would be 'call focusLost on the keyboard whenever any menu shows' but that's probably tricky too.

gonna pause on this one on friday night. some thinking

PR #8550, opened 2026-09-07
Allow wavetable oneshot, looping and re-slicing to be changed at runtime

mkruselj:
The .wt format carries metadata saying whether a wavetable should play back as a wavetable or as a sample, and how it is cut into frames, but none of it could be changed once loaded - and for a WAV or a patch wavetable it could not be expressed at all. This adds a runtime path for all of it, hung off the oscillator display's context menu.

`Wavetable::Reslice()` rebuilds a table in place from its own contents: the top mipmap level is flattened back into the sample stream it was built from, then handed to `BuildWT()` with a new frame size, frame count and flag set. That works for every wavetable regardless of where it came from - file, patch blob or Lua script - and re-slicing a script's output is a legitimate effect in its own right, so the menu never regenerates. Script resolution stays where it already lives, in the script editor.

Frame size is a power of two between 16 and 4096, since both oscillators mask with size - 1. Frame count is a free integer: when the requested count will not fit at the current frame size the size is halved until it does, and samples past the last whole frame are truncated, so growing the count past what the table holds works the way you would expect.

Sample mode is handled by `SourceFrameCount`, which sees through the three silent frames `BuildWT()` appends - and through however many of those survived a patch round trip, since load_patch rebuilds with `AppendSilence` false and they arrive as ordinary data. Without that, every toggle would grow the table.

`wtf_loop_sample` was declared but never read. It now pins `sampleloop` to the value at which it stops counting down, overriding rather than extending the long-standing behaviour where the unison voice count doubles as a play count for samples. Existing patches do not set the flag, so they are unaffected.

A re-sliced table no longer matches the file it came from, so it detaches: `current_id` is cleared, which makes undo capture the full wavetable rather than reloading the original, while `wavetable_display_name` is deliberately left alone so the oscillator keeps showing the name it was loaded under instead of falling back to `"(Patch Wavetable)"`. The new `wtf_user_modified` bit records this acrossa patch round trip, so `load_patch()` does not re-attach the table to the `wt_list` entry its name matches. The bit is masked out of .wt export.

The rebuild is queued and performed in `perform_queued_wtloads()`, on the audio thread at a block boundary, the same way a queued wavetable load is.

Unison on samples is deliberately left out: `tableid`, `last_tableid` and `sampleloop` are scalars shared across unison voices, so letting more than one voice run would need them promoted to per-voice arrays. That is a separate change coming in a separate PR.

### DIFF
--- a/src/common/dsp/Wavetable.h
+++ b/src/common/dsp/Wavetable.h
@@ -22,6 +22,7 @@
 #ifndef SURGE_SRC_COMMON_DSP_WAVETABLE_H
 #define SURGE_SRC_COMMON_DSP_WAVETABLE_H
 #include <string>
+#include <vector>
 #include <StringOps.h>
 const int max_wtable_size = 4096;
 const int max_subtables = 512;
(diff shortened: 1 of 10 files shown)


src/common/dsp/Wavetable.h
```diff
@@ -50,9 +51,36 @@ class Wavetable
 
     void allocPointers(size_t newSize);
 
+    // Smallest frame size the runtime re-slicer will produce. BuildWT copes with anything
+    // down to 1, but below this a "frame" stops being a waveform in any useful sense, and
+    // the auto-shrink in Reslice would happily grind all the way to the bottom.
+    static constexpr int min_reslice_size = 16;
```
baconpaul, line 57: I don’t quite get this 16 and if you use a shorter wavegable dont quite see where the error gets reported to the user. I’m imagining a user taking a 20 frame wavetables and wanting to slice it as a 10 (to say get 2 cycles per cycle).

src/surge-xt/gui/UndoManager.cpp
```diff
@@ -444,14 +456,17 @@ struct UndoManagerImpl
 
     void doCleanup()
     {
-        while (undoStackMem > maxUndoStackMem)
+        // Now that a wavetable record is accounted for honestly, a single one can be a
+        // sizeable fraction of the budget on its own - so these have to stop at empty
+        // rather than trusting the budget to be reachable by evicting.
+        while (undoStackMem > maxUndoStackMem && !undoStack.empty())
```
baconpaul, line 462: Hmm if this is really true that a single item can be more than the budget now we probably want to increase the max undo stack mem some

baconpaul (commented):
Looks generally good. One question one comment

src/common/dsp/Wavetable.h
baconpaul: Oh right this is samples not frames. Ignore this comment.

src/surge-xt/gui/UndoManager.cpp
mkruselj: Will be handled in upcoming PR.

src/common/dsp/Wavetable.h
mkruselj: 16 (samples per frame) is just ultra cruddy! But sure we could take it up to 128 samples at minimum, or 64 (which is what PPG did).

PR #8554, opened 2026-09-07
Add periodic timestamped patch backups

### LINKED ISSUE #7217: Periodically saved timestamped backups
It would be a pretty cool option to add. Enable saving timestamped backups in User Data/Backups folder (so that they don't clutter up the patch browser), by default save every 5 minutes, but also allow changing the interval (1 minute should be the lowest value IMO).

Also we shouldn't save a backup if transport is running.

mkruselj:
Surge could lose an unsaved patch to a crash or a host going down with no recovery path at all. This adds an opt-in periodic backup that streams the current patch to `User Data/Backups` whenever it is dirty.

The backup writer reuses `savePatchToPath()` with `refreshPatchList = false`, so it writes a normal fxp without renumbering `patchid`, retargeting the current category, or clearing `isDirty`. Backups are laid out as:

```
Backups/YYYY-MM-DD/<Category>/<PatchName> (HH-MM-SS).fxp
```

with patch name and category scrubbed of characters the filesystem rejects. A category can itself be nested, as third party patches are, and those levels are mirrored as real directories rather than flattened into one name. The timestamp also goes into the streamed patch name, so a restored backup says what it is and when it was taken rather than looking like the patch it came from.

The interval timer runs off the editor `idle()`, which means backups only happen with the UI open. Two further gates keep the folder useful: the patch has to be dirty, and the host transport has to be stopped. `isDirty` is sticky once set, so the serialized patch is also hashed - before the timestamp goes into the name, or every serialization would differ from the last - and an unchanged patch does not produce a second identical file. Neither the dirty nor the transport gate re-arms the timer, so a backup happens promptly once the condition clears rather than waiting out another full interval.

The timer, the hash and the failure latch live on the synth rather than on the editor, because the editor is constructed afresh every time the plugin window is opened. Editor-side state would restart the interval, forget what was last backed up, and re-report a failed backup on every open and close - so someone who reopens the window regularly would never get a backup at all. Seeding the timer at synth construction is also what stops a window opened right after instantiation from writing one immediately.

Closing the window stops the interval, so the editor destructor takes a backup on the way out whenever the patch is dirty, which covers tweaking for a few minutes and then closing the window. The hash check keeps that from producing duplicates when nothing has actually changed.

Transport state needed a new `SurgeSynthesizer::transportRunning` atomic rather than `time_data.isPlaying`, because the standalone deliberately pins `isPlaying = true` so formula modulators keep running. `transportRunning` is only ever set from a real host playhead, so backups are never transport gated in standalone.

The backups folder is a sibling of `User Data/Patches`, not a child, so `refresh_patchlist()` never walks it and backups stay out of the patch browser categories and out of the patch database. To keep them reachable anyway, the patch browser grows a Backups submenu built live off disk when the menu opens, above Tutorials, plus an Open Backups Folder entry alongside the other folder shortcuts. It walks the tree recursively, so nested categories show up as nested submenus, and it natural sorts each level the way the rest of the patch browser does. The loaded backup is ticked and preselected through `lastLoadedPatch`, and since a file loaded patch leaves `current_category` at -1, the patch list side of the menu now stops ticking whatever stale entry `patchid` still points at - it was claiming a database patch was loaded when a backup was. Loading goes through the usual dirty check and lands on the undo stack like any other patch load.

Backups do have to be loaded as a restore rather than as a preset, though. `loadPatchByPath` defaults forceIsPreset to true, which throws away the name,
category, master volume and FX bypass streamed in the file and keeps whatever name the caller passed - correct for dropping an arbitrary fxp onto a running session, but it left a restored backup sitting in a Drag & Drop category and named after its own file, and resaving it kept that. So the queued file load carries the flag through to both load paths, and the backups menu asks for a restore. Everything else keeps the existing preset behavior, which the new test in UnitTestsIO pins down in both directions.

Both settings live in Main Menu > Patch Settings > Patch Backups. Off by default, with a five minute default interval clamped to one to thirty minutes. Retention is left to the user.

The two trailing commits are pre-existing bugs uncovered along the way and are independently rebase-mergeable.

Closes #7217

### DIFF
--- a/src/common/SurgeStorage.cpp
+++ b/src/common/SurgeStorage.cpp
@@ -663,6 +663,7 @@ void SurgeStorage::initializeUserDataPaths()
     userMidiMappingsPath = userDataPath / "MIDI Mappings";
     userModulatorSettingsPath = userDataPath / "Modulator Presets";
     userSkinsPath = userDataPath / "Skins";
+    userBackupsPath = userDataPath / "Backups";
 }
 
 void SurgeStorage::createUserDirectory()
--- a/src/common/SurgeStorage.h
+++ b/src/common/SurgeStorage.h
@@ -1766,6 +1766,9 @@ class alignas(16) SurgeStorage
     fs::path userWavetablesExportPath;
     fs::path userWavetableScriptsPath;
     fs::path userSkinsPath;
+    // Timestamped periodic patch backups. Deliberately a sibling of userPatchesPath rather than
+    // a child, so refresh_patchlist() never scans it and backups stay out of the patch database.
+    fs::path userBackupsPath;
     fs::path userMidiMappingsPath;
     fs::path extraThirdPartyWavetablesPath; // used by rack
     fs::path extraUserWavetablesPath;       // used by rack
--- a/src/common/SurgeSynthesizer.cpp
+++ b/src/common/SurgeSynthesizer.cpp
@@ -4347,7 +4347,8 @@ void loadPatchInBackgroundThread(SurgeSynthesizer *sy)
         }
         else
         {
-            synth->loadPatchByPath(synth->patchid_file, -1, path_to_string(ppath).c_str(),
+            // The stem, not the whole path: this is the name the patch will be displayed under.
+            synth->loadPatchByPath(synth->patchid_file, -1, path_to_string(ppath.stem()).c_str(),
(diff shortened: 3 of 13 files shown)


mkruselj:
Rebase merge, don't squash!

PR #8555, opened 2026-09-08
Clear unreachable routings when an LFO leaves the formula shape

### LINKED ISSUE #6705: Problems on switch from Formula to other LFO shapes
Assign Formula output >1 to modulate something. Switch to Sine.

If you modulated from 2nd or 3rd Formula output, those assignments are now coming from the Raw or EG-only outputs. Everything is fine.

If you modulated from the 4th output or higher, the mod assignment sticks around in the UI even though it won't do anything. When right clicking the modded param it displays as "LFO 3.8" for example. 

What's more. If you were viewing the 8th formula output before switching, the display will clamp down to LFO output three (EG output), but if you arm and assign from there, the assignment will actually come from the first LFO output. See picture. That's a bit borked. 

Slightly Borked

mkruselj:
A formula LFO has `max_formula_outputs` outputs; every other shape has three. Switching away from formula left any routing above the third output in the patch, where it did nothing - a non-formula LFO never writes those slots of output_multi - but still showed up in menus as "LFO 3.8" and survived save and load.

Worse, the output index was tracked in two places which the shape change only half updated. `lfoShapeChanged` reset the editor's `modsource_index` to
zero, but `setupAlternates()` then ran and `ModulationSourceButton::setModList()` restored the button's index from the DAW extra state and clamped it to the new maximum. So the button read "LFO 3" while assignment, the mod rings and the infowindows all used output one.

Fix both. `setLFOModulationIndex()` is now the single place which moves an LFO's output index, keeping the editor state, the button and the DAW extra state in agreement. On leaving formula we then clear the routings at or above the new maximum, the way an oscillator or FX type change clears the modulations it invalidates.

Those routings come back in a single undo press: `UndoLFOShape` records them alongside the shape, mirroring how `UndoOscillator` and `UndoFX` record what they blow away, and undo and redo both restore the recorded set wholesale rather than trying to work out which routings to put back.

Second commit fixes a bug found along the way, this PR should be rebase-merged.

Fixes #6705

### DIFF
--- a/src/common/SurgeSynthesizer.cpp
+++ b/src/common/SurgeSynthesizer.cpp
@@ -3974,6 +3974,44 @@ int SurgeSynthesizer::getMaxModulationIndex(int scene, modsources modsource) con
     return 1;
 }
 
+std::vector<std::pair<int, int>> SurgeSynthesizer::getModulationsFromSource(int modsourceScene,
+                                                                            modsources modsource,
+                                                                            int minIndex) const
+{
+    std::vector<std::pair<int, int>> res;
+
+    auto collect = [&](const std::vector<ModulationRouting> &ml, int idBase, int listScene) {
+        for (const auto &mr : ml)
+        {
+            if (mr.source_id != modsource || mr.source_index < minIndex)
+            {
+                continue;
+            }
+
+            // A global routing carries its originating scene explicitly, since the target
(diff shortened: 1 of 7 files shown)


mkruselj:
Rebase merge, don't squash!

PR #8556, opened 2026-09-08
Add a keytracking trigger mode to the oscilloscope

### LINKED ISSUE #6962: Keytracking mode in Oscilloscope
User Benis67 on Discord says: "Why not add a sync option controlled by the lowest or highest note? Great for analyzing monophonic sounds".

...which is a good idea! So I'm noting it here so we remember it (in 1.3 for now, but may move later).

mkruselj:
The waveform scope could only be made to stand still by hand: pick Internal Trigger and dial the trigger frequency until the trace stopped moving. For a monophonic sound that frequency is already known - it is the note being held.

Add a Keytrack trigger mode that derives the trigger rate from the played note, and with it the time base, so a chosen number of note cycles always fills the display no matter which note is played. Note Priority selects which held key is followed (lowest, highest or latest, unioned across both scenes since the scope taps the master output), and Cycles sets how many periods are shown, which doubles as the zoom control. Time Scaling has nothing to say in this mode and is greyed out rather than hidden, so the panel does not reshuffle. The last key is held onto after release, so lifting your hands leaves the trace in place instead of collapsing it just as the release tail gets interesting.

The trigger oscillator can only fire on sample boundaries, which at high notes put the trace several pixels off from one sweep to the next. The overshoot left in the accumulator is the sub-sample error, so pass it to the interpolated draw path and start that fraction of a sample early. This steadies the existing
Internal Trigger mode as well.

Note tracking follows the key only, not the sounding pitch, so glide, pitch bend and vibrato will still slide the trace.

The trigger mode switch is now horizontal, lined up under the Sync Redraw button, with the per-mode controls laid out horizontally alongside Time Scaling. That needs a little more room, so the overlay's minimum width goes from 500 to 560.

Closes #6962

Trailing commits are bugfixes found along the way that fall into the same blast area. This PR should be rebase-merged as such.

### DIFF
--- a/src/common/SurgePatch.cpp
+++ b/src/common/SurgePatch.cpp
@@ -3671,6 +3671,7 @@ void SurgePatch::load_xml(const void *data, int datasize, bool is_preset)
                         node->QueryIntAttribute("mode", &oos->mode);
                         node->QueryFloatAttribute("trigger_speed", &oos->trigger_speed);
                         node->QueryFloatAttribute("trigger_level", &oos->trigger_level);
+                        node->QueryFloatAttribute("trigger_limit", &oos->trigger_limit);
                         node->QueryFloatAttribute("time_window", &oos->time_window);
                         node->QueryFloatAttribute("amp_window", &oos->amp_window);
                         node->QueryIntAttribute("trigger_type", &oos->trigger_type);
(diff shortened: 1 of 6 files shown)


mkruselj:
Rebase merge, don't squash!

Issue #2285, opened 2020-07-04
Scene B S-LFO 1-6 can still modulate Scene A/B effect parameters even if it's not suppose to work.


K0rrid0r:
**Describe the bug**
Scene B S-LFO 1-6 can still modulate Scene A/B effect parameters even if it's not suppose to work according to the manual.

**To Reproduce**
![161](https://user-images.githubusercontent.com/56724286/86521831-fe875100-be55-11ea-86e3-1c2761e3b1bf.gif)
These LFO links wont do anything, still it is possible to do them. Very confusing.

**Expected behavior**
https://surge-synthesizer.github.io/manual/#modulation-routing-in-depth

VST3 hash 75f7bc1

baconpaul:
Oh very good find.

K0rrid0r:
It would be good if we later on could make it work this way though, this seems like a strange limitation. But it is probably deeper then that.

baconpaul:
Yeah
or maybe we don't even fix depending on how deep it is.
Still need to look.

mkruselj:
So wait only scene A SLFOs can modulate FX? Hmmm. Bit of an arbitrary limitation, no?

baconpaul:
Well I think @K0rrid0r 's request is: make the manual, the synth, and the UI do the same thing for 1.7.0; and then we can think about what it is

I think A SLFO should be able to modulate AFX and MasterFX; BSLFO for BFX and MasterFX, myself.

I don't know what the synth actually does

I may ship 1.7.0 with this break in place though, since it feels like it could be a bigger fix

baconpaul:
In my test I was able to modulate A and B fx with both A and B SLFO.
So I'm not sure what's up here?
Gonna push out of 1.7.0 though until I understand it better. If we do need a change from looking at the code it would be a bigger one. But not entirely sure we do.

K0rrid0r:
Let me double check!

baconpaul:
No you are right
The reason is

SurgeSynthesizer::setModulation

uses the parameter->scene as the scene param for the source list

For FX there is no scene so it uses the modlist_global

but the modlist is keyed uniquely on mod source, not on mod source and scene

so mod source 23 gets collapsed in the global routings

and blammo. there's only one LFO.

What we need to do, alas, is key the mod source routings including the source scene

which is painful.

K0rrid0r:
Great find and explanation of the inner working of this!

baconpaul:
[...]

PR #8557, opened 2026-09-08
Show macro modulations of FX parameters in the macro menu

### LINKED ISSUE #8053: Context menu of a macro does not show macro-to-FX modulations
**Bug Description:**
Context menu of a macro does not show macro-to-FX modulations.

**Surge XT Version**
Version: Surge XT 1.3.4.f7b97c6
Build Info: 2024-08-11 @ 16:02:24 on 'fv-az292-728/pipeline' with 'MSVC-19.40.33813.0' using JUCE 7.0.12
System Info: Windows 64-bit Standalone on Intel(R) Core(TM) i7-7700HQ CPU @ 2.80GHz, 32 GB RAM
Sample Rate: 48000 Hz
Processing Block: 32 samples

**Reproduction Steps:**
1. Create an FX. Ensure it has a slider you can modulate.
2. Create a modulation from Macro 1 (cc 41 by default) to a slider the FX.
3. Right-click the Macro to view its modulations.

I expect the modulation to appear in the list
However, the modulation is not displayed.

Screenshot shows an existing patch. In the **Modulation List** to the right, all modulations from the macro display as expected: four FX parameters, **Scene B Waveshaper Drive**, and **LFO Amplitude**. However, these are not displayed on the context menu for the macro.

Image

**Additional Information:**
Not critical; the bug presents an inconsistency in the UI, and the workaround is less convenient than full functionality.

mkruselj:
Fixes #8053.

### What was wrong

`ModulationRouting::source_scene` was added in #4960 to disambiguate a routing onto a global (FX) target when the modulator exists once per scene — LFOs and envelopes. Macros and MIDI controllers are a *single object aliased into both scenes* (`SurgeSynthesizer.cpp:196-200`), so the field carries no information for them.

Every global lookup compares it anyway (`getModulationIndicesBetween`, `getModRouting`, `getModDepth`, `clearModulation`, `setModDepth01`), and the modulator context menu always asks for scene A because `isModulatorDistinctPerScene()` is false for a macro. So a macro→FX routing tagged scene B is invisible in that menu — while still sounding, and still listed in the Modulation List, which reads the routing vectors directly.

Two ways patches end up in that state:

- **Streams written before #4960** tagged these with whichever scene happened to be active. j5v's patch on the issue has exactly four such routings from Macro 7 onto `fx8_p1` / `fx15_p1` / `fx15_p4` / `fx15_p7`, all `source_scene="1"`, which is precisely what his screenshot shows missing. **Seven factory patches** are affected: *Pipe Organ Flutes & Reeds*, *Solo Violin*, *String Cello*, *String Contrabass*, *Violin Section*, *Massive Pad 1*, *Musical Pulses*.
- **Copy Scene A → Paste Scene B in a current build**, because the paste stamped the target scene onto every global routing it carried across.

### The fix

Promote `isModulatorDistinctPerScene` to a free function beside the other modulator predicates in `ModulationSource.h` (it was already a pure function of the enum) and use it to force `source_scene` to zero for shared modulators, both when a routing is unstreamed and when one is pasted. Unstreaming normalizes older patches on load, so the field is only ever non-zero for modulators it actually describes, and no version gate is needed — a correctly streamed patch already has zero there.

Note this does rewrite patch data on load, so an affected patch changes on the user's next save. That seemed clearly preferable to teaching six read sites to special-case the field and leaving the bad data to round-trip.

### Second commit

`Carry macro to FX modulations when copying a scene` is the copy side of the same confusion and is separable — it can be dropped without affecting the fix above.

Scene copy picked the global routings belonging to a scene by comparing `source_scene`, so a macro routing onto scene B's insert FX was never collected and copying scene B lost it. It now decides by the modulator for a per-scene modulator and by the destination FX slot for a shared one.

That commit also fixes an uninitialized read: a routing onto a *send* or *global* FX parameter took the `cg_FX` branch but matched no insert slot, so the clipboard entry was pushed with an unset `destination_id`, which the scene paste then wrote into `modulation_global` with no validity check.

### Testing

New `TEST_CASE("Global Modulation Routings Are Not Scene Tagged For Shared Modulators", "[io][mod]")` with four sections: the legacy stream normalizing on load, a scene LFO correctly *keeping* its scene tag (so #2285 still holds), the paste path, and the copy path. Each bug section was verified to fail with the fix reverted.

### DIFF
--- a/src/common/ModulationSource.h
+++ b/src/common/ModulationSource.h
@@ -221,6 +221,26 @@ inline bool canModulateMonophonicTarget(modsources ms)
     return isScenelevel(ms) || ms == ms_aftertouch;
 }
 
+/*
+ * Does this modulator have a separate instance per scene? Voice and scene LFOs and the
+ * envelopes do; macros, MIDI controllers and the like are single objects shared by both
+ * scenes. Only modulators for which this is true need ModulationRouting::source_scene to
+ * disambiguate a routing onto a global (FX) target. See #2285 and #8053.
+ */
+inline bool isModulatorDistinctPerScene(modsources ms)
+{
+    if (ms >= ms_lfo1 && ms <= ms_slfo6)
+        return true;
+
+    if (ms == ms_ampeg || ms == ms_filtereg)
+        return true;
+
+    if (ms == ms_lowest_key || ms == ms_highest_key || ms == ms_latest_key)
+        return true;
+
+    return false;
+}
+
 inline bool isCustomController(modsources ms) { return (ms >= ms_ctrl1) && (ms <= ms_ctrl8); }
 
 inline bool isEnvelope(modsources ms) { return (ms == ms_ampeg) || (ms == ms_filtereg); }
--- a/src/common/SurgePatch.cpp
+++ b/src/common/SurgePatch.cpp
@@ -2218,6 +2218,15 @@ void SurgePatch::load_xml(const void *data, int datasize, bool is_preset)
                             // Explicitly set scene to A. See #2285
                             t.source_scene = 0;
                         }
+
+                        // source_scene only disambiguates modulators which exist once per
+                        // scene. Streams written before #4960 tagged macros and MIDI
+                        // controllers with whichever scene happened to be active, which then
+                        // hides the routing from that modulator's menu. See #8053.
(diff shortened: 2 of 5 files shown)


baconpaul:
yup

Issue #6959, opened 2023-04-19
Rename user preset appears database lock error


Wandelwolf:
1. Rename a user preset and save
2. The error appears: PatchDB - Loading Favorites SQL Error[5]. Database is locked (This Error appears two times immediataly)
3. After confirm, the preset is finally renamed
4. If i want to perform now the following i got the same error: PatchDB - SQL Error[5]. Database is locked (without the "Loading Favorites"), and this time up to 20+ times immediataly, that i have to confirm now
- save a new preset, delete a preset, favorite a preset
- again after confirming all the errors, the action was performed, and i see the favorite, the preset is deleted or a new is created!
- Refresh patch browser doesn't help
- Restart the surge standalone session or the DAW host helps and all actions can performed without an error as long as no preset rename is performed! Then the error appears again for all the actions above!

* Version: Linux x64 Verison 1.1.2, 1.2.0, and the nightly downloaded on 2023.04.18
* Plugin Type: VST3, Clap, LV2 in Reaper and Standalone

 - OS: openSUSE Tumbleweed 20230413, amd64
 - Host: Reaper 6.73
- Another User has the same issue in Arch Linux too.

baconpaul:
I was able to reproduce this on ubuntu with the standalone.

baconpaul:
Ahh the problem is that rename == save then delete
the save happens and rebuilds the database
so then the delete happens and also rebuilds the database
so we have dueling rebuilds

annoying. need to ponder.

mkruselj:
Can we not save then delete and only then update database?

baconpaul:
Yes but I think that’s not it anyway 

also that’s an annoying code change for an annoying reason

baconpaul:
OK I've found it. It was slightly different than I thought

basically a threading error. A couple of defensive fixes coming in next 30 minutes.

baconpaul:
OK all merged.

PR #8558, opened 2026-09-08
Make PatchDB patch indexing robust against vanishing files

### LINKED ISSUE #6973: PatchDB::parseFXPIntoDB isn't robust if the file is deleted while parsing
This is basically the cause of #6959 but fixing it is quite hard so in #6975 I just added a wait-until-prior-traversal for a more immediate fix. But we should return to this one day.

mkruselj:
`parseFXPIntoDB` runs on the patch database worker thread, and the file it is pointed at can disappear at any moment: a rename, a file sync client, or a second Surge instance can remove it between any two statements. It handled that badly in two distinct ways.

**It could take the process down.** The mtime was read with the throwing `fs::last_write_time` overload, guarded only by an `fs::exists()` call three lines earlier, so a file removed in that window raised `std::filesystem_error`. The worker loop caught only SQL exceptions, so that escaped `loadQueueFunction` and the thread lambda entirely and hit `std::terminate`. This is the mechanism behind #6959.

**It wrote to the database before it had proven it could read the file.** The old order dropped the existing rows for the path and inserted the new `Patches` row, and only then opened the stream, validated the magic and read the XML. All five failure paths after that insert were bare returns, so a file which went away mid-parse left behind a row with no features and a null search string, and a merely transient read failure destroyed a good patch's existing index entry and replaced it with that stub.

So the parse is now strictly read-then-write: all the file IO happens first into local state, and the database is only touched once a complete and valid patch is in hand. Any file-side failure returns having changed nothing, which is the correct outcome since a genuinely deleted file gets reaped by the directory scan instead. Because the search string is known before the insert, the row now goes in complete and the trailing `UPDATE search_over` is gone.

Alongside that: the non-throwing filesystem overloads are used and the `fs::exists()` pre-check is dropped, since it could never close the race and only widened it; the worker loop gets catch-all arms so no future stray exception can kill the application over one bad patch file; and the queue chunk is held by `unique_ptr` so an exception part way through it no longer leaks the items that had not run yet.

This does not make the index transactionally correct against a concurrently changing filesystem. A file deleted just after its transaction commits still leaves a stale row until the next rescan. Closing that properly needs a filesystem watcher or an mtime revalidation pass, which does not seem worth it. What this does is reduce the failure from "crash or corrupt the index" to "one stale row until the next scan".

Fixes #6973

## Also on this branch

**Tell PatchDB when a patch file is deleted instead of rescanning.** Removing a patch file is something the UI knows about as it happens, but the database could only find out via a full `initializePatchDb(true)` rescan that diffs the whole patch tree in order to rediscover the one path we just deleted ourselves. That indirection is why the rename in #6975 had to defer its `fs::remove` until the indexer drained, and bounce the work from the GUI thread to the patch database thread and back.

A new `erasePatchByPath` work item says so directly. Since it is enqueued behind the indexing that `savePatch` has already queued for the new file, the ordering the drain was buying comes for free, so the rename can delete the old file immediately and drop the round trip entirely. Deleting a patch takes the same path, so neither operation forces a full re-index any more. `doAfterCurrentQueueDrained` and `waitForJobsOutstandingComplete` have no callers left, but are kept deliberately as reasonable primitives to have on hand.

The rename's `fs::remove` is also now guarded the way the delete already was, and both operations stop on a filesystem error rather than continuing to update state as though the file had gone away.

**Fix broken patch header tag check.** `parseFXPIntoDB` used `memcpy` where `memcmp` was intended, so the `"sub3"` header check never fired and in fact overwrote the tag rather than comparing it. This is a pre-existing bug that the restructuring merely happens to touch, so it is first on the branch and cherry-picks on its own.

## Testing

Unit tests pass and the standalone builds and links. The rename and delete flows are GUI driven and no test covers the indexer, so the ordering was verified by reading rather than exercised end to end in the running app.

### DIFF
--- a/src/common/PatchDB.cpp
+++ b/src/common/PatchDB.cpp
@@ -652,7 +652,7 @@ CREATE TABLE IF NOT EXISTS Favorites (
         int lock_retries{0};
         while (keepRunning)
         {
-            std::vector<EnQAble *> doThis;
+            std::vector<std::unique_ptr<EnQAble>> doThis;
             {
                 std::unique_lock<std::mutex> lk(qLock);
 
@@ -671,7 +671,10 @@ CREATE TABLE IF NOT EXISTS Favorites (
                     const auto &e = (pathQ.size() < transChunkSize)
                                         ? pathQ.end()
                                         : pathQ.begin() + transChunkSize;
-                    std::copy(b, e, std::back_inserter(doThis));
+                    for (auto it = b; it != e; ++it)
+                    {
+                        doThis.emplace_back(*it);
+                    }
                     pathQ.erase(b, e);
                 }
             }
@@ -688,10 +691,10 @@ CREATE TABLE IF NOT EXISTS Favorites (
                     {
                         SQL::TxnGuard tg(dbh);
 
-                        for (auto *p : doThis)
+                        for (auto &p : doThis)
                         {
                             p->go(*this);
-                            delete p;
+                            p.reset();
                         }
 
                         tg.end();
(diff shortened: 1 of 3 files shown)


mkruselj:
This PR should be rebase-merged!

PR #8559, opened 2026-09-09
Publish our tuning as soon as we become an MTS-ESP source

### LINKED ISSUE #7812: MTS-ESP source inits without tuning editor changes
UserMinusOne reports if you open the tuning editor and make some changes, then enable the instance as MTS-ESP source, it initially sets the frequencies wrong (but applying in the tuning editor afterwards works). Investigate some day!

mkruselj:
`MTS_RegisterMaster()` starts the session off at 12-TET, and Surge only pushed its own table afterwards if no editor was open. With the tuning
editor open the push was left to `send_tuning_update()` on the next UI idle tick, so every client in the session was retuned to 12-TET for a frame, and any client latching tuning at note on kept the wrong pitches until the next note. Becoming a source is always a deliberate main thread act, so publish right there; the deferral only exists because a retune can arrive on the audio thread, which is still handled by the idle path.

Restoring a session had the same problem for a different reason: the `oddsound_mts_active_as_main flag` was acted on while the DAW extra state XML was still being parsed, before the scale and mapping were read. Stage it in `dawExtraState` like every other field and connect in `loadFromDawExtraState()` once the tuning is in place.

Also fold the three copies of the broadcast into one `publish_tuning_as_oddsound_main()`, which now sends the table with a single `MTS_SetNoteTunings()` call instead of 128 individual ones.

Fixes #7812

### DIFF
--- a/src/common/SurgePatch.cpp
+++ b/src/common/SurgePatch.cpp
@@ -3814,18 +3814,21 @@ void SurgePatch::load_xml(const void *data, int datasize, bool is_preset)
             }
 
             auto mts_main = TINYXML_SAFE_TO_ELEMENT(de->FirstChild("oddsound_mts_active_as_main"));
+            dawExtraState.oddsoundMTSActiveAsMain = false;
+
             if (mts_main)
             {
                 int tv;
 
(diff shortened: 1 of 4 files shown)


mkruselj:
Rebase-merge this PR!

Issue #8569, opened 2026-09-14
Channel-to-octave shift is wrong when Surge XT is an MTS-ESP client


mkruselj:
When **Use MIDI Channel for Octave Shift** is enabled and Surge XT is connected to an MTS-ESP source, each MIDI channel should transpose by one period of the incoming tuning. It does not, and in some setups it transposes in the wrong direction.

The shift is computed in `SurgeVoice::channelKeyEquivalent()`, and its MTS-ESP handling has two problems:

- **The MTS-ESP branch is checked after `storage->isStandardTuning`.** That flag describes Surge's own internal tuning, not the MTS-ESP one. So in the usual MTS-ESP setup, where no internal tuning is loaded, the shift is always a flat 12 semitones regardless of the source's period.
- **When the MTS-ESP branch is reached, it shifts by `MTS_GetMapSize()` semitones.** Map size is a number of keys per period, not a pitch distance. The value being shifted is already in semitones (key plus MTS-ESP retuning), so this is only correct when the two happen to coincide, as with 12-note octave scales. For Bohlen-Pierce (13 keys per period, a period of about 19.02 semitones) each channel lands roughly 6 semitones short. Worse, `MTS_GetMapSize()` returns -1 when the source does not supply a map size, which makes every channel shift one semitone **down**.

#8143 intended to use the scale period in MTS-ESP mode. `SurgeStorage::tuningPeriodSemitones()`, added since, already returns the correct quantity for every tuning mode including MTS-ESP (via `MTS_GetPeriodSemitones()`, which falls back to 12 when the source supplies no period), and the non-MTS branches of `channelKeyEquivalent()` duplicate it exactly.

### Steps to reproduce

1. Run an MTS-ESP source with a non-octave scale, for example a Surge XT instance with a Bohlen-Pierce scale loaded and **Act as MTS-ESP Source** enabled.
2. In a second Surge XT instance connected as a client, with no tuning of its own loaded, enable **Use MIDI Channel for Octave Shift**.
3. Play the same note on MIDI channels 1 and 2.

**Expected:** channel 2 sounds one period (3/1) above channel 1.
**Actual:** channel 2 sounds 12 semitones above channel 1.

Loading any internal tuning into the client instance first switches to the map size code path instead, giving a 13-semitone shift for Bohlen-Pierce, or a 1-semitone downward shift with a source that does not publish its map size.

iibaranov-IG:
I opened draft PR #8572 with the focused repair and regression test.

[...]

PR #8572, opened 2026-09-14
Use the MTS-ESP tuning period for MIDI channel shifts

iibaranov-IG:
With Use MIDI Channel for Octave Shift enabled, an MTS-ESP client uses 12 semitones when the source period differs and can shift down by one semitone when the source supplies no map size.

## Change

Replace the duplicated tuning-mode branch in channelKeyEquivalent() with the existing tuningPeriodSemitones() helper and add a C++ regression for an MTS source without map metadata.

Related issue: https://github.com/surge-synthesizer/surge/issues/8569

## Reproduction and acceptance

Mark a Surge storage instance as an MTS-ESP client whose source has no map size, then shift MIDI key 60 by channel index 1. The old channelKeyEquivalent path uses MTS_GetMapSize()=-1 and returns key 59 instead of the MTS period fallback at key 72.

Channel shifts delegate to tuningPeriodSemitones(), which uses MTS_GetPeriodSemitones() for an active client and preserves the existing behavior for standard and internal tunings; the missing-map regression returns 72 and the complete tuning suite remains green.

## BCL evidence

Run: https://github.com/iibaranov-IG/broadcast-control-lab/actions/runs/34843954537

Artifact: `batch-evidence/surge-8569`
BCL revision: `83809f34c29edf47992cb6686167f33f18310d7e`
Tested baseline: `surge-synthesizer/surge@e2b3494b0b12172b7582e340cac182606bf36f9f`
Candidate SHA-256: `b72a628f7fbf7050cb6a133e408e02d219a9027098eddac0830606d9ba27fbf5`

- baseline-red: PASS; command: `node --test tests/test_mts_channel_period_bcl.mjs`; exit: 1; log: `upstream-2-baseline-red.log`
- candidate-green: PASS; command: `node --test tests/test_mts_channel_period_bcl.mjs`; exit: 0; log: `upstream-3-candidate-green.log`
- candidate-suite: PASS; command: `node --test tests/test_mts_channel_period_bcl.mjs`; exit: 0; log: `upstream-4-candidate-suite.log`

## Owner check

Connect Surge XT as an MTS-ESP client to a source using a non-octave scale such as Bohlen-Pierce, enable Use MIDI Channel for Octave Shift, and play the same note on channels 1 and 2. Confirm channel 2 is one source period (3:1 for Bohlen-Pierce) above channel 1. Repeat with a source that omits map size and confirm channel 2 moves up one 12-semitone fallback period rather than down one semitone.

BCL verifies the exact source transformation, the MTS-period dispatch, and presence of the executable upstream C++ regression. A local macOS arm64 Release build passed the targeted regression and all 20 tuning test cases (61,304 assertions), but that local run is not treated as immutable BCL Actions evidence. Upstream cross-platform CI remains the publication gate.

Hardware verified: false. Full application verified: false.

The evidence artifact contains HARDWARE-CHECK.md and owner-result.json. Automated checks do not certify hardware.


mkruselj:
Yup, I already have a fix for this issue on a branch, just haven't pushed it yet! And that branch fixes more than one thing, so I would prefer that one to go instead of this one, if you don't mind!

baconpaul:
this fix is I think not appropriate

MTS-ESP has per channel tuning. Surge should consume those and ignore per channel octave shift if it is an MTS client

If you would like surge as an MTS source to provide this, we can do that as a separate project

Andreya-Autumn:
That's not quite right. Almost no MTS-ESP sources populate the per-channel tables, in fact only Oddsund Mini does. Channel/octave (or more correctly per period) working with MTS-ESP makes Surge more useful for Lumatone users who want to use a source other than Mini. I discussed this with out Lumatone friends and we agreed the correct choice is choct should be a choice and if you want to use mts-esp multichannel you turn it off. 

Lemme take a closer look later and see which of these changes, if any, do make sense.

mkruselj:
@Andreya-Autumn Can you copy that response to the actually opened PR #8574 ?

PR #8573, opened 2026-09-14
Show MTS-ESP tunings in full in the Tuning Editor

### LINKED ISSUE #8148: Expand the Tuning Analyzer capabilities in MTS-ESP mode
Now that we can, why not?

mkruselj:
Infer a Scala scale and mapping from what an MTS-ESP source sends, checked
against every note, and give that to the Tuning Editor as is. All its views
work while connected as a client, but read-only.

Closes #8148


baconpaul:
Older versions of MTS-ESP don't havre the period / scale length API. Have we made sure this succesfully degrades when using a runtime with that case?

mkruselj:
We already have tests which test the degraded values:

  - "An Octave Scale Without Map Information Is Found From The Frequencies"
  - "A Non-Octave Scale Without Any Shape Information Falls Back To The Whole Keyboard"

So this all comes out cleanly then!

baconpaul:
cool then i would just try it with a few non-repetition-interval 2 scales (b-p is a good one) and a few one note scales (like the carlos greek letter ones) and if that works then feel free to merge!

mkruselj:
B-P already works beautifully, I tested that one first. 🙂

baconpaul:
Great. So just test one of the one note scales like https://github.com/surge-synthesizer/surge/blob/main/resources/data/tuning_library/SCL/35.1%20CET%20-%20Wendy%20Carlos%20-%20Gamma.scl

mkruselj:
That one works too! This pic is from MTS cliented Surge:

image

baconpaul:
Awesome!! merge that sucker.

PR #8574, opened 2026-09-14
Shift channel octaves by the tuning period under MTS-ESP

### LINKED ISSUE #8569: Channel-to-octave shift is wrong when Surge XT is an MTS-ESP client

### LINKED ISSUE #8570: Portamento glides from the wrong pitch with channel-to-octave shift under MTS-ESP
With **Use MIDI Channel for Octave Shift** enabled and Surge XT connected to an MTS-ESP source, portamento starts its glide from the wrong pitch on any MIDI channel other than channel 1.

A glide interpolates from `state.portasrc_key` to `state.getPitch()`, and the start point is set in `SurgeVoice::resetPortamentoFrom()`. Without MTS-ESP, that function passes the prior key through `channelKeyEquivalent()`, so it gets the same per-channel period shift as the target pitch. The MTS-ESP branch returns early with only the retuning applied:

```cpp
lk += MTS_RetuningInSemitones(storage->oddsound_mts_client, lk, channel);
state.portasrc_key = lk;
```

The target pitch from `getPitch()` does include the channel shift, so on a channel shifted by *n* periods every glide starts *n* periods too low (or too high on channels 9-16, which shift downwards). A small legato step played on channel 3 turns into a sweep up from two periods below. With **Constant Rate** portamento, the glide time is derived from that distance too, so the glide also takes far longer than it should. Channel 1 has no shift, which makes this easy to miss.

The same branch has a second, related mismatch. It always passes the note's MIDI channel to `MTS_RetuningInSemitones()`, whereas `getPitch()` passes -1 whenever `mtsUseChannelWhenRetuning` is false (MPE, channel split, **Use MIDI Channels 2 and 3 to Play Scenes Individually**, or channel octave shift itself). With an MTS-ESP source that uses multichannel tuning tables, the glide can therefore start from one table and end on another.

### Steps to reproduce

1. Run an MTS-ESP source, for example a Surge XT instance with **Act as MTS-ESP Source** enabled (a 12-TET tuning is enough).
2. In a second Surge XT instance connected as a client, enable **Use MIDI Channel for Octave Shift**, set the scene to **Mono** with a noticeable **Portamento** time.
3. On MIDI channel 3, play C4 and then legato to D4.

**Expected:** a short glide from C to D, both two octaves up.
**Actual:** the glide sweeps up from roughly two octaves below.

Repeating step 3 on MIDI channel 1, or with MTS-ESP disconnected, glides correctly.

This is related to #8569, which fixes the size of the per-channel shift itself under MTS-ESP.

mkruselj:
`channelKeyEquivalent()` checked Surge's own isStandardTuning before MTS-ESP, so clients always shifted 12 semitones, and otherwise shifted by the MTS-ESP map size, a key count that is -1 when the source doesn't supply it. Use `tuningPeriodSemitones()`, which already covers every tuning mode.

Fixes #8569
Fixes #8570

Also:

`resetPortamentoFrom()` skipped the channel octave shift under MTS-ESP, and always retuned with the note's channel where `getPitch()` may pass -1, so glides started from a different pitch than the prior key actually sounded.


baconpaul:
Please don't merge this. See the discussion in 8569

mkruselj:
Ohh gotcha. Well I'll be damned. 😁

I guess the second commit about portamento reset is still valid, though. That issue is real.

baconpaul:
Right. So I think the thing to do is

1. make the surge-as-master with octave shift publish RI shifts onto the per channel mts bus
2. make sure the surge-as-client puts channel through and works in omni mode
3. fix the porta

mkruselj:
Two further things turned up after tracing (c/o @Andreya-Autumn):

1. Turning octave shift off isn't enough to get per-channel tables. As a client, Surge only passes the MIDI channel to MTS-ESP when `mtsUseChannelWhenRetuning` is true (SurgeVoice.cpp). That flag is also false whenever "Use MIDI Channels 2 and 3 to Play Scenes Individually" is on, and that preference defaults to on. So with default settings, Surge never uses per-channel tables, even with octave shift off. MTS-ESP's guidance is to always pass the channel except with MPE. For Andreya's plan to work, that condition probably needs loosening. This is Paul's point 2 either way.
2. The portamento problem is deeper than #8570 describes. Surge remembers the previous note's key per scene, but not its channel. When a new voice starts a glide, it retunes the previous key on the new note's channel, and with octave shift on it applies the new channel's shift as well. Going from C on channel 1 to D on channel 2 therefore starts from C an octave too high. This happens with or without MTS-ESP; legato across channels is already correct. Commit 843afac0 only makes the MTS-ESP branch consistent with the non-MTS one, so both still share this problem. A full fix would store the previous note's channel alongside its key.

Andreya-Autumn:
Point 1 is on purpose. Combining the MTS-ESP multichannel tuning tables with literally any of Surges other usages of MIDI channel seemed so nonsensical that I purposefully disallowed it. See longer comment on the issue.

Andreya-Autumn:
The portamento thing seems real and the fix looks correct!

PR #8575, opened 2026-09-14
Hardening against poisoned audio streams

### LINKED ISSUE #8240: Volume Spike
**Bug Description:**
Volume Spiked and maxed out the ableton volume meter for a minute or so. Very painful for the ears. 
**Surge XT Version**
Surge XT 1.3.4.f7b97c6

* Version: 1.3.4.f7b97c6
* Plugin Type: VST3
* Bitness: 64-bit

**Reproduction Steps:**
i did not reproduce the error but the error happened when:;
1. Factory preset played 
2. Audio device plugged out mid session 
3. Restarted Ableton because i needed to do that to get my audio device to be noticed in Ableton again 
4. Playback 
5. Boom! volume spike and ear pain 

**Expected Behavior:**
Volume spike 
The factory sound changed a bit, might have been the volume but sounded way brighter and super distorted

 - OS: Windows 11
 - Host: Ableton

mkruselj:
A host can report a rate of zero or garbage while an audio device is being torn down or re-initialized. Applying it filled every rate-derived table with inf and NaN, pinning oscillators at Nyquist and making envelopes instantaneous. The only range check lived in the SurgeStorage constructor, where it could never trigger.

JUCE does not zero the buffer handed to `processBlock()`, so returning early on a bad bus configuration played back stale or uninitialized memory. Clear it on every early return, and clear the scene buses whenever they aren't written.

Conditioner never reset its filters, Exciter never reset its oversampler, and Flanger never cleared its delay lines, so a single non-finite sample reaching any of them recirculated forever, even across a panic. Bumps `sst-effects` to latest for the Flanger fix.

Hardclip mode was cast straight from the patch file, and a value outside the enum matched no case at clip time, silently disabling the clipper.

When no audio input was processed, the right channel was cleared twice and the left one never, so Audio Input and Vocoder read stale or uninitialized memory.

The not-bypassed flag was set at the top of every block, so the countdown was restarted each time and `stopSound()` ran on every bypassed block. Let it actually complete now.

In Surge XT Effects plugin, `std::clamp()` passes NaN straight through, so the output hard clip did not catch it. Now it does!

Fixes #8240


mkruselj:
Probably makes sense to rebase-merge this one!

PR #8576, opened 2026-09-14
Ramp the Sine/FM2/FM3 oscillator's phase increment across each block

### LINKED ISSUE #3784: Sine Oscillator Aliasing with LFO on Pitch
**Is your feature request related to a problem? Please describe.**
Sometimes, when using osciallators at very high-frequencies (>10kHz) I can hear aliasing occuring, even on a sine wave. This is especially apparent when modulating the pitch.

I think that maybe adding an option to oversample the oscillators (to avoid aliasing) and then lowpassing them before feeding the signal down the line could solve this issue. 

Maybe including an oversampling option for the entire signal chain?

mkruselj:
Pitch reaches the oscillator once per block, and `omega` was held constant for the whole block. Under pitch modulation that steps the instantaneous frequency at the block rate, which puts sidebands spaced at the block rate around the fundamental. That is what the issue reports: with the waveshaper off, an LFO sweep carried spurs at -55 to -70 dBc.

`omega` now ramps linearly across the block, with the slope of the change since the prior block, bringing those spurs down to around -100 dBc. The ramp is centered on the current block's `omega` rather than starting from the prior one, so each block advances the phase exactly as before. A ramp from the prior value lagged by half a block, and that lag accumulated into a phase offset proportional to the total pitch change, which noticeably changed pitch enveloped kicks, so that's why it was decided not to start ramping from the prior value, but to center it within the current block.

Static pitch is bit-identical, and the legacy FM behavior is untouched.

Fixes #3784


baconpaul:
Awesome

PR #8579, opened 2026-09-18
Fix wave table save failure

### LINKED ISSUE #8548: Saving a wavetable script truncates the existing file before the save can fail
`SurgeGUIEditor::saveWavetableScript` opens the destination for writing before it does any of the work that can fail:

```cpp
std::ofstream outFile(fullLocation, std::ios::binary);   // SurgeGUIEditor.cpp:7324
```

That truncates whatever was already at `fullLocation`. Every failure after that point returns having destroyed the previous file:

- `SurgeGUIEditor.cpp:7347` — `ZSTD_compress` fails, we report "Failed to compress snapshot data." and return. The file is now zero bytes.
- `SurgeGUIEditor.cpp:7368` — the post-`close()` check reports "Failed to write file." after a short or failed write. The file is now empty, or truncated part way through the blob.

The user gets there through the overwrite prompt: they confirm "yes, overwrite", get an error dialog, and their previous script is gone. The dialog says the save failed, which reads as "nothing happened" rather than "the file you had is now empty".

The compress path is unlikely in practice, since `compBound` comes from `ZSTD_compressBound`. The write path is the realistic one — a full disk, or an I/O error on a network or removable volume, truncates the file and only then reports failure.

Fix would be to write to a temporary file in the same directory and rename it over the target once everything is written and closed cleanly, so a failure leaves the existing file alone. Assembling the whole payload (header + XML + blob) in memory before opening would close the compress path, but not the partial-write one.

Spotted while reviewing #8529, which touches this function but deliberately leaves this alone.

yuing798:
When a user attempts to save a wave table and the operation fails, the original file is deleted. I resolved issue #8548 by creating a temporary file and overwriting the original file after the save was successful.

Fixes #8548


mkruselj:
@yuing798 Thank you! Two things:

1. Was this done with AI assistance? If yes, please update your commit message with `Assisted-By: <name of the model>`. If not, no problem!
2. Please add yourself to AUTHORS file!

